PR #75, opened 2021-08-05
Explicitly import etcdraft

yacovm:


adecaro:
Why do we need this?

yacovm:
> Why do we need this?

Otherwise the integration tests don't run when it attempts to build the fabric nodes via `go build`

adecaro:
Why is that?

adecaro:
Does this create a dependency to a specific fabric version?

yacovm:
> Does this create a dependency to a specific fabric version?

Of course it does.

PR #174, opened 2022-02-25
NFT support

adecaro:
This goal of this Pull Request is manyfold.
- Introduce a service package for NFT (`nftcc`). This package will augment a `ttxcc` transaction with methods targeting NFTs. 
- Introduce an integration test for a DvP between fungible and non-fungible tokens
- Update the documentation

In a subsequent PR, the integration test will be used to create a sample app.


docs/token-api.md
```diff
@@ -9,11 +9,16 @@ A token consists of the following triplet:
...
-- `Quantity`: The amount stored by this token. It is a positive number
+- `Quantity`: The amount stored by this token. It is a positive number, larger or equal to zero, 
   encoded as a string containing a number in base 16. The string starts with the prefix `0x`.
 
-These tokens are `fungible` with the respect to the same type. In particular,
-tokens with the same denomination can be merged and split, if not otherwise forbidden.
+These tokens are `fungible` with the respect to the same type. 
+In particular, tokens with the same denomination can be merged and split, if not otherwise forbidden.
+
+It looks like then that the above definition allows the developers to define non-fungible tokens as well.
```
mbrandenburger, line 18:
```suggestion
It looks like that the above definition allows the developers to define non-fungible tokens as well.
```

integration/token/tcc/dvp/views/auditor.go
```diff
@@ -0,0 +1,45 @@
...
+*/
+
+package views
+
+import (
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/assert"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
+
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/ttxcc"
+)
```
mbrandenburger, line 14: maybe adding goimports (and gofmt?) checks to the CI pipeline would help to automate to detect these formatting issues?!

token/services/nftcc/uniqueness/uniqueness.go
```diff
@@ -0,0 +1,84 @@
...
+		n, err := rand.Read(key)
+		if err != nil {
+			return "", errors.Wrap(err, "error getting random bytes")
+		}
+		if n != size {
+			return "", errors.New("error getting random bytes")
+		}
+		if err := kvs.Put(k, key); err != nil {
+			return "", errors.WithMessagef(err, "failed to put key %s", k)
+		}
```
mbrandenburger, line 58: What about the use of `errors.Wrap` and `errors.WithMessagef`. For all those errors here, does it make sense to use `errors.Wrap` to also extend the stack trace, rather than only extending the error message?

token/services/nftcc/transaction.go
```diff
@@ -0,0 +1,144 @@
...
+func (t *Transaction) Transfer(wallet *token.OwnerWallet, state interface{}, recipient view.Identity, opts ...token.TransferOption) error {
+	// marshal state to json
+	stateJSON, err := json.Marshal(state)
+	if err != nil {
+		return errors.Wrap(err, "failed to marshal state")
+	}
+	stateJSONStr := base64.StdEncoding.EncodeToString(stateJSON)
+
+	return t.Transaction.Transfer(wallet, stateJSONStr, []uint64{1}, []view.Identity{recipient}, opts...)
+}
```
mbrandenburger, line 102: Would it make sense to refactor this code a bit in order to move the json specific calls to another file/package/component?! This would allow us to switch the serialization easier in case we want/need at some point?

integration/token/tcc/dvp/topology.go
```diff
@@ -0,0 +1,102 @@
...
+
+	tokenTopology := token.NewTopology()
+	tokenTopology.SetDefaultSDK(fscTopology)
+	tms := tokenTopology.AddTMS(fabricTopology, tokenSDKDriver)
+	tms.SetNamespace([]string{"Org1"}, "100", "2")
+
+	// Monitoring
+	//monitoringTopology := monitoring.NewTopology()
+	//monitoringTopology.EnableHyperledgerExplorer()
+	//monitoringTopology.EnablePrometheusGrafana()
```
mbrandenburger, line 94: just commented out for the moment?

mbrandenburger (commented):
I quickly skimmed this PR. It's huge, so it's hard to judge if something is really bad :) I left a few comments.

integration/token/tcc/dvp/views/auditor.go
adecaro: I have applied them but it is not considered a formatting issue that one.

integration/token/tcc/dvp/topology.go
adecaro: forgot to remove

token/services/nftcc/uniqueness/uniqueness.go
adecaro: I'm using WithMessagef because I expect the kvs to already have the stack attached. So, I just want to append the message. But, I will adjust the other lines that are not in sync

token/services/nftcc/transaction.go
adecaro: good point. Addressed via functions in marshaller.go that can be changed at anytime later

adecaro:
Thanks, Marcus. I have addressed all your comments.

KElkhiyaoui (approved):

PR #182, opened 2022-03-03
remove dvp integration test

adecaro:
remove dvp integration test, it will be replaced by another dvp test that combines fungile and non-fungible tokens (#174)


mbrandenburger (commented):
LGTM. For me it is hard to judge if there are anywhere some dvp related leftovers.

Issue #186, opened 2022-03-04
token selector: streamline filter interfaces


adecaro:
The Token Selector takes in input a filter to filter the tokens to be selected. This filter mixes different things.
This issue proposes to streamline the filter interface and split it in multiple filter types with different behaviour.

PR #187, opened 2022-03-04
[WIP] token selector: streamline filter interfaces #186

adecaro:

### DIFF
--- a/token/selector.go
+++ b/token/selector.go
@@ -3,6 +3,7 @@ Copyright IBM Corp. All Rights Reserved.
 
 SPDX-License-Identifier: Apache-2.0
 */
+
 package token
 
 import (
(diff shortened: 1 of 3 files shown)


token/selector.go
```diff
@@ -25,21 +26,31 @@ var (
...
+	// ID returns the ID of the identity the selection should select for.
 	ID() string
+}
+
+// SelectorFilterByUnspentToken is a SelectorFilter that selects tokens that are not yet spent.
+type SelectorFilterByUnspentToken interface {
+	SelectorFilter
 	// ContainsToken returns true if the passed token is recognized, false otherwise.
 	ContainsToken(token *token2.UnspentToken) bool
 }
```
mbrandenburger, line 44: Mhhh. The filter used here is only there to dispatch the select call to either to selectByOwnerID or selectByFilter.

My understanding of a filter is something that once applies on a data structure and there is a predicate to be true in order to get the item or not. this is not the case with the selectorFilter. Also, the filter may have more predicates. For instance, tokenType could be one. We maybe wanna add more filters, like minTokenSize, maxTokenSize, oldestToken, …. In order to model is nicely, we could throw away the filter interface it is right now and make it just a function, or a list of functions.  WDYT?

mbrandenburger (approved):
This looks already better than before. See my comment on the filter, but let's address this later.

token/selector.go
mbrandenburger: @adecaro is the ID of the identity always a string? In some code I saw that this requires to do something some casting of the actual identity using the "raw material" `string(Owner.Raw)`.

token/services/selector/selector.go
```diff
@@ -7,11 +7,11 @@ SPDX-License-Identifier: Apache-2.0
 package selector
 
 import (
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
 	"strconv"
 	"time"
 
 	"github.com/hashicorp/go-uuid"
-	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
 	"github.com/pkg/errors"
```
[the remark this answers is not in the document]
adecaro: if I run goimports, it gets worse 😆

adecaro:
I will withdraw this waiting for a deeper discussion on the selector API :)

PR #193, opened 2022-03-15
F 181

KElkhiyaoui:


token/core/zkatdlog/crypto/transfer/sender.go
```diff
@@ -169,6 +169,9 @@ func (t *TransferAction) IsGraphHiding() bool {
 	return false
 }
 
+func (t *TransferAction) GetMetadata() []byte {
+	return nil
+}
```
adecaro, line 174: can we add a new line to separate the function, please?

adecaro (changes requested):
LGTM. Just a minor comment to have the code better formatted :)

adecaro (approved):

PR #194, opened 2022-03-16
Test the token transaction release function

adecaro:
This PR tests the token transaction release function that is triggered
when a context is cancelled due to a panic or an error returned by a view.
The test first locks a token and trigger the release, then
it tries to relock again the same token. The latter operation should succeed.


mbrandenburger (approved):
LGTM!

PR #207, opened 2022-03-30
network.fabric: support FPC

adecaro:
If the token chaincode is a FPC, then we must pass the token request as an input not via transient.
This, at least, until FPC supports transient: https://github.com/hyperledger/fabric-private-chaincode/issues/666

### DIFF
--- a/token/services/network/fabric/network.go
+++ b/token/services/network/fabric/network.go
@@ -8,19 +8,18 @@ package fabric
 
 import (
 	"encoding/json"
-	"strings"
-	"sync"
-
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric"
 	idemix2 "github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/msp/idemix"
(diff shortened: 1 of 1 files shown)


token/services/network/fabric/network.go
```diff
@@ -309,3 +319,43 @@ func (n *Network) GetEnrollmentID(raw []byte) (string, error) {
...
+	).WithSignerIdentity(
+		signer,
+	).WithTransientEntry(
+		"token_request", requestRaw,
+	).WithTxID(
+		fabric.TxID{
+			Nonce:   txID.Nonce,
+			Creator: txID.Creator,
+		},
+	).Endorse(context)
```
mbrandenburger, line 360: I believe you missed to remove the transient entry and pass the `requestRaw` as parameter :)

mbrandenburger (changes requested):

token/services/network/fabric/network.go
adecaro: good catch, fixed :)

mbrandenburger (approved):
LGTM now

PR #208, opened 2022-03-30
Add token request extractor to TCC

mbrandenburger:
Refactor TCC that the tokenRequest extract logic can be injected. This
allows to use alternativ implmentations of the extracting logic other
than the default transient-based logic. A use case for this is, the
FPC-TCC. Currently, FPC does not support transient data, so the
tokenRequest must be provided via function args.

Signed-off-by: Marcus Brandenburger <bur@zurich.ibm.com>


mbrandenburger:
An example of an alternative extraction method can be found in the FPC-TCC sample. 
https://github.com/mbrandenburger/fabric-secure-chaincode/blob/another-example/samples/chaincode/token-validation/main.go#L77

adecaro (approved):
LGTM

PR #211, opened 2022-04-01
nwo token enhancements 

adecaro:
- ability to set custom crypto material generators
- ability to set token chaincode params replace suffix


integration/nwo/token/fabric/tcc.go
```diff
@@ -71,7 +71,7 @@ func (p *NetworkHandler) tccSetup(tms *topology3.TMS, cc *topology.ChannelChainc
 		func(s string, s2 string) (string, []byte) {
 			// logger.Infof("replace [%s:%s]?", s, s2)
 			// Is the public params?
-			if strings.HasSuffix(s, "/token/services/network/fabric/tcc/params.go") {
+			if strings.HasSuffix(s, p.TokenChaincodeParamsReplaceSuffix) {
```
mbrandenburger, line 74: what is the variable `s` here?

adecaro: renamed the variables to match their meanings

PR #231, opened 2022-05-03
Fix data race on TMS and expose TMSProvider

mffrench:


token/core/identity/tms/wm.go
```diff
@@ -447,8 +447,8 @@ func (lm *LocalMembership) addResolver(Name string, Type string, EnrollmentID st
 }
 
 func (lm *LocalMembership) getAnonymousResolver(label string) *Resolver {
-	lm.resolversMutex.RLock()
-	defer lm.resolversMutex.RUnlock()
+	lm.resolversMutex.Lock()
```
adecaro, line 450: but here read lock should be enough because we are not modifying the data structues.

token/core/identity/tms/wm.go
```diff
@@ -411,8 +411,8 @@ func (lm *LocalMembership) registerIdentity(id string, typ string, path string,
 }
 
 func (lm *LocalMembership) addResolver(Name string, Type string, EnrollmentID string, defaultID bool, IdentityGetter GetIdentityFunc) {
-	lm.resolversMutex.RLock()
-	defer lm.resolversMutex.RUnlock()
+	lm.resolversMutex.Lock()
```
adecaro, line 414: good catch

adecaro (changes requested):

token/core/identity/tms/wm.go
mffrench: 👍

mffrench: @adecaro : fixed ;)

adecaro (approved):

PR #235, opened 2022-05-05
tokengen: add support for custom generator #234

adecaro:

### DIFF
--- a/token/core/cmd/pp/gen.go
+++ b/token/core/cmd/pp/gen.go
@@ -59,6 +59,9 @@ var (
 )
 
 func AddGenerator(label string, generator GeneratorFunc) {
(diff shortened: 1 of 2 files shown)


token/core/cmd/pp/gen.go
```diff
@@ -51,8 +53,24 @@ var (
...
+
+	// public parameters generators
+	generators map[string]GeneratorFunc
 )
 
+func AddGenerator(label string, generator GeneratorFunc) {
+	if generators == nil {
+		generators = make(map[string]GeneratorFunc)
+	}
+	generators[label] = generator
```
mbrandenburger, line 65: better check that the function is not `nil`

Issue #238, opened 2022-05-16
tokengen: make GeneratorFunc independent from the global variables


adecaro:
Currently, an implementation of the GeneratorFunc template must use the global variable to access the input variables passed to `tokengen`. 

To simplify the implementation of new generators, we will introduce a struct containing all the inputs passed to `tokengen`. This struct will then passed to the generator.

PR #239, opened 2022-05-16
tokengen: make GeneratorFunc independent from the global variables #238

adecaro:

### DIFF
--- a/token/core/cmd/pp/gen.go
+++ b/token/core/cmd/pp/gen.go
@@ -26,12 +26,30 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type GeneratorFunc = func(args []string) ([]byte, error)
+type GeneratorArgs struct {
+	// Driver is the Token-SDK driver to use
+	Driver string
+	// IdemixMSPDir is the directory containing the Idemix MSP config (Issuer Key Pair)
+	IdemixMSPDir string
+	// OutputDir is the directory to output the generated files
+	OutputDir string
+	// GenerateCCPackage is whether to generate the chaincode package
+	GenerateCCPackage bool
+	// Issuers is the list of issuers to include in the public parameters.
+	// Each issuer should be specified in the form of <MSP-Dir>:<MSP-ID>
+	Issuers []string
+	// Auditors is the list of auditors to include in the public parameters.
+	// Each auditor should be specified in the form of <MSP-Dir>:<MSP-ID>
+	Auditors []string
 
-const (
-	IdemixConfigDirMsp              = "msp"
-	IdemixConfigFileIssuerPublicKey = "IssuerPublicKey"
-)
+	// Base is a dlog driver related parameter
+	Base int64
+	// Exponent is a dlog driver related parameter
+	Exponent int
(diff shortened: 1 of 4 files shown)


token/core/cmd/pp/gen.go
```diff
@@ -90,8 +115,8 @@ func Cmd() *cobra.Command {
 	flags.StringSliceVarP(&Auditors, "auditors", "a", nil, "list of auditor keys in the form of <MSP-Dir>:<MSP-ID>")
 	flags.StringSliceVarP(&Issuers, "issuers", "s", nil, "list of issuer keys in the form of <MSP-Dir>:<MSP-ID>")
 
-	flags.Int64VarP(&Base, "base", "b", 100, "base field used by the dlog driver")
-	flags.IntVarP(&Exponent, "exponent", "e", 2, "exponent field used by the dlog driver")
+	flags.Int64VarP(&Base, "base", "b", 100, "tt is used to define the maximum quantity a token can contain as Base^Exponent")
+	flags.IntVarP(&Exponent, "exponent", "e", 2, "tt is used to define the maximum quantity a token can contain as Base^Exponent")
```
mbrandenburger, line 119: What is `tt`?

mbrandenburger (commented):
This looks like we are heading in the right direction. Let me propose some more improvements:
- What about having a cobra subcommand for each available generator (e.g. `tokengen gen fabtoken params`, `tokengen gen dlog params`, ...). By doing this, each subcommand can deal with the specific parameters the generator needs. The subcommand assembles the arguments and passes it to the generator function. That is, each generator receives the inputs it needs and no others. For instance, the `Fabtoken` generator will not receive any esoteric inputs needed for `dlog`. :) Thus, the subcommand can also enforce that the user provides all needed parameters with and even checks validity. Sorry for suggesting the common `GeneratorArgs` struct in the first place. :)

Future changes to a specific generator will not impact other generators. CMD documentation for each generator is available in the corresponding subcommand.

- The tokengen (`main.go`) just loads the available generator subcommands. In the future we can easily add or remove generators.   

- No need anymore for package variables consumed by the generator impl. 

All the tokengen commands will be implemented in `cmd/tokengen/generators/` and loaded with something like ...

```go
func init() {
        rootCmd.AddCommand(genCmd)
	genCmd.AddCommand(fabtokenGenCmd)
	genCmd.AddCommand(dlogGenCmd)
...
```
The actual generator implementations are then placed in `token/core/cmd/pp/`, having a package for each available generator. If there is some common functionality we would like to share, we can put it in `token/core/cmd/pp/common`. All the generator impls are free from any cobra code and thus can be easily used somewhere else (e.g., some unit/integration tests or another application).

token/core/cmd/pp/gen.go
adecaro: Thanks @mbrandenburger , I have moved to subcommands :)

mbrandenburger (approved):
The refactoring using cobra subcommands looks good to me. Now we can add "new" generators by just adding them to `token/core/cmd/pp/gen.go`

PR #254, opened 2022-06-10
fix AuditDB redeem bug

adecaro:
Currently, AuditDB fails to record redeem transactions because it does consider only the enrollment IDs of the outputs.
In a redeem, the output has no enrollment ID.
To make sure the fix works, we augment the CheckBalance function in the integration test to enforce that the holding of the party, at the auditor, is the same as the balance at the party.
We also introduce additional checks for the current spending.

### DIFF
--- a/integration/token/fungible/tests.go
+++ b/integration/token/fungible/tests.go
@@ -523,7 +523,7 @@ func CheckSpending(network *integration.Infrastructure, id string, wallet string
 	Expect(err).NotTo(HaveOccurred())
 	spending, err := strconv.ParseUint(common.JSONUnmarshalString(spendingBoxed), 10, 64)
 	Expect(err).NotTo(HaveOccurred())
-	Expect(spending).To(Equal(int(expected)))
+	Expect(spending).To(Equal(expected))
 }
 
 func ListIssuerHistory(network *integration.Infrastructure, wallet string, typ string) *token2.IssuedTokens {
--- a/integration/token/fungible/tests.go
+++ b/integration/token/fungible/tests.go
@@ -508,7 +508,7 @@ func CheckBalance(network *integration.Infrastructure, id string, wallet string,
 	Expect(holding).To(Equal(int(expected)))
 }
 
-func CheckSpending(network *integration.Infrastructure, id string, wallet string, typ string, expected uint64) {
+func CheckSpending(network *integration.Infrastructure, id string, wallet string, tokenType string, expected uint64) {
 	// check spending
 	// first get the enrollment id
 	eIDBoxed, err := network.Client(id).CallView("GetEnrollmentID", common.JSONMarshall(&views.GetEnrollmentID{
@@ -518,10 +518,10 @@ func CheckSpending(network *integration.Infrastructure, id string, wallet string
 	eID := common.JSONUnmarshalString(eIDBoxed)
 	spendingBoxed, err := network.Client("auditor").CallView("spending", common.JSONMarshall(&views.CurrentSpending{
 		EnrollmentID: eID,
-		TokenType:    typ,
+		TokenType:    tokenType,
 	}))
 	Expect(err).NotTo(HaveOccurred())
-	spending, err := strconv.Atoi(common.JSONUnmarshalString(spendingBoxed))
+	spending, err := strconv.ParseUint(common.JSONUnmarshalString(spendingBoxed), 10, 64)
 	Expect(err).NotTo(HaveOccurred())
 	Expect(spending).To(Equal(int(expected)))
 }
--- a/integration/token/fungible/tests.go
+++ b/integration/token/fungible/tests.go
@@ -10,6 +10,7 @@ import (
 	"crypto/rand"
 	"fmt"
 	"math/big"
+	"strconv"
 	"strings"
 	"time"
 
@@ -179,6 +180,7 @@ func TestAll(network *integration.Infrastructure) {
 	TransferCash(network, "alice", "", "USD", 111, "bob")
 	t9 := time.Now()
 	CheckAuditedTransactions(network, TestAllTransactions[5:7], &t8, &t9)
+	CheckSpending(network, "alice", "", "USD", 111)
 	ut := ListUnspentTokens(network, "alice", "", "USD")
 	Expect(ut.Count() > 0).To(BeTrue())
 	Expect(ut.Sum(64).ToBigInt().Cmp(big.NewInt(9))).To(BeEquivalentTo(0))
@@ -195,6 +197,7 @@ func TestAll(network *integration.Infrastructure) {
 	t11 := time.Now()
 	CheckAuditedTransactions(network, TestAllTransactions[9:10], &t10, &t11)
 	CheckAuditedTransactions(network, TestAllTransactions[:], &t0, &t11)
+	CheckSpending(network, "bob", "", "USD", 11)
 
 	IssueCash(network, "", "USD", 1, "alice")
 
@@ -210,9 +213,12 @@ func TestAll(network *integration.Infrastructure) {
 	CheckBalance(network, "alice", "", "EUR", 10)
 	CheckBalance(network, "bob", "", "EUR", 20)
 	CheckBalance(network, "bob", "", "USD", 120)
+	CheckSpending(network, "alice", "", "USD", 121)
+	CheckSpending(network, "bob", "", "EUR", 10)
 
 	RedeemCash(network, "bob", "", "USD", 10)
 	CheckBalance(network, "bob", "", "USD", 110)
+	CheckSpending(network, "bob", "", "USD", 21)
 
 	// Check self endpoints
 	IssueCash(network, "", "USD", 110, "issuer")
@@ -261,6 +267,7 @@ func TestAll(network *integration.Infrastructure) {
 	CheckBalance(network, "manager", "manager.id3", "USD", 10)
 
 	TransferCash(network, "manager", "manager.id1", "USD", 10, "manager.id2")
+	CheckSpending(network, "manager", "manager.id1", "USD", 10)
 	CheckBalance(network, "manager", "", "USD", 20)
 	CheckBalance(network, "manager", "manager.id1", "USD", 0)
 	CheckBalance(network, "manager", "manager.id2", "USD", 20)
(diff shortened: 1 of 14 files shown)


integration/token/fungible/views/auditor.go
```diff
@@ -148,3 +149,79 @@ func (p *RegisterAuditorViewFactory) NewView(in []byte) (view.View, error) {
...
+	auditor := ttx.NewAuditor(context, w)
+
+	aqe := auditor.NewQueryExecutor()
+	defer aqe.Done()
+
+	filter, err := aqe.Holdings().ByEnrollmentId(r.EnrollmentID).ByType(r.TokenType).Execute()
+	assert.NoError(err, "failed retrieving holding for [%s][%s]", r.EnrollmentID, r.TokenType)
+	currentHolding := filter.Sum()
+	decimal := currentHolding.Decimal()
+	logger.Infof("Current Holding: [%s][%s][%s]", r.EnrollmentID, r.TokenType, decimal)
```
mbrandenburger, line 176: Make this a debug?

integration/token/fungible/views/auditor.go
```diff
@@ -148,3 +149,79 @@ func (p *RegisterAuditorViewFactory) NewView(in []byte) (view.View, error) {
...
+	auditor := ttx.NewAuditor(context, w)
+
+	aqe := auditor.NewQueryExecutor()
+	defer aqe.Done()
+
+	filter, err := aqe.Payments().ByEnrollmentId(r.EnrollmentID).ByType(r.TokenType).Execute()
+	assert.NoError(err, "failed retrieving spending for [%s][%s]", r.EnrollmentID, r.TokenType)
+	currentSpending := filter.Sum()
+	decimal := currentSpending.Decimal()
+	logger.Infof("Current Spending: [%s][%s][%s]", r.EnrollmentID, r.TokenType, decimal)
```
mbrandenburger, line 214: Make this a debug?

integration/token/fungible/views/auditor.go
```diff
@@ -148,3 +149,79 @@ func (p *RegisterAuditorViewFactory) NewView(in []byte) (view.View, error) {
...
+	w := ttx.MyAuditorWallet(context)
+	assert.NotNil(w, "failed getting default auditor wallet")
+
+	auditor := ttx.NewAuditor(context, w)
+
+	aqe := auditor.NewQueryExecutor()
+	defer aqe.Done()
+
+	filter, err := aqe.Holdings().ByEnrollmentId(r.EnrollmentID).ByType(r.TokenType).Execute()
+	assert.NoError(err, "failed retrieving holding for [%s][%s]", r.EnrollmentID, r.TokenType)
```
mbrandenburger, line 173: mh ... Angelo, I must say that I don't like this way of error handling. The call function API allows us to return an error but here we use an assertion ... panic / maybe recover somewhere ... or crash.

integration/token/fungible/views/auditor.go
```diff
@@ -148,3 +149,79 @@ func (p *RegisterAuditorViewFactory) NewView(in []byte) (view.View, error) {
 	f := &RegisterAuditorView{}
 	return f, nil
 }
+
+type CurrentHolding struct {
+	EnrollmentID string `json:"enrollment_id"`
+	TokenType    string `json:"token_type"`
+}
```
mbrandenburger, line 156: It seems that `CurrentHolding` and `CurrentSpending` contain the same fields. Will we add more fields in the future which will make these two types different? Otherwise can't we use a common struct?

mbrandenburger: Also, as this is part of the `views` package, the usage of this type feels not intuitive for me. 
```go
	spendingBoxed, err := network.Client("auditor").CallView("spending", common.JSONMarshall(&views.CurrentSpending{
		EnrollmentID: eID,
		TokenType:    typ,
	}))
```

This type is used to serialize the input of the view, so should we give it a name like `SpendingRequest`, `SpendingInput`, or similar?

integration/token/fungible/tests.go
```diff
@@ -477,6 +490,40 @@ func CheckBalance(network *integration.Infrastructure, id string, wallet string,
 	Expect(err).NotTo(HaveOccurred())
 	expectedQ := token2.NewQuantityFromUInt64(expected)
 	Expect(expectedQ.Cmp(q)).To(BeEquivalentTo(0), "[%s]!=[%s]", expected, q)
+
+	// check holding, it must be equal to the balance
+	// first get the enrollment id
+	eIDBoxed, err := network.Client(id).CallView("GetEnrollmentID", common.JSONMarshall(&views.GetEnrollmentID{
```
mbrandenburger, line 496: is there a particular reason why this view name uses camelcase whereas the auditor views are in lower case?

token/services/auditor/auditdb/db/memory/memory.go
```diff
@@ -18,11 +19,11 @@ type Persistence struct {
 	transactionRecords []*driver.TransactionRecord
 }
 
-func (p *Persistence) QueryMovements(ids []string, types []string, status []driver.TxStatus, direction driver.SearchDirection, value driver.MovementDirection, numRecords int) ([]*driver.MovementRecord, error) {
+func (p *Persistence) QueryMovements(enrollmentIDs []string, tokenTypes []string, txStatuses []driver.TxStatus, searchDirection driver.SearchDirection, movementDirection driver.MovementDirection, numRecords int) ([]*driver.MovementRecord, error) {
```
mbrandenburger, line 22: Love this renaming :)

mbrandenburger (commented):
Thanks Angelo, I see a few minor things for improvement. :)

token/services/auditor/auditdb/auditdb.go
```diff
@@ -333,8 +333,13 @@ func (db *AuditDB) Unlock(eIDs ...string) {
...
 	inputs := record.Inputs
 	outputs := record.Outputs
-	eIDs := outputs.EnrollmentIDs()
+	// we need to consider both inputs and outputs enrollment IDs because the record can refer to a redeem
+	iEIDs := inputs.EnrollmentIDs()
+	oEIDs := outputs.EnrollmentIDs()
+	eIDs := append(iEIDs, oEIDs...)
+	eIDs = deduplicate(eIDs)
 	tokenTypes := outputs.TokenTypes()
+
```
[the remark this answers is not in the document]
adecaro: okay

token/services/auditor/auditdb/db/memory/memory.go
adecaro: :)

integration/token/fungible/tests.go
```diff
@@ -477,6 +490,40 @@ func CheckBalance(network *integration.Infrastructure, id string, wallet string,
...
+		EnrollmentID: eID,
+		TokenType:    typ,
+	}))
+	Expect(err).NotTo(HaveOccurred())
+	holding, err := strconv.Atoi(common.JSONUnmarshalString(holdingBoxed))
+	Expect(err).NotTo(HaveOccurred())
+	Expect(holding).To(Equal(int(expected)))
+}
+
+func CheckSpending(network *integration.Infrastructure, id string, wallet string, typ string, expected uint64) {
```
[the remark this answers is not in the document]
adecaro: will rename to `tokenType`

integration/token/fungible/tests.go
```diff
@@ -477,6 +490,40 @@ func CheckBalance(network *integration.Infrastructure, id string, wallet string,
...
+		Wallet: wallet,
+	}))
+	Expect(err).NotTo(HaveOccurred())
+	eID := common.JSONUnmarshalString(eIDBoxed)
+	spendingBoxed, err := network.Client("auditor").CallView("spending", common.JSONMarshall(&views.CurrentSpending{
+		EnrollmentID: eID,
+		TokenType:    typ,
+	}))
+	Expect(err).NotTo(HaveOccurred())
+	spending, err := strconv.Atoi(common.JSONUnmarshalString(spendingBoxed))
```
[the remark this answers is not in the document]
adecaro: good point.

integration/token/fungible/views/auditor.go
adecaro: okay

integration/token/fungible/views/auditor.go
adecaro: I understand, indeed developers can handle error as they better prefer. If a panic is generated, the context catches it and send back to the caller of the view the error generated.
We use this approach in all integration tests. 
Can we keep it?

integration/token/fungible/tests.go
adecaro: good point, there is no rule. Shall we decide one and then submit a PR that enforces that?

integration/token/fungible/views/auditor.go
mbrandenburger: ah I didn't recognized that this is in the integration test, therefore yes let's keep it here.

integration/token/fungible/tests.go
mbrandenburger: I would just define the rule we using consistently throughout the token sdk. Other users of FSC will maybe have different opinions / naming conventions. Therefore, enforcing something like this doesn't seem to be desirable - I believe.

integration/token/fungible/views/auditor.go
adecaro: okay

mbrandenburger (approved):
LGTM now! Thanks Angelo

integration/token/fungible/views/auditor.go
KElkhiyaoui: If we are to merge the structs, we need to come up with a better name. This struct corresponds to a filter?

integration/token/fungible/views/info.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+// GetEnrollmentIDView is a view that returns the enrollment ID of a wallet.
+type GetEnrollmentIDView struct {
+	*GetEnrollmentID
+}
+
+func (r *GetEnrollmentIDView) Call(context view.Context) (interface{}, error) {
+	w := ttx.GetWallet(context, r.Wallet)
+	assert.NotNil(w, "failed getting default auditor wallet")
```
KElkhiyaoui, line 28: The view seems to be generic and not auditor specific. Should we change the error message.

integration/token/fungible/views/auditor.go
adecaro: That struct just contain the input to the view.
```In the current integration tests, we use struct X for the input to a view, XView for the view, XViewFactory for the factory. So, they all have the same prefix.```.
Do we still need to change the name?

PR #275, opened 2022-06-28
F 252: zkatdlog crypto library clean up

KElkhiyaoui:


token/core/zkatdlog/crypto/common/schnorr.go
```diff
@@ -10,67 +10,116 @@ import (
...
 
+// ComputePedersenCommitment returns the commitment to opening relative to base in passed curve.
 func ComputePedersenCommitment(opening []*math.Zr, base []*math.G1, c *math.Curve) (*math.G1, error) {
 	if len(opening) != len(base) {
 		return nil, errors.Errorf("can't compute Pedersen commitment [%d]!=[%d]", len(opening), len(base))
 	}
+	if c == nil {
+		return nil, errors.Errorf("can't compute Pedersen commitment: please initialize curve")
+	}
 	com := c.NewG1()
```
yacovm, line 66: does mathlib guarantee that NewG1() is initialized to be the identity element?

token/core/zkatdlog/crypto/common/nym.go
```diff
@@ -12,118 +12,146 @@ import (
...
 }
 
-// get verifier
+// Serialize returns NYMSigner bytes
 func (s *NYMSigner) Serialize() ([]byte, error) {
 	return s.NYMVerifier.Serialize()
 }
 
-// sign message anonymously using Schnorr signature
+// Sign produces a SOK signature for (SK, BF) such that NYM = P^SK*Q^BF
```
yacovm, line 28: what's a SOK? Should it be PoK?

token/core/zkatdlog/crypto/common/nym.go
```diff
@@ -12,118 +12,146 @@ import (
...
 func (s *NYMSigner) Sign(message []byte) ([]byte, error) {
+	if s.Curve == nil {
```
yacovm, line 30: I think in general, checking that the "self" object is correctly initialized doesn't add enough to justify its existence. 

Unlike inputs, this is the object's configuration. If we can reach a state where the object is misconfigured, we better find it in tests rather than check it at runtime, because in most cases, if we're in a real environment and we reached a forbidden state, this is an irrecoverable state.

token/core/zkatdlog/crypto/common/schnorr.go
```diff
@@ -10,67 +10,116 @@ import (
...
 	}
 	return com, nil
 }
+
+// RecomputeCommitment is called by the verifier.
+// It takes a SchnorrProof and returns the corresponding randomness commitment.
+func (v *SchnorrVerifier) RecomputeCommitment(zkp *SchnorrProof) (*math.G1, error) {
+	// safety checks
+	if zkp.Challenge == nil || zkp.Statement == nil {
+		return nil, errors.Errorf("invalid zero-knowledge proof")
```
yacovm, line 81: probably better to distinguish line 80 from line 97.

token/core/zkatdlog/crypto/common/schnorr.go
```diff
@@ -10,67 +10,116 @@ import (
...
+		if v.PedParams == nil {
+			return nil, errors.New("please initialize Pedersen parameters")
+		}
+		if p == nil {
+			return nil, errors.Errorf("invalid zero-knowledge proof")
+		}
+		com.Add(v.PedParams[i].Mul(p))
+	}
+
+	com.Sub(zkp.Statement.Mul(zkp.Challenge))
```
yacovm, line 102: probably worth commenting what is subtracted here.

token/core/zkatdlog/crypto/o2omp/3omp.go
```diff
@@ -9,17 +9,30 @@ import (
...
-	"github.com/IBM/mathlib"
+	math "github.com/IBM/mathlib"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/crypto/common"
 	"github.com/pkg/errors"
 )
 
-// prover for the one out of many proofs
+// Prover produces a one-out-of-many proof
+// A one-out-of-many proof allows a prover who is given a list of commitments
+// (c_1, ..., c_n) shows that they know that there is i \in [N] and r_i \in Z_p
```
yacovm, line 19: 1. shows --> to show
2. did you mean i.e as in e.g ?

token/core/zkatdlog/crypto/common/schnorr.go
```diff
@@ -10,67 +10,116 @@ import (
 	"github.com/pkg/errors"
 )
 
-// Struct for Schnorr proofs
+// SchnorrProof carries a ZKP for statement (w_1, ..., w_n): Com = \prod_{i=1}^n P_i^w_i
```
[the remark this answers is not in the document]
adecaro: do you mean to include a Makefile entry to pass it over the code?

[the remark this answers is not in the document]
yacovm: I wouldn't go that far, I think we can just take a comment snippet, apply mathfmt on it, and copy the text...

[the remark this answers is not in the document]
yacovm: Also, it's not perfect. It has things it cannot handle, or at least back in the day it couldn't...  maybe they improved it since. 

https://github.com/SmartBFT-Go/randomcommittees/blob/master/internal/pvss.go#L311
https://github.com/SmartBFT-Go/randomcommittees/blob/master/internal/pvss.go#L43

token/core/zkatdlog/crypto/audit/auditor.go
```diff
@@ -262,6 +317,9 @@ func getAuditInfoForTransfers(des Deserializer, transfers [][]byte, metadata []d
 		for i := 0; i < len(tr.SenderAuditInfos); i++ {
 			var matcher driver.Matcher
 			var err error
+			if inputs[k][i] == nil {
+				return nil, nil, errors.Errorf("input[%d][%d] is nil", k, i)
+			}
 			if !inputs[k][i].IsRedeem() {
 				matcher, err = des.GetOwnerMatcher(tr.SenderAuditInfos[i])
 				if err != nil {
```
adecaro, line 325: shall we open a github issue to return always a message with the error?

token/core/zkatdlog/crypto/common/nym.go
adecaro: makes sense indeed. What about adding a constructor to NYMSigner and make it a private struct?

adecaro (approved):
Please, just address Yacov changes. The rest is good to me.

token/core/zkatdlog/crypto/audit/auditor.go
KElkhiyaoui: I will address it in this PR.

token/core/zkatdlog/crypto/audit/auditor.go
```diff
@@ -24,13 +24,20 @@ import (
 
 //go:generate counterfeiter -o mock/signing_identity.go -fake-name SigningIdentity . SigningIdentity
 
+var logger = flogging.MustGetLogger("token-sdk.zkatdlog.audit")
```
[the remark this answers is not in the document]
KElkhiyaoui: I will remove this logger as the crypto library does not use any loggers except here. We could have another PR to add/enhance logging.

token/core/zkatdlog/crypto/common/schnorr.go
KElkhiyaoui: I believe it is the case for the available implementations.

token/core/zkatdlog/crypto/common/schnorr.go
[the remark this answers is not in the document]
KElkhiyaoui: I just tried it, it does not work properly for power or indices.

Issue #255, opened 2022-06-14
Introduce HTLC support


adecaro:
Current token drivers do not support any form of scripting. The drivers will always interpreter the token owner field as containing the public key of the token owner.

This github issue proposes to augment the token owner field to support also HTLC. Namely, the token owner field can encode the relevant information for an HTLC, instead of just a public key, and the validator will run appropriate code to validate such ownership.

More details to come.

PR #290, opened 2022-07-10
htlc #255

HagarMeir:


integration/token/interop/views/exchange/exchange.go
```diff
@@ -0,0 +1,291 @@
...
+	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric"
+	view3 "github.com/hyperledger-labs/fabric-smart-client/platform/view"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/assert"
+	view2 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/view"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
+	"github.com/hyperledger-labs/fabric-token-sdk/token"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/interop/exchange"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/ttx"
+)
+
```
KElkhiyaoui, line 22: could you explain the difference between a fast exchange and a normal exchange.

integration/token/interop/views/exchange/exchange.go
```diff
@@ -0,0 +1,291 @@
...
+func (v *FastExchangeInitiatorView) Call(context view.Context) (interface{}, error) {
+	// Preliminary:
+	// 1. Exchange recipient identities
+	// 2. Agree on the terms
+	// 3. Lock assets
+	// 4. Initiator claims
+	// 5. Responder retrieves pre-images and claims
+
+	// We assume that the initiator and the responder are on the both networks
```
KElkhiyaoui, line 54: are on both networks

integration/token/interop/views/exchange/exchange.go
```diff
@@ -0,0 +1,291 @@
...
+	// Initiator's Leg (HTLC)
+	var preImage []byte
+	_, err = view2.RunCall(context, func(context view.Context) (interface{}, error) {
+		tx, err := exchange.NewTransaction(
+			context,
+			fabric.GetIdentityProvider(context, v.TMSID1.Network).DefaultIdentity(),
+			ttx.WithAuditor(view3.GetIdentityProvider(context).Identity("auditor")),
+			ttx.WithTMSID(v.TMSID1),
+		)
+		assert.NoError(err, "failed created an exchange transaction")
```
KElkhiyaoui, line 80: failed to create an exchange transaction

integration/token/interop/views/exchange/exchange.go
```diff
@@ -0,0 +1,291 @@
...
+	// Preliminary:
+	// 1. Exchange recipient identities
+	// 2. Agree on the terms
+	// 3. Lock assets
+	// 4. Initiator claims
+	// 5. Responder retrieves pre-images and claims
+
+	// We assume that the initiator and the responder are on the both networks
+	me1, recipient, err := exchange.ExchangeRecipientIdentities(context, "", v.Recipient, token.WithTMSID(v.TMSID1))
+	assert.NoError(err, "failed getting recipient identity")
```
KElkhiyaoui, line 56: of initiator?

KElkhiyaoui, line 56: of participants in the first network

integration/token/interop/views/exchange/exchange.go
```diff
@@ -0,0 +1,291 @@
...
+	me2, sender, err := exchange.ExchangeRecipientIdentities(context, "", v.Recipient, token.WithTMSID(v.TMSID2))
+	assert.NoError(err, "failed getting recipient identity 2")
```
KElkhiyaoui, line 58: of participants in the second network

integration/token/interop/views/exchange/exchange.go
```diff
@@ -0,0 +1,291 @@
...
+}
+
+type FastExchangeResponderView struct{}
+
+func (v *FastExchangeResponderView) Call(context view.Context) (interface{}, error) {
+	// Preliminary:
+	// 1. Exchange recipient identities
+	// 2. Agree on the terms
+	me1, sender, err := exchange.RespondExchangeRecipientIdentities(context)
+	assert.NoError(err, "failed to respond to identity request")
```
KElkhiyaoui, line 183: in the first network

integration/token/interop/views/exchange/exchange.go
```diff
@@ -0,0 +1,291 @@
...
+
+	me2, recipient, err := exchange.RespondExchangeRecipientIdentities(context)
+	assert.NoError(err, "failed to respond to identity request 2")
```
KElkhiyaoui, line 186: in the second network

integration/token/interop/views/exchange/lock.go
```diff
@@ -0,0 +1,163 @@
...
+	Amount uint64
+	// Recipient is the identity of the recipient's FSC node
+	Recipient view.Identity
+	// Hash is the hash to use in the script, if nil, a fresh one is generated
+	Hash []byte
+	// HashFunc is the hash function to use in the script
+	HashFunc crypto.Hash
+}
+
+type LockResult struct {
```
KElkhiyaoui, line 44: LockInfo?

integration/token/interop/views/exchange/lock.go
```diff
@@ -0,0 +1,163 @@
...
+	PreImage []byte
+	Hash     []byte
+}
+
+type LockView struct {
+	*Lock
+}
+
+func (hv *LockView) Call(context view.Context) (interface{}, error) {
+	// As a first step operation, the sender contacts the recipient's FSC node
```
KElkhiyaoui, line 55: As a first step, the sender...

integration/token/interop/views/exchange/lock.go
```diff
@@ -0,0 +1,163 @@
...
+	// to ask for the identity to use to assign ownership of the freshly created token.
+	// Notice that, this step would not be required if the sender knew already which
```
KElkhiyaoui, line 57: if the sender knows already which...

integration/token/interop/views/exchange/lock.go
```diff
@@ -0,0 +1,163 @@
...
+
+	// The sender is ready to collect all the required signatures.
+	// In this case, the sender's and the auditor's signatures.
+	// Invoke the Token Chaincode to collect endorsements on the Token Request and prepare the relative transaction.
+	// This is all done in one shot running the following view.
+	// Before completing, all recipients receive the approved transaction.
+	// Depending on the token driver implementation, the recipient's signature might or might not be needed to make
+	// the exchange transaction valid.
+	_, err = context.RunView(exchange.NewCollectEndorsementsView(tx))
+	assert.NoError(err, "failed to collect endorsements on exchange transaction")
```
KElkhiyaoui, line 98: for exchange transaction

integration/token/interop/views/exchange/lock.go
```diff
@@ -0,0 +1,163 @@
...
+
+	// Last but not least, the locker sends the transaction for ordering and waits for transaction finality.
+	_, err = context.RunView(exchange.NewOrderingAndFinalityView(tx))
+	assert.NoError(err, "failed to commit issue transaction")
```
KElkhiyaoui, line 102: exchange transaction

integration/token/interop/views/exchange/reclaim.go
```diff
@@ -0,0 +1,76 @@
...
+
+// ReclaimAll contains the input information to reclaim tokens
+type ReclaimAll struct {
+	// TMSID identifies the TMS to use to perform the token operation
+	TMSID token.TMSID
+	// Wallet is the identifier of the wallet that owns the tokens to reclaim
+	Wallet string
+}
+
+// ReclaimAllView is the initiation view of a party who wants to reclaim all reclaimable exchange
```
KElkhiyaoui, line 29: can this  be rephrased?

integration/token/interop/views/exchange/reclaim.go
```diff
@@ -0,0 +1,76 @@
...
+	*ReclaimAll
+}
+
+func (r *ReclaimAllView) Call(context view.Context) (interface{}, error) {
+	// The sender will select tokens owned by this wallet
+	senderWallet := exchange.GetWallet(context, r.Wallet, token.WithTMSID(r.TMSID))
+	assert.NotNil(senderWallet, "sender wallet [%s] not found", r.Wallet)
+
+	expired, err := exchange.Wallet(context, senderWallet, token.WithTMSID(r.TMSID)).ListExpired()
+	assert.NoError(err, "cannot retrieve list of expired exchange")
```
KElkhiyaoui, line 40: list of expired exchanges.

integration/token/interop/views/exchange/reclaim.go
```diff
@@ -0,0 +1,76 @@
...
+	assert.True(len(expired) > 0, "no exchange script has expired")
+
+	tx, err := exchange.NewTransaction(
+		context,
+		fabric.GetDefaultIdentityProvider(context).DefaultIdentity(),
+		ttx.WithAuditor(view2.GetIdentityProvider(context).Identity("auditor")),
+		ttx.WithTMSID(r.TMSID),
+	)
+	assert.NoError(err, "failed created an exchange transaction")
```
KElkhiyaoui, line 49: failed to create an exchange transaction

integration/token/interop/views/exchange/reclaim.go
```diff
@@ -0,0 +1,76 @@
...
+	// The sender is ready to collect all the required signatures.
+	// In this case, the sender's and the auditor's signatures.
+	// Invoke the Token Chaincode to collect endorsements on the Token Request and prepare the relative transaction.
+	// This is all done in one shot running the following view.
+	_, err = context.RunView(exchange.NewCollectEndorsementsView(tx))
+	assert.NoError(err, "failed to collect endorsements on exchange transaction")
+
+	// Last but not least, the transaction is sent for ordering, and we wait for transaction finality.
+	_, err = context.RunView(exchange.NewOrderingAndFinalityView(tx))
+	assert.NoError(err, "failed to commit issue transaction")
```
KElkhiyaoui, line 63: to commit exchange transaction

integration/token/interop/views/exchange/scan.go
```diff
@@ -0,0 +1,54 @@
...
+	"time"
+
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/assert"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
+	"github.com/hyperledger-labs/fabric-token-sdk/token"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/interop/encoding"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/interop/exchange"
+)
+
+// Scan contains the input information for a scan
```
KElkhiyaoui, line 21: what do we scan exactly?

integration/token/interop/views/issue.go
```diff
@@ -0,0 +1,96 @@
...
+
+// IssueCash contains the input information to issue a token
+type IssueCash struct {
+	// TMSID identifies the TMS to use to perform the token operation
+	TMSID token.TMSID
+	// IssuerWallet is the issuer's wallet to use
+	IssuerWallet string
+	// TokenType is the type of token to issue
+	TokenType string
+	// Quantity represent the number of units of a certain token type to issue
```
KElkhiyaoui, line 28: represents

integration/token/interop/views/issue.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+type IssueCashView struct {
+	*IssueCash
+}
+
+func (p *IssueCashView) Call(context view.Context) (interface{}, error) {
+	// As a first step operation, the issuer contacts the recipient's FSC node
+	// to ask for the identity to use to assign ownership of the freshly created token.
+	// Notice that, this step would not be required if the issuer knew already which
```
KElkhiyaoui, line 41: knows already ...

token/core/fabtoken/actions.go
```diff
@@ -8,13 +8,19 @@ package fabtoken
 import (
 	"encoding/json"
 
-	token2 "github.com/hyperledger-labs/fabric-token-sdk/token/token"
-
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
-
 	"github.com/hyperledger-labs/fabric-token-sdk/token/driver"
+	token2 "github.com/hyperledger-labs/fabric-token-sdk/token/token"
```
KElkhiyaoui, line 13: fix imports

token/core/fabtoken/validator.go
```diff
@@ -9,24 +9,95 @@ package fabtoken
...
+			}
+
+			// check that owner field in output is correct
+			if err := interop.VerifyTransferFromExchangeScript(inputTokens[0].Owner.Raw, out.Owner.Raw); err != nil {
+				return errors.Wrap(err, "failed to verify transfer from exchange script")
+			}
+		}
+	}
+
+	for _, o := range tr.GetOutputs() {
```
KElkhiyaoui, line 60: it is not clear to me why we check here for redeem ? this check should also be peformed when there is not exchange script validator.

token/core/fabtoken/validator.go
```diff
@@ -54,7 +125,20 @@ func (v *Validator) VerifyTokenRequest(ledger driver.Ledger, signatureProvider d
 	for _, action := range ta {
 		actions = append(actions, action)
 	}
-
+	for _, sig := range signatureProvider.Signatures() {
```
KElkhiyaoui, line 128: why is this not performed by the extravalidator?

token/core/interop/deserializer.go
```diff
@@ -0,0 +1,65 @@
...
+		return d.getExchangeVerifier(si.Identity)
+	}
+	return nil, errors.Errorf("failed to deserialize RawOwner: Unknown owner type %s", si.Type)
+}
+
+func (d *Deserializer) getExchangeVerifier(raw []byte) (driver.Verifier, error) {
+	script := &exchange.Script{}
+	err := json.Unmarshal(raw, script)
+	if err != nil {
+		return nil, errors.Errorf("failed to unmarshal RawOwner as a exchange script")
```
KElkhiyaoui, line 49: as an exchange script

token/core/interop/info.go
```diff
@@ -0,0 +1,77 @@
...
+		auditInfo, err := view2.GetSigService(s).GetAuditInfo(raw)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed getting audit info for recipient identity [%s]", view.Identity(raw).String())
+		}
+		return auditInfo, nil
+	}
+
+	sender, recipient, err := GetScriptSenderAndRecipient(owner)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed getting script sender and recipient")
```
KElkhiyaoui, line 46: exchange script

token/core/zkatdlog/crypto/validator/validator.go
```diff
@@ -8,31 +8,99 @@ package validator
...
+			}
+
+			// check that owner field in output is correct
+			if err := interop.VerifyTransferFromExchangeScript(tokens[0].Owner, out.Owner); err != nil {
+				return errors.Wrap(err, "failed to verify transfer from exchange script")
+			}
+		}
+	}
+
+	for _, o := range tr.GetOutputs() {
```
KElkhiyaoui, line 63: why is redeem checked here?

token/core/zkatdlog/crypto/validator/validator.go
```diff
@@ -103,6 +179,21 @@ func (v *Validator) VerifyTokenRequest(ledger driver.Ledger, signatureProvider d
 		actions = append(actions, action)
 	}
 
+	for _, sig := range signatureProvider.Signatures() {
```
KElkhiyaoui, line 182: why is this not performed by the extravalidator?

token/services/interop/exchange/script.go
```diff
@@ -0,0 +1,81 @@
...
+
+// ScriptOwnership implements the Ownership interface for scripts
+type ScriptOwnership struct{}
+
+// AmIAnAuditor returns false for script ownership
+func (s *ScriptOwnership) AmIAnAuditor(tms *token.ManagementService) bool {
+	return false
+}
+
+// IsMine returns true it there exists an owner wallet for the token's owner
```
KElkhiyaoui, line 44: This is not very clear. According to the code, IsMine checks if one is either a sender or recipient in exchange script.

token/services/interop/exchange/signer.go
```diff
@@ -0,0 +1,111 @@
...
+			Hash:         v.HashInfo.Hash,
+			HashFunc:     v.HashInfo.HashFunc,
+			HashEncoding: v.HashInfo.HashEncoding,
+		}}
+		if err := cv.Verify(msg, sigma); err != nil {
+			return errors.WithMessagef(err, "failed verifying exchange claim signature")
+		}
+		return nil
+	}
+	if err := v.Sender.Verify(msg, sigma); err != nil {
```
KElkhiyaoui, line 107: Add comment to the fact that if timeout has elapsed, only a reclaim is possible.

token/services/interop/exchange/signer.go
```diff
@@ -0,0 +1,111 @@
...
+type ExchangeVerifier struct {
+	Recipient driver.Verifier
+	Sender    driver.Verifier
+	Deadline  time.Time
+	HashInfo  HashInfo
+}
+
+// Verify verifies the claim signature, if the deadline hasn't passed, or the reclaim signature
+func (v *ExchangeVerifier) Verify(msg []byte, sigma []byte) error {
+	if time.Now().Before(v.Deadline) {
```
KElkhiyaoui, line 96: Add comment that if timeout has not elapsed, only claim is allowed.

token/services/interop/exchange/transaction.go
```diff
@@ -0,0 +1,342 @@
...
+		deadline = defaultDeadlineOffset
+	}
+	if recipient.IsNone() {
+		return nil, errors.Errorf("must specify a recipient")
+	}
+
+	if sender == nil {
+		sender, err = wallet.GetRecipientIdentity()
+		if err != nil {
+			return nil, errors.WithMessagef(err, "failed getting owner identity")
```
KElkhiyaoui, line 126: sender identity

integration/token/interop/views/auditor.go
```diff
@@ -0,0 +1,93 @@
...
+	eIDs := inputs.EnrollmentIDs()
+	tokenTypes := inputs.TokenTypes()
+	fmt.Printf("Limits on inputs [%v][%v]\n", eIDs, tokenTypes)
+	for _, eID := range eIDs {
+		assert.NotEmpty(eID, "enrollment id should not be empty")
+		for _, tokenType := range tokenTypes {
+			// compute the payment done in the transaction
+			sent := inputs.ByEnrollmentID(eID).ByType(tokenType).Sum().ToBigInt()
+			received := outputs.ByEnrollmentID(eID).ByType(tokenType).Sum().ToBigInt()
+			fmt.Printf("Payment Limit: [%s] Sent [%d], Received [%d], type [%s]\n", eID, sent.Int64(), received.Int64(), tokenType)
```
mbrandenburger, line 55: We should use a logger here. I guess this comment applies to all files.

token/services/interop/exchange/transaction.go
```diff
@@ -0,0 +1,342 @@
...
+	}
+	script := &Script{}
+	switch owner.Type {
+	case ScriptTypeExchange:
+		err := json.Unmarshal(owner.Identity, script)
+		if err != nil {
+			return errors.Errorf("failed to unmarshal RawOwner as an exchange script")
+		}
+
+		// TODO: does the pre-image match?
```
KElkhiyaoui, line 237: this should be handled by the exchange verifier. the pre-image is part of the signature that is checked against a verifier that encodes the hash.

integration/token/interop/views/auditor.go
```diff
@@ -0,0 +1,93 @@
...
+
+			diff := big.NewInt(0).Sub(sent, received)
+			if diff.Cmp(big.NewInt(0)) <= 0 {
+				continue
+			}
+			fmt.Printf("Payment Limit: [%s] Diff [%d], type [%s]\n", eID, diff.Int64(), tokenType)
+
+			assert.True(diff.Cmp(big.NewInt(int64(Limit))) <= 0, "payment limit reached [%s][%s][%s]", eID, tokenType, diff.Text(10))
```
mbrandenburger, line 63: what about returning an error instead? I personally think the use of `assert` that results in a panic and than let it caught by the FSC runtime is not nice.

mbrandenburger: anyway, using `assert` is consistent with most of our view implementations.

KElkhiyaoui (changes requested):
The PR looks good to me overall. There are changes that need to be made for approval.

token/core/fabtoken/actions.go
```diff
@@ -8,13 +8,19 @@ package fabtoken
...
 )
 
+type Signature struct {
+	metadata map[string][]byte // metadata may include for example the preimage of an exchange script
+}
+
+func (s *Signature) Metadata() map[string][]byte {
+	return s.metadata
+}
+
```
mbrandenburger, line 23: should this return a copy of the metadata?

token/core/fabtoken/deserializer.go
```diff
@@ -55,5 +58,18 @@ func NewEnrollmentIDDeserializer() *enrollmentService {
 }
 
 func (e *enrollmentService) GetEnrollmentID(auditInfo []byte) (string, error) {
+	if len(auditInfo) == 0 {
+		return "", nil
+	}
+
+	// Try to unmarshal it as ScriptInfo
+	si := &interop.ScriptInfo{}
+	err := json.Unmarshal(auditInfo, si)
```
mbrandenburger, line 67: what about using our json marshaller in `common`?

token/services/interop/encoding/logger.go
```diff
@@ -0,0 +1,11 @@
...
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package encoding
+
+import "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/flogging"
+
+var logger = flogging.MustGetLogger("token-sdk.encoding")
```
mbrandenburger, line 11: should this logger point to `token-sdk.interop.encoding`?

token/services/interop/exchange/distribute.go
```diff
@@ -0,0 +1,92 @@
...
+func (v *termsReceiverView) Call(context view.Context) (interface{}, error) {
+	_, payload, err := session.ReadFirstMessage(context)
+	if err != nil {
+		return nil, err
+	}
+	terms := &Terms{}
+	if err := terms.FromBytes(payload); err != nil {
+		return nil, errors.Wrapf(err, "failed unmarshalling terms")
+	}
+	// TODO review terms and accept
```
mbrandenburger, line 90: will this be part of a follow up PR?

token/services/interop/exchange/scanner.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
+	"github.com/hyperledger-labs/fabric-token-sdk/token"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/interop/encoding"
+	fabric2 "github.com/hyperledger-labs/fabric-token-sdk/token/services/network/fabric"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/vault/translator"
+	"go.uber.org/zap/zapcore"
+)
+
+// ScanForPreImage scans for a preimage of the passed image, taking into account the timeout
+func ScanForPreImage(ctx view.Context, image []byte, hashFunc crypto.Hash, hashEncoding encoding.Encoding, timeout time.Duration, opts ...token.ServiceOption) ([]byte, error) {
```
mbrandenburger, line 26: this function seems to be a beast :)

token/core/zkatdlog/crypto/validator/validator.go
HagarMeir: the extravalidator is there so it would be easier to introduce different types of scripts 
since this applies only to the exchange script no reason to have an extravalidator for it

token/core/zkatdlog/crypto/validator/validator.go
HagarMeir: Just to make sure we don't create an error when trying to unmarshall later

token/core/interop/info.go
HagarMeir: actually this should remain generic enough for any script with a sender and a recipient

token/core/fabtoken/validator.go
HagarMeir: the extravalidator is there so it would be easier to introduce different types of scripts
since this applies only to the exchange script no reason to have an extravalidator for it

token/core/fabtoken/validator.go
HagarMeir: We check here just to make sure we don't create an error when trying to unmarshall later
There was no check like this performed by the non-script case

integration/token/interop/views/auditor.go
HagarMeir: so maybe this should be a separate issue to clean up all of the code

token/core/fabtoken/actions.go
HagarMeir: this is only used in one place and so far there was no problem with using it this way without copying

token/core/fabtoken/deserializer.go
HagarMeir: ok
but why?

token/services/interop/exchange/distribute.go
HagarMeir: This could be as part of a new issue
Or maybe just put a comment instead stating that terms could be examined at this point

token/services/interop/exchange/scanner.go
HagarMeir: true but I rather leave it as is at this point

token/core/fabtoken/deserializer.go
mbrandenburger: ah! I had the `common` package in mind we have in FSC. However, this is part of the integration tests. Not applicable at this point.

KElkhiyaoui (commented):
Looks good to me. If all comments are addressed, I will approve.

integration/token/interop/tests.go
```diff
@@ -0,0 +1,167 @@
...
+	Expect(h.Sum(64).ToBigInt().Cmp(big.NewInt(120))).To(BeEquivalentTo(0))
+	Expect(h.ByType("USD").Count()).To(BeEquivalentTo(h.Count()))
+
+	h = listIssuerHistory(network, "", "EUR")
+	Expect(h.Count() > 0).To(BeTrue())
+	Expect(h.Sum(64).ToBigInt().Cmp(big.NewInt(30))).To(BeEquivalentTo(0))
+	Expect(h.ByType("EUR").Count()).To(BeEquivalentTo(h.Count()))
+
+	checkBalance(network, "alice", "", "USD", 120)
+	checkBalance(network, "bob", "", "EUR", 30)
```
adecaro, line 54: let's check that Bob's USD balance is 0

integration/token/interop/tests.go
```diff
@@ -0,0 +1,167 @@
...
+
+	// exchange (lock, reclaim)
+	exchangeLock(network, token.TMSID{}, "alice", "", "USD", 10, "bob", 10*time.Second, nil, crypto.SHA512)
+	time.Sleep(15 * time.Second)
+	checkBalance(network, "alice", "", "USD", 110)
```
adecaro, line 59: let's check that Bob's USD balance is still 0 here and after the reclaim

integration/token/interop/tests.go
```diff
@@ -0,0 +1,167 @@
...
+	exchangeReclaimAll(network, "alice", "")
+	checkBalance(network, "alice", "", "USD", 120)
+
+	// exchange (lock, claim)
+	defaultTMSID := token.TMSID{}
+	preImage, _ := exchangeLock(network, defaultTMSID, "alice", "", "USD", 20, "bob", 1*time.Hour, nil, crypto.SHA3_256)
```
adecaro, line 65: same here, check Bob's USD balance

adecaro (changes requested):

adecaro:
Just to keep track of next items to track:
- How are these new operations tracked in the transaction db `token/services/ttxdb`. Recall that the auditors and each participants in the assembling of the transaction store traces in the transaction db.
- Is the auditor able to audit the script? I don't see code in `integration/token/interop/views/auditor.go` for that. 
- How can the auditor track current holding when scripts are involved?
- Is the acceptor able to inspect the script? I don't see code in `integration/token/interop/views/accept.go`
- `ScanForPreImage` actually works only when the backed is Fabric. But it should be backed agnostic by using the Network API (token/services/network)

token/services/interop/exchange/scanner.go
```diff
@@ -0,0 +1,110 @@
...
+	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
+	"github.com/hyperledger-labs/fabric-token-sdk/token"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/interop/encoding"
+	fabric2 "github.com/hyperledger-labs/fabric-token-sdk/token/services/network/fabric"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/vault/translator"
+	"go.uber.org/zap/zapcore"
+)
+
+// ScanForPreImage scans for a preimage of the passed image, taking into account the timeout
```
adecaro, line 25: Please, clarify that this functions scans the ledger and not the vault. It is therefore a heavy function. Just to alert the developer.

token/core/fabtoken/validator.go
KElkhiyaoui: Claim is part of the exchange script and it seems it should be validated by the `exchangeScriptValidator'. I also see that we have action Signature to carry the Claim Preimage. Action is reserved for tms actions (issue, transfer and setup). Signature itself is not an action. I see that having it this way, allow for an easy lookup of preimages, but we could probably find another way, using `TransferAction.GetMetadata()` instead.

token/core/fabtoken/validator.go
```diff
@@ -82,6 +154,20 @@ func (v *Validator) VerifyTokenRequest(ledger driver.Ledger, signatureProvider d
 	for _, action := range ta {
 		actions = append(actions, action)
 	}
+	for _, sig := range signatureProvider.Signatures() {
+		claim := &exchange.ClaimSignature{}
+		if err = json.Unmarshal(sig, claim); err != nil {
+			continue
+		}
+		if len(claim.Preimage) == 0 || len(claim.RecipientSignature) == 0 {
```
KElkhiyaoui, line 162: Shouldn't this be in-acceptable. If the unmarshalling is successful, shouldn't `Pre-image` and  `RecipientSignature` not be empty.

KElkhiyaoui (changes requested):

mbrandenburger (approved):
Overall looks good to me. Can we check that this does not break any of our internal prototypes? Let's discuss offline.

KElkhiyaoui (approved):

token/core/fabtoken/validator.go
HagarMeir: Will be addressed in another PR

PR #297, opened 2022-07-15
`token\core` improvements

adecaro:
This PR brings the following improvements:
- Simplified wallet management `core\identity`
- extensible `zkatdlog\nogh` auditor implementation
- token request metadata: add support for extra signers, in case a driver needs it.
- add more logging to token processors
 
Plus:
- dvp test: added delay to wait for the network to be ready.
- left a TODO to be addressed in a future PR


token/core/identity/msp/wf.go
```diff
@@ -0,0 +1,136 @@
...
+	// IssuerMSPID is the default MSP ID for the issuer wallet
+	IssuerMSPID = "IssuerMSPID"
+	// AuditorMSPID is the default MSP ID for the auditor wallet
+	AuditorMSPID = "AuditorMSPID"
+	// CertifierMSPID is the default MSP ID for the certifier wallet
+	CertifierMSPID = "CertifierMSPID"
+)
+
+// RoleToMSPID maps the role to the MSP ID
+var RoleToMSPID = map[driver.IdentityRole]string{
```
yacovm, line 36: this is a global variable. How do you plan to support multi-tenancy?

token/services/ttx/endorse.go
```diff
@@ -745,6 +734,35 @@ func (s *endorseView) requestsToBeSigned() ([]*token.Transfer, error) {
...
+func (s *endorseView) receiveTransaction(context view.Context) error {
+	if logger.IsEnabledFor(zapcore.DebugLevel) {
+		logger.Debugf("Receive transaction with envelope...")
+	}
+	// TODO: this might also happen multiple times because of the pseudonym. Avoid this by identity resolution at the sender
+	tx, err := ReceiveTransaction(context)
+	if err != nil {
+		return errors.Wrapf(err, "failed receiving transaction")
+	}
+	agent := metrics.Get(context)
```
yacovm, line 755: probably should open an issue about making this not a singleton.

token/services/ttx/endorse.go
```diff
@@ -745,6 +734,35 @@ func (s *endorseView) requestsToBeSigned() ([]*token.Transfer, error) {
...
+	agent.EmitKey(0, "ttx", "received", "env", tx.ID())
+
+	// Check the envelope exists
+	if logger.IsEnabledFor(zapcore.DebugLevel) {
+		logger.Debugf("Processes Fabric Envelope with ID [%s]", tx.ID())
+	}
+	env := tx.Payload.Envelope
```
yacovm, line 762: what if `Payload` is nil?

token/services/ttx/endorse.go
```diff
@@ -745,6 +734,35 @@ func (s *endorseView) requestsToBeSigned() ([]*token.Transfer, error) {
...
+	if env == nil {
+		return errors.Errorf("expected fabric envelope")
+	}
+	s.tx.Payload.Envelope = tx.Payload.Envelope
```
yacovm, line 766: I assume since each view is initialized from a view factory, there is no issue in concurrency because this `s.tx.Payload.Envelope` assignment acts only on a single dedicated view instance correct?

integration/token/dvp/tests.go
```diff
@@ -21,6 +21,11 @@ import (
 )
 
 func TestAll(network *integration.Infrastructure) {
+	// Give some time for the network to start
```
yacovm, line 24: the NWO framework allows you to wait for a string to be printed in the output of the process when it starts. This implicitly makes the test "wait" for the process to start, etc.

token/services/ttx/endorse.go
adecaro: what do you mean here? agent is taken from the context.

token/services/ttx/endorse.go
adecaro: let's check

token/services/ttx/endorse.go
adecaro: right

integration/token/dvp/tests.go
adecaro: unfortunately, there is no way that I know that gives us the believe the the p2p has discovered all nodes.

token/core/identity/msp/wf.go
yacovm: But I thought you have a mapping from role to MSP ID? What if the MSP IDs are different across channels?

adecaro: but here is the point, these are not Fabric identities. Their management is different.

PR #298, opened 2022-07-18
Add recipient signature as acknowledgment 

sapthasurendran:


integration/token/fungible/topology.go
```diff
@@ -38,7 +38,7 @@ func Topology(backend string, tokenSDKDriver string, auditorAsIssuer bool) []api
 
 	// FSC
 	fscTopology := fsc.NewTopology()
-	//fscTopology.SetLogging("debug", "")
+	fscTopology.SetLogging("debug", "")
```
adecaro, line 41: please, remove the debug here for a faster CI execution

token/services/ttx/accept.go
```diff
@@ -7,6 +7,10 @@ SPDX-License-Identifier: Apache-2.0
 package ttx
 
 import (
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/hash"
+
```
adecaro, line 11: please, remove the newlines here and run goimports. Thanks :)

token/services/ttx/accept.go
```diff
@@ -51,18 +64,25 @@ func (s *acceptView) Call(context view.Context) (interface{}, error) {
...
+		return nil, errors.WithMessage(err, "failed to sign ack response")
+	}
 	// Ack for distribution
+	// Send the signature back
 	session := context.Session()
-	// Send the proposal response back
-	err = session.Send([]byte("ack"))
-	if err != nil {
-		return nil, err
+	logger.Debugf("ack response: [%s] from [%s]", hash.Hashable(sigma), view2.GetIdentityProvider(context).DefaultIdentity())
```
adecaro, line 81: can we check first if debug log is enabled using `if logger.IsEnabledFor(zapcore.DebugLevel) {` ?
This should improve performance when debug is disbaled

integration/token/fungible/views/signature.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+type SignatureView struct {
+	*Signature
+}
+
+func (t *SignatureView) Call(context view.Context) (interface{}, error) {
+
+	//generate compositekey
+	k := kvs.GetService(context)
```
adecaro, line 32: shall we add a function to the ttx package that checks this? I fear that some refactoring can break this and it becomes hard to track

integration/token/fungible/tests.go
```diff
@@ -716,6 +719,51 @@ func TransferCash(network *integration.Infrastructure, id string, wallet string,
...
+	} else {
+		Expect(err).To(HaveOccurred())
+		for _, msg := range errorMsgs {
+			Expect(err.Error()).To(ContainSubstring(msg))
+		}
+		time.Sleep(5 * time.Second)
+	}
+	CheckRecipientSignature(network, id, common.JSONUnmarshalString(txid), receiver)
+}
+func CheckRecipientSignature(network *integration.Infrastructure, id string, txnId string, receiver string, errorMsgs ...string) {
```
adecaro, line 747: shall we add this check directly to TransferCash instead of introducing TransferCashAndCheckRecepientSignature?

adecaro (changes requested):

Issue #299, opened 2022-07-18
Tokengen expects a full msp directory, but should also work without signing key of auditor and issuer


arner:
Tokengen takes msp directories as input for the auditor and issuer.

It only needs some specific files to operate, but it's not clear which files these are, and tokengen could be more tolerant of files missing that are not needed. Especially the private key would not be available in a production scenario.

Example:

```
issuerwalletuser
├── cacerts
│   └── cert.pem
├── keystore
│   └── priv_sk
└── signcerts
    └── cert.pem
```
```
tokengen gen dlog --base 100 --exponent 2 --issuers /input/issuerwalletuser --idemix /input/idemix/ --auditors /input/auditorwalletuser --output /artifacts

Error: failed to generate public parameters: failed to get issuer identity [/input/issuerwalletuser]: failed to create x509 provider for [/input/issuerwalletuser]: KeyMaterial not found in SigningIdentityInfo
```

PR #300, opened 2022-07-19
Tokengen should also work without signing key of auditor and issuer #299

adecaro:


token/core/cmd/pp/common/common.go
```diff
@@ -68,3 +87,54 @@ func SetupIssuersAndAuditors(pp PP, Auditors, Issuers []string) error {
...
+
+	return bytes, nil
+}
+
+func GetPemMaterialFromDir(dir string) ([][]byte, error) {
+	_, err := os.Stat(dir)
+	if os.IsNotExist(err) {
+		return nil, err
+	}
+	content := make([][]byte, 0)
```
mbrandenburger, line 119:
```suggestion
	var content [][]byte
```

token/core/cmd/pp/common/common.go
```diff
@@ -68,3 +87,54 @@ func SetupIssuersAndAuditors(pp PP, Auditors, Issuers []string) error {
...
+func ReadPemFile(file string) ([]byte, error) {
+	bytes, err := ReadFile(file)
+	if err != nil {
+		return nil, errors.Wrapf(err, "reading from file %s failed", file)
+	}
+
+	b, _ := pem.Decode(bytes)
+	if b == nil { // TODO: also check that the type is what we expect (cert vs key..)
+		return nil, errors.Errorf("no pem content for file %s", file)
+	}
```
mbrandenburger, line 109: Here you just want to check if `bytes` contains at least a valid pem decoded block, right? What is multiple blocks are available?

token/core/cmd/pp/common/common.go
```diff
@@ -37,13 +46,23 @@ func GetMSPIdentity(entry string, mspID string) (view.Identity, error) {
...
 		}
 	}
-	provider, err := x509.NewProvider(entries[0], mspID, nil)
+
+	// read certificate from entries[0]/signcerts
+	signcertDir := filepath.Join(entries[0], signcerts)
+	content, err := GetPemMaterialFromDir(signcertDir)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to load certificates from %s", signcertDir)
+	}
```
mbrandenburger, line 55: redundant error check?

token/core/cmd/pp/common/common.go
```diff
@@ -7,14 +7,23 @@ SPDX-License-Identifier: Apache-2.0
...
 import (
+	"encoding/pem"
+	"io/ioutil"
+	"os"
+	"path/filepath"
 	"strings"
 
-	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/msp/x509"
+	msp3 "github.com/hyperledger/fabric/msp"
+
```
mbrandenburger, line 17: empty line

cmd/tokengen/main_test.go
```diff
@@ -13,6 +13,12 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/hyperledger-labs/fabric-token-sdk/token/core/identity/msp"
+
+	"github.com/hyperledger-labs/fabric-token-sdk/token/core/cmd/pp/common"
+
+	"github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/crypto"
+
```
mbrandenburger, line 21: empty lines

mbrandenburger (changes requested):

token/core/cmd/pp/common/common.go
adecaro: the folder might contain no files

mbrandenburger: What I am saying is that we check the error twice! in line 52 and 58.

token/core/cmd/pp/common/common.go
```diff
@@ -68,3 +86,60 @@ func SetupIssuersAndAuditors(pp PP, Auditors, Issuers []string) error {
...
+func ReadFile(file string) ([]byte, error) {
+	fileCont, err := ioutil.ReadFile(file)
+	if err != nil {
+		return nil, errors.Wrapf(err, "could not read file %s", file)
+	}
+
+	return fileCont, nil
+}
+
+func ReadPemFile(file string) ([]byte, error) {
```
mbrandenburger, line 99: Let's give the function some documentation and even change the name. `ReadPemFile` is a very generic name, however, the current impl returns errors if the content in the pem file it not a cert, or there are more than one entry.

Maybe call it `ReadCertFromPemFile(file string)` explaining that it returns the first item from a pem file if exists. If needed we can later add `ReadCertsFromPemFile(...)` to parse the entire file.

token/core/cmd/pp/common/common.go
```diff
@@ -68,3 +86,60 @@ func SetupIssuersAndAuditors(pp PP, Auditors, Issuers []string) error {
...
+		return nil, errors.Wrapf(err, "reading from file %s failed", file)
+	}
+
+	b, rest := pem.Decode(bytes)
+	if b == nil {
+		return nil, errors.Errorf("no pem content for file %s", file)
+	}
+	if len(rest) != 0 {
+		return nil, errors.Errorf("extra content after pem file %s", file)
+	}
```
mbrandenburger, line 111: We could just ignore the rest. If there are more certs in the file, we don't care, do we?

token/core/cmd/pp/common/common.go
```diff
@@ -68,3 +86,60 @@ func SetupIssuersAndAuditors(pp PP, Auditors, Issuers []string) error {
...
+	if b.Type != "CERTIFICATE" {
+		return nil, errors.Errorf("pem file %s is not a certificate", file)
+	}
+
+	return bytes, nil
```
mbrandenburger, line 116: should we just return b?

token/core/cmd/pp/common/common.go
adecaro: my bad, sorry

token/core/cmd/pp/common/common.go
adecaro: I would say, let's constraint to a single certificate in the file

token/core/cmd/pp/common/common.go
adecaro: if we force only one cert, we can return directly the bytes

token/core/cmd/pp/common/common.go
adecaro: okay

mbrandenburger (approved):
LGTM

PR #304, opened 2022-07-21
Optimize Idemix pre-provision cache

yacovm:
This commit:

- Replaces time.After() with time.NewTimer()
- Only provisions the background workers if needed
- Provisions runtime.NumCPU() background workers
  to reduce waiting time in case of high bursts


mbrandenburger (approved):
LGTM

PR #305, opened 2022-07-21
Expose idemix identity cache size in config

yacovm:

PR #306, opened 2022-07-21
use time.NewTimer instead of time.After

yacovm:
Using time.After enqueues an entry in Go's runtime scheduler and doesn't garbage collect it even if the function the timer was fired from terminates.
As a result, there can be a temporary high memory pressure and it degrades performance.

Replaced all places where time.After was used except in places where we use it for timing periodical processes.

### DIFF
--- a/token/core/identity/msp/idemix/cache.go
+++ b/token/core/identity/msp/idemix/cache.go
@@ -102,13 +102,16 @@ func NewWalletIdentityCache(backed WalletIdentityCacheBackendFunc, size int) *Wa
 }
 
 func (c *WalletIdentityCache) Identity() (view.Identity, error) {
+	timeout := time.NewTimer(c.timeout)
+	defer timeout.Stop()
+
 	select {
 	case entry := <-c.ch:
 		if logger.IsEnabledFor(zapcore.DebugLevel) {
 			logger.Debugf("fetch identity from producer channel done [%s][%d]", entry)
 		}
 		return entry, nil
-	case <-time.After(c.timeout):
+	case <-timeout.C:
 		if logger.IsEnabledFor(zapcore.DebugLevel) {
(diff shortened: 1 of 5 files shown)


mbrandenburger (approved):
LGTM! Would be nice to see this enhancement on the FSC code as well. :)

yacovm:
> LGTM! Would be nice to see this enhancement on the FSC code as well. :)

https://github.com/hyperledger-labs/fabric-smart-client/pull/320

adecaro (changes requested):
there are still two places where time.After is used:
- token/core/identity/msp/idemix/cache.go
- token/services/certifier/interactive/client.go

yacovm:
> there are still two places where time.After is used:
> 
>     * token/core/identity/msp/idemix/cache.go
> 
>     * token/services/certifier/interactive/client.go

Correct:

In [token/services/certifier/interactive/client.go](https://github.com/hyperledger-labs/fabric-token-sdk/blob/main/token/services/certifier/interactive/client.go#L146) the time.After is used to make the loop iterate. It means that when the `i`'th time.After fires, all previous `time.After` have been garbage collected, and that there can only be a single `time.After` at a time. This is when can should use time.After without any implications. 

In token/core/identity/msp/idemix/cache.go I didn't change [here](https://github.com/hyperledger-labs/fabric-token-sdk/pull/306/commits/c427c96a6ff7a9e0fd76ebcc9041ced84c288060#diff-43f058af561c7b85aa8df37f53a3e7d09f3049fad33d3df64283d9dc50c13062R54), because [this](https://github.com/hyperledger-labs/fabric-token-sdk/pull/305/commits/639669affc01f2730507fffde873efa1b8f79760) PR already changes this and I don't want a merge conflict.

Issue #309, opened 2022-07-26
Data Race on public params 


ltxyz:
When we run tests in parallel with a same wallet, we sometime get a data race: here is an example

``` 
WARNING: DATA RACE
Write at 0x00c01c5cac78 by goroutine 1128:
  github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/crypto/ppm.(*PublicParamsManager).PublicParams()
      /go/src/github.com/banquedefrance/dl3s/backend/dl3s/vendor/github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/crypto/ppm/ppm.go:67 +0x9c
  github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/crypto/ppm.(*PublicParamsManager).PublicParameters()
      /go/src/github.com/banquedefrance/dl3s/backend/dl3s/vendor/github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/crypto/ppm/ppm.go:40 +0x2e
  github.com/hyperledger-labs/fabric-token-sdk/token.(*PublicParametersManager).Precision()
      /go/src/github.com/banquedefrance/dl3s/backend/dl3s/vendor/github.com/hyperledger-labs/fabric-token-sdk/token/publicparams.go:26 +0x42
  github.com/hyperledger-labs/fabric-token-sdk/token/services/query.(*BalanceView).Call()
      /go/src/github.com/banquedefrance/dl3s/backend/dl3s/vendor/github.com/hyperledger-labs/fabric-token-sdk/token/services/quer
...

Previous read at 0x00c01c5cac78 by goroutine 699:
  github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/crypto/ppm.(*PublicParamsManager).PublicParams()
      /go/src/github.com/banquedefrance/dl3s/backend/dl3s/vendor/github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/crypto/ppm/ppm.go:62 +0x32
  github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/crypto/ppm.(*PublicParamsManager).PublicParameters()
      /go/src/github.com/banquedefrance/dl3s/backend/dl3s/vendor/github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/crypto/ppm/ppm.go:40 +0x2e
[...]

PR #310, opened 2022-07-26
fix data race in ppm #309

Grimnir777:
Signed-off-by: Thibaud Germain <thibaud.germain1@ibm.com>

### DIFF
--- a/token/core/zkatdlog/crypto/ppm/ppm.go
+++ b/token/core/zkatdlog/crypto/ppm/ppm.go
@@ -14,7 +14,6 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/flogging"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/crypto"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/driver"
-	"github.com/pkg/errors"
 )
 
 var logger = flogging.MustGetLogger("token-sdk.zkatdlog")
@@ -67,7 +66,7 @@ func (v *PublicParamsManager) ForceFetch() error {
 }
 
 func (v *PublicParamsManager) PublicParams() *crypto.PublicParams {
-	v.mutex.Lock()
-	defer v.mutex.Unlock()
+	v.mutex.RLock()
+	defer v.mutex.RUnlock()
 	return v.pp
 }
--- a/token/core/zkatdlog/crypto/ppm/ppm.go
+++ b/token/core/zkatdlog/crypto/ppm/ppm.go
@@ -7,6 +7,10 @@ SPDX-License-Identifier: Apache-2.0
 package ppm
 
 import (
+	"sync"
+
+	"github.com/pkg/errors"
+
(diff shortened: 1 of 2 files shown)


token/core/zkatdlog/crypto/ppm/ppm.go
```diff
@@ -59,6 +65,9 @@ func (v *PublicParamsManager) ForceFetch() error {
 }
 
 func (v *PublicParamsManager) PublicParams() *crypto.PublicParams {
+	v.mutex.Lock()
```
adecaro, line 68: What about using first a read lock to check if v.pp is not nil and then a full lock in the case the public params must be fetched?

adecaro (changes requested):

adecaro:
@Grimnir777 , thanks for submitting this PR. 
Please, could you open a github issue that describes the problem?
Also, does this apply also to the fabtoken driver? Shall we apply the same patch there as well?

Grimnir777:
Well we didn't see any issues in the fabtoken driver but for sure I can apply the changes also there

token/core/zkatdlog/crypto/ppm/ppm.go
Grimnir777: Is this working ? I'm not really familiar with the RWmutex 

```
	v.mutex.RLock()
	defer v.mutex.RUnlock()
	if v.pp == nil {
		v.mutex.Lock()
		defer v.mutex.Unlock()
		if v.publicParamsLoader == nil {
			panic("public parameters loaded not set")
		}
		var err error
		v.pp, err = v.publicParamsLoader.Load()
		if err != nil {
			panic(err)
		}
	}
```

adecaro: What about this?

```go	
        v.mutex.RLock()
	if v.pp != nil {
		v.mutex.RUnlock()
		return v.pp
	}
	v.mutex.RUnlock()

	v.mutex.Lock()
	defer v.mutex.Unlock()
	if v.pp == nil {
		if v.publicParamsLoader == nil {
			panic("public parameters loaded not set")
		}
		var err error
		v.pp, err = v.publicParamsLoader.Load()
		if err != nil {
			panic(err)
		}
	}
	return v.pp
```

adecaro:
Yes, please, can you apply the change also to fabtoken? Thanks :)

adecaro:
Hi @Grimnir777, I think we need a simpler locking strategy now.
A PR merged yesterday simplified the public parameter manager.
Can you rebase, please?

Grimnir777:
Hello @adecaro I did the update from the rebase

adecaro:
Hi @Grimnir777 , thanks for updating the PR.
One of checks fails:

```
The following files contain goimports errors
token/core/fabtoken/ppm.go
make: *** [checks.mk:25: goimports] Error 1
The goimports command 'goimports -l -w' must be run for these files```
```

If you want to test locally on your machine, you can run `make checks`. 
Thanks :)

adecaro:
@Grimnir777 , thanks a lot. I'll merge as soon as the CI passes :)

adecaro (approved):

PR #330, opened 2022-08-03
check that the tokens accepted appear in the vault

adecaro:
- token.Quantity must contain a number in hex format.

### DIFF
--- a/integration/token/fungible/views/accept.go
+++ b/integration/token/fungible/views/accept.go
@@ -87,7 +87,7 @@ func (a *AcceptCashView) Call(context view.Context) (interface{}, error) {
 		} else {
 			// check it does not exist
 			_, _, err := qe.GetTokens(tokenID)
-			assert.Equal(err, "token [%s] should not exist", tokenID)
+			assert.Error(err, "token [%s] should not exist", tokenID)
 			assert.True(strings.Contains(err.Error(), "token not found"))
 		}
 	}
--- a/integration/token/fungible/views/accept.go
+++ b/integration/token/fungible/views/accept.go
@@ -7,8 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 package views
 
 import (
-	"strings"
-
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/assert"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/network"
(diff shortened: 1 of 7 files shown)


integration/token/fungible/views/accept.go
```diff
@@ -72,5 +73,24 @@ func (a *AcceptCashView) Call(context view.Context) (interface{}, error) {
...
+			assert.Equal(output.Quantity.Hex(), toks[0].Quantity, "token quantity mismatch")
+			assert.Equal(output.Type, toks[0].Type, "token type mismatch")
+		} else {
+			// check it does not exist
+			_, _, err := qe.GetTokens(tokenID)
+			assert.Error(err, "token [%s] should not exist", tokenID)
+			assert.True(strings.Contains(err.Error(), "token not found"))
+		}
+	}
+
```
mbrandenburger, line 94: It seems that this code is added 3x in this PR. Please refactor

token/stream.go
```diff
@@ -19,11 +19,22 @@ type QueryService interface {
...
-	Quantity     token2.Quantity
+	// Type is the type of token
+	Type string
+	// Quantity is the quantity of tokens
+	Quantity token2.Quantity
+}
+
+func (o Output) ID(txID string) *token2.ID {
+	return &token2.ID{TxId: txID, Index: o.Index}
 }
```
mbrandenburger, line 38: the name of the method suggests to return the `ID` of an output (if something like that exists) even though it returns a tokenID.

I am wondering if this helper must be part of Output or can we make it a simple utility function?

integration/token/fungible/views/swap.go
```diff
@@ -184,5 +204,26 @@ func (t *SwapResponderView) Call(context view.Context) (interface{}, error) {
 	assert.NoError(err, "failed to retrieve vault status for transaction [%s]", tx.ID())
 	assert.Equal(network.Valid, vc, "transaction [%s] should be in valid state", tx.ID())
 
+	// Check that the tokens are or are not in the vault
+	qe := vault.TokenVault().QueryEngine()
+	outputs, err := tx.Outputs()
+	assert.NoError(err, "failed to retrieve outputs")
+	for _, output := range outputs.Outputs() {
+		tokenID := output.ID(tx.ID())
```
mbrandenburger, line 212: See my comment below regarding ID. This line confuses me a bit.
Why not just:
```go
tokenID := &token2.ID{TxId: tx.ID, Index: output.Index}
```
Not sure if we need a constructor function for that

mbrandenburger (changes requested):

integration/token/fungible/views/accept.go
```diff
@@ -72,5 +71,16 @@ func (a *AcceptCashView) Call(context view.Context) (interface{}, error) {
 	assert.NoError(err, "failed to retrieve vault status for transaction [%s]", tx.ID())
 	assert.Equal(network.Valid, vc, "transaction [%s] should be in valid state", tx.ID())
 
+	// Check that the tokens are in the vault
+	qe := vault.TokenVault().QueryEngine()
+	for _, output := range outputs.ByRecipient(id).Outputs() {
+		tokenID := &token2.ID{TxId: tx.ID(), Index: output.Index}
+		_, toks, err := qe.GetTokens(tokenID)
```
yacovm: It might be more efficient to aggregate the token identifiers and then do a query for all of them.

integration/token/fungible/views/swap.go
```diff
@@ -184,5 +196,18 @@ func (t *SwapResponderView) Call(context view.Context) (interface{}, error) {
 	assert.NoError(err, "failed to retrieve vault status for transaction [%s]", tx.ID())
 	assert.Equal(network.Valid, vc, "transaction [%s] should be in valid state", tx.ID())
 
+	// Check that the tokens are in the vault
+	qe := vault.TokenVault().QueryEngine()
+	outputs, err := tx.Outputs()
+	assert.NoError(err, "failed getting outputs")
+	for _, output := range outputs.ByRecipient(me).Outputs() {
+		tokenID := &token2.ID{TxId: tx.ID(), Index: output.Index}
+		_, toks, err := qe.GetTokens(tokenID)
```
yacovm: ditto

token/stream.go
adecaro: I'm sure I get your comment. An output has a well defined ID with the respect to a transaction id.

integration/token/fungible/views/accept.go
adecaro: this is just an example. I think it is okay, no?

integration/token/fungible/views/accept.go
adecaro: fixed

yacovm (approved):

Issue #317, opened 2022-07-30
Enable staticcheck in CI


mbrandenburger:
An assessment of the current code is needed and staticchecks issues should be resolved before enabling staticcheck in CI.

alexandrosfilios:
Checks to be addressed:

- [ ] ST1001 https://github.com/hyperledger-labs/fabric-token-sdk/pull/323
- [ ] S1023 https://github.com/hyperledger-labs/fabric-token-sdk/pull/324
- [ ] S1011 https://github.com/hyperledger-labs/fabric-token-sdk/pull/326
- [ ] S1039 https://github.com/hyperledger-labs/fabric-token-sdk/pull/327
- [ ] SA4010 https://github.com/hyperledger-labs/fabric-token-sdk/pull/328
- [ ] SA9003 https://github.com/hyperledger-labs/fabric-token-sdk/pull/329
- [ ] S1038, S1012, S1021, SA4001 PR to open
- [ ] SA5007 PR to open
- [ ] SA1019 PR to open
- [ ] SA4022
- [ ] SA4011
- [ ] U1000

PR #336, opened 2022-08-04
Reduce staticcheck warnings (U1000)

alexandrosfilios:
As part of issue #317, removed warnings related to:
* U1000: Removed unused variables/functions


token/core/cache/secondcache/second_chance.go
```diff
@@ -49,13 +49,6 @@ func New(cacheSize int) *secondChanceCache {
 	return &cache
 }
 
-func (cache *secondChanceCache) len() int {
```
mbrandenburger, line 52: uiiii we hance the `secondChanceCache` also in FTS. I would suggest that we just use it from there instead of maintaining here a copy.

https://github.com/hyperledger-labs/fabric-smart-client/tree/main/platform/view/services/cache/secondcache

token/core/zkatdlog/crypto/sigproof/membership.go
```diff
@@ -51,8 +51,6 @@ type MembershipWitness struct {
 	value *math.Zr
 	// hash is hash of value and it is also signed with Pointcheval-Sandes signature
 	hash *math.Zr
-	// sigBlindingFactor is the randomness used to obfuscate signature
-	sigBlindingFactor *math.Zr
```
adecaro, line 55: Kaoutar, please, can you check this?

adecaro (changes requested):

token/core/cache/secondcache/second_chance.go
alexandrosfilios: Unfortunately it is private and I can't import it (unless I make it public of course).
This code is also copied from `hyperledger/fabric`.

Fortunately, I don't see any usage of this whole file, so most probably I could delete it.

mbrandenburger: ah yes, it seems we got it from Fabric :) I believe we could make it public in FSC if we really needed it here in FTS.

If not needed here .... let's remove

token/core/zkatdlog/crypto/sigproof/membership.go
mbrandenburger: Let's mark this as unused and add a TODO for @KElkhiyaoui to check when she is back from vacation, so this does not block the entire PR

mbrandenburger: cc @alexandrosfilios

mbrandenburger (approved):

PR #339, opened 2022-08-04
Reduce staticcheck warnings (SA4011)

alexandrosfilios:
As part of issue #317, removed warnings related to:
* SA4011: Ineffective break statement

### DIFF
--- a/token/services/certifier/interactive/client.go
+++ b/token/services/certifier/interactive/client.go
@@ -138,24 +138,26 @@ func (d *CertificationClient) Scan() {
 			logger.Debugf("request certification of [%v] satisfied with no error", toBeCertified)
 		}
 
-		done := false
-		for {
-			select {
-			case <-d.ctx.Done():
-				return
-			case <-time.After(2 * time.Second):
-				txid, err := d.vault.GetLastTxID()
-				if err != nil {
-					continue
-				}
-				if txid != lastTXID {
-					lastTXID = txid
-					done = true
-					break
-				}
-			}
-			if done {
-				break
+		nextTxId, ok := d.getNextCommittedTxID(lastTXID)
+		if !ok {
+			return
+		}
+		lastTXID = nextTxId
+	}
+}
+
+func (d *CertificationClient) getNextCommittedTxID(lastTXID string) (string, bool) {
+	timeout := time.NewTimer(2 * time.Second)
+	defer timeout.Stop()
+
+	for {
+		select {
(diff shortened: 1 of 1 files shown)


token/services/certifier/interactive/client.go
```diff
@@ -138,25 +138,17 @@ func (d *CertificationClient) Scan() {
 			logger.Debugf("request certification of [%v] satisfied with no error", toBeCertified)
 		}
 
-		done := false
-		for {
+		for done := false; !done; {
```
mbrandenburger, line 141: mhhhhh .... while this maybe fixes the break .... the entire methods should be improved to get rid of the "for in for" situation.

mbrandenburger (changes requested):

token/services/certifier/interactive/client.go
adecaro: can we move this for to another accessory function?

alexandrosfilios: Sure I also try to avoid nested for's.

token/services/certifier/interactive/client.go
```diff
@@ -138,25 +138,32 @@ func (d *CertificationClient) Scan() {
...
-				break
+		nextTxId := <-d.getNextCommittedTxID(lastTXID)
+		if !nextTxId.ok {
+			return
+		}
+		lastTXID = nextTxId.value
+	}
+}
+
+func (d *CertificationClient) getNextCommittedTxID(lastTXID string) <-chan channelResult {
```
mbrandenburger, line 149: while I really like channels, not sure why we use one here :)

we are not running `getNextCommittedTxID` in a new go routine, actually what we want is a blocking behaviour, right?

So we could just return (string, bool) and use it like that:

```go
nextTxId, ok := d.getNextCommittedTxID(lastTXID)
if !ok {
    return
}
```

In that case there is also no need for `channelResult`

mbrandenburger (changes requested):

token/services/certifier/interactive/client.go
alexandrosfilios: I think that we want to run the outer for-loop once per new TX ID. So we need to block until the vault gives us a new TX ID value. We could still skip the channel within the method `getNextCommittedTxID` in order to return `(string, bool)` but I think this would make the inner for loop run non stop (now we poll every 2 seconds for the result).

mbrandenburger (approved):

PR #342, opened 2022-08-05
Add manual CI test trigger

mbrandenburger:
This PR adds support to manually start CI test and set the FSC dependency. Thereby, we can easily check if open PR on FSC is working fine with FTS.

Signed-off-by: Marcus Brandenburger <bur@zurich.ibm.com>

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -5,6 +5,12 @@ on:
     branches: [ main ]
   pull_request:
     branches: [ main ]
+  workflow_dispatch:
+    inputs:
+      fsc-version:
+        default: 'github.com/hyperledger-labs/fabric-smart-client@main'
+        required: true
+        description: 'FSC dependency'
 
 env:
   GOPATH: ${{ github.workspace }}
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -26,6 +32,17 @@ jobs:
         with:
           go-version-file: "${{ env.FTS_PATH }}/go.mod"
 
+      - name: Replace FSC dep
+        working-directory: ${{ env.FTS_PATH }}
+        run: |
+          if [ -n "${{ github.event.inputs.fsc-version }}" ]; then
+            echo "Replace FSC dependency"
+            go mod edit -replace=github.com/hyperledger-labs/fabric-smart-client=${{ github.event.inputs.fsc-version }}
```
alexandrosfilios, line 40: Maybe it would make sense to also make `hyper-ledger-labs` configurable, so that we can test the integration while its on our fork still and not when it reaches the master

alexandrosfilios (approved):

PR #362, opened 2022-08-19
Cleanup deps.go

alexandrosfilios:
Re-using deps.go from fabric-smart-client and removing redundant deps from blank imports in deps.go


adecaro (approved):
LGTM

PR #414, opened 2022-10-27
check that output's quantity is less or equal to max_value. #411

adecaro:


integration/token/fungible/views/accept.go
```diff
@@ -41,6 +41,9 @@ func (a *AcceptCashView) Call(context view.Context) (interface{}, error) {
 	// Just a fancy query to show the capabilities of the services we are using.
 	precision := tx.TokenService().PublicParametersManager().Precision()
 	for _, output := range outputs.ByRecipient(id).Outputs() {
+		if output.Type == "MAX" {
```
KElkhiyaoui, line 44: what does "MAX" refer to?

adecaro: it is just a token type 😅

KElkhiyaoui (approved):

PR #431, opened 2022-12-01
load public params from vault when possible

adecaro:
This PR brings the following: At bootstrap, the stack first tries to load the public parameters from the local vault. 
If the public parameters are not found and are nevertheless required, they are fetched from the remote network.


integration/token/fungible/tests.go
```diff
@@ -209,6 +209,9 @@ var BobAcceptedTransactions = []*ttxdb.TransactionRecord{
 func TestAll(network *integration.Infrastructure, auditor string) {
 	RegisterAuditor(network, auditor)
 
+	// give some time to the nodes to get the public parameters
+	time.Sleep(10 * time.Second)
```
HagarMeir, line 213: Is there a way to move this sleep somewhere so it doesn't have to appear at the beginning of each test?

HagarMeir (approved):

PR #440, opened 2022-12-13
remove htlc TODOs

HagarMeir:
Signed-off-by: Hagar Meir <hagar.meir@ibm.com>


token/services/interop/htlc/transaction.go
```diff
@@ -251,12 +253,6 @@ func (t *Transaction) Claim(wallet *token.OwnerWallet, tok *token2.UnspentToken,
 		return errors.New("failed to unmarshal RawOwner as an htlc script")
 	}
```
adecaro, line 255: let's check if the preImage matches the image in the script. We avoid possible usage mistake :)

adecaro (changes requested):

Issue #459, opened 2023-02-24
Swap view - asserting transaction output fails


IgorSimj:
For example, Alice have 10 TOK and Bob have 5 KOT

Swap view is initiated (from Alice's node), 1 TOK for 1 KOT. 
After Bob add his own transfers execution continue at Alice's node where tx outputs will be checked.
```

outputs, err := tx.Outputs()
os := outputs.ByRecipient(other)
assert.Equal(0, os.Sum().Cmp(big.NewInt(int64(t.FromQuantity))))
```
'os' object contain Bob's outputs (1 TOK, 4 KOT) and assert will fail. 
Here, we need to analyze action by action.

adecaro:
done

PR #460, opened 2023-02-28
#459 - filter tx outputs by type as well

IgorSimj:
Transaction outputs contain information about all tokens that owner has, therefore when validating transaction additional filter (by type) is applied that will return number of tokens which are part of the swap.


adecaro:
Hi @IgorSimj , any chance you can rebase the PR, please?

adecaro:
ported to #781

PR #473, opened 2023-05-31
various improvements

adecaro:
This PR brings the following:
- Endorse options for accept view
- Options for ttx endorsement
- Check token request if you are not an auditor
- Open new transaction to get token request
- Expose methods to get the token request
- ttxdb: store token request when storing transaction records
- Load public params from file if the configuration says so


adecaro:
embedded in #478

PR #479, opened 2023-07-06
Mailman

mbrandenburger:
This PR introduces a new token selector implementation that targets high concurrent use.


token/services/network/fabric/tcc/main/main
adecaro, line 1: this should not be checked in, right?

mbrandenburger: removed

PR #508, opened 2023-10-16
add tokengen function to update auditor and issuer certs in dlog pp files

arner:
If the auditor or issuer certificate expires, we have to update it in the token chaincode. This action should *not* change the public parameters, because this would render existing tokens/proofs invalid.

This PR adds a command to the tokengen cli utility to update the auditor and/or issuers certificates in an existing `zkatdlog_pp.json`, without changing the public parameters.


adecaro (changes requested):
Arne, please, could you update also the README under `cmd/tokengen`?

PR #516, opened 2023-11-12
store mine key in tokenstore instead of processor directly

arner:
Moving the storage of the mine key to the TokenStore makes the code more flexible to change; now the processor only uses the TokenStore for any storage instead of using the rwSet directly. An implementation of the TokenStore interface would be independent of the processor to decide how to store tokens and lookup values.

Note that the code assumes that tokens stored via `StoreFabToken` are always 'mine'. If that's not true we could change the signature of the method to add a boolean flag for 'mine'.


adecaro (approved):
LGTM

PR #523, opened 2023-12-14
refactoring services

adecaro:
This PR does the following: 
- Refactors the package `services/vault`. The goal is to allow different implementations of the token vault.
- Make TransactionInfoProvider private


alexandrosfilios (approved):
There are some imports that can be re-formatted.

PR #524, opened 2023-12-15
allow empty prefix

arner:
Table prefixes for the sql driver are useful for data segregation if you have multiple instances of the token sdk running with different configuration / owner wallets and want to connect them to the same database instance (e.g. to save cost or ease operations in a non-production environment).

The current code defaults to an empty prefix if the tablePrefix configuration is not set, but then fails because the regex does not allow an empty prefix. This PR fixes that, so that the user can not set the tablePrefix if they don't need it.


adecaro:
Hi @arner , why do we need this? Could you explain it in the PR's description, please? Thanks 😄

adecaro (approved):
LGTM

PR #529, opened 2024-01-10
remove kvs from ttxdb

adecaro:


token/services/ttxdb/db/badger/badger.go
```diff
@@ -512,6 +521,91 @@ func (db *Persistence) GetStatus(txID string) (driver.TxStatus, error) {
...
+
+	txn := db.db.NewTransaction(true)
+	defer txn.Discard()
+	err = txn.Set([]byte(key), b)
+	if err != nil {
+		return errors.Wrapf(err, "could not set value for key %s", key)
+	}
+	err = txn.Commit()
+	if err != nil {
+		return errors.Wrapf(err, "could not committ transaction for [%s]", txID)
```
arner, line 550: committ = typo

token/services/ttxdb/db/badger/badger.go
```diff
@@ -512,6 +521,91 @@ func (db *Persistence) GetStatus(txID string) (driver.TxStatus, error) {
 	return driver.Unknown, errors.Errorf("transaction [%s] not found", txID)
 }
 
+func (db *Persistence) AddTransactionEndorsementAck(txID string, id view.Identity, sigma []byte) error {
+	key, err := db.transactionEndorseAckKey(txID, id)
+	if err != nil {
+		return errors.Wrapf(err, "could not get key for token request %s", txID)
+	}
+	logger.Debugf("Adding token request [%s] with key [%s]", txID, key)
```
arner, line 529: I think we start logs with lowercase?

token/services/ttxdb/db/sql/sql.go
```diff
@@ -329,6 +330,63 @@ func (db *Persistence) QueryValidations(params driver.QueryValidationRecordsPara
...
+				logger.Warnf("tried to get status for non-existent tx %s, returning unknown", txID)
+				continue
+			}
+			return nil, errors.Wrapf(err, "error querying db")
+		}
+		fmt.Printf(" found [%s] [%s]\n", id, sigma)
+		acks[view.Identity(id).String()] = sigma
+	}
+
+	fmt.Printf(" result [%v]\n", acks)
```
arner, line 386: remove this (or change to debug log)

token/services/ttxdb/db/sql/sql.go
```diff
@@ -329,6 +330,63 @@ func (db *Persistence) QueryValidations(params driver.QueryValidationRecordsPara
...
+		var sigma []byte
+		if err := rows.Scan(&id, &sigma); err != nil {
+			if errors.Is(err, sql.ErrNoRows) {
+				// not an error for compatibility with badger.
+				logger.Warnf("tried to get status for non-existent tx %s, returning unknown", txID)
+				continue
+			}
+			return nil, errors.Wrapf(err, "error querying db")
+		}
+		fmt.Printf(" found [%s] [%s]\n", id, sigma)
```
arner, line 382: Remove or change to debug log

token/services/ttxdb/db/sql/random.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/pkg/errors"
+)
+
+const (
+	// NonceSize is the default NonceSize
+	NonceSize = 24
+)
+
+// GetRandomBytes returns len random looking bytes
+func GetRandomBytes(len int) ([]byte, error) {
```
arner, line 31: In other places, the SQL driver uses go-uuid to generate a unique ID (see https://github.com/hyperledger-labs/fabric-token-sdk/blob/ad5e98882ce2cad448e1e4f397f912e416203552/token/services/ttxdb/db/sql/sql.go#L177). I propose to stick to that.

adecaro: Perfect, I'll use that.

adecaro:
@arner thanks. I have addressed all your comments :)

arner (approved):

token/services/ttxdb/db/badger/badger.go
```diff
@@ -512,6 +521,91 @@ func (db *Persistence) GetStatus(txID string) (driver.TxStatus, error) {
...
+	type Entry struct {
+		key   string
+		value []byte
+	}
+	var entries []Entry
+	stream := db.db.NewStream()
+	stream.NumGo = db.numGoStream
+	stream.LogPrefix = streamLogPrefixStatus
+	txIdAsBytes := []byte(txID)
+	stream.ChooseKey = func(item *badger.Item) bool {
```
alexandrosfilios, line 569: If I understand correctly, you only choose the "tea"-prefixed entries later. In this case I see that badger has a `Prefix` property that I think filters out more efficiently. This way you skip the check later and you iterate over less keys.

token/services/ttxdb/db/badger/badger.go
```diff
@@ -512,6 +521,91 @@ func (db *Persistence) GetStatus(txID string) (driver.TxStatus, error) {
...
+
+func (db *Persistence) GetTransactionEndorsementAcks(txID string) (map[string][]byte, error) {
+	acks := make(map[string][]byte)
+
+	// search for all matching keys
+	type Entry struct {
+		key   string
+		value []byte
+	}
+	var entries []Entry
```
alexandrosfilios, line 564: I see that another method `SetStatus()` and `GetStatus()` use the same logic for the streams. Maybe you could extract this functionality into a method:
```
func (db *Persistence) GetEntriesByTxID(txID string, prefix string) []Entries {
 var entries []Entry
	stream := db.db.NewStream()
	stream.NumGo = db.numGoStream
	stream.LogPrefix = streamLogPrefixStatus
	txIdAsBytes := []byte(txID)
	stream.ChooseKey = func(item *badger.Item) bool {
		return bytes.HasSuffix(item.Key(), txIdAsBytes)
	}
if len(prefix) > 0 { stream.Prefix = []byte(prefix)}
	stream.Send = func(buf *z.Buffer) error {
		list, err := badger.BufferToKVList(buf)
		if err != nil {
			return err
		}
		for _, kv := range list.Kv {
			entries = append(entries, Entry{key: string(kv.Key), value: kv.Value})
		}
		return nil

	}
	if err := stream.Orchestrate(context.Background()); err != nil {
		return err
	}
return entries
}
```

token/services/ttxdb/db/badger/marshaller.go
```diff
@@ -28,6 +28,11 @@ func MarshalTokenRequest(mr *TokenRequest) ([]byte, error) {
 	return json.Marshal(mr)
 }
 
+// MarshalTransactionEndorseAck marshals a TransactionEndorseAck into a byte array
```
alexandrosfilios, line 31: The wrapper methods for marshalling make it a bit harder to re-use code. Do we need to have them separate?

token/services/ttxdb/db/badger/badger.go
```diff
@@ -512,6 +521,91 @@ func (db *Persistence) GetStatus(txID string) (driver.TxStatus, error) {
...
+
+	if len(entries) == 0 {
+		// nothing to update
+		logger.Debugf("no entries found for txID %s, skipping", txID)
+		return nil, nil
+	}
+
+	for _, entry := range entries {
+		switch {
+		case strings.HasPrefix(entry.key, "tea"):
```
alexandrosfilios, line 596: Maybe we could have a string-backed enum for these prefixes

PR #531, opened 2024-01-15
identity storage

adecaro:
This PR extends the `StorageProvider` to add support for `WalletPathStorage` and `keystore.KVS`.
This further reduces direct dependency to the FSC's KVS.

### DIFF
--- a/token/core/identity/kvs/storage.go
+++ b/token/core/identity/kvs/storage.go
@@ -87,3 +87,56 @@ func (s *IdentityStorage) IdentityExists(identity view.Identity, wID identity.Wa
 func (s *IdentityStorage) walletPrefix(wID identity.WalletID) string {
 	return fmt.Sprintf("%s-%s-%s-%s", s.tmsID.Network, s.tmsID.Channel, s.tmsID.Namespace, wID)
 }
+
+type WalletPathStorage struct {
+	kvs    KVS
+	prefix string
+}
+
+func NewWalletPathStorage(kvs KVS, prefix string) *WalletPathStorage {
+	return &WalletPathStorage{kvs: kvs, prefix: prefix}
+}
+
+func (w *WalletPathStorage) AddWallet(id string, path string) error {
+	k, err := kvs.CreateCompositeKey("token-sdk", []string{"msp", w.prefix, "registeredIdentity", id})
+	if err != nil {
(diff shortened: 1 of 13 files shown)


token/core/identity/storage.go
```diff
@@ -26,8 +38,15 @@ type Storage interface {
 	LoadMeta(identity view.Identity, meta any) error
 }
 
+type WalletPathStorage interface {
+	AddWallet(id string, path string) error
```
alexandrosfilios, line 42: We could simplify this signature to
```
Add(w WalletPath) error
GetAll() (Iterator[WalletPath], error)
```
Then we could even use generics, if this kind of storage is used in other places

token/sdk/identity/provider.go
```diff
@@ -22,6 +23,14 @@ func NewKVSStorageProvider(sp view.ServiceProvider) *KVSStorageProvider {
 	return &KVSStorageProvider{sp: sp}
 }
 
-func (s *KVSStorageProvider) New(tmsID token.TMSID) (identity.Storage, error) {
+func (s *KVSStorageProvider) NewStorage(tmsID token.TMSID) (identity.Storage, error) {
```
alexandrosfilios, line 26: Maybe we could skip the error since it is always null. This way we will have a nil check less.

token/core/identity/kvs/storage.go
```diff
@@ -88,3 +87,56 @@ func (s *IdentityStorage) IdentityExists(identity view.Identity, wID identity.Wa
...
+		return nil, errors.WithMessage(err, "failed to get registered identities from kvs")
+	}
+	return &WalletPathStorageIterator{Iterator: it}, nil
+}
+
+type WalletPathStorageIterator struct {
+	kvs.Iterator
+}
+
+func (w *WalletPathStorageIterator) HasNext() bool {
```
alexandrosfilios, line 120: You can remove `HasNext()` and `Close()`. They are "inherited" from `kvs.Iterator`.

token/sdk/identity/provider.go
adecaro: In the current implementation, yes. But another might return an error 😓

token/core/identity/storage.go
adecaro: 👍

PR #537, opened 2024-01-26
reduce dependency on service provider

adecaro:
This PR replaces dependency to `ServiceProvider` with the actual needed dependencies.


alexandrosfilios (approved):
LGTM

PR #538, opened 2024-01-29
new public params management

adecaro:
This PR introduces the following changes:
- Token Driver: public parameters must be available at time of instantiation
- AuditorDB does not require a tms anymore
- if the vault contains the public params, the corresponding tms is instantiated immediately

### DIFF
--- a/token/core/fabtoken/ppm/ppm.go
+++ b/token/core/fabtoken/ppm/ppm.go
@@ -54,7 +54,11 @@ func NewPublicParamsManagerFromParams(pp *fabtoken.PublicParams) (*PublicParamsM
 
 // PublicParameters returns the public parameters of PublicParamsManager
 func (v *PublicParamsManager) PublicParameters() driver.PublicParameters {
-	return v.PublicParams()
+	pp := v.PublicParams()
+	if pp == nil {
+		return nil
+	}
+	return pp
 }
 
 // SerializePublicParameters returns the public params in a serialized form
--- a/token/core/fabtoken/ppm/ppm.go
+++ b/token/core/fabtoken/ppm/ppm.go
@@ -38,7 +38,10 @@ type PublicParamsManager struct {
 
 // NewPublicParamsManager initializes a PublicParamsManager with the passed PublicParamsLoader
 func NewPublicParamsManager(PPLabel string, vault Vault) *PublicParamsManager {
-	return &PublicParamsManager{PPLabel: PPLabel, Vault: vault, Mutex: sync.RWMutex{}}
+	return &PublicParamsManager{
+		PPLabel: PPLabel,
+		Vault:   vault,
+	}
 }
 
 // NewPublicParamsManagerFromParams initializes a PublicParamsManager with the passed PublicParams
@@ -46,12 +49,16 @@ func NewPublicParamsManagerFromParams(pp *fabtoken.PublicParams) (*PublicParamsM
 	if pp == nil {
 		return nil, errors.Errorf("public parameters must be non-nil")
 	}
-	return &PublicParamsManager{PP: pp, Mutex: sync.RWMutex{}}, nil
+	return &PublicParamsManager{PP: pp, PPLabel: pp.Label}, nil
 }
 
 // PublicParameters returns the public parameters of PublicParamsManager
 func (v *PublicParamsManager) PublicParameters() driver.PublicParameters {
-	return v.PublicParams()
+	pp := v.PublicParams()
+	if pp == nil {
+		return nil
+	}
+	return pp
 }
 
 // SerializePublicParameters returns the public params in a serialized form
@@ -89,7 +96,6 @@ func (v *PublicParamsManager) SetPublicParameters(raw []byte) error {
 	if err != nil {
 		return err
 	}
-	logger.Debugf("setting new public parameters...")
 
 	if err := pp.Validate(); err != nil {
 		return errors.WithMessage(err, "invalid public parameters")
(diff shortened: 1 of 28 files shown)


token/core/fabtoken/ppm/ppm.go
```diff
@@ -47,21 +37,28 @@ type PublicParamsManager struct {
...
 	}
-	return &PublicParamsManager{PP: pp, Mutex: sync.RWMutex{}}, nil
+	return &PublicParamsManager{PP: pp, PPLabel: pp.Label}, nil
 }
 
 // PublicParameters returns the public parameters of PublicParamsManager
 func (v *PublicParamsManager) PublicParameters() driver.PublicParameters {
-	return v.PublicParams()
+	pp := v.PublicParams()
+	if pp == nil {
```
alexandrosfilios, line 58: What is the difference?

token/core/tms.go
```diff
@@ -56,87 +56,151 @@ func NewTMSProvider(sp view.ServiceProvider, configProvider ConfigProvider, vaul
...
+	if len(opts.Namespace) == 0 {
+		return errors.Errorf("namespace not specified")
+	}
+	if len(opts.PublicParams) == 0 {
+		return errors.Errorf("public params not specified")
+	}
+	m.lock.Lock()
+	defer m.lock.Unlock()
+
+	key := opts.Network + opts.Channel + opts.Namespace
```
alexandrosfilios, line 125: I would extract this into another method because this logic for the creation of the key is used on other places. Not to avoid code repetition but to be consistent on the construction of the key when we read and when we write.

token/core/tms.go
```diff
@@ -56,87 +56,151 @@ func NewTMSProvider(sp view.ServiceProvider, configProvider ConfigProvider, vaul
...
+	logger.Debugf("update tms for [%s:%s:%s] with key [%s]", opts.Network, opts.Channel, opts.Namespace, key)
+	service, ok := m.services[key]
+	if ok {
```
alexandrosfilios, line 128:
```
if !ok {
panic("not implemented yet")
}
```

token/driver/driver.go
```diff
@@ -25,5 +27,5 @@ type Driver interface {
 // ExtendedDriver is the interface that models additional services a token driver may offer
 type ExtendedDriver interface {
 	// NewWalletService returns an instance of the WalletService interface for the passed arguments
-	NewWalletService(sp view.ServiceProvider, network string, channel string, namespace string, params PublicParameters) (WalletService, error)
+	NewWalletService(sp ServiceProvider, networkID string, channel string, namespace string, params PublicParameters) (WalletService, error)
```
alexandrosfilios, line 30: tmsid

token/core/tms.go
```diff
@@ -56,87 +56,151 @@ func NewTMSProvider(sp view.ServiceProvider, configProvider ConfigProvider, vaul
...
-			return "", errors.WithMessagef(err, "failed to identify driver from the configuration of [%s:%s:%s], loading driver from public parameters failed too [%s]", networkID, channel, namespace, err)
+			return "", errors.WithMessagef(err, "failed to identify driver from the configuration of [%s:%s:%s], loading driver from public parameters failed too [%s]", opts.Network, opts.Channel, opts.Namespace, err)
 		}
 
 		driverName := tmsConfig.TMS().Driver
 		if len(driverName) != 0 {
 			return driverName, nil
 		}
-		return "", errors.WithMessagef(err, "failed to identify driver for [%s:%s:%s]", networkID, channel, namespace)
+		return "", errors.WithMessagef(err, "failed to identify driver for [%s:%s:%s]", opts.Network, opts.Channel, opts.Namespace)
```
alexandrosfilios, line 180:
```
errors.WithMessagef(err, "failed to identify driver for %s", opts)
```

token/provider.go
```diff
@@ -86,14 +96,23 @@ func (p *ManagementServiceProvider) GetManagementService(opts ...ServiceOption)
...
+	var tokenService driver.TokenManagerService
+	driverOpts := driver.ServiceOptions{
+		Network:             opt.Network,
+		Channel:             opt.Channel,
+		Namespace:           opt.Namespace,
+		PublicParamsFetcher: opt.PublicParamsFetcher,
+		PublicParams:        opt.PublicParams,
+		Params:              opt.Params,
+	}
+	if aNew {
```
alexandrosfilios, line 109: Is it possible to have two different TMSs having the same TMSID?

token/services/owner/manager.go
```diff
@@ -50,99 +43,66 @@ func NewManager(tmsProvide TokenManagementServiceProvider, np NetworkProvider, t
...
-	}
-}
-
-func (cm *Manager) newOwner(tms *token.ManagementService) (*Owner, error) {
-	db, err := cm.ttxdbProvider.DB(&tmsWallet{tms: tms})
+func (cm *Manager) newOwner(tmsID token.TMSID) (*Owner, error) {
+	db, err := cm.ttxdbProvider.DBByTMSId(tmsID)
 	if err != nil {
-		return nil, errors.WithMessagef(err, "failed to get ttxdb for [%s:%s]", tms.ID().Network, tms.ID().Channel)
+		return nil, errors.WithMessagef(err, "failed to get ttxdb for [%s:%s]", tmsID.Network, tmsID.Channel)
```
alexandrosfilios, line 71: you don't have the namespace. you could use directly "[%s]", tmsID, since it implements `String()`

alexandrosfilios (approved):
LGTM

token/core/fabtoken/ppm/ppm.go
adecaro: pp is a struct, if pp is nil and return as an interface, what you get is a pointer to a nil struct which is not nil.

PR #540, opened 2024-02-02
RWS-based vault

adecaro:
This PR refactors the code to better split the RWS-based token vault. This will make more clear that a different implementation can be plugged.


token/sdk/vault/rws/rws.go
```diff
@@ -0,0 +1,110 @@
...
+		return nil, errors.WithMessagef(err, "failed to get token store")
+	}
+
+	storageProvider, err := certification.GetStorageProvider(v.sp)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get storage provider")
+	}
+
+	// Create new vault
+	if fns := fabric.GetFabricNetworkService(v.sp, network); fns != nil {
```
arner, line 71: This couples the Vault implementation with only two possible Networks (Fabric and Orion), but the code is almost the same between the two. Could it be an idea to hide the implementation behind an interface, so that as few components as possible care about the underlying network implementation?

arner (approved):

token/sdk/vault/rws/rws.go
adecaro: indeed, the responsibility of the provider is to return an instance of the interface vault.TokenVault.
One can change this provider in the sdk.go for full customisation. 
So, we need to care about the underlying backend only there. 
Anyhow, let's refine even more this part :)

PR #542, opened 2024-02-04
token and audit db

arner:
This PR introduces two additional databases:
- The `TokenDB` used to store token information. It is populated starting from either transactions or directly from token requests (future)
- The `AuditDB` is similar to the current transaction db, but it is the database the auditor uses to store information relative to transactions.

This PR also removes support from badger db.

Things that remain to do and will address in subsequent PRs:
- Ability to remove the entries relative to a given transaction from the token db
- Rework the SQL table model of the token db
- Add parallelism to the token db, meaning more transaction can be processed at the same time


token/services/sql/init.go
```diff
@@ -0,0 +1,139 @@
...
+
+var logger = flogging.MustGetLogger("token-sdk.sql")
+
+var Transactions *TransactionDB
+var Tokens *TokenDB
+
+func Init(driverName, dataSourceName, tablePrefix, name string, createSchema bool, maxOpenConns int) error {
+	logger.Infof("connecting to sql database [%s:%s]", driverName, tablePrefix) // dataSource can contain a password
+	if Transactions != nil || Tokens != nil {
+		// return errors.New("database can only be initialized once") // TODO: how do we handle this?
```
arner, line 28: This one is about having several TMS and database instances. We generate a suffix for the table names based on the 'name' (i.e. we're hashing the TMS id), see getTableNames in this same file. It's an ugly solution and I hope we can get rid of it. More so because there is only one global instance of each DB in the package, so you're not even able to instantiate a second one for another TMS.

More logical is if the user configures the prefix in the TMS config (like we discussed earlier). I'm not sure what that means for the global instances in this package, we should discuss.

token/services/sql/init.go
```diff
@@ -0,0 +1,139 @@
...
+	Certifications        string
+	Tokens                string
+	Ownership             string
+	AuditTokens           string
+	IssuedTokens          string
+	PublicParams          string
+	Ledger                string
+}
+
+// TODO get rid of the suffix
```
arner, line 107: See above

token/services/sql/tokens.go
```diff
@@ -0,0 +1,886 @@
...
+	var quantity string
+	if err := row.Scan(&tokenOwner, &tokenType, &quantity); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, nil, nil
+		}
+		return nil, nil, err
+	}
+
+	// select owners
+	query = fmt.Sprintf("SELECT enrollment_id FROM %s WHERE %s", db.table.Ownership, where)
```
arner, line 157: token and owners can be retrieved in one query. See UnspentTokensIteratorBy(ns, ownerEID, typ string), it's quite similar (would be good to have a test!).

token/services/sql/tokens.go
```diff
@@ -0,0 +1,886 @@
...
+	}
+	logger.Debugf("found [%d] records, expected [%d]", counter, len(inputs))
+	if err = rows.Err(); err != nil {
+		return nil, isSpent, err
+	}
+	if counter == 0 {
+		return nil, nil, errors.Errorf("token not found for key [%s:%d]", inputs[0].TxId, inputs[0].Index)
+	}
+	if counter != len(inputs) {
+		return nil, nil, errors.Errorf("record missings")
```
arner, line 726: Maybe we can be more descriptive, similar to the debug of line 716?

integration/nwo/token/fabric/fabric.go
```diff
@@ -169,10 +170,36 @@ func (p *NetworkHandler) GenerateArtifacts(tms *topology2.TMS) {
...
+	Expect(os.MkdirAll(p.TokensDBSQLDataSourceDir(node), 0775)).ToNot(HaveOccurred(), "failed to create [%s]", p.TokensDBSQLDataSourceDir(node))
+	Expect(os.MkdirAll(p.AuditDBSQLDataSourceDir(node), 0775)).ToNot(HaveOccurred(), "failed to create [%s]", p.AuditDBSQLDataSourceDir(node))
+
 	t, err := template.New("peer").Funcs(template.FuncMap{
 		"TMSID":   func() string { return tms.ID() },
 		"TMS":     func() *topology2.TMS { return tms },
 		"Wallets": func() *generators.Wallets { return p.GetEntry(tms).Wallets[node.Name] },
+		"SQLDataSource": func() string {
+			return "file:" +
+				filepath.Join(
```
alexandrosfilios, line 183: We could extract this to a method

integration/nwo/token/fabric/fabric.go
```diff
@@ -295,6 +322,15 @@ func (p *NetworkHandler) GenOwnerCryptoMaterial(tms *topology2.TMS, nodeID strin
 	return ""
 }
 
+func (p *NetworkHandler) DeleteDBs(node *sfcnode.Node) {
+	//logger.Infof("remove all [%s]", p.TTXDBSQLDataSourceDir(node))
+	//Expect(os.RemoveAll(p.TTXDBSQLDataSourceDir(node))).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
adecaro: 👍

integration/nwo/token/platform.go
```diff
@@ -284,18 +282,22 @@ func (p *Platform) StartSession(cmd *exec.Cmd, name string) (*gexec.Session, err
...
 	return p.Topology
 }
 
+func (p *Platform) DeleteDBs(tms *topology2.TMS, id string) {
+	logger.Infof("delete dbs for [%s:%s]", tms.ID(), id)
+	for _, node := range tms.FSCNodes {
+		if node.Name == id {
+			// get the network handler for this TMS
+			nh := p.NetworkHandlers[p.Context.TopologyByName(tms.Network).Type()]
+			// generate artifacts
```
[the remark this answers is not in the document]
adecaro: 👍

integration/token/fungible/views/checks.go
```diff
@@ -111,14 +118,14 @@ func (m *CheckTTXDBView) Call(context view.Context) (interface{}, error) {
...
 
-		txInfo, err := ttxDB.TransactionInfo(transactionRecord.TxID)
-		if err != nil {
-			errorMessages = append(errorMessages, fmt.Sprintf("failed to load transaction info for transaction record [%s]: [%s]", transactionRecord.TxID, err))
-		}
-		assert.NotEmpty(txInfo.TokenRequest, "token request not found in database")
-		tokenRequest, err := ttxDB.GetTokenRequest(transactionRecord.TxID)
-		assert.NoError(err, "failed to retrieve token request for [%s]", transactionRecord.TxID)
-		assert.Equal(txInfo.TokenRequest, tokenRequest, "token requests do not match")
+		//txInfo, err := ttxDB.TransactionInfo(transactionRecord.TxID)
```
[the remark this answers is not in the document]
adecaro: 👍

token/sdk/vault/db/vault.go
```diff
@@ -4,21 +4,24 @@ Copyright IBM Corp. All Rights Reserved.
...
 	fabric2 "github.com/hyperledger-labs/fabric-token-sdk/token/services/network/fabric"
 	orion2 "github.com/hyperledger-labs/fabric-token-sdk/token/services/network/orion"
-	"github.com/hyperledger-labs/fabric-token-sdk/token/services/network/processor/rws"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/network/processor/db"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/tokendb"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/ttxdb"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/vault"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/vault/certification"
-	rws2 "github.com/hyperledger-labs/fabric-token-sdk/token/services/vault/rws"
+	db2 "github.com/hyperledger-labs/fabric-token-sdk/token/services/vault/db"
```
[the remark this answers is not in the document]
adecaro: 👍

token/services/sql/init.go
adecaro: Notice that we use this only for the tests. For the rest, each TMS defines it is own databases with their locations. 
I think we should be good. What do you think?

token/services/sql/tokens.go
adecaro: isn't the `join` more expensive?

token/services/sql/init.go
arner: Right, I took your PR for a spin now, and noticed that the user can define all the settings per db, including the tableprefix. This removes the need for the suffix, so let's remove it.

I still have my doubts about this level of configurability. Why don't we always use the same database, tableprefix, maxOpenConnections for all the sql stores in a TMS? Setting it in one place will make it easier to configure and understand, and harder to make mistakes.

token/services/sql/tokens.go
arner: No, the database optimizes for queries like that, especially because we have indexes on the txid on the ownership and token tables. I mean, in this case it's not going to matter that much in terms of performance but I'd prefer to use the db functionalities where we can. Let me know if you want me to write the query.

token/services/network/processor/db/tokenstore.go
```diff
@@ -0,0 +1,140 @@
...
+			TxID:           txID,
+			Index:          index,
+			Namespace:      ns,
+			IssuerRaw:      nil,
+			OwnerRaw:       tok.Owner.Raw,
+			Ledger:         tokenOnLedger,
+			LedgerMetadata: tokenOnLedgerMetadata,
+			Quantity:       tok.Quantity,
+			Type:           tok.Type,
+			Amount:         0,
```
arner, line 68: Should we still add the Amount as number to the function signature? We have it for issuedTokens and it'd allow for some nice queries if we add it (get all balances, for instance, and business reporting). By the way shall we also add a StoredAt? That kind of info is always good for debugging or auditability.

token/services/sql/init.go
adecaro: Yes, indeed. I was planning to introduce a unified driver that offers instantiations for all databases backing them with a single sql db instance. Shall I add it here?

token/services/sql/tokens.go
adecaro: Perfect, I'll change the code 👍

token/services/sql/querybuilder.go
```diff
@@ -202,6 +200,25 @@ func in(args *[]interface{}, field string, searchFor []interface{}) (where strin
 	return fmt.Sprintf("(%s)", strings.Join(argnum, " OR "))
 }
 
+func whereTokenIDs(args *[]interface{}, namespace string, ids []*token.ID) (where string) {
+	switch len(ids) {
+	case 0:
+		return ""
```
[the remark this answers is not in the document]
adecaro: if you look at the way we call this function, the assumption is that the list is never empty.
So, I would say we are good there. What do you think?

token/services/sql/tokens.go
adecaro: @arner , actually, shall we address this when we rework the table organization in another PR?

token/services/sql/tokens.go
adecaro: 👍

token/services/network/processor/db/tokenstore.go
arner: We could introduce a limit to the size of a transaction... 64 bit is already 18 billion billion, do we need more?

arner:
I can't approve on github because I started the PR originally, but for the record: I approve of these changes. Still it would be good if we soon (in a new PR) address:

- Removal of the suffix (or you could do that as part of this PR, it's a small code change and not backwards compatible)
- Amount and StoredAt
- GetOwners
- the SDK instantiation and dependencies

adecaro:
thanks @arner . I have added now the conversion to uint. We should be good there. I'll address the rest in other smaller PRs

adecaro:
So, now we have `ttxdb`, `tokendb`, and `auditdb`. `ttxdb` and `auditdb` are identical (code wise), shall we differentiate them somehow or keep as it is because in the future they might diverge?

token/services/db/driver/audit.go
```diff
@@ -0,0 +1,66 @@
...
+
+package driver
+
+import (
+	view2 "github.com/hyperledger-labs/fabric-smart-client/platform/view"
+	token2 "github.com/hyperledger-labs/fabric-token-sdk/token"
+)
+
+// AuditTransactionDB defines the interface for a token transactions database
+type AuditTransactionDB interface {
```
alexandrosfilios, line 15: Could this extend the TransactionDB interface?

token/services/db/driver/audit.go
```diff
@@ -0,0 +1,66 @@
...
+*/
+
+package driver
+
+import (
+	view2 "github.com/hyperledger-labs/fabric-smart-client/platform/view"
+	token2 "github.com/hyperledger-labs/fabric-token-sdk/token"
+)
+
+// AuditTransactionDB defines the interface for a token transactions database
```
alexandrosfilios, line 14: Maybe we could have a more complete description of what these DB's are used for.

token/services/db/sql/driver/driver.go
```diff
@@ -0,0 +1,170 @@
...
+		return nil, nil, "", errors.WithMessagef(err, "failed to load configuration for tms [%s]", tmsID)
+	}
+	if err := tmsConfig.UnmarshalKey(OptsKey, opts); err != nil {
+		return nil, nil, "", errors.Wrapf(err, "failed getting opts for vault")
+	}
+	if opts.Driver == "" {
+		panic(fmt.Sprintf("%s.driver not set. See https://github.com/golang/go/wiki/SQLDrivers", OptsKey))
+	}
+
+	name := fmt.Sprintf("%s-%s-%s", tmsID.Network, tmsID.Channel, tmsID.Namespace)
```
alexandrosfilios, line 91: Does this need to be in this method? We could have a separate static function `DatasourceName(tmsID)` and not even call it in here.

token/services/db/driver/audit.go
adecaro: I would prefer to keep them distinct to favor independent evolution

token/services/db/driver/audit.go
adecaro: 👍

token/services/db/sql/init.go
```diff
@@ -0,0 +1,88 @@
...
+	Ownership             string
+	AuditTokens           string
+	IssuedTokens          string
+	PublicParams          string
+	Ledger                string
+}
+
+func getTableNames(prefix, name string) (tableNames, error) {
+	if prefix != "" {
+		r := regexp.MustCompile("^[a-zA-Z_]+$")
```
alexandrosfilios, line 59: can be a static var

token/services/db/sql/querybuilder.go
```diff
@@ -202,6 +200,25 @@ func in(args *[]interface{}, field string, searchFor []interface{}) (where strin
 	return fmt.Sprintf("(%s)", strings.Join(argnum, " OR "))
 }
 
+func whereTokenIDs(args *[]interface{}, ids []*token.ID) (where string) {
```
alexandrosfilios, line 203: do you need to name the output var?

token/services/db/sql/querybuilder.go
```diff
@@ -202,6 +200,25 @@ func in(args *[]interface{}, field string, searchFor []interface{}) (where strin
...
+	case 1:
+		*args = append(*args, ids[0].TxId, ids[0].Index)
+		l := len(*args)
+		return fmt.Sprintf("tx_id = %s AND idx = %s", fmt.Sprintf("$%d", l-1), fmt.Sprintf("$%d", l))
+	default:
+		in := make([]string, len(ids))
+		for i, id := range ids {
+			*args = append(*args, id.TxId, id.Index)
+			l := len(*args)
+			in[i] = fmt.Sprintf("(%s, %s)", fmt.Sprintf("$%d", l-1), fmt.Sprintf("$%d", l))
```
alexandrosfilios, line 216: This could also be
```
fmt.Sprintf("($%d, $%d)", l-1, l)
```

token/services/db/sql/querybuilder.go
```diff
@@ -202,6 +200,25 @@ func in(args *[]interface{}, field string, searchFor []interface{}) (where strin
 	return fmt.Sprintf("(%s)", strings.Join(argnum, " OR "))
 }
 
+func whereTokenIDs(args *[]interface{}, ids []*token.ID) (where string) {
+	switch len(ids) {
+	case 0:
+		return ""
+	case 1:
```
alexandrosfilios, line 207: Do we need to handle this as a special case? These produce the same execution plan: `tx_id = ...` versus `tx_id IN (...)` for one value. (Do not know about `tx_id = ... AND idx = ...` versus `(tx_id, idx) = (..., ...)` though

alexandrosfilios (approved):

token/services/vault/db/vault.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (v *Vault) CertificationStorage() certification.Storage {
+	return v.certificationStorage
+}
+
+func (v *Vault) DeleteTokens(ids ...*token.ID) error {
+	return v.tokenDB.DeleteTokens(ids...)
+}
+
+type QueryEngine struct {
```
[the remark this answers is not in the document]
adecaro: 👍

token/services/db/sql/querybuilder.go
adecaro: not sure, honestly. We can handle this in another PR

token/services/db/sql/querybuilder.go
adecaro: 👍

token/services/db/sql/init.go
adecaro: isn't this memory that we waste though? We invoke getTableNames only few times.

PR #548, opened 2024-02-21
Bullet proofs: implementation + integration in the token-sdk.

KElkhiyaoui:
This PR implements the range proofs in the token-sdk using Bulletproofs.


adecaro (approved):
LGTM

token/core/zkatdlog/crypto/common/array.go
```diff
@@ -33,6 +35,7 @@ func (a *G1Array) Bytes() ([]byte, error) {
 			return nil, errors.Errorf("failed to marshal array of G1")
 		}
 		bytes := e.Bytes()
+		raw = append(raw, []byte(Seperator)...)
```
yacovm: General comment - when you consecutively append to an empty slice like that, it's `O(n*log(n))` complexity, because each time the slice's capacity is not enough to hold the required bytes, so Go's runtime allocates a new array twice the size of the old array and copies over the bytes. So it's better to pre-allocate the `raw`'s capacity to be an upper bound on the cumulative size of the elements, and then iterate and append.

token/core/zkatdlog/crypto/common/array.go
```diff
@@ -10,6 +10,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+const Seperator = "%%"
+
```
yacovm, line 19: Not related to this PR, but, FYI - it's easier to just have a definition like:

```
type G1Array []*math.G1`
```

This makes it easier to initialize and it can be easily converted from and to `[]*math.G1`.

token/core/zkatdlog/crypto/rp/bulletproof.go
```diff
@@ -0,0 +1,422 @@
...
+		left = append(left, p.Curve.NewZrFromInt(int64(b)))
+		// this is an array of b_i - 1
+		right = append(right, p.Curve.ModSub(left[i], p.Curve.NewZrFromInt(1), p.Curve.GroupOrder))
+		// these are randomly generated arrays
+		randomLeft = append(randomLeft, p.Curve.NewRandomZr(rand))
+		randomRight = append(randomRight, p.Curve.NewRandomZr(rand))
+	}
+
+	// C commits to L = (b_0, ..., b_{p.BitLength}) and R = (b_0 - 1 , ..., b_{p.BitLength} - 1)
+	C := commitVector(left, right, p.LeftGenerators, p.RightGenerators, p.Curve)
```
yacovm, line 290: `C`,`D` correspond to `A`,`S` in the `BP` paper and `P` is `h`?

token/core/zkatdlog/crypto/rp/bulletproof.go
```diff
@@ -0,0 +1,422 @@
...
+	D := commitVector(randomLeft, randomRight, p.LeftGenerators, p.RightGenerators, p.Curve)
+	// D is a hiding commitment thanks to eta
+	D.Add(p.P.Mul(eta))
+
+	array := common.GetG1Array([]*math.G1{C, D})
+	// compute challenges y and z
+	bytesToHash, err := array.Bytes()
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
```
yacovm, line 304: I noticed you don't hash the public parameters here as you did in other places. Is there a reason?

token/core/zkatdlog/crypto/rp/bulletproof.go
```diff
@@ -0,0 +1,422 @@
...
+	zPrime := make([]*math.Zr, len(left))
+	// z^2
+	zSquare := z.PowMod(p.Curve.NewZrFromInt(2))
+
+	for i := 0; i < len(left); i++ {
+		// compute L_i - z
+		leftPrime[i] = p.Curve.ModSub(left[i], z, p.Curve.GroupOrder)
+		// compute R_i + z
+		rightPrime[i] = p.Curve.ModAdd(right[i], z, p.Curve.GroupOrder)
+		// compute y^i
```
yacovm, line 322: nit: you can take the previous y^i and just multiply to get the current one, probably faster?

token/core/zkatdlog/crypto/rp/bulletproof.go
```diff
@@ -0,0 +1,422 @@
...
+	tau2 := p.Curve.NewRandomZr(rand)
+	T2 := p.CommitmentGenerators[0].Mul(t2)
+	T2.Add(p.CommitmentGenerators[1].Mul(tau2))
+
+	// compute challenge x
+	array = common.GetG1Array([]*math.G1{T1, T2})
+	bytesToHash, err = array.Bytes()
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
```
yacovm, line 355: maybe hash also public parameters?

token/core/zkatdlog/crypto/rp/bulletproof_test.go
```diff
@@ -0,0 +1,56 @@
...
+			It("Succeeds", func() {
+				curve := math.Curves[1]
+				nr := 3
+				l := int(math2.Pow(2, float64(nr)))
+				leftGens := make([]*math.G1, l)
+				rightGens := make([]*math.G1, l)
+
+				rand, err := curve.Rand()
+				Expect(err).NotTo(HaveOccurred())
+
```
yacovm, line 32: why are they all the same?

token/core/zkatdlog/crypto/rp/bulletproof.go
```diff
@@ -0,0 +1,422 @@
...
+	// Commitment is a hiding Pedersen commitment to value: Commitment = G^vH^r
+	Commitment *math.G1
+	// CommitmentGenerators are the generators (G, H) used to compute Commitment
+	CommitmentGenerators []*math.G1
+	// LeftGenerators are the generators that will be used to commit to
+	// the bits (b_0,..., b_{BitLength-1}) of value
+	LeftGenerators []*math.G1
+	// RightGenerators are the generators that will be used to commit to (b_i-1)
+	RightGenerators []*math.G1
+	// P is a random generator of G1
```
yacovm, line 54: I noticed that these public parameters are all injected through `setup.go`.
What's the reason for this? Why not instantiate them in this package instead globally?

token/core/zkatdlog/crypto/setup.go
```diff
@@ -193,19 +204,24 @@ func (pp *PublicParams) GeneratePedersenParameters() error {
...
+		P:              curve.HashToG1([]byte(strconv.Itoa(0))),
+		Q:              curve.HashToG1([]byte(strconv.Itoa(1))),
+		BitLength:      bitLength,
+		NumberOfRounds: int(math.Log2(float64(bitLength))),
+	}
+	pp.RangeProofParams.LeftGenerators = make([]*mathlib.G1, bitLength)
+	pp.RangeProofParams.RightGenerators = make([]*mathlib.G1, bitLength)
+
+	for i := 0; i < bitLength; i++ {
+		pp.RangeProofParams.LeftGenerators[i] = curve.HashToG1([]byte(strconv.Itoa(2 * (i + 1))))
```
yacovm, line 223: Maybe it's worth to add some prefix "range proof" to the hash's input, otherwise in other protocols they would be the same generators?

token/core/zkatdlog/crypto/rp/bulletproof.go
```diff
@@ -0,0 +1,422 @@
...
+	bytesToHash, err = array.Bytes()
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+	x := p.Curve.HashToZr(bytesToHash)
+
+	// compute vectors left and right against which an IPA will be produced
+	// if p.Value is within the authorized range, then L_iR_i =0 and L_i-R_i-1 = 0
+	// the inner product <left, right> = p.Value*z^2+t1x+t2x^2+f(z, y)
+	// f(z, y) = \sum (z-z^2)*y^i + z^3*2^i
```
yacovm, line 361: in the paper `f(z, y) = \sum (z-z^2)*y^i + z^3*2^i` is `f(z, y) = \sum (z-z^2)*y^i - z^3*2^i`, why the `+`?  What am I missing?

token/core/zkatdlog/crypto/rp/ipa.go
```diff
@@ -0,0 +1,317 @@
...
+// value c is the inner product of two committed vectors a = (a_1, ..., a_n)
+// (leftVector) and b = (b_1, ..., b_n) (rightVector)
+type ipaProver struct {
+	// rightVector is one of the committed vectors in Commitment
+	rightVector []*mathlib.Zr
+	// leftVector is one of the committed vectors in Commitment
+	leftVector []*mathlib.Zr
+	// InnerProduct is the inner product of leftVector and rightVector
+	InnerProduct *mathlib.Zr
+	// Q is a random generators of G1
```
yacovm, line 42: `Q` is `u` right?

token/core/zkatdlog/crypto/rp/bulletproof.go
```diff
@@ -0,0 +1,422 @@
...
+	var left, right []*math.Zr
+	var randomLeft, randomRight []*math.Zr
+	rand, err := p.Curve.Rand()
+	if err != nil {
+		return nil, nil, nil, nil, err
+	}
+	rho := p.Curve.NewRandomZr(rand)
+	eta := p.Curve.NewRandomZr(rand)
+	for i := 0; i < int(p.BitLength); i++ {
+		b := 1 << uint(i) & p.value
```
yacovm, line 276: instead of:
```
b := 1 << uint(i) & p.value
		if b > 0 {
			b = 1
		}
```

Maybe you can do: `(p.value >> i) & 1 ?

KElkhiyaoui: I will fix this on another branch dedicated to cleaning up the code

token/core/zkatdlog/crypto/rp/bulletproof.go
KElkhiyaoui: Yes.

token/core/zkatdlog/crypto/rp/bulletproof.go
KElkhiyaoui: There is no need to hash the public parameters.  
However, what's missing here is the commitment to the value which must be added. 
In the branch dedicated to cleaning up the code including the public parameters in the hash will be removed.

token/core/zkatdlog/crypto/rp/bulletproof.go
KElkhiyaoui: OK

token/core/zkatdlog/crypto/rp/bulletproof.go
KElkhiyaoui: This is a typo. If you look at the verifier code what's implemented it is what's in the paper.

token/core/zkatdlog/crypto/rp/bulletproof_test.go
KElkhiyaoui: Not clear what this comment refers to?

token/core/zkatdlog/crypto/rp/bulletproof.go
KElkhiyaoui: This package is used within the token library where all the public parameters are produced. As of now it is not used in a standalone capacity. If the need arises, this will be changed accordingly.

token/core/zkatdlog/crypto/rp/ipa.go
KElkhiyaoui: Yes.

token/core/zkatdlog/crypto/setup.go
KElkhiyaoui: OK.

token/core/zkatdlog/crypto/common/array.go
KElkhiyaoui: I will address this in the PR dedicated to cleaning up the code.

token/core/zkatdlog/crypto/common/array.go
KElkhiyaoui: I will address this in a PR dedicated to cleaning up the code.

token/core/zkatdlog/crypto/rp/bulletproof_test.go
yacovm: you selected all points Q,P,H,G with the same exponent

yacovm (approved):

PR #551, opened 2024-02-23
simplified token processor and db

adecaro:
This PR simplify the `TokenStore` interface unifying the function to append new tokens. As a byproduct, this PR also reworks the data model used by the tokendb's sql implementation.


token/sdk/tms/db/tms.go
```diff
@@ -137,9 +146,18 @@ func (p *PostInitializer) ConnectNetwork(networkID, channel, namespace string) e
...
 		namespace,
 		fabric2.NewTokenRWSetProcessor(
-			n,
+			n.Name(),
 			namespace,
-			p.sp,
+			func() *token3.ManagementServiceProvider {
+				return token3.GetManagementServiceProvider(p.sp)
+			},
+			func(tms *token3.ManagementService, txID string) ([]byte, error) {
```
alexandrosfilios, line 154: I may be missing something, but I think we could create a function and reuse it for both constructors.

token/services/db/sql/tokens.go
```diff
@@ -884,43 +837,13 @@ func (db *TokenDB) GetSchema() string {
 			is_deleted BOOL NOT NULL DEFAULT false,
 			spent_by TEXT NOT NULL DEFAULT '',
 			spent_at TIMESTAMP,
+			owner BOOL NOT NULL DEFAULT false,
+			auditor BOOL NOT NULL DEFAULT false,
+			issuer BOOL NOT NULL DEFAULT false,
 			PRIMARY KEY (tx_id, idx)
 		);
 		CREATE INDEX IF NOT EXISTS idx_spent_%s ON %s ( is_deleted );
```
alexandrosfilios, line 845: Maybe you want to include owner in the index.

Also you may want to try CREATE BITMAP INDEX instead of index. In theory it should work well with such cases.

token/services/network/fabric/processor.go
```diff
@@ -174,6 +164,13 @@ func (r *RWSetProcessor) tokenRequest(req fabric.Request, tx fabric.ProcessTrans
 		}
 	}
 
+	auditorFlag := r.ownership.AmIAnAuditor(tms)
+	if auditorFlag {
+		if logger.IsEnabledFor(zapcore.DebugLevel) {
```
alexandrosfilios, line 169: I think you don't need this if, unless you have some calculation inside.

token/services/network/fabric/processor.go
```diff
@@ -256,37 +253,37 @@ func (r *RWSetProcessor) tokenRequest(req fabric.Request, tx fabric.ProcessTrans
 			continue
 		}
 
+		issuerFlag := !issuer.IsNone() && r.issued.Issued(tms, issuer, tok)
 		ids, mine := r.ownership.IsMine(tms, tok)
-		if mine {
-			if logger.IsEnabledFor(zapcore.DebugLevel) {
+		if logger.IsEnabledFor(zapcore.DebugLevel) {
```
alexandrosfilios, line 258: maybe not needed.

PR #553, opened 2024-02-26
add dependency to the hash of public params in the rws

adecaro:
This PR adds an additional dependency to the current public parameters in the RWSet. 
This is helpful every time one wants to update the public parameters without upgrading the chaincode.

### DIFF
--- a/token/services/vault/rws/keys/keys.go
+++ b/token/services/vault/rws/keys/keys.go
@@ -20,9 +20,10 @@ const (
 
 	numComponentsInKey = 2 // 2 components: txid, index, excluding TokenKeyPrefix
 
-	TokenKeyPrefix      = "ztoken"
-	TokenMineKeyPrefix  = "mine"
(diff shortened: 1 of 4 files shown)


token/services/vault/rws/keys/keys.go
```diff
@@ -60,6 +61,10 @@ func CreateSetupKey() (string, error) {
 	return CreateCompositeKey(TokenKeyPrefix, []string{TokenSetupKeyPrefix})
 }
 
+func CreateSetupHashKey() (string, error) {
```
alexandrosfilios, line 64: This could be a constant, right?

alexandrosfilios (approved):
LGTM

token/services/vault/rws/keys/keys.go
adecaro: indeed 👍

PR #554, opened 2024-02-27
tokens service

adecaro:
This PR introduces the `tokens` service used to update the `tokendb` starting from a `TokenRequest`. 
The fabric and orion processors will use this service to commit locally the transactions.

In the following PRs, we will address the following points:
- Atomicity of  `AppendTransaction`. If a failure happens, no traces should be left in the `tokendb`.
- Ability to remove all the records related to a given transaction
- Fix  the TODOs in `EndorseView`'s `receiveTransaction` function


token/sdk/tms/db/tms.go
```diff
@@ -134,28 +118,16 @@ func (p *PostInitializer) ConnectNetwork(networkID, channel, namespace string) e
...
-	tokenStore, err := db.NewTokenStore(notifier, tokenDB, tmsID)
-	if err != nil {
-		return errors.WithMessagef(err, "failed to get token store")
-	}
 	if err := n.ProcessorManager().AddProcessor(
 		namespace,
 		fabric2.NewTokenRWSetProcessor(
 			n.Name(),
 			namespace,
+			func() *tokens2.Tokens {
```
alexandrosfilios, line 126: I may be missing it, but if there is no other way of getting the tokens, then it would make more sense to pass the service directly

token/services/tokens/tokens.go
```diff
@@ -0,0 +1,193 @@
...
+		issuerFlag := !issuer.IsNone() && t.Issued.Issued(tms, issuer, tok)
+		ids, mine := t.Ownership.IsMine(tms, tok)
+		if logger.IsEnabledFor(zapcore.DebugLevel) {
+			if mine {
+				logger.Debugf("transaction [%s], found a token and it is mine", txID)
+			} else {
+				logger.Debugf("transaction [%s], found a token and it is NOT mine", txID)
+			}
+		}
+		if logger.IsEnabledFor(zapcore.DebugLevel) {
```
alexandrosfilios, line 151: I don't think we need this check if logger.IsEnabledFor if we don't do any calculations. But in any case we could at least put them all together.

token/sdk/tms/db/tms.go
adecaro: I just want to postone this call to when it will be really needed so to reduce resource allocation.
Wdyt?

token/services/tokens/tokens.go
adecaro: 👍

token/services/tokens/tokens.go
```diff
@@ -0,0 +1,193 @@
...
+	// parse the inputs
+	for _, input := range is.Inputs() {
+		if input.Id == nil {
+			logger.Infof("transaction [%s] found an input that is not mine, skip it", tx.ID())
+			continue
+		}
+		if logger.IsEnabledFor(zapcore.DebugLevel) {
+			logger.Debugf("transaction [%s] delete input [%s]", tx.ID(), input.Id)
+		}
+		if err := t.TokenStore.DeleteToken(input.Id.TxId, input.Id.Index, tx.ID()); err != nil {
```
alexandrosfilios, line 113: maybe for better rollback it would make sense to collect all tokens to be deleted and do them in one transaction at the end?

adecaro: yes, I mentioned this in the description of the PR as a point to address in a subsequent PR 👍

token/sdk/tms/db/tms.go
alexandrosfilios: Then you will call this method every time you need the tokens. We could use some lazy initialisation to combine th best of both worlds. We could have a generic solution like
```
type lazyGetter[V] struct {
v V
provider func() V
once sync.Once 
}

func NewLazyGetter[V](provider func() V) *lazyGetter[V] {
 return &lazyGetter[V]{provider: provider}
}

func (g *lazyGetter[V]) Get() V {
 g.once.Do(func() {
  g.v = g.provider()
 })
 return g.v
}
```

alexandrosfilios (approved):
LGTM modulo comment

token/sdk/tms/db/tms.go
adecaro: 👍

PR #555, opened 2024-02-28
move owner into ttx

adecaro:
This PR moves the owner service inside the ttx service. This is in alignment with `audit` and `auditdb` services, and `tokens` and `tokendb` services.

### DIFF
--- a/token/services/ttx/manager.go
+++ b/token/services/ttx/manager.go
@@ -25,20 +25,20 @@ type DBProvider interface {
 // Manager handles the databases
 type Manager struct {
 	networkProvider NetworkProvider
-	ttxdbProvider   DBProvider
+	dbProvider      DBProvider
 
 	storage storage.DBEntriesStorage
 	mutex   sync.Mutex
-	owners  map[string]*DB
+	dbs     map[string]*DB
 }
 
 // NewManager creates a new DB manager.
-func NewManager(np NetworkProvider, ttxdbManager DBProvider, storage storage.DBEntriesStorage) *Manager {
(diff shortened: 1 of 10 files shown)


token/services/ttx/manager.go
```diff
@@ -18,32 +18,32 @@ import (
...
-// NewManager creates a new Owner manager.
-func NewManager(np NetworkProvider, ttxdbManager TTXDBProvider, storage storage.DBEntriesStorage) *Manager {
+// NewManager creates a new DB manager.
+func NewManager(np NetworkProvider, ttxdbManager DBProvider, storage storage.DBEntriesStorage) *Manager {
 	return &Manager{
 		networkProvider: np,
 		storage:         storage,
 		ttxdbProvider:   ttxdbManager,
-		owners:          map[string]*Owner{},
+		owners:          map[string]*DB{},
```
alexandrosfilios, line 41: Maybe we can rename this field to txDbs or something similar.

token/services/ttx/manager.go
```diff
@@ -65,12 +65,12 @@ func (cm *Manager) Owner(tmsID token.TMSID) (*Owner, error) {
 	return c, nil
 }
 
-func (cm *Manager) newOwner(tmsID token.TMSID) (*Owner, error) {
+func (cm *Manager) newOwner(tmsID token.TMSID) (*DB, error) {
```
alexandrosfilios, line 68: and this one to newDB.

alexandrosfilios (approved):
LGTM

token/services/ttx/manager.go
adecaro: 👍

PR #557, opened 2024-02-28
`tokendb` and `tokens`'s support for transactions

adecaro:


token/services/db/sql/tokens.go
```diff
@@ -893,6 +879,136 @@ func (db *TokenDB) Close() {
...
+	return NewTokenTransaction(db, tx), nil
+}
+
+type TokenTransaction struct {
+	db *TokenDB
+	tx *sql.Tx
+}
+
+func (t *TokenTransaction) TransactionExists(id string) (bool, error) {
+	query := fmt.Sprintf("SELECT tx_id FROM %s WHERE tx_id=$1;", t.db.table.Tokens)
```
alexandrosfilios, line 896: Although you are filtering on an indexed field, you could add LIMIT 1 in your result, just in case you think there might be many rows affected.

token/services/tokens/tokens.go
```diff
@@ -66,24 +69,45 @@ func (t *Tokens) AppendTransaction(tx Transaction) error {
...
+	}
 	if tms.PublicParametersManager().PublicParameters().GraphHiding() {
-		ids := metadata.SpentTokenID()
+		ids := md.SpentTokenID()
 		if logger.IsEnabledFor(zapcore.DebugLevel) {
 			logger.Debugf("transaction [%s] with graph hiding, delete inputs [%v]", txID, ids)
 		}
 		for _, id := range ids {
-			if err := t.TokenStore.DeleteToken(id.TxId, id.Index, txID); err != nil {
+			if err := ts.DeleteToken(id.TxId, id.Index, txID); err != nil {
```
alexandrosfilios, line 110: I think we could have an optimisation in the query if we had a method `TokenStore.DeleteTokens`.

token/services/db/sql/tokens.go
```diff
@@ -150,13 +128,20 @@ func (db *TokenDB) OwnersOf(txID string, index uint64) (*token.Token, []string,
...
-	now := time.Now().UTC()
-	query := fmt.Sprintf("UPDATE %s SET is_deleted = true, spent_by = $1, spent_at = $2 WHERE tx_id = $3 AND idx = $4;", db.table.Tokens)
-	logger.Debug(query, deletedBy, now, txID, index)
-	if _, err := db.db.Exec(query, deletedBy, now, txID, index); err != nil {
-		return errors.Wrapf(err, "error setting token to deleted [%s]", txID)
+	tx, err := db.NewTokenDBTransaction()
+	if err != nil {
+		return err
+	}
+	defer func() {
```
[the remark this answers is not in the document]
alexandrosfilios: Is it because if we have committed, the rollback won't take effect? If so, we can still spare the cycles of calling rollback by making a check whether `err` is set or something.

token/sdk/vault/db/vault.go
```diff
@@ -74,7 +69,7 @@ func (v *VaultProvider) Vault(network string, channel string, namespace string)
 	if err != nil {
 		return nil, errors.WithMessagef(err, "failed to get token db")
 	}
-	tokenStore, err := tokens.NewTokenStore(notifier, tokenDB, tmsID)
+	tokens, err := tokens.Get(v.sp, tmsID)
```
alexandrosfilios, line 72: I think tokenStore helps more to give context. and `tokens.GetStore` or something like this.

token/sdk/common/lazy.go
```diff
@@ -10,17 +10,18 @@ import "sync"
...
 }
 
-func NewLazyGetter[V any](provider func() V) *LazyGetter[V] {
+func NewLazyGetter[V any](provider func() (V, error)) *LazyGetter[V] {
 	return &LazyGetter[V]{provider: provider}
 }
 
-func (g *LazyGetter[V]) Get() V {
+func (g *LazyGetter[V]) Get() (V, error) {
 	g.once.Do(func() {
```
alexandrosfilios, line 23: If you think that the consumer of this getter is thread safe, you can use an `if` instead of `once`.

alexandrosfilios (approved):
LGTM modulo comments

token/sdk/vault/db/vault.go
adecaro: this is the token service, though. Shall I call GetService?

token/services/tokens/tokens.go
adecaro: okay

token/sdk/vault/db/vault.go
alexandrosfilios: yes that would make more sense (and be more consistent with other similar cases we have in the code)

token/services/tokens/tokens.go
```diff
@@ -66,26 +69,45 @@ func (t *Tokens) AppendTransaction(tx Transaction) error {
...
 		logger.Debugf("transaction [%s], no metadata found, skip it", txID)
 		return nil
 	}
-	metadata, err := request.GetMetadata()
+
+	ts, err := t.Storage.NewTransaction()
+	if err != nil {
+		return err
+	}
+	defer func() {
```
alexandrosfilios, line 85: We still have an unconditional rollback.

token/services/tokens/tokens.go
```diff
@@ -183,9 +205,39 @@ func (t *Tokens) AppendTransaction(tx Transaction) error {
...
+}
+
+// DeleteToken marks the entries corresponding to the passed token ids as deleted.
+// The deletion is attributed to the passed deletedBy argument.
+func (t *Tokens) DeleteToken(deletedBy string, ids ...*token2.ID) error {
+	ts, err := t.Storage.NewTransaction()
+	if err != nil {
+		return err
+	}
+	defer func() {
```
alexandrosfilios, line 226: Unconditional rollback

token/services/db/sql/tokens.go
adecaro: 👍

token/sdk/vault/db/vault.go
adecaro: 👍

PR #560, opened 2024-03-01
network service cleanup

adecaro:
This PR proposes the following:
- remove transient-related function from the network service, they are not needed anymore
- cleanup sql db logs

### DIFF
--- a/token/services/db/sql/init.go
+++ b/token/services/db/sql/init.go
@@ -20,28 +20,30 @@ import (
 
 var logger = flogging.MustGetLogger("token-sdk.sql")
 
-func initSchema(db *sql.DB, schemas ...string) error {
+func initSchema(db *sql.DB, schemas ...string) (err error) {
 	logger.Info("creating tables")
 	tx, err := db.Begin()
 	if err != nil {
 		return err
 	}
 	defer func() {
-		if err := tx.Rollback(); err != nil {
-			logger.Errorf("failed to rollback [%s][%s]", err, debug.Stack())
+		if err != nil && tx != nil {
+			if err := tx.Rollback(); err != nil {
+				logger.Errorf("failed to rollback [%s][%s]", err, debug.Stack())
+			}
 		}
 	}()
 
 	for _, schema := range schemas {
 		logger.Debug(schema)
-		if _, err := db.Exec(schema); err != nil {
+		if _, err = db.Exec(schema); err != nil {
 			return errors.Wrap(err, "error creating schema")
 		}
 	}
 	if err = tx.Commit(); err != nil {
 		return err
 	}
-	return nil
+	return
 }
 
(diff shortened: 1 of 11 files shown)


token/services/db/sql/init.go
```diff
@@ -20,28 +20,30 @@ import (
 
 var logger = flogging.MustGetLogger("token-sdk.sql")
 
-func initSchema(db *sql.DB, schemas ...string) error {
+func initSchema(db *sql.DB, schemas ...string) (err error) {
```
alexandrosfilios, line 23: Since you always return the error, you don't have to declare the result var in the signature.

token/services/db/sql/transactions.go
```diff
@@ -243,33 +243,35 @@ func (db *TransactionDB) AddTransaction(r *driver.TransactionRecord) error {
 	return err
 }
 
-func (db *TransactionDB) SetStatus(txID string, status driver.TxStatus) error {
+func (db *TransactionDB) SetStatus(txID string, status driver.TxStatus) (err error) {
```
alexandrosfilios, line 246: Same here for err.

token/services/db/sql/transactions.go
```diff
@@ -354,7 +356,7 @@ func (db *TransactionDB) QueryValidations(params driver.QueryValidationRecordsPa
 	return &ValidationRecordsIterator{txs: rows, filter: params.Filter}, nil
 }
 
-func (db *TransactionDB) AddTransactionEndorsementAck(txID string, endorser view.Identity, sigma []byte) error {
+func (db *TransactionDB) AddTransactionEndorsementAck(txID string, endorser view.Identity, sigma []byte) (err error) {
```
alexandrosfilios, line 359: Same here

token/services/ttx/info.go
```diff
@@ -46,37 +46,39 @@ func (a *TransactionInfoProvider) TransactionInfo(txID string) (*TransactionInfo
...
 
-func (a *TransactionInfoProvider) loadTransient(txID string) (map[string][]byte, error) {
-	tm, err := network.GetInstance(a.sp, a.tms.Network(), a.tms.Channel()).GetTransient(txID)
-	if err != nil {
-		return nil, errors.WithMessagef(err, "failed to load transient for [%s]", txID)
-	}
-	if !tm.Exists(TokenRequestMetadata) {
+func (a *TransactionInfoProvider) loadTransient(trRaw []byte, txID string) (map[string][]byte, error) {
+	if len(trRaw) == 0 {
+		if logger.IsEnabledFor(zapcore.DebugLevel) {
```
alexandrosfilios, line 67: You don't have to make the check isEnabledFor unless you have any calculations to do.

alexandrosfilios (approved):
LGTM modulo comments

PR #562, opened 2024-03-03
identity types

adecaro:
This PR does the following: replace the `SerializedIdentity` type with more specific identity types.
This allows a driver to support multiple owner identity types


alexandrosfilios (approved):
LGTM

PR #563, opened 2024-03-03
remove suffix, use prefixes consistently, only rollback on error

arner:
We don't need the tablesuffix because the db is configured by the user per TMS. Even if they choose to use the same database for multiple TMS, they can use the prefix to have different tables per TMS.


adecaro (approved):
LGTM module the rebase

PR #564, opened 2024-03-05
identity stack cleanup

adecaro:
This PR cleans up the identity stack by cutting the dependency from the FSC's sig and deserialisation service.


token/services/db/sql/identity.go
```diff
@@ -48,25 +67,139 @@ func NewIdentityDB(db *sql.DB, tablePrefix, name string, createSchema bool) (*Id
...
 
-	_, err := db.db.Exec(query, wp.ID, wp.URL)
+	_, err := db.db.Exec(query, wp.ID, wp.Type, wp.URL)
 	return err
 }
 
-func (db *IdentityDB) IteratorConfigurations() (driver.Iterator[driver.IdentityConfiguration], error) {
-	query := fmt.Sprintf("SELECT identity_id, url FROM %s", db.table.Identities)
+func (db *IdentityDB) IteratorConfigurations(configurationType string) (driver.Iterator[driver.IdentityConfiguration], error) {
+	query := fmt.Sprintf("SELECT identity_id, url FROM %s WHERE type = $1", db.table.IdentityConfigurations)
```
alexandrosfilios, line 78: I also wanted to add the type in the table so that we only have one table for all, but the queries never concern two different types. Are we sure that the table doesn't get too big? I suppose it's 4 times as big now.

token/services/db/sql/identity.go
```diff
@@ -79,17 +212,39 @@ func (w *WalletPathStorageIterator) HasNext() bool {
...
 	return c, err
 }
 
 func (db *IdentityDB) GetSchema() string {
 	return fmt.Sprintf(`
-		-- Identities
+		-- IdentityConfigurations
 		CREATE TABLE IF NOT EXISTS %s (
 			identity_id TEXT NOT NULL PRIMARY KEY,
+            type TEXT NOT NULL,
```
alexandrosfilios, line 225: Maybe an index on the type might speed up a bit the queries.

token/services/identity/msp/idemix/lm.go
```diff
@@ -155,7 +156,11 @@ func (lm *LocalMembership) RegisterIdentity(id string, path string) error {
 	lm.resolversMutex.Lock()
 	defer lm.resolversMutex.Unlock()
 
-	if err := lm.storage.AddConfiguration(driver3.IdentityConfiguration{ID: id, URL: path}); err != nil {
+	if err := lm.storage.AddConfiguration(driver3.IdentityConfiguration{
+		ID:   id,
+		Type: "idemix",
```
alexandrosfilios, line 161: We could use an enum for the type since it can only be x509 or idemix. Same for the DB.

token/services/identity/registry.go
```diff
@@ -45,6 +45,9 @@ func NewWalletsRegistry(identityProvider driver.IdentityProvider, identityRole d
 // If no wallet is found, Lookup returns the identity info and a potential wallet identifier for the passed id.
 // The identity info can be nil meaning that nothing has been found bound to the passed identifier
 func (r *WalletsRegistry) Lookup(id interface{}) (driver.Wallet, driver.IdentityInfo, string, error) {
+	if logger.IsEnabledFor(zapcore.DebugLevel) {
```
alexandrosfilios, line 48: No need for if

alexandrosfilios (approved):
LGTM modulo comments

token/services/db/sql/identity.go
adecaro: I think it is okay, this table will not have too many entries.

token/services/db/sql/identity.go
adecaro: will do.

token/services/identity/msp/idemix/lm.go
adecaro: will do

token/services/identity/registry.go
adecaro: printing id can be heavy

PR #568, opened 2024-03-10
Postgres compatibility

arner:
This PR improves the compatibility with postgres by fixing some bugs encountered while running a token sdk application:

-`identity_id` has non-utf8 characters which is not supported in a `TEXT` column by postgres.
- `raw` cannot be the primary key for params if the params are too large. This PR changes the primary key to the time stored in the database, which is also unique.
- postgres does not have `ON CONFLICT REPLACE` on columns
- the 'restore auditor db' feature was doing writes while iterating over the dbentries, which causes a deadlock with a single threaded KVS.


adecaro (approved):
LGTM

PR #574, opened 2024-03-14
introduce a common wallet service

adecaro:


token/core/common/wallet.go
```diff
@@ -0,0 +1,278 @@
...
+	return s.auditorWallet(id)
+}
+
+func (s *WalletService) AuditorWalletByIdentity(identity view.Identity) (driver.AuditorWallet, error) {
+	return s.auditorWallet(identity)
+}
+
+func (s *WalletService) auditorWallet(id interface{}) (driver.AuditorWallet, error) {
+	s.Logger.Debugf("get auditor wallet for [%v]", id)
+	s.AuditorWalletsRegistry.Lock()
```
alexandrosfilios, line 219: In the first case we only read so we could probably just acquire a Rock

token/core/fabtoken/wallet.go
```diff
@@ -346,12 +177,12 @@ func (w *ownerWallet) Remote() bool {
 }
 
 type issuerWallet struct {
-	tokenService *WalletService
+	tokenService *common.WalletService
```
alexandrosfilios, line 180: Passing a whole service into the wallet adds a dependency that would maybe better be avoided.
I checked why we need to pass the wallet service, and if I saw correctly, it's only because of `nogh.Service.Issue()` that calls `w.GetSigner()`. If this is the case, we could pass the walletService into `nogh.Service` directly and avoid passing the walletService altogether.
But it might be that it's more complex than that.

adecaro:
@alexandrosfilios , I did all the changes you suggested. Please, have a look again. Thanks much :)

token/core/common/wallet.go
```diff
@@ -0,0 +1,225 @@
...
+
+func (s *WalletService) GetRevocationHandler(auditInfo []byte) (string, error) {
+	return s.IdentityProvider.GetRevocationHandler(auditInfo)
+}
+
+func (s *WalletService) RegisterRecipientIdentity(data *driver.RecipientData) error {
+	if data == nil {
+		return errors.Errorf("cannot recigest empty recipient")
+	}
+	s.Logger.Debugf("register recipient identity [%s] with audit info [%s]", data.Identity.String(), hash.Hashable(data.AuditInfo).String())
```
alexandrosfilios, line 106: If you define `%s` it is going to call `String()` lazily. This way, it is going to call it anyways, so maybe it's better to leave `.String()` out.

token/core/common/wallet.go
```diff
@@ -0,0 +1,225 @@
...
+	return s.IdentityProvider.GetEnrollmentID(auditInfo)
+}
+
+func (s *WalletService) GetRevocationHandler(auditInfo []byte) (string, error) {
+	return s.IdentityProvider.GetRevocationHandler(auditInfo)
+}
+
+func (s *WalletService) RegisterRecipientIdentity(data *driver.RecipientData) error {
+	if data == nil {
+		return errors.Errorf("cannot recigest empty recipient")
```
alexandrosfilios, line 104: Not for performance (it's a negligible gain), but maybe as aroule we could define such errors as "constants"

adecaro: the issue here is that `errors.Errorf` records also the stack trace. I can create an error and then I will have to wrap it there to add the stack trace. Would that be okay?

token/core/common/wallet.go
adecaro: 👍

token/core/common/wallet.go
```diff
@@ -0,0 +1,225 @@
...
+
+func (s *WalletService) IssuerWallet(id any) (driver.IssuerWallet, error) {
+	w, err := s.walletByID(driver.IssuerRole, id)
+	if err != nil {
+		return nil, err
+	}
+	return w.(driver.IssuerWallet), nil
+}
+
+func (s *WalletService) AuditorWallet(id any) (driver.AuditorWallet, error) {
```
alexandrosfilios, line 167: Not really important but you could skip the repetition of these lines if you change the signature of `walletByID`:

```
func walletByID[T driver.Wallet](entry RegistryEntry, id any) (T, error)
```
Then all your methods above would become like this:
```
func (s *WalletService) AuditorWallet(id any) (driver.AuditorWallet, error) {
 return walletByID[driver.AuditorWallet](s.Registries[driver.AuditorRole], id)
}
```

token/core/fabtoken/driver/driver.go
```diff
@@ -111,21 +114,26 @@ func (d *Driver) NewTokenService(sp driver.ServiceProvider, networkID string, ch
...
+	publicParamsManager := fabtoken.NewPublicParamsManager(
+		fabtoken.PublicParameters,
 		qe,
-		NewDeserializer(),
+	)
+	deserializer := NewDeserializer()
+	ws := common.NewWalletService(
+		logger,
+		ip,
+		func() (driver.Deserializer, error) {
```
alexandrosfilios, line 125: If you think that this getter might not be used, you could use `NewLazyGetter(NewDeserializer)`

token/core/common/wallet.go
adecaro: but then I have to pass also `s.WalletFactory`. I thought about it, but it gets ugly in other ways. No?

token/core/fabtoken/driver/driver.go
adecaro: if you are okay, I'll change this in a later PR where I'll extract common parts of the token service.

token/core/zkatdlog/nogh/driver/driver.go
```diff
@@ -122,23 +128,24 @@ func (d *Driver) NewTokenService(sp driver.ServiceProvider, networkID string, ch
...
+		nil,
 	)
 	service, err := zkatdlog.NewTokenService(
 		ws,
 		ppm,
 		&zkatdlog.VaultTokenLoader{TokenVault: qe},
 		zkatdlog.NewVaultTokenCommitmentLoader(qe, 3, 3*time.Second),
 		ip,
-		NewDeserializerProvider().Deserialize,
+		NewDeserializerProvider(ppm).Deserialize,
```
alexandrosfilios, line 148: You may want to use the `deserializerProvider` you declared above (I suppose they are thread safe).

token/core/fabtoken/wallet.go
```diff
@@ -7,251 +7,61 @@ SPDX-License-Identifier: Apache-2.0
...
 }
 
-func newOwnerWallet(tokenService *WalletService, identity view.Identity, id string, identityInfo driver.IdentityInfo) *ownerWallet {
+func newOwnerWallet(IdentityProvider driver.IdentityProvider, TokenVault TokenVault, identity view.Identity, id string, identityInfo driver.IdentityInfo) *ownerWallet {
 	return &ownerWallet{
-		tokenService: tokenService,
-		id:           id,
-		identity:     identity,
-		identityInfo: identityInfo,
+		IdentityProvider: IdentityProvider,
```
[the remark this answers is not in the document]
adecaro: 👍

token/core/zkatdlog/nogh/driver/driver.go
adecaro: 👍

token/driver/wallet.go
```diff
@@ -139,31 +139,19 @@ type WalletService interface {
...
-	IssuerWalletByIdentity(identity view.Identity) (IssuerWallet, error)
-
-	// AuditorWalletByIdentity returns an instance of the AuditorWallet interface that contains the passed identity.
-	AuditorWalletByIdentity(identity view.Identity) (AuditorWallet, error)
+	IssuerWallet(id any) (IssuerWallet, error)
 
 	// AuditorWallet returns an instance of the AuditorWallet interface bound to the passed id.
 	// The id can be: the wallet identifier or a unique id of a view identity belonging to the wallet.
-	AuditorWallet(id string) (AuditorWallet, error)
+	AuditorWallet(id any) (AuditorWallet, error)
```
alexandrosfilios, line 150: Maybe it would be better to create another type instead of any, even if it is just an alias. So that we know what kind of `any` we expect to find.

token/core/fabtoken/wallet.go
```diff
@@ -7,251 +7,61 @@ SPDX-License-Identifier: Apache-2.0
...
-	if w != nil {
-		return w
+	var newWallet driver.Wallet
+	switch role {
+	case driver.OwnerRole:
+		newWallet = newOwnerWallet(w.IdentityProvider, w.TokenVault, idInfoIdentity, id, info)
+	case driver.IssuerRole:
+		newWallet = newIssuerWallet(w.IdentityProvider, w.TokenVault, id, idInfoIdentity)
+	case driver.AuditorRole:
+		newWallet = newAuditorWallet(w.IdentityProvider, id, idInfoIdentity)
```
alexandrosfilios: not sure if it makes sense, but maybe for the sake of completion/clarity I would add a case for driver.CertifierRole and panic because it's not supported. Because it's not clear to everyone why it's left out.

token/core/common/wallet.go
alexandrosfilios: aaah sorry missed that one!

token/core/common/wallet.go
alexandrosfilios: aaah ok fair!

token/driver/wallet.go
alexandrosfilios: OK I can't find any way we can further restrict it or have some common functionality, but we could do the following:
* Try to replace also other points in code where this ID is passed (currently as `any` or `interface{}`, e.g. in `WalletRegistry.Lookup`, `Role.MapToID`)
* Write a comment above this type that the ID can be either `string` or `view.Identity`

adecaro: no, because it is the driver that must decide.

alexandrosfilios (approved):
LGTM

PR #577, opened 2024-03-18
extract common structs among drivers

adecaro:


token/core/common/loaders.go
```diff
@@ -88,45 +98,48 @@ func (s *VaultTokenCommitmentLoader) GetTokenOutputs(ids []*token3.ID) ([]*token
...
 // tokens in clear text and the identities of their owners
 // LoadToken returns an error in case of failure
-func (s *VaultTokenLoader) LoadTokens(ids []*token3.ID) ([]string, []*token.Token, []*token.Metadata, []view.Identity, error) {
-	var tokens []*token.Token
+func (s *VaultLedgerTokenAndMetadataLoader[T, M]) LoadTokens(ids []*token.ID) ([]string, []T, []M, []view.Identity, error) {
+	var tokens []T
 	var inputIDs []string
-	var inputInf []*token.Metadata
+	var inputInf []M
 	var signerIds []view.Identity
```
alexandrosfilios, line 118: the capacity of these arrays is not `len(ids)`?

token/core/common/ppm.go
```diff
@@ -0,0 +1,143 @@
...
+func NewPublicParamsManagerFromParams[T driver.PublicParameters](logger *flogging.FabricLogger, PublicParamsDeserializer PublicParamsDeserializer[T], pp T) (*PublicParamsManager[T], error) {
+	return &PublicParamsManager[T]{
+		Logger:                   logger,
+		publicParameters:         pp,
+		PublicParamsDeserializer: PublicParamsDeserializer,
+	}, nil
+}
+
+func (v *PublicParamsManager[T]) PublicParameters() driver.PublicParameters {
+	v.Mutex.RLock()
```
alexandrosfilios, line 67: do we need this lock?

token/core/common/ppm.go
```diff
@@ -0,0 +1,143 @@
...
+	defer v.Mutex.RUnlock()
+
+	return v.publicParameters
+}
+
+// SerializePublicParameters returns the public params in a serialized form
+func (v *PublicParamsManager[T]) SerializePublicParameters() ([]byte, error) {
+	return v.PublicParams().Serialize()
```
alexandrosfilios, line 75: ok now I see probably you wanted to use the RLock here

token/core/common/ppm.go
```diff
@@ -0,0 +1,143 @@
...
+func (v *PublicParamsManager[T]) Load() error {
+	ppRaw, err := v.Vault.PublicParams()
+	if err != nil {
+		return errors.WithMessage(err, "failed to fetch public params from vault")
+	}
+	if len(ppRaw) == 0 {
+		return nil
+	}
+
+	v.Logger.Debugf("fetched public parameters [%s], set them...", hash.Hashable(ppRaw).String())
```
alexandrosfilios, line 91: better not to use `String()` for lazy eval

token/core/common/ppm.go
```diff
@@ -0,0 +1,143 @@
...
+	return v.SetPublicParameters(ppRaw)
+}
+
+// SetPublicParameters updates the public parameters with the passed value
+func (v *PublicParamsManager[T]) SetPublicParameters(raw []byte) error {
+	v.Mutex.Lock()
```
alexandrosfilios, line 97: Maybe we could lock in line 120 or 113 (not sure what the callbacks do)?

token/core/common/ppm.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+	}
+
+	v.Logger.Debugf("set public parameters [%s]", hash.Hashable(raw).String())
+	v.publicParameters = pp
+
+	return nil
+}
+
+func (v *PublicParamsManager[T]) PublicParams() T {
```
alexandrosfilios, line 125: This method is a duplicate I think

token/core/fabtoken/driver/deserializer.go
```diff
@@ -7,169 +7,45 @@ SPDX-License-Identifier: Apache-2.0
...
-// GetOwnerMatcher is not needed in fabtoken, as identities are in the clear
-func (d *Deserializer) GetOwnerMatcher(raw []byte) (driver.Matcher, error) {
-	ai := &x509.AuditInfo{}
-	err := ai.FromBytes(raw)
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to unmarshal")
-	}
-	return &x509.AuditInfoDeserializer{CommonName: ai.EnrollmentId}, nil
+// EIDRHDeserializer returns enrollment ID and revocation handle behind the owners of token
+type EIDRHDeserializer struct {
```
alexandrosfilios, line 42: You can also define an alias
```
type EIDRHDeserializer = *common.EIDRHDeserializer[*x509.AuditInfo]
```

token/core/zkatdlog/nogh/driver/deserializer.go
```diff
@@ -165,61 +110,14 @@ func (d *DeserializerProvider) Deserialize() (driver.Deserializer, error) {
...
-	ai, err := e.getAuditInfo(auditInfo)
-	if err != nil {
-		return "", err
-	}
-	if ai == nil {
-		return "", nil
-	}
-	return ai.RevocationHandle(), nil
+// EIDRHDeserializer returns enrollment ID and revocation handle behind the owners of token
+type EIDRHDeserializer struct {
```
alexandrosfilios, line 114: Same as above

token/core/zkatdlog/nogh/driver/driver.go
```diff
@@ -139,11 +137,20 @@ func (d *Driver) NewTokenService(sp driver.ServiceProvider, networkID string, ch
...
-		&zkatdlog.VaultTokenLoader{TokenVault: qe},
-		zkatdlog.NewVaultTokenCommitmentLoader(qe, 3, 3*time.Second),
+		common.NewVaultLedgerTokenAndMetadataLoader[*token3.Token, *token3.Metadata](
+			qe,
+			tokDeserializer,
+		),
+		common.NewLedgerTokenLoader[*token3.Token](
+			qe,
+			tokDeserializer,
+			3,
```
alexandrosfilios, line 151: We can use constants for these.

token/core/fabtoken/driver/deserializer.go
adecaro: 👍

token/core/fabtoken/driver/deserializer.go
```diff
@@ -7,169 +7,45 @@ SPDX-License-Identifier: Apache-2.0
...
 
-// VerifierDES is the interface for verifiers' deserializer
-// A verifier checks the validity of a signature against the identity
-// associated with the verifier
-type VerifierDES interface {
-	DeserializeVerifier(id view.Identity) (driver.Verifier, error)
-}
-
 // Deserializer deserializes verifiers associated with issuers, owners, and auditors
 type Deserializer struct {
```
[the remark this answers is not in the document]
adecaro: no need anymore

token/core/common/ppm.go
adecaro: there is an interface that needs to be satisfied and requires a function that returns driver.PublicParamters

token/core/zkatdlog/nogh/driver/driver.go
adecaro: 👍

token/core/common/certifier.go
```diff
@@ -13,21 +13,21 @@ import (
 
 // NewCertificationRequest creates a request to certify the tokens identified by the passed identifiers
 // zkatdlog does not make use of the certification service
-func (s *Service) NewCertificationRequest(ids []*token.ID) ([]byte, error) {
+func (s *Service[T]) NewCertificationRequest(ids []*token.ID) ([]byte, error) {
```
alexandrosfilios, line 16: Maybe there are reasons, but we could split the Service into smaller services, because these methods maybe don't necessarily share something in common with the rest of the struct.

token/core/common/loaders.go
```diff
@@ -0,0 +1,223 @@
...
+		})
+		if err == nil {
+			return tokens, nil
+		}
+
+		// check if there is any token id whose corresponding transaction is pending
+		// if there is, then wait a bit and retry to load the outputs
+		retry := false
+		for _, id := range ids {
+			pending, err := s.TokenVault.IsPending(id)
```
alexandrosfilios, line 80: Maybe we could do a batch request here by changing the signature of `TokenVault`
```
IsPending(ids ...TokenID) map[TokenID]bool
```
or at least add a todo

token/core/common/deserializer.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// AuditMatcherDeserializer deserializes raw bytes into a matcher, which allows an auditor to match an identity to an enrollment ID
+type AuditMatcherDeserializer interface {
+	GetOwnerMatcher(raw []byte) (driver.Matcher, error)
+}
+
+// Deserializer deserializes verifiers associated with issuers, owners, and auditors
+type Deserializer struct {
+	identityType        string
```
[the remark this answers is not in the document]
adecaro: this is the driver to decide

token/core/common/loaders.go
```diff
@@ -0,0 +1,223 @@
...
+		tok, err := s.Deserializer.DeserializeToken(comm)
+		if err != nil {
+			return errors.Wrapf(err, "failed deserializing token for id [%v][%s]", id, string(comm))
+		}
+		ti, err := s.Deserializer.DeserializeMetadata(info)
+		if err != nil {
+			return errors.Wrapf(err, "failed deserializeing token info for id [%v]", id)
+		}
+
+		inputIDs[counter] = key
```
alexandrosfilios, line 147: I would use an atomic counter for this

token/core/common/loaders.go
```diff
@@ -0,0 +1,223 @@
...
+func (s *VaultLedgerTokenAndMetadataLoader[T, M]) LoadTokens(ids []*token.ID) ([]string, []T, []M, []view.Identity, error) {
+	n := len(ids)
+	tokens := make([]T, n)
+	inputIDs := make([]string, n)
+	inputInf := make([]M, n)
+	signerIds := make([]view.Identity, n)
+
+	// return token outputs and the corresponding opening
+	counter := 0
+	if err := s.TokenVault.GetTokenInfoAndOutputs(ids, func(id *token.ID, key string, comm, info []byte) error {
```
alexandrosfilios, line 129: Maybe we should change the signature of this method to return directly the slices. Because now we get the slices, and we call a callback function on them. and then with this callback we put them back into slices.

alexandrosfilios: Same also for the other vault operations

adecaro: we need a deserialization step here, no?

token/core/common/loaders.go
adecaro: to support parallelism we would need a different approach because we need to guarantee the order, no?

token/core/common/loaders.go
adecaro: 👍

token/core/common/wallets.go
```diff
@@ -0,0 +1,175 @@
...
+	Logger           *flogging.FabricLogger
+	IdentityProvider driver.IdentityProvider
+	TokenVault       IssuerTokenVault
+	id               string
+	identity         view.Identity
+}
+
+func NewIssuerWallet(Logger *flogging.FabricLogger, IdentityProvider driver.IdentityProvider, TokenVault IssuerTokenVault, id string, identity view.Identity) *IssuerWallet {
+	return &IssuerWallet{
+		Logger:           Logger,
```
alexandrosfilios, line 73: we can keep these fields private

token/core/tms.go
```diff
@@ -67,39 +67,22 @@ func (m *TMSProvider) GetTokenManagerService(opts driver.ServiceOptions) (servic
...
-		m.lock.Unlock()
-		// invoke callback
-		if invokeCallback && m.callbackFunc != nil {
-			err = m.callbackFunc(service, opts.Network, opts.Channel, opts.Namespace)
-			if err != nil {
-				logger.Fatalf("failed to initialize tms for [%s]: [%s]", opts, err)
-			}
-		}
-	}()
+	defer m.lock.Unlock()
```
alexandrosfilios, line 70: we can work with a RLock and then in the `if !ok` clause acquire a write Lock

token/core/common/certifier.go
adecaro: 👍

token/core/tms.go
```diff
@@ -127,37 +110,46 @@ func (m *TMSProvider) Update(opts driver.ServiceOptions) (err error) {
...
+
+	// create the service for the new public params
+	newService, err := m.getTokenManagerService(opts)
+	if err == nil {
+		// unload the old service, if set
+		if service != nil {
+			if err := service.Done(); err != nil {
+				return errors.WithMessage(err, "failed to unload token service")
+			}
+			delete(m.services, key)
```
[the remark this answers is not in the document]
adecaro: ah, we can just replace? right 👍

token/core/tms.go
```diff
@@ -67,39 +67,22 @@ func (m *TMSProvider) GetTokenManagerService(opts driver.ServiceOptions) (servic
...
-		if invokeCallback && m.callbackFunc != nil {
-			err = m.callbackFunc(service, opts.Network, opts.Channel, opts.Namespace)
-			if err != nil {
-				logger.Fatalf("failed to initialize tms for [%s]: [%s]", opts, err)
-			}
-		}
-	}()
+	defer m.lock.Unlock()
 
 	key := tmsKey(opts)
```
alexandrosfilios, line 72: We can avoid most locks in such cases with double-checked locking:
```
key := tmsKey(opts)
if service, ok := m.services[key]; ok {
 return service, nil
}
m.lock.Lock()
defer m.lock.Unlock()
if service, ok := m.services[key]; ok {
 return service, nil
}
service, err := m.getTokenManagerService(opts)
if err != nil {
 return nil, err
}
m.services[key] = service
return service, nil
```

token/core/tms.go
```diff
@@ -127,37 +110,46 @@ func (m *TMSProvider) Update(opts driver.ServiceOptions) (err error) {
...
 	service, ok := m.services[key]
 	if !ok {
 		logger.Debugf("no service found, instantiate token management system for [%s:%s:%s] for key [%s]", opts.Network, opts.Channel, opts.Namespace, key)
-		return
+	} else {
+		logger.Debugf("service found, unload token management system for [%s:%s:%s] for key [%s] and reload it", opts.Network, opts.Channel, opts.Namespace, key)
+	}
+
+	// create the service for the new public params
+	newService, err := m.getTokenManagerService(opts)
```
alexandrosfilios, line 125: Same comment about double-check locking

token/core/common/wallets.go
alexandrosfilios: aaah ok I understand!

alexandrosfilios (approved):
LGTM modulo comments

token/core/tms.go
adecaro: 👍

token/core/tms.go
adecaro: no, here we always write.

PR #580, opened 2024-03-20
register wallets dynamically

adecaro:
This PR brings the following:
- The `WalletService` allows you to register an `IdentityConfiguration`. This struct can carry directly the identity's cryptographic content. This means that no access to disk might be required. This identity configuration is then stored in the token-sdk's identity db and can be reload later.
- New contract for the token drivers: The driver must be instantiated with the given public params and the wallets can be loaded immediately.


token/services/identity/msp/idemix/bccsp.go
```diff
@@ -7,34 +7,82 @@ SPDX-License-Identifier: Apache-2.0
...
+func NewBCCSP(keyStore bccsp.KeyStore, curveID math.CurveID, aries bool) (bccsp.BCCSP, error) {
+	curve, tr, err := GetCurveAndTranslator(curveID)
 	if err != nil {
 		return nil, err
 	}
-	cryptoProvider, err := idemix2.NewAries(&keystore.Dummy{}, curve, tr, true)
+
+	var cryptoProvider bccsp.BCCSP
+	if aries {
+		cryptoProvider, err = idemix.NewAries(keyStore, curve, tr, true)
```
alexandrosfilios, line 41: Maybe a check for the curve in the case of Aries

token/services/identity/msp/idemix/deserializer.go
```diff
@@ -20,25 +20,14 @@ import (
 
 // NewDeserializer returns a new deserializer for the idemix ExpectEidNymRhNym verification strategy
 func NewDeserializer(ipk []byte, curveID math.CurveID) (*DeserializerWrapper, error) {
-	if curveID == math.BLS12_381_BBS {
-		return NewDeserializerAries(ipk)
-	}
 	logger.Infof("new deserialized for dlog idemix")
-	cryptoProvider, err := NewBCCSP(curveID)
+	cryptoProvider, err := NewBCCSPWithDummyKeyStore(curveID, curveID == math.BLS12_381_BBS)
```
alexandrosfilios, line 24: Is this curve only usable with Aries?

token/services/identity/msp/x509/deserializer.go
```diff
@@ -80,3 +84,51 @@ func (a *AuditInfoDeserializer) DeserializeAuditInfo(raw []byte) (*AuditInfo, er
...
 	return ai, nil
 }
+
+func GetEnrollmentID(id []byte) (string, error) {
+	si := &msp.SerializedIdentity{}
+	err := proto.Unmarshal(id, si)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to unmarshal to msp.SerializedIdentity{}")
+	}
+	block, _ := pem.Decode(si.IdBytes)
```
alexandrosfilios, line 94: We could re-use PemDecodeCert(si.IdBytes) to avoid some duplication

token/services/identity/msp/x509/deserializer.go
```diff
@@ -80,3 +84,51 @@ func (a *AuditInfoDeserializer) DeserializeAuditInfo(raw []byte) (*AuditInfo, er
...
+	if block == nil {
+		return nil, errors.New("bytes are not PEM encoded")
+	}
+	switch block.Type {
+	case "CERTIFICATE":
+		cert, err := x509.ParseCertificate(block.Bytes)
+		if err != nil {
+			return nil, errors.WithMessage(err, "pem bytes are not cert encoded ")
+		}
```
alexandrosfilios, line 125: Up to this point, the method is similar to GetEnrollmentId. We could have a function:
```
func getCertificate(id []byte) (*x509.Certificate, error)
```

token/services/identity/msp/x509/lm.go
```diff
@@ -333,6 +332,24 @@ func (lm *LocalMembership) getResolver(label string) *common.Resolver {
 	return nil
 }
 
+func (lm *LocalMembership) keyStorePath(translatedPath string) (keyStorePath string) {
+	if lm.ignoreVerifyOnlyWallet {
```
alexandrosfilios, line 336: Maybe it's easier if we invert the logic:
```
if !lm.ignoreVerifyOnlyWallet {
 return ""
}

path := filepath.Join(translatedPath, KeystoreFullFolder)
if _, err := os.Stat(path); err == nil {
 return path
}

path = filepath.Join(translatedPath, "msp", KeystoreFullFolder)
if _, err := os.Stat(path); err == nil {
 return path
}

return ""

alexandrosfilios (approved):
LGTM

token/services/identity/msp/idemix/bccsp.go
adecaro: in reality aries now supports any curve

token/services/identity/msp/idemix/deserializer.go
adecaro: yes

token/services/identity/msp/x509/deserializer.go
adecaro: 👍

token/services/identity/msp/x509/deserializer.go
adecaro: 👍

token/services/identity/msp/x509/lm.go
adecaro: 👍

PR #582, opened 2024-03-21
Delete tokens: move nilcheck and simplify the call used by HTLC

arner:
This moves the nilcheck to make sure it gets used (Delete does not fail if the token doesn't exist).

When running two replicas of Token SDK, one instance will delete the token from the database and notify its mailman. A current limitation is that the mailman of the second replica will not be notified (if tok turns out to be nil we don't post the event). This is not new in this PR, but we should consider it when working on the statelessness of the Token SDK instance.


adecaro:
@arner , I was wondering the following: GetToken returns nil (without error) only if the sql select returns nothing. The token might still be there but marked as deleted already. So, instead of using GetToken internally, why don't we introduce a private `getToken` function that allows us to select the token even if it is marked as deleted. This would allow other replicas to send the event to mailman anyway. What do you think?

adecaro (approved):
LGTM

PR #584, opened 2024-03-25
improved token transaction flow

adecaro:
The goal of this PR is to remove the need to distributed the fabric envelope to the participant nodes of a transaction.
It is enough that each node receives the token request.
Moreover, if the recipients must sign the token request, then we can save one round of interaction.


token/core/common/validator.go
```diff
@@ -121,7 +135,7 @@ func (v *Validator[P, T, TA, IA]) VerifyTokenRequest(ledger driver.Ledger, signa
 	for _, action := range ta {
 		actions = append(actions, action)
 	}
-	return actions, nil, nil
+	return actions, attributes, nil
```
alexandrosfilios, line 138: I wouldn't return the attributes. It is confusing because it seems that this method could modify the map and return it. So tracking what the actual values of the map are becomes difficult (one needs to track all methods where this is passed). I would return only actions and error instead.

token/core/common/validator.go
```diff
@@ -93,23 +101,29 @@ func (v *Validator[P, T, TA, IA]) VerifyTokenRequestFromRaw(getState driver.GetS
...
 
+	attributes := make(map[string][]byte)
+	attributes[TokenRequestToSign], err = v.Serializer.MarshalTokenRequestToSign(req, nil)
+	if err != nil {
+		return nil, nil, errors.Wrap(err, "failed to marshal signed token request")
+	}
+
 	backend := NewBackend(getState, signed, signatures)
-	return v.VerifyTokenRequest(backend, backend, anchor, tr)
+	return v.VerifyTokenRequest(backend, backend, anchor, tr, attributes)
```
alexandrosfilios, line 111: Do we expect a variable number/type of attributes to be passed? otherwise we could pass a typed struct and if we see that it becomes too big, then we can switch to the map.

integration/token/fungible/views/redeem.go
```diff
@@ -84,7 +84,7 @@ func (t *RedeemView) Call(context view.Context) (interface{}, error) {
 	net := network.GetInstance(context, tx.Network(), tx.Channel())
 	vault, err := net.Vault(tx.Namespace())
 	assert.NoError(err, "failed to retrieve vault [%s]", tx.Namespace())
-	vc, err := vault.Status(tx.ID())
+	vc, _, err := vault.Status(tx.ID())
```
alexandrosfilios, line 87: We can remove the second output param from `Status`, as it seems it isn't used anywhere and the signature is not part of an interface.

token/core/common/serializer.go
```diff
@@ -0,0 +1,20 @@
...
+*/
+
+package common
+
+import "github.com/hyperledger-labs/fabric-token-sdk/token/driver"
+
+type Serializer struct {
+}
+
+func (s Serializer) MarshalTokenRequestToSign(request *driver.TokenRequest, meta *driver.TokenRequestMetadata) ([]byte, error) {
```
alexandrosfilios, line 14: We could remove the metadata from the signature (it is the only implementation), unless other future implementations are supposed to use it.

token/services/db/sql/transactions.go
```diff
@@ -478,16 +491,14 @@ func (db *TransactionDB) GetSchema() string {
 		CREATE INDEX IF NOT EXISTS idx_tx_id_%s ON %s ( tx_id );
```
alexandrosfilios, line 493: also not needed.

PR #588, opened 2024-04-01
return identities slice of len 0 for auditors instead of len 1 with <…

arner:
…empty> id when no auditors are set.

This caused tokenchaincode to fail when there are no auditors.


adecaro (commented):
LGTM

PR #592, opened 2024-04-08
change max token calculation to use big int because high float64 is h…

arner:
…andled unreliably across architectures.

Found this out when creating the tokengen params on arm64 and then running the code on amd64... See also https://groups.google.com/g/golang-nuts/c/YxpOo02fT9s?pli=1. I also noticed the original code does -1, whereas I had to -2 to get the same result for a bitlength of 64.


adecaro (approved):
LGTM

PR #597, opened 2024-04-11
store token request hash in rwsets

adecaro:
The translation of the token request for fabric and orion contain the hash of the token request to later be checked again the local copy.

### DIFF
--- a/token/core/common/validator.go
+++ b/token/core/common/validator.go
@@ -13,6 +13,10 @@ import (
 	"github.com/pkg/errors"
 )
 
+const (
+	TokenRequestToSign = "trs"
+)
+
 type Context[P driver.PublicParameters, T any, TA driver.TransferAction, IA driver.IssueAction] struct {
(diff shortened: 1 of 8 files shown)


token/core/common/validator.go
```diff
@@ -23,6 +27,7 @@ type Context[P driver.PublicParameters, T any, TA driver.TransferAction, IA driv
 	IssueAction       IA
 	Ledger            driver.Ledger
 	MetadataCounter   map[string]int
+	Attributes        map[string][]byte
```
alexandrosfilios, line 30: If we really need to pass a map and not a struct with a stricter structure, I would suggest at least we pass a `map[AttributeType][]byte` where `AttributeType` is an enum, so that we know what to expect.

adecaro: 👍

alexandrosfilios (approved):
LGTM

PR #599, opened 2024-04-15
Centralize database initialization logic

arner:
Simplifies opening of SQL databases by centralizing the configuration in the database package, similar to how identitydb was already doing it.


adecaro (approved):
LGTM

PR #600, opened 2024-04-16
add atomic write to ttxdb

arner:
if this is ok, we can do the same for the auditdb and remove the functions that are now on the AtomicWrite (as well as the 'txn' field) from the database struct.


arner:
updated to include SetStatus (as discussed offline) and the audit db update.

PR #602, opened 2024-04-17
core logging

adecaro:
This PR does the following: It removes from the `core` package the package level logging to introduce struct level logging.


token/core/common/meta/metadata.go
```diff
@@ -0,0 +1,30 @@
...
+)
+
+const (
+	TransferMetadataPrefix = "TransferMetadataPrefix"
+)
+
+// SetTransferAction extracts the transfer metadata from the passed attributes and
+// sets them to the passed metadata
+func SetTransferAction(attrs map[interface{}]interface{}, metadata map[string][]byte) {
+	for key, value := range attrs {
```
alexandrosfilios, line 20: Modifying an input parameter is confusing for the logic flow. Instead I would return a new map:
```
func GetTransferActions(attrs map[any]any) map[string][]byte {
metadata := map[string][]byte
for ...
return metadata
}
```
and then loop through the map and set the values. The big O complexity is the same.

token/core/logging/logger.go
```diff
@@ -0,0 +1,44 @@
...
+	Infof(format string, args ...interface{})
+	Panic(args ...interface{})
+	Panicf(format string, args ...interface{})
+	Warn(args ...interface{})
+	Warnf(format string, args ...interface{})
+	IsEnabledFor(level zapcore.Level) bool
+}
+
+func DriverLogger(prefix string, networkID string, channel string, namespace string) Logger {
+	return flogging.MustGetLogger(prefix + extension(networkID) + extension(channel) + extension(namespace))
```
alexandrosfilios, line 32: We can also do it like this
```
func isEmptyString(s string) bool { return len(s) == 0 }

func loggerName(parts ...string) string {
 return strings.Join(slices.DeleteFunc(parts, isEmptyString), ".")}
```
but it's a matter of taste

alexandrosfilios (approved):
LGTM modulo comments

token/core/logging/logger.go
adecaro: 👍

token/core/common/meta/metadata.go
adecaro: 👍

PR #603, opened 2024-04-18
driver cleanup

adecaro:
This PR does the following: It refactors the `TokenManagerService` to avoid interface imports.
This makes more modular the implementation of a new driver.


alexandrosfilios (approved):
LGTM

PR #604, opened 2024-04-18
remove query executor

arner:
Breaking API change: instead of opening and closing a query executor:

```
aqe := auditor.NewQueryExecutor()
defer aqe.Done()
 aqe.NewPaymentsFilter()
```
You have to directly:
```
auditor.NewPaymentsFilter()
```


adecaro (approved):
LGTM

PR #612, opened 2024-04-24
add ability to query for token details

arner:
This PR:

- Adds the ability to query for token with details, to allow the user to enrich the data about a transaction
- Splits the Raw Owner into its two components for more efficient queries. This can later be extended to the Owner struct in the Token (replacing the raw version)
- Stores the ApplicationMetadata as a separate field of the token request and adds it to retrieved transactions (where they are usually used)
- Returns owners when doing GetToken
- Simplifies the code of the querybuilder a bit


token/services/tokens/storage.go
```diff
@@ -114,12 +113,20 @@ func (t *transaction) AppendToken(
 	if err != nil {
 		return errors.Wrapf(err, "cannot covert [%s] with precision [%d]", tok.Quantity, precision)
 	}
+	id, err := identity.UnmarshalTypedIdentity(tok.Owner.Raw)
+	if err != nil {
+		logger.Errorf("could not unmarshal identity when storing token: %s", err.Error())
+		return fmt.Errorf("could not unmarshal identity when storing token: %w", err)
```
adecaro, line 119: please, use "github.com/pkg/errors"

token/services/tokens/storage.go
```diff
@@ -114,12 +113,20 @@ func (t *transaction) AppendToken(
 	if err != nil {
 		return errors.Wrapf(err, "cannot covert [%s] with precision [%d]", tok.Quantity, precision)
 	}
+	id, err := identity.UnmarshalTypedIdentity(tok.Owner.Raw)
```
adecaro, line 116: I'm wondering: Should we replace this with an `unmarshaller` that given in input the raw []bytes gives the type and the identity? 
It is just to detach the tokens package from the identity package.
In theory, the content of an identity is driver dependant, but we can make in a future PR explicit the presence of the type.

arner: I like the idea of decoupling them on this level, but where should the unmarshaler be? Let me know if you have an example that I can reuse.

PR #613, opened 2024-04-25
remove orion commit processor

adecaro:
This PR does the following: It removes the orion transaction processor. Transactions are committed into the local databases upon an event received via a finality listener. 
Next step is to remove the fabric transaction processor too.


alexandrosfilios (approved):
LGTM

PR #614, opened 2024-04-25
query directly token requests

adecaro:


token/services/auditor/manager.go
```diff
@@ -110,89 +109,27 @@ func (cm *Manager) restore(tmsID token.TMSID, walletID string) error {
...
-		return errors.Errorf("failed to get vault for [%s:%s]", tmsID, walletID)
-	}
-	type ToBeUpdated struct {
-		TxID    string
-		Status  auditdb.TxStatus
-		Message string
-	}
-	var toBeUpdated []ToBeUpdated
-	var pendingTXs []string
 	for {
```
alexandrosfilios, line 121: A more fluent way of writing this would be
```
var record SomeType
var err error
for record, err = it.Next(); record != nil && err == nil; record, err = it.Next() {
 if err := net.AddFinalityListener(record.TxID, &FinalityListener{net, auditor.db}); err != nil {
  return errors.WithMessagef(err, "failed to subscribe event listener to network [%s] for [%s]", tmsID, record.TokenRequest)
 }
}
if err != nil {
 return errors.Errorf("failed to get next tx record for [%s:%s]", tmsID, walletID)
}
```

token/services/db/sql/querybuilder.go
```diff
@@ -182,6 +182,27 @@ func certificationsQuerySql(ids []*token.ID) (string, []any, error) {
 	return builder.String(), tokenIDs, nil
 }
 
+func tokenRequestConditionsSql(params driver.QueryTokenRequestsParams) (string, []interface{}) {
```
alexandrosfilios, line 185: You can check in fsc `unversioned.go`
```
query := fmt.Sprintf("SELECT pkey, val FROM %s WHERE ns = $1 AND pkey IN %s", db.table, generateParamSet(2, len(keys)))
	logger.Debug(query, ns, keys)

	rows, err := db.readDB.Query(query, append([]any{ns}, castAny(keys)...)...)
	
```
We could extract these two methods to utils to re-use them

alexandrosfilios (approved):
LGTM modulo changes

token/services/auditor/manager.go
adecaro: but the error generated by `record, err = it.Next()` is not checked

token/services/db/sql/querybuilder.go
adecaro: sure, we can do that in another PR

PR #615, opened 2024-04-25
Fts crypto code refactoring

KElkhiyaoui:
* Remove un-necessary crypto code. 
* Remove anonymous issuance for the time being. Replace it instead with `type_is_hidden` or `not`. In systems with a single issuer hiding the type can be required. For transparent accounting, systems may require the type to be revealed at issuance time. In systems where a single issuer issues a single type, the type is implicitly revealed. 
* We plan to support anonymous issuance if applications call for it.


token/core/zkatdlog/crypto/audit/auditor.go
```diff
@@ -413,3 +409,11 @@ func GetAuditInfoForTransfers(transfers [][]byte, metadata []driver.TransferMeta
 	}
 	return auditableInputs, outputs, nil
 }
+
+func commit(vector []*math.Zr, generators []*math.G1, c *math.Curve) *math.G1 {
+	com := c.NewG1()
```
yacovm, line 414: In Gnark this produces the point to infinity, but are we sure this is true for all mathlib implementations?

KElkhiyaoui: I will add a test to check that `c.NewG1()` always returns a `point_at_infinity`.

adecaro:
@KElkhiyaoui , following our offline discussion, please remove the IsTypeHidden flag. Thanks much :)

PR #619, opened 2024-04-29
fabric commit processor doesn't handle anymore tokens

adecaro:
This job is left to the audit and transaction dbs.

### DIFF
--- a/token/services/network/fabric/processor.go
+++ b/token/services/network/fabric/processor.go
@@ -20,23 +20,20 @@ var logger = flogging.MustGetLogger("token-sdk.vault.processor")
 
 type GetTokensFunc = func() (*tokens.Tokens, error)
 type GetTMSProviderFunc = func() *token.ManagementServiceProvider
-type GetTokenRequestFunc = func(tms *token.ManagementService, txID string) ([]byte, error)
 
 type RWSetProcessor struct {
-	network         string
-	nss             []string
-	GetTokens       GetTokensFunc
-	GetTMSProvider  GetTMSProviderFunc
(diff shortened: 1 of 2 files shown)


token/sdk/tms/db/tms.go
```diff
@@ -117,15 +109,9 @@ func (p *PostInitializer) ConnectNetwork(networkID, channel, namespace string) e
...
-			n.Name(),
-			namespace,
-			common.NewLazyGetter[*tokens2.Tokens](func() (*tokens2.Tokens, error) {
-				return tokens2.GetService(p.sp, tmsID)
-			}).Get,
-			GetTMSProvider,
-			GetTokenRequest,
-		),
+		fabric2.NewTokenRWSetProcessor(n.Name(), namespace, common.NewLazyGetter[*tokens2.Tokens](func() (*tokens2.Tokens, error) {
+			return tokens2.GetService(p.sp, tmsID)
```
alexandrosfilios, line 113: We can inject the tokensManager and remove the usage of sp

token/services/network/fabric/processor.go
```diff
@@ -65,7 +62,7 @@ func (r *RWSetProcessor) Process(req fabric.Request, tx fabric.ProcessTransactio
 	case "init":
```
alexandrosfilios, line 62: Maybe we can add a comment to explain why init stays (because we don't know the txid of the chain code-upgrade tx's) and why default returns nil (because it's processed by the filter)

alexandrosfilios (approved):
LGTM modulo comment

PR #620, opened 2024-04-29
check status and tokens from the dbs

adecaro:


token/sdk/vault/provider.go
```diff
@@ -0,0 +1,81 @@
...
+	if ok {
+		return res, nil
+	}
+
+	tmsID := token.TMSID{
+		Network:   network,
+		Channel:   channel,
+		Namespace: namespace,
+	}
+	tokenDB, err := tokendb.GetByTMSId(v.sp, tmsID)
```
alexandrosfilios, line 58: Let's avoid using SP and pass the token db provider instead. This is injected directly into the constructor.

token/services/network/common/vault/vault.go
```diff
@@ -54,15 +52,27 @@ func (v *Vault) DeleteTokens(ids ...*token.ID) error {
 
 type QueryEngine struct {
 	*tokendb.DB
-	backend driver2.Vault
+	auditDB *auditdb.DB
+	ttxdb   *ttxdb.DB
 }
 
 func (q *QueryEngine) IsPending(id *token.ID) (bool, error) {
-	vc, _, err := q.backend.TransactionStatus(id.TxId)
```
alexandrosfilios, line 61: We can do
```
vd, _, err := q.GetStatus(id.TxId)
return vd == ttxdb.Pending, err
```

token/sdk/vault/provider.go
adecaro: okay

token/services/network/common/vault/vault.go
adecaro: okay

token/sdk/vault/provider.go
```diff
@@ -0,0 +1,97 @@
...
+		ttxDBProvider:   ttxDBProvider,
+		tokenDBProvider: tokenDBProvider,
+		auditDBProvider: auditDBProvider,
+		fabricNSP:       fabricNSP,
+		orionNSP:        orionNSP,
+		vaultCache:      make(map[string]vault.Vault),
+	}
+}
+
+func (v *Provider) Vault(network string, channel string, namespace string) (vault.Vault, error) {
```
alexandrosfilios, line 49: We could use TMSID

alexandrosfilios (commented):
LGTM

PR #625, opened 2024-05-07
remove envelope

adecaro:
This PR does the following: It removes the need to register the envelope at the network interface level. This reduces further the dependencies with the underlying backend.

### DIFF
--- a/token/services/ttx/transaction.go
+++ b/token/services/ttx/transaction.go
@@ -7,10 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 package ttx
 
 import (
-	"encoding/asn1"
-
-	view2 "github.com/hyperledger-labs/fabric-smart-client/platform/view"
-	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/hash"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
 	"github.com/hyperledger-labs/fabric-token-sdk/token"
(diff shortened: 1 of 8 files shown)


token/services/ttx/accept.go
```diff
@@ -40,18 +40,6 @@ func (s *AcceptView) Call(context view.Context) (interface{}, error) {
...
 		return nil, errors.Wrap(err, "failed to marshal token request")
 	}
-	// Check the envelope exists
-	env := s.tx.Payload.Envelope
-	if env == nil {
-		return nil, errors.Errorf("expected fabric envelope")
-	}
-
-	// Store envelope
-	if !s.options.SkipApproval {
```
alexandrosfilios, line 50: We can remove the field `SkipApproval`

token/services/ttx/transaction.go
```diff
@@ -291,25 +290,13 @@ func (t *Transaction) TMSID() token.TMSID {
 }
 
 func (t *Transaction) setEnvelope(envelope *network.Envelope) error {
```
alexandrosfilios, line 292: We don't need this method I think

token/services/ttx/accept.go
adecaro: it is still used in other places.

alexandrosfilios (approved):
LGTM

token/services/ttx/transaction.go
adecaro: 👍

PR #626, opened 2024-05-08
cleaup service provider's use in services

adecaro:
This PR does the following:
- Use token.ServiceProvider rather than view.ServiceProvider.
- Remove ServiceProvider as a field where possible.


alexandrosfilios (approved):
LGTM

PR #627, opened 2024-05-08
add an option to stop the htlc scan for preimage when all blocks are …

arner:
…searched until the tip of the blockchain.

Can be used for a better experience if triggered by a user and you want to show them immediately whether the preimage is available on the current ledger, instead of having to wait until timing out.

### DIFF
--- a/integration/token/interop/support.go
+++ b/integration/token/interop/support.go
@@ -390,7 +390,7 @@ func fastExchange(network *integration.Infrastructure, id string, recipient stri
 	time.Sleep(10 * time.Second)
 }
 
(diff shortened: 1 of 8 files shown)


adecaro:
Hi @arner , thanks for submitting this PR. I was wondering if we could test the new options enhancing the test `TestHTLCNoCrossClaimTwoNetworks`. That test calls a few times the `scan` function that invokes a view that calls `htlc.ScanForPreImage`. We could try to have a positive and a negative case with the use of `WithStopOnLastTransaction`. What do you think?

adecaro (approved):
LGTM

PR #628, opened 2024-05-10
identity deserialization

adecaro:
This PR does the following:
- It introduces an alias of `view.Identity` from the smart client in the Token and Driver API. In future PR, the alias will be used everywhere and then the dependency to the smart client removed.
- It refactors the identity deserialization infrastructure to be more customizable. The identity type tells which deserializer to use in all circumstances.

Side node: This PR leaves a TODO to be addressed in another PR. The auditor of the zkatdlog driver needs to be updated to better handle the htlc identities and their relative audit information.


token/core/common/validator.go
```diff
@@ -94,7 +93,7 @@ func (v *Validator[P, T, TA, IA]) VerifyTokenRequestFromRaw(getState driver.GetS
 		return nil, nil, errors.Wrap(err, "failed to marshal signed token request")
 	}
 
-	v.Logger.Debugf("cc tx-id [%s][%s]", hash.Hashable(raqRaw).String(), anchor)
+	v.Logger.Debugf("cc tx-id [%s][%s]", Hashable(raqRaw).String(), anchor)
```
alexandrosfilios, line 96: Avoid to call String() in debug logs. It is called lazily

token/core/fabtoken/transfer.go
```diff
@@ -125,14 +136,15 @@ func (s *TransferService) Transfer(txID string, wallet driver.OwnerWallet, ids [
 	}
 
 	metadata := &driver.TransferMetadata{
-		Outputs:            outputs,
```
alexandrosfilios, line 128: I think it would give more clarity to have a struct:
```
type OutToken struct {
 TokenID
 Sender
 Output
 Metadata
 AuditInfo
 Receiver
 ReceiverAuditInfo
 ReceiverIsSender
}
```

This way it's not necessary to read the code to figure out which slice elements are related, and we don't need to check that the lengths are correct.

token/services/identity/deserializer/typed.go
```diff
@@ -0,0 +1,128 @@
...
+	GetOwnerAuditInfo(id driver.Identity, typ string, raw []byte, p driver.AuditInfoProvider) ([][]byte, error)
+}
+
+// AuditMatcherDeserializer deserializes raw bytes into a matcher, which allows an auditor to match an identity to an enrollment ID
+type AuditMatcherDeserializer interface {
+	GetOwnerMatcher(raw []byte) (driver.Matcher, error)
+}
+
+type TypedVerifierDeserializerMultiplex struct {
+	deserializers            map[string]TypedVerifierDeserializer
```
alexandrosfilios, line 29: We could use an alias to link this value to the `Type` property and even provide constants for it

token/services/identity/interop/htlc/deserializer.go
```diff
@@ -60,3 +52,68 @@ func (d *Deserializer) getHTLCVerifier(raw []byte) (driver.Verifier, error) {
...
+}
+
+func NewAuditDeserializer(auditInfoDeserializer deserializer.AuditInfoDeserializer) *AuditDeserializer {
+	return &AuditDeserializer{AuditInfoDeserializer: auditInfoDeserializer}
+}
+
+func (a *AuditDeserializer) DeserializeAuditInfo(bytes []byte) (deserializer.AuditInfo, error) {
+	si := &ScriptInfo{}
+	err := json.Unmarshal(bytes, si)
+	if err == nil && (len(si.Sender) != 0 || len(si.Recipient) != 0) {
```
alexandrosfilios, line 108: We could inverse the condition to reduce nesting.

token/services/identity/interop/htlc/deserializer.go
```diff
@@ -60,3 +52,68 @@ func (d *Deserializer) getHTLCVerifier(raw []byte) (driver.Verifier, error) {
...
+		if len(si.Recipient) != 0 {
+			ai, err := a.AuditInfoDeserializer.DeserializeAuditInfo(si.Recipient)
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed unamrshalling audit info [%s]", bytes)
+			}
+			return ai, nil
+		}
+		return nil, errors.Errorf("no recipient defined")
+	}
+	return nil, errors.Errorf("ivalid audit info, failed unmarshal [%s]", string(bytes))
```
alexandrosfilios, line 118: Maybe it makes sense to include also the length of sender/recipient for better debugging.

token/services/identity/sig/deserializer.go
```diff
@@ -0,0 +1,126 @@
...
+		if logger.IsEnabledFor(zapcore.DebugLevel) {
+			logger.Debugf("trying signer deserialization with [%s] failed [%s]", des, err)
+		}
+		errs = append(errs, err)
+	}
+
+	return nil, errors.Errorf("failed signer deserialization [%v]", errs)
+}
+
+func (d *multipler) Info(raw []byte, auditInfo []byte) (string, error) {
```
alexandrosfilios, line 95: To avoid duplication we could have a method:
```
func deserialize[V](extractor func() (V, err)) {
	var errs []error

	copyDeserial := d.threadSafeCopyDeserializers()
	for _, des := range copyDeserial {
		if logger.IsEnabledFor(zapcore.DebugLevel) {
			logger.Debugf("trying signer deserialization with [%s]", des)
		}
		v, err := extractor()
		if err == nil {
			if logger.IsEnabledFor(zapcore.DebugLevel) {
				logger.Debugf("trying signer deserialization with [%s] succeeded", des)
			}
			return v, nil
		}

		if logger.IsEnabledFor(zapcore.DebugLevel) {
			logger.Debugf("trying signer deserialization with [%s] failed [%s]", des, err)
		}
		errs = append(errs, err)
	}

	return nil, errors.Errorf("failed signer deserialization [%v]", errs)
}
```

Then your three methods are almost the same:
```
func (d *multipler) DeserializeSigner(raw []byte) (driver.Signer, error) {
 return deserialize(d.threadSafeCopyDeserializers(), func(des driver.Deserializer) (driver.Signer, error) {return des.DeserializeSigner(raw)})
}
```

alexandrosfilios (approved):
LGTM modulo comments

token/core/fabtoken/transfer.go
adecaro: 👍

token/services/identity/interop/htlc/deserializer.go
adecaro: 👍

token/services/identity/interop/htlc/deserializer.go
adecaro: 👍

token/services/identity/sig/deserializer.go
adecaro: beautiful 👍

token/services/identity/deserializer/typed.go
adecaro: 👍

PR #631, opened 2024-05-21
cleanup

adecaro:
This PR does the following:
- It replaces view.Identity with token.Identity or driver.Identity.
- Use services/logging
- TokenRequest#Issue appends the metadata as produced by the driver

### DIFF
--- a/token/services/ttx/collect.go
+++ b/token/services/ttx/collect.go
@@ -9,6 +9,8 @@ import (
 	"fmt"
 	"time"
 
+	view2 "github.com/hyperledger-labs/fabric-smart-client/platform/view"
+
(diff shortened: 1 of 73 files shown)


token/services/ttx/collect.go
```diff
@@ -83,7 +85,12 @@ func (c *collectActionsView) collectLocal(context view.Context, actionTransfer *
 	}
 
 	// Binds identities
-	if err := c.tx.TokenRequest.BindTo(context, party); err != nil {
+	es := view2.GetEndpointService(context)
```
alexandrosfilios, line 88: We could extract this to a method and re-use it for lines 157-162

alexandrosfilios (approved):
LGTM

PR #632, opened 2024-05-22
Sherdlock

alexandrosfilios:


token/services/db/driver/token.go
```diff
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"time"
 
+	"github.com/hyperledger-labs/fabric-smart-client/platform/common/core"
```
adecaro, line 13: please, remove this dependency

integration/nwo/token/orion/template.go
```diff
@@ -37,6 +37,9 @@ token:
       auditdb:
         persistence:
           type: unity
+      tokenlockdb:
```
adecaro, line 40: This should not be needed. The driver should switch to the shared configuration under `db`, no?

token/vault.go
```diff
@@ -59,7 +59,7 @@ func (q *QueryEngine) UnspentTokensIterator() (*UnspentTokensIterator, error) {
 }
 
 // UnspentTokensIteratorBy is an iterator over all unspent tokens in this vault owned by passed wallet id and whose token type matches the passed token type
-func (q *QueryEngine) UnspentTokensIteratorBy(walletID, tokenType string) (*UnspentTokensIterator, error) {
+func (q *QueryEngine) UnspentTokensIteratorBy(walletID, tokenType string) (driver.UnspentTokensIterator, error) {
```
adecaro, line 62: please, rollback this to respect the pattern that we return only structs here.

token/services/drivers/dbholder.go
```diff
@@ -0,0 +1,72 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package drivers
+
+import (
+	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils"
```
adecaro, line 10: please, remove the deps to the smart client from the driver interfaces.

adecaro (changes requested):

integration/nwo/token/orion/template.go
alexandrosfilios: For Orion it was needed. It failed otherwise and the other db's also have it.

adecaro: so, why doesn't the fabric template need it?

alexandrosfilios: Because you take the following configs in this order:
```
dbconfig.NewConfig(configProvider, "tokenlockdb.persistence.type", "db.persistence.type"))
```
For Orion, the latter is not defined, for Fabric it is.

adecaro (approved):
LGTM. The rest will be addressed in other PRs

PR #635, opened 2024-05-23
clean up dependencies of sdk.go

adecaro:


token/provider.go
```diff
@@ -140,6 +141,73 @@ func (p *ManagementServiceProvider) managementService(aNew bool, opts ...Service
 	return ms, nil
 }
 
+func (p *ManagementServiceProvider) normalize(opt *ServiceOptions) (*ServiceOptions, error) {
+	// lookup configurations
+	configs, err := p.tmsProvider.Configs()
```
alexandrosfilios, line 146: Maybe it's better to create another struct that implements the normaliser interface and pass the slice of normalisers.

arner (approved):
LGTM!

token/sdk/sdk.go
```diff
@@ -205,9 +179,18 @@ func (p *SDK) Start(ctx context.Context) error {
 		logger.Infof("start token management service [%s]...", tmsID)
 
 		// connect network
-		if err := p.postInitializer.ConnectNetwork(tmsID.Network, tmsID.Channel, tmsID.Namespace); err != nil {
+		net := network.GetInstance(p.registry, tmsID.Network, tmsID.Channel)
```
alexandrosfilios, line 182: Since we are in the sdk and we have access to the network provider, the tms provider etc, maybe we could use those instead of calling the registry

token/services/network/common/filter.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/auditdb"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/ttxdb"
+	"github.com/pkg/errors"
+)
+
+type TransactionFilterProvider[F driver2.TransactionFilter] interface {
+	New(tmsID token3.TMSID) (F, error)
+}
+
+type AcceptTxInDBFilterProvider struct {
```
alexandrosfilios, line 21: Maybe a comment to describe more or less what this filter is used for

alexandrosfilios (approved):
LGTM

token/sdk/sdk.go
adecaro: 👍

token/services/network/common/filter.go
adecaro: 👍

PR #636, opened 2024-05-23
remove dependencies to core and driver packages

adecaro:


token/services/identity/msp/x509/msp/config.go
```diff
@@ -82,3 +82,29 @@ func ToPKCS11OptsOpts(o *PKCS11) *pkcs11.PKCS11Opts {
...
+		SW: &SoftwareProvider{
+			Hash:     "SHA2",
+			Security: 256,
+		},
+		PKCS11: &PKCS11{
+			Hash:     "SHA2",
+			Security: 256,
+		},
+	}
+	if defaultProvider == "PKCS11" {
```
alexandrosfilios, line 100: We could also have a constant field
```
var pkcs11Bccsp = ...
```

alexandrosfilios (approved):
LGTM

PR #640, opened 2024-05-27
Refactor drivers

alexandrosfilios:


token/services/db/driver/audit.go
```diff
@@ -8,6 +8,7 @@ package driver
 
 import (
 	token2 "github.com/hyperledger-labs/fabric-token-sdk/token"
```
adecaro, line 10: no need for the import alias

token/services/db/driver/audit.go
```diff
@@ -49,9 +50,4 @@ type AuditDBDriver interface {
 	Open(cp ConfigProvider, tmsID token2.TMSID) (AuditTransactionDB, error)
 }
 
-type ConfigProvider interface {
-	UnmarshalKey(key string, rawVal interface{}) error
-	GetString(key string) string
-	IsSet(key string) bool
-	TranslatePath(path string) string
-}
+type ConfigProvider = core.ConfigProvider
```
adecaro, line 53: I think it is better to not link to core because it does not guarantee that this interface will remain there

token/services/drivers/dbholder.go
```diff
@@ -7,12 +7,14 @@ SPDX-License-Identifier: Apache-2.0
 package drivers
```
adecaro, line 7: can we move this package under db?

adecaro (changes requested):

token/services/drivers/dbholder.go
alexandrosfilios: I can move the db drivers. but the holders contain generic drivers too

adecaro (approved):
LGTM

PR #641, opened 2024-05-28
Integration tests for replicas

alexandrosfilios:


token/services/tokens/storage.go
```diff
@@ -147,7 +149,7 @@ func (t *transaction) AppendToken(
 		},
 		ids,
 	)
-	if err != nil {
+	if err != nil && !errors2.HasCause(err, driver.UniqueKeyViolation) {
```
adecaro, line 152: I think we can use the `errors.Is` function in the Go SDK.

token/services/network/fabric/network.go
```diff
@@ -515,7 +515,13 @@ func (t *FinalityListener) OnStatus(txID string, status int, message string) {
 	if err != nil {
 		panic(fmt.Sprintf("can't get query executor [%s]", txID))
 	}
-	tokenRequestHash, err := qe.GetState(t.namespace, key)
+	var tokenRequestHash []byte
```
adecaro, line 518: why do we need this? Can we avoid these magic numbers?

token/services/tokens/storage.go
```diff
@@ -7,6 +7,8 @@ SPDX-License-Identifier: Apache-2.0
 package tokens
 
 import (
+	errors2 "github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
```
adecaro, line 10: can we avoid to import "github.com/pkg/errors" then?

adecaro (approved):
LGTM, modulo the comments. Please, apply before merging. Thanks :)

PR #642, opened 2024-05-28
configuration service

adecaro:
This PR does the following: It introduces a new configuration service that is used by core and other services (i.e. identity service) to obtain various configuration parameters.
The Driver API has been therefore simplified.


token/services/config/config.go
```diff
@@ -0,0 +1,198 @@
...
+
+// configuration is the configuration of a given configuration
+type configuration struct {
+	cp      configProvider
+	keyID   string
+	tmsID   driver.TMSID
+	version string
+}
+
+func NewConfiguration(cp configProvider, version string, keyID string, tmsID driver.TMSID) *configuration {
```
alexandrosfilios, line 40: This is the same as `PrefixConfig`. We could reuse it

alexandrosfilios (approved):
LGTM modulo comments

PR #646, opened 2024-06-02
db extensions

adecaro:
This PR introduces a mechanism to extend the TokenDB to store additional information about a token. 
For example, when a token's owner field contains an HTLC (`service/interop/htlc`) script, it would be convenient to store script-related information for more targeted queries and data analysis.

### DIFF
--- a/token/services/interop/htlc/db/sql/tokendbext.go
+++ b/token/services/interop/htlc/db/sql/tokendbext.go
@@ -1,0 +1,91 @@
+package sql
+
+import (
+	"database/sql"
+	"encoding/json"
+	"fmt"
+	"runtime/debug"
+
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/db/driver"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/interop/htlc"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/logging"
+	"github.com/pkg/errors"
+)
+
+var logger = logging.MustGetLogger("token-sdk.sql.htlc")
+
+type htlcTokenDBExtensionTables struct {
+	Tokens string
+}
+
+type HTLCTokenDBExtension struct {
+	table htlcTokenDBExtensionTables
+}
+
+func NewHTLCTokenDBExtension(table htlcTokenDBExtensionTables) *HTLCTokenDBExtension {
+	return &HTLCTokenDBExtension{table: table}
+}
+
+func (e *HTLCTokenDBExtension) GetSchema() string {
+	return fmt.Sprintf(`
+		-- Tokens
+		CREATE TABLE IF NOT EXISTS %s (
+			tx_id TEXT NOT NULL,
+			idx INT NOT NULL,
+			sender_raw BYTEA NOT NULL,
+			recipient_raw BYTEA NOT NULL,
+			deadline TIMESTAMP NOT NULL,
+			hash BYTEA NOT NULL,
+			hash_function INT NOT NULL,
+			hash_encoding INT NOT NULL,
+			PRIMARY KEY (tx_id, idx)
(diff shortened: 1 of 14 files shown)


token/services/interop/htlc/db/sql/tokendbext.go
```diff
@@ -0,0 +1,97 @@
...
+	table htlcTokenDBExtensionTables
+}
+
+func NewHTLCTokenDBExtension(table htlcTokenDBExtensionTables) *HTLCTokenDBExtension {
+	return &HTLCTokenDBExtension{table: table}
+}
+
+func (e *HTLCTokenDBExtension) GetSchema() string {
+	return fmt.Sprintf(`
+		-- Tokens
```
arner, line 37: I guess Tokens should be 'HTLC' or 'Interop' here. Or wait, do you actually mean the existing 'tokens' table here, or a new one that exists next to the 'tokens' table?

If it's the first, that could become an issue with versioning and updates (our code doesn't handle db migrations yet, but should in the near future and extending an existing table adds another dimension that makes it harder).

token/services/interop/htlc/db/sql/tokendbext.go
```diff
@@ -0,0 +1,97 @@
...
+)
+
+var logger = logging.MustGetLogger("token-sdk.sql.htlc")
+
+type htlcTokenDBExtensionTables struct {
+	Tokens string
+}
+
+type HTLCTokenDBExtension struct {
+	table htlcTokenDBExtensionTables
```
arner, line 28: There is a dependency on the 'tables' struct here. We can make our lives much easier by removing the feature of the 'table prefixes' everywhere (which I think we should do anyway!). Then, every db just decides by itself in the code what its tables are called.

token/services/interop/htlc/db/sql/tokendbext.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (e *HTLCTokenDBExtension) GetSchema() string {
+	return fmt.Sprintf(`
+		-- Tokens
+		CREATE TABLE IF NOT EXISTS %s (
+			tx_id TEXT NOT NULL,
+			idx INT NOT NULL,
+			sender_raw BYTEA NOT NULL,
+			recipient_raw BYTEA NOT NULL,
+			deadline TIMESTAMP NOT NULL,
```
arner, line 43: All of these fields are already very nice to have! Now after this, it would be amazing to have the following two fields as well:

- preimage (easy; after scanning the blockchain it can be just updated in the table)
- status (trickier; depends on heuristics of whether the deadline is past and the preimage is available...)

I'm not sure how the process of these should work, like when they would be updated and if the 'interop' package already has the right hooks to manage that. But let's consider it in the design; if there is a way, it would help with the transparency and integrity of htlc transactions.

In terms of use cases, I'll share some diagrams and code with you that I used to approach this state on the application level.

token/services/interop/htlc/db/sql/tokendbext.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+	// store the script
+	logger.Debugf("store htlc record [%s:%d,%v] in table [%s]", tr.TxID, tr.Index, owners, e.table.Tokens)
+
+	// Store token
+	query := fmt.Sprintf("INSERT INTO %s (tx_id, idx, sender_raw, recipient_raw, deadline, hash, hash_function, hash_encoding) VALUES ($1, $2, $3, $4, $5, $6, $7, $8)", e.table.Tokens)
+	logger.Debug(query,
+		tr.TxID,
+		tr.Index,
+		script.Sender,
```
arner, line 77: sender and recipient are probably too large to debug?

adecaro:
to be addressed by another Pr

PR #648, opened 2024-06-04
move types to utils

adecaro:


token/services/db/driver/token.go
```diff
@@ -177,9 +178,9 @@ type TokenDBDriver interface {
 // - unlock tokens that were locked by a process that exited unexpectedly
 type TokenLockDB interface {
 	// Lock locks a specific token for the consumer TX
-	Lock(tokenID *token.ID, consumerTxID core.TxID) error
+	Lock(tokenID *token.ID, consumerTxID utils.TxID) error
```
alexandrosfilios, line 181: Just for the naming maybe it would be better to put it under `utils/core`? But I'm fine with it, too

alexandrosfilios (approved):
LGTM modulo comments

token/services/db/driver/token.go
adecaro: I'll move it under `utils/types/transaction` and rename the type to `ID`.

PR #649, opened 2024-06-04
metrics

adecaro:


token/core/common/metrics/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+		return
+	}
+	m.FailedIssues.With("token_type", tokenType).Add(1)
+}
+
+func (m *Metrics) ObserveIssueDuration(duration time.Duration) {
+	m.IssueDuration.Observe(float64(duration.Milliseconds()))
+}
+
+func (m *Metrics) AddTransfer(noErr bool) {
```
alexandrosfilios, line 80: I think it would be easier to understand if we used two labels instead. So instead of `m.AddIssue("CHF", err == nil)` we could do
```
m.Issues.With("token_type", tokenType, "failed", err != nil).Add(1)
```

token/core/common/metrics/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+func (m *Metrics) ObserveTransferDuration(duration time.Duration) {
+	m.TransferDuration.Observe(float64(duration.Milliseconds()))
+}
+
+func (m *Metrics) AddAudit(noErr bool) {
+	if noErr {
+		m.Audits.With().Add(1)
+		return
+	}
+	m.FailedAudits.With().Add(1)
```
alexandrosfilios, line 97: Do we need `With`?

token/core/common/metrics/metrics.go
```diff
@@ -0,0 +1,138 @@
...
+	return c.Counter.With(l...)
+}
+
+type histogram struct {
+	rootLabels []string
+	Histogram
+}
+
+func (c *histogram) With(labels ...string) Histogram {
+	l := make([]string, len(c.rootLabels)+len(labels))
```
alexandrosfilios, line 134: the length needs to be 0. But it maybe even more concise to write
```return c.Histogram.With(append(c.rootLabels, labels...)...```

**Comment by alexandrosfilios** [comment]
In the last commit, I "hid" the two metrics (operations and duration) behind a tracer.

The tracer is a more general standard that traces processes by registering their start, some events that may happen and then the end. These can then be visualised with tools like Jaeger. These events have timestamps to register the duration of each stage, as well as optional properties we can attach (e.g. tokenType, success).

A span is a series of events that captures the lifecycle of a request from its creation to its end. The main methods of the span interface are:
```
	End(options ...SpanEndOption)
	AddEvent(name string, options ...EventOption)
	SetAttributes(kv ...attribute.KeyValue)
```
The spans are created by a tracer every time we want to record a new request. The interface of the tracer contains only one method, that of creating and starting a span.
```
	Start(ctx context.Context, spanName string, opts ...SpanStartOption) (context.Context, Span)
```
In the example of the observable services of this PR, the issue operation would be re-written as follows:
```
func (o *ObservableIssueService) Issue(issuerIdentity driver.Identity, tokenType string, values []uint64, owners [][]byte, opts *driver.IssueOptions) (driver.IssueAction, *driver.IssueMetadata, error) {
	_, span := o.Metrics.issueTracer.Start(context.Background(), "issue", trace.WithAttributes(attribute.String(TokenTypeLabel, tokenType)))
	defer span.End()

	action, meta, err := o.IssueService.Issue(issuerIdentity, tokenType, values, owners, opts)
	span.SetAttributes(attribute.Bool(SuccessfulLabel, err == nil))

// Maybe some more operations could be added here with new events
...
span.AddEvent("start_signing", WithAttributes(String("signer_id", "..."))), WithTimestamp(...))
...

	return action, meta, err
}
```

In the code above, we only used interfaces that can have different implementations, e.g.
* no-op tracers/spans
* tracers that print their results in the logs
* tracers that send the results to Jaeger

In this PR, I assumed a backing implementation of a tracer (by default the no-op unless defined otherwise) and I enhanced it by adding the two metrics we are interested in: operation count and operation duration. The main advantage of this is that we can now use a unified interface and depending on our backing implementation, we can extract and visualise the spans in Jaeger without any further change in the code (only the correct exporter has to be defined in the sdk).

Hence, the new tracer has two components:
* backing tracer provider (registers the spans for use in the logs, Jaeger etc)
* backing metrics provider (updates the counter and the histogram for the operations)

As with the tracer provider, also the implementation of the metrics provider can vary, hence also be disabled, if the backing metrics provider is the no-op provider (this way it will be as if we had only integrated the original tracer).
---

Another change I made is the creation of a new metrics provider that integrates and attaches the TMSID automatically. Instead of adding the labels network/channel/namespace for every counter/histogram operation, we decorate the existing metric provider with the new one that attaches them at the beginning. This way we don't need to pass the tmsid every time (we don't need to even pass it into the observable service anymore) and it is a bit faster, because we don't need to attach the tmsid labels every time.

---

Putting the two parts together we have a tracer that uses
* a backing trace provider (currently the no-op)
* a decorated version of the backing metrics provider (that attaches the TMSID labels always)

For the instantiation we do:
```
// The original metrics provider (can be no-op, statsd, prometheus)
backingMetricsProvider := metrics.GetProvider(sp)
// The decorated metrics provider that attaches the TMSID labels
metricsProvider := metrics.NewTMSProvider(tmsID, backingMetricsProvider)

// The original tracer provider (can be no-op, logger exporter, jaeger exporter and further variants)
backingTracerProvider := trace.NewNoOpTracerProvider()
// The decorated tracer provider that combines the traces above with the new metrics (that also contain the TMSID in the labels)
tracerProvider := tracing.NewProvider(metricsProvider)
```
Now the metrics struct will contain all tracers and metrics as follows:
```
type Metrics struct {
	issueTracer trace.Tracer
        validateTracer trace.Tracer

	otherMetric     metrics.Counter
        anotherMetric metrics.Histogram
        ...
}

func New(tp trace.TracerProvider, mp metrics.Provider) *Metrics {
	return &Metrics{
		issueTracer: tp.Tracer("issue", tracing.WithMetricsOpts(tracing.MetricsOpts{
			Namespace:    "token_sdk",
			LabelNames:   metrics.AllLabelNames(TokenTypeLabel, SuccessfulLabel),
			StatsdFormat: metrics.StatsdFormat(TokenTypeLabel, SuccessfulLabel),
		})),
		verifyTracer: tp.Tracer(...),
		otherMetric: mp.NewCounter(...),
		anotherMetric: mp.NewHistogram(...),
		...
	}
}
```
Keep in mind that instead of a tracer provider and a metrics provider we could have an enhanced provider that contains both, so we don't need to pass both every time.

PR #651, opened 2024-06-05
Refactor integration tests and add web socket integration tests

alexandrosfilios:
* Split the tests into more parts
* Added web socket integration tests
* Extracted common code from integration tests to avoid duplication
* Added selectors to distinguish when we need the ID and when the name of a node


adecaro (approved):
LGTM modulo the comments that can be addressed in another PR too.

PR #653, opened 2024-06-10
remove append from crypto library

KElkhiyaoui:


token/core/zkatdlog/crypto/audit/auditor.go
```diff
@@ -334,6 +334,7 @@ func GetAuditInfoForIssues(issues [][]byte, metadata []driver.IssueMetadata) ([]
 		if len(ia.OutputTokens) != len(md.ReceiversAuditInfos) || len(ia.OutputTokens) != len(md.OutputsMetadata) {
 			return nil, errors.Errorf("number of output does not match number of provided metadata")
 		}
+		outputs[k] = make([]*AuditableToken, len(md.ReceiversAuditInfos))
```
adecaro, line 337: if you allocate the array as 
```go
outputs[k] = make([]*AuditableToken, 0, len(md.ReceiversAuditInfos))
```
Then you can keep `append`. The above call creates an array of size len(md.ReceiversAuditInfos) and 0 length.

token/core/zkatdlog/crypto/common/array.go
```diff
@@ -22,18 +23,16 @@ type G1Array []*math.G1
...
 		}
 		st := hex.EncodeToString(e.Bytes())
-		raw = append(raw, []byte(Separator)...)
-		raw = append(raw, []byte(st)...)
-
+		raw[2*i] = []byte(st)
+		raw[2*i+1] = []byte(Separator)
 	}
-	return raw, nil
+	return json.Marshal(raw)
```
adecaro, line 35: better to avoid json here given that it is slow.
What you could do is to obtain the 3 slices you need and then aggregate them by allocating a new slice of size the sum of the 3 slices. Then you can use append.

adecaro (changes requested):
LGTM, please address the json-related comment 👍

token/core/zkatdlog/crypto/common/array.go
KElkhiyaoui: Hi @adecaro, I will use `bytes.Join(raw, separator)` instead. It seems to be fit for purpose.

adecaro (approved):
LGTM

PR #655, opened 2024-06-13
fabric network adaptor: support for FSC-based endorsement

adecaro:
This PR does the following: It allows the developer to configure FSC nodes as endorsers of the Token Chaincode.
In this case, the FSC node will not run directly the chaincode but its logic directly.

### DIFF
--- a/docs/core-token.md
+++ b/docs/core-token.md
@@ -52,6 +52,28 @@ token:
             maxOpenConns: 10
             dataSource: /some/path/tokendb
 
+      services:
+        # This section contains network specific configuration
+        network:
+          # Configuration related to the Fabric network
+          fabric:
+            # In Fabric, the execution of the token chaincode can be endorsed by any node equipped with
+            # a proper endorsement key.
+            # Therefore, also FSC nodes equipped with proper endorsement keys can perform the same function.
+            # This section is dedicated to the configuration of the endorsement of the token chaincode by
+            # other FSC nodes.
+            endorsement:
+              # Is this node an endorser?: true/false
+              endorser: true
+              # If this node is an endorser, which Fabric identity should be used to sign the endorsement?
+              # If empty, the default identity will be used
+              id: 
+              # A list of FSC node identifiers that must be contacted to obtain the endorsement 
+              endorsers:
+              - endorser1
+              - endorser2
+              - endorser2
+
       # sections dedicated to the definition of the wallets
       wallets:
         # Default cache size reference that can be used by any wallet that support caching
--- a/docs/core-token.md
+++ b/docs/core-token.md
@@ -67,7 +67,11 @@ token:
               endorser: true
               # If this node is an endorser, which Fabric identity should be used to sign the endorsement?
               # If empty, the default identity will be used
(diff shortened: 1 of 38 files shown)


docs/core-token.md
```diff
@@ -51,6 +51,25 @@ token:
...
+            # This section is dedicated to the configuration of the endorsement of the token chaincode by
+            # other FSC nodes.
+            endorsement:
+              # Is this node an endorser?: true/false
+              endorser: true
+              # If this node is an endorser, which Fabric identity should be used to sign the endorsement?
+              # If empty, the default identity will be used
+              id: 
+              # A comma-separated list of FSC node identifiers that must be contacted to obtain the endorsement 
+              endorsers: {{ Endorsers }}
```
alexandrosfilios, line 71: Why comma separated and not a list?

integration/nwo/token/fabric/opts.go
```diff
@@ -21,3 +26,35 @@ func IsFabricCA(tms *topology.TMS) bool {
...
+	v, ok := tms.BackendParams["endorsements"]
+	return ok && v.(bool)
+}
+
+func EndorsersToString(tms *topology.TMS) string {
+	v, ok := tms.BackendParams["endorsers"]
+	if !ok {
+		return ""
+	}
+	endorsersStr := strings.Builder{}
```
alexandrosfilios, line 46: We could also use strings.Join I guess, but still i suppose using a list is a cleaner option, unless I'm missing something

token/services/network/fabric/endorsement.go
```diff
@@ -0,0 +1,72 @@
...
+
+type Endorsement interface {
+	Endorse(context view.Context, tms *token2.ManagementService, requestRaw []byte, signer view.Identity, txID driver.TxID) (driver.Envelope, error)
+}
+
+type FSCEndorsement struct {
+	Endorsers []view.Identity
+}
+
+func (e *FSCEndorsement) Endorse(context view.Context, tms *token2.ManagementService, requestRaw []byte, signer view.Identity, txID driver.TxID) (driver.Envelope, error) {
```
alexandrosfilios, line 28: We could make this a service and inject tms and viewmanager

adecaro: This has the same signature of the method in the Network interface. We need tms at the time of the invocation. The viewManager con be injected, sure.

integration/token/common/topology.go
```diff
@@ -32,6 +32,7 @@ type Opts struct {
 	SDKs            []api.SDK
 	WebEnabled      bool
 	Monitoring      bool
+	WithEndorsers   bool
```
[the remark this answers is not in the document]
adecaro: renamed

integration/nwo/token/fabric/opts.go
adecaro: okay

docs/core-token.md
adecaro: changed 👍

alexandrosfilios (approved):

PR #656, opened 2024-06-13
DIG for tokensdk

alexandrosfilios:


token/dig/sdk.go
```diff
@@ -0,0 +1,229 @@
...
+	fabricsdk "github.com/hyperledger-labs/fabric-smart-client/platform/fabric/dig"
+	view2 "github.com/hyperledger-labs/fabric-smart-client/platform/view"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/driver"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/flogging"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/kvs"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/tracing"
+	"github.com/hyperledger-labs/fabric-token-sdk/token"
+	core2 "github.com/hyperledger-labs/fabric-token-sdk/token/core"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/core/common/logging"
+	_ "github.com/hyperledger-labs/fabric-token-sdk/token/core/fabtoken/driver"
```
adecaro, line 25: the other sdk.go doesn't have any blind imports. Why do we need them here?

token/dig/sdk.go
```diff
@@ -0,0 +1,229 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package sdk
```
adecaro, line 7: can we move this package under sdk?

adecaro (changes requested):

token/dig/sdk.go
alexandrosfilios: Oh no we don't need them. This version was based on a previous version of the sdk that had the blind imports. Will remove them.

adecaro (approved):
LGTM

PR #658, opened 2024-06-19
stress test update

adecaro:
This PR adds also transfers to itself.
Notice that this integration test is not included in the CI. It is uses locally for rapid stress checks.


integration/token/fungible/tests.go
```diff
@@ -1088,8 +1089,23 @@ func TestStress(network *integration.Infrastructure, auditorId string, selector
 		})
 	}()
 
+	// start transfers from Charlie
+	charlieTransferPool := dlog.NewPool(10)
+	go func() {
```
alexandrosfilios, line 1094: schedule task could already launch a goroutine

adecaro:
To be resubmitted

PR #666, opened 2024-07-01
move authorization to the driver space

adecaro:
This PR moves the responsibility of the implementation of the Authorization interface to the driver that better knows the relation between tokens and wallets.


alexandrosfilios (approved):
LGTM

PR #668, opened 2024-07-02
unity driver for everything + table renaming

adecaro:
This PR does the following: 
- Allows you to choose unity as the single provider for all databases used in a test.
- Better naming for sql tables.


arner (commented):
I like the clearer table names.

For unity, how about we make it the default for all databases? Then, if you only configure `db.persistence.type`, you are done. Not just for NWO but for normal core.yaml too.

(PS I'd like to get rid of the configurable table prefix in another PR)

adecaro:
Hi @arner , for NWO, it is nice to test also when you have unity and another database. Then, for any other application you can do whatever you want. Not sure if I understood your request though. Can you clarify, please?

arner:
Yes, I agree. Maybe I'm wrong, but I thought the config currently requires you to do something like this if you want unity everywhere:

```
      db:
        persistence:
          type: unity
          opts:
            driver: sqlite
            dataSource: file:/var/dc/data/bank1/db.sqlite?_pragma=journal_mode(WAL)&_pragma=busy_timeout(5000)&_txlock=immediate
      ttxdb:
        persistence:
          type: unity
      auditdb:
        persistence:
          type: unity
      tokendb:
        persistence:
          type: unity
      identitydb:
        persistence:
          type: unity
```

And I thought it would be nice if unity would be the default, so that you don't need the bottom 12 lines in my example. Or is that already the case?

Anyway I see now that this is separate from the configurability in NWO.

adecaro:
HI @arner , no, that is not needed anymore.  This is enough: 
```yaml
       db:
        persistence:
          type: unity
          opts:
            driver: sqlite
            dataSource: file:/var/dc/data/bank1/db.sqlite?_pragma=journal_mode(WAL)&_pragma=busy_timeout(5000)&_txlock=immediate
```

arner (approved):

PR #687, opened 2024-07-12
fix data race in local membership

Grimnir777:
Prevent some data race for the function `GetDefaultIdentifier` (reading list of resolvers) when called from `mapStringToID`


adecaro:
@Grimnir777 , please, could you rebase?

Grimnir777:
@adecaro I'll be off for 2 weeks, I'll check the failing tests when I came back

adecaro:
@Grimnir777 , do you want to check if your fix is still needed, please? Thanks :)

adecaro:
close, it is addressed by the current code

PR #692, opened 2024-07-15
add config flag for retries for selectors

arner:


token/services/selector/simple/service.go
```diff
@@ -33,11 +29,19 @@ type SelectorService struct {
 	managerLazyCache utils.LazyProvider[*token.ManagementService, token.SelectorManager]
 }
 
-func NewService(lockerProvider LockerProvider) *SelectorService {
+func NewService(tms *token.ManagementService, lockerProvider LockerProvider, cfg viewdriver.ConfigService) *SelectorService {
+	retryInterval := 5 * time.Second
```
alexandrosfilios, line 33: Maybe we could have a struct with this config that we unmarshal
```
type SelectorConfig struct {
 RetryInterval time.Duration
 NumRetries int
}

func (c *SelectorConfig) GetNumRetries() int {
 if c.NumRetries > 0 return c.NumRetries else return 3
}
```

adecaro: Hi @arner , did you address this?

arner: Yes @adecaro I rebased and added the SelectorConfig.

docs/core-token.md
```diff
@@ -16,6 +16,8 @@ token:
   # if empty, the default selector is used
   selector:
     driver: sherdlock
+    retryInterval: 5s
+    numRetries: 5
```
adecaro, line 20: please, can you add some information about what are these retries about?

token/sdk/dig/sdk.go
```diff
@@ -87,7 +86,7 @@ func (p *SDK) Install() error {
 
 	logger.Infof("Token platform enabled, installing...")
 
-	fmt.Printf("token selector [%s]\n", p.ConfigService().GetString("token.selector.driver"))
+	logger.Infof("token selector [%s]\n", p.ConfigService().GetString("token.selector.driver"))
```
adecaro, line 89: do we need Info here? Maybe debug?

token/services/selector/sherdlock/service.go
```diff
@@ -17,17 +18,23 @@ import (
...
 )
 
-const retrySelectionBackoff = 5 * time.Second
-
 type SelectorService struct {
 	managerLazyCache utils.LazyProvider[*token.ManagementService, token.SelectorManager]
 }
 
-func NewService(tokenDBManager *tokendb.Manager, tokenLockDBManager *tokenlockdb.Manager, metricsProvider metrics.Provider) *SelectorService {
+func NewService(tokenDBManager *tokendb.Manager, tokenLockDBManager *tokenlockdb.Manager, cfg driver.ConfigService, metricsProvider metrics.Provider) *SelectorService {
```
adecaro, line 25: Please, instead of importing driver.ConfigService, could you define a local interface with just the methods you need?

token/services/selector/sherdlock/service.go
```diff
@@ -17,17 +18,23 @@ import (
...
+	c := SelectorConfig{}
+	err := cfg.UnmarshalKey("token.selector", &c)
+	if err != nil {
+		panic("invalid config for key [token.selector]: expect retryInterval (duration) and numRetries (integer))")
```
adecaro, line 29: do you think panicing here is the right way to handle the error? Can we unmarshall outside the function and pass directly SelectorConfig?

adecaro (changes requested):

token/services/selector/driver/driver.go
```diff
@@ -0,0 +1,61 @@
...
+type SelectorConfig interface {
+	GetDriver() Driver
+	GetNumRetries() int
+	GetRetryInterval() time.Duration
+}
+
+type Driver string
+
+type Config struct {
+	Driver        Driver
```
alexandrosfilios, line 24: It is going to work, but maybe it is better if we add the json annotations

alexandrosfilios (approved):
LGTM

adecaro (approved):
LGTM. Please rebase.

token/sdk/dig/sdk.go
```diff
@@ -103,7 +104,10 @@ func (p *SDK) Install() error {
 		p.Container().Provide(func(ttxdbManager *ttxdb.Manager) *network2.LockerProvider {
 			return network2.NewLockerProvider(ttxdbManager, 2*time.Second, 5*time.Minute)
 		}, dig.As(new(selector.LockerProvider))),
-		p.Container().Provide(selectorProviders[p.ConfigService().GetString("token.selector.driver")], dig.As(new(token.SelectorManagerProvider))),
+		p.Container().Provide(func() (sdriver.SelectorConfig, error) {
+			return sdriver.New(p.ConfigService())
```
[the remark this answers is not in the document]
alexandrosfilios: This could be directly
```
p.Container().Provide(sdriver.New)
```

alexandrosfilios (approved):
LGTM

PR #699, opened 2024-07-17
hide pkcs11 behind a build tag so that token sdk is pure go by default

arner:
Companion PR to https://github.com/hyperledger-labs/fabric-smart-client/pull/621.

Goal is to make HSM support 'opt-in' by using a build tag: `CGO_ENABLED=1 go build -tags pkcs11`. If you don't supply the tag, it will not require CGO. This makes it easier to have cross-platform builds or to use alpine containers, for instance.


adecaro (approved):
LGTM

PR #700, opened 2024-07-17
TokensDB: Support for update

adecaro:
This commit does the following: When deleting a token, if the token does not exists, an entry in the tokens table is added and marked as deleted. When the token is later added, the entry is updated


adecaro:
the goal of this PR is not clear anymore if ever will be needed. Closing it for now.

PR #712, opened 2024-07-23
orion network: cached status

adecaro:


token/services/network/orion/approval.go
```diff
@@ -242,6 +249,11 @@ func (r *RequestApprovalResponderView) validate(context view.Context, request *A
 	if err != nil {
 		return nil, true, errors.Wrapf(err, "failed to sign and close transaction [%s]", request.TxID)
 	}
+	// update the cache
+	r.statusCache.Add(request.TxID, &TxStatusResponse{
```
alexandrosfilios, line 253: Do we need to update the cache? We can add the tx only when it is valid/invalid later. This way we put less elements in the cache and we can use also `GetOrLoad`, in order to leverage the locking mechanism during update.

adecaro: it is helpful because we store already h in the cache so we don't have to query it after when the status is retrieved.

PR #723, opened 2024-08-05
remove table prefix

arner:
Without a customizable table prefix, it will be easier to handle schema migrations.

(breaking change: datamodel)


adecaro:
to be addressed later

PR #734, opened 2024-08-13
include wallet_id into tokens table

adecaro:
This PR includes a new column into the tokens table to track the wallet id the token is bound to.
Notice that this can be empty depending on the nature of the token's owner (see the htlc script ownership, for an example).
In this way, we can avoid the join in the sql tokens db implementation when selecting tokens to spend.

This PR includes also unit-tests for the token lock db.

### DIFF
--- a/token/core/common/authrorization.go
+++ b/token/core/common/authrorization.go
@@ -16,7 +16,7 @@ import (
 type Authorization interface {
 	// IsMine returns true if the passed token is owned by an owner wallet.
 	// It returns the ID of the owner wallet and any additional owner identifier, if supported.
-	// It is possible that the wallet ID is empty an the additional owner identifier list is not.
+	// It is possible that the wallet ID is empty and the additional owner identifier list is not.
 	IsMine(tok *token2.Token) (string, []string, bool)
 	// AmIAnAuditor return true if the passed TMS contains an auditor wallet for any of the auditor identities
 	// defined in the public parameters of the passed TMS.
@@ -44,7 +44,7 @@ func NewTMSAuthorization(publicParameters driver.PublicParameters, walletService
 }
 
 // IsMine returns true if the passed token is owned by an owner wallet.
-// It returns the ID of the owner wallet and no additional owner identifiers
+// It returns the ID of the owner wallet and no additional owner identifiers.
 func (w *WalletBasedAuthorization) IsMine(tok *token2.Token) (string, []string, bool) {
 	wallet, err := w.WalletService.OwnerWallet(tok.Owner.Raw)
 	if err != nil {
--- a/token/core/common/authrorization.go
+++ b/token/core/common/authrorization.go
@@ -14,8 +14,10 @@ import (
 )
 
 type Authorization interface {
-	// IsMine returns true if the passed token is owned by an owner wallet in the passed TMS
-	IsMine(tok *token2.Token) ([]string, bool)
+	// IsMine returns true if the passed token is owned by an owner wallet.
+	// It returns the ID of the owner wallet and any additional owner identifier, if supported.
+	// It is possible that the wallet ID is empty an the additional owner identifier list is not.
+	IsMine(tok *token2.Token) (string, []string, bool)
 	// AmIAnAuditor return true if the passed TMS contains an auditor wallet for any of the auditor identities
 	// defined in the public parameters of the passed TMS.
 	AmIAnAuditor() bool
@@ -41,13 +43,14 @@ func NewTMSAuthorization(publicParameters driver.PublicParameters, walletService
 	return &WalletBasedAuthorization{PublicParameters: publicParameters, WalletService: walletService, amIAnAuditor: amIAnAuditor}
 }
 
-// IsMine returns true if the passed token is owned by an owner wallet in the passed TMS
-func (w *WalletBasedAuthorization) IsMine(tok *token2.Token) ([]string, bool) {
+// IsMine returns true if the passed token is owned by an owner wallet.
+// It returns the ID of the owner wallet and no additional owner identifiers
+func (w *WalletBasedAuthorization) IsMine(tok *token2.Token) (string, []string, bool) {
 	wallet, err := w.WalletService.OwnerWallet(tok.Owner.Raw)
 	if err != nil {
-		return nil, false
+		return "", nil, false
 	}
-	return []string{wallet.ID()}, true
+	return wallet.ID(), nil, true
 }
 
 // AmIAnAuditor return true if the passed TMS contains an auditor wallet for any of the auditor identities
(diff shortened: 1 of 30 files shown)


token/services/db/sql/common/querybuilder.go
```diff
@@ -145,16 +145,17 @@ func movementConditionsSql(params driver.QueryMovementsParams) (where string, ar
...
 	and := []string{"owner = true"}
 	if params.OwnerType != "" {
 		args = append(args, params.OwnerType)
 		and = append(and, fmt.Sprintf("owner_type = $%d", len(args)))
 	}
-	if params.OwnerEnrollmentID != "" {
-		args = append(args, params.OwnerEnrollmentID)
-		and = append(and, fmt.Sprintf("enrollment_id = $%d", len(args)))
+	if params.WalletID != "" {
+		args = append(args, params.WalletID)
```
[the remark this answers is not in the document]
adecaro: I add only once here to get the right index used after

token/services/db/sql/common/querybuilder.go
```diff
@@ -183,6 +184,39 @@ func tokenQuerySql(params driver.QueryTokenDetailsParams, tokenTable, ownerTable
...
+		and = append(and, fmt.Sprintf("owner_wallet_id = $%d", len(args)))
+	}
+
+	if params.TokenType != "" {
+		args = append(args, params.TokenType)
+		and = append(and, fmt.Sprintf("token_type = $%d", len(args)))
+	}
+
+	if len(params.TransactionIDs) > 0 {
+		colTxID := "tx_id"
```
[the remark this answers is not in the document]
adecaro: yes

token/services/db/driver/token.go
```diff
@@ -128,12 +130,12 @@ type TokenDB interface {
 	IsMine(txID string, index uint64) (bool, error)
 	// UnspentTokensIterator returns an iterator over all owned tokens
 	UnspentTokensIterator() (driver.UnspentTokensIterator, error)
-	// UnspentTokensIteratorBy returns an iterator over all tokens owned by the passed identifier of a given type
-	UnspentTokensIteratorBy(ctx context.Context, id, tokenType string) (driver.UnspentTokensIterator, error)
-	// MinTokenInfoIteratorBy returns the minimum information about the tokens needed for the selector
-	MinTokenInfoIteratorBy(ctx context.Context, ownerEID string, typ string) (driver.MinTokenInfoIterator, error)
+	// UnspentTokensIteratorBy returns an iterator over all tokens owned by the passed identifier and of a given type
+	UnspentTokensIteratorBy(ctx context.Context, walletID, tokenType string) (driver.UnspentTokensIterator, error)
```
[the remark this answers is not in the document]
adecaro: yes

token/driver/wallet.go
```diff
@@ -132,8 +132,10 @@ type WalletLookupID = any
 // Authorization defines method to check the relation between a token
 // and wallets (owner, auditor, etc.)
 type Authorization interface {
-	// IsMine returns true if the passed token is owned by an owner wallet in the passed TMS
-	IsMine(tok *token.Token) ([]string, bool)
+	// IsMine returns true if the passed token is owned by an owner wallet.
+	// It returns the ID of the owner wallet and any additional owner identifier, if supported.
+	// It is possible that the wallet ID is empty an the additional owner identifier list is not.
+	IsMine(tok *token.Token) (string, []string, bool)
```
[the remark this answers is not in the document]
adecaro: rewritten

token/services/db/driver/token.go
```diff
@@ -128,12 +130,12 @@ type TokenDB interface {
 	IsMine(txID string, index uint64) (bool, error)
 	// UnspentTokensIterator returns an iterator over all owned tokens
```
arner, line 131: Maybe it's best to also make clear that the UnspentTokensIterator also includes partially owned tokens. If we don't want to change the API, it would suffice to do it in the comment (and also in the UnspentTokensIteratorBy). Maybe something like this?

```
	// UnspentTokensIterator returns an iterator over all tokens that are fully or partially owned by any wallet.
```

token/services/tokens/tokens_test.go
```diff
@@ -27,8 +27,8 @@ type authMock struct{}
 func (a authMock) Issued(issuer driver.Identity, tok *token2.Token) bool {
 	return false
 }
-func (a authMock) IsMine(tok *token2.Token) ([]string, bool) {
-	return []string{string(tok.Owner.Raw)}, true
+func (a authMock) IsMine(tok *token2.Token) (string, []string, bool) {
+	return "", []string{string(tok.Owner.Raw)}, true
```
arner, line 31: is IsMine supposed to return the Raw owner or the wallet id?

arner (approved):
LGTM. One small comment left that would hopefully make it a bit more clear still.

I think it's worthwhile to take another look at the datamodel for tokens in a next iteration. At least we can update the table and column names of the current 'ownership' table for clarity, and maybe do some other refactoring like moving the owner type there and see which fields we really need in the tokens table itself. Then we can have consistent naming throughout, and hopefully someone looking at the database or code would quickly understand the difference between full and shared or 'special' ownership. But we can have that discussion with some diagrams instead of on a PR :)

token/core/common/authrorization.go
```diff
@@ -72,14 +75,14 @@ func NewAuthorizationMultiplexer(ownerships ...Authorization) *AuthorizationMult
 }
 
 // IsMine returns true it there exists an authorization checker that returns true
-func (o *AuthorizationMultiplexer) IsMine(tok *token2.Token) ([]string, bool) {
+func (o *AuthorizationMultiplexer) IsMine(tok *token2.Token) (string, []string, bool) {
```
alexandrosfilios, line 78: We could have an alias `WalletID` to make it clear what it returns

token/services/db/sql/common/condition.go
```diff
@@ -0,0 +1,233 @@
+/*
```
alexandrosfilios, line 1: This file would be in FSC

token/core/common/authrorization.go
```diff
@@ -14,8 +14,10 @@ import (
 )
 
 type Authorization interface {
-	// IsMine returns true if the passed token is owned by an owner wallet in the passed TMS
-	IsMine(tok *token2.Token) ([]string, bool)
+	// IsMine returns true if the passed token is owned by an owner wallet.
+	// It returns the ID of the owner wallet and any additional owner identifier, if supported.
+	// It is possible that the wallet ID is empty and the additional owner identifier list is not.
+	IsMine(tok *token2.Token) (string, []string, bool)
```
alexandrosfilios, line 20: * Would it be a problem if we only passed the owner as a parameter and not the whole token?
* Maybe we could create a new struct
```
type TokenOwner struct {
 Owner view.Id
 AdditionalOwners []view.Id
}
```
and then define methods like `Equals` or `CanSpend` on this?

token/services/db/sql/common/tokens.go
```diff
@@ -866,18 +869,26 @@ func (t *TokenTransaction) GetToken(ctx context.Context, txID string, index uint
 	var tokenType string
 	var quantity string
 	owners := []string{}
+	var walletID *string
```
alexandrosfilios, line 872: Do you mean `string`? Because later you use `&walletID`

alexandrosfilios (approved):
LGTM

token/core/common/authrorization.go
adecaro: It is important that the driver gets the entire token to take a proper decision

token/core/common/authrorization.go
adecaro: okay

token/services/db/driver/token.go
adecaro: okay

token/services/db/sql/common/condition.go
adecaro: what's this?

PR #740, opened 2024-08-22
Demo fixes

alexandrosfilios:


integration/token/fungible/views/issue.go
```diff
@@ -115,7 +115,7 @@ func (p *IssueCashView) Call(context view.Context) (interface{}, error) {
 	// Depending on the token driver implementation, the recipient's signature might or might not be needed to make
 	// the token transaction valid.
 	_, err = context.RunView(ttx.NewCollectEndorsementsView(tx))
-	assert.NoError(err, "failed to sign issue transaction")
+	assert.NoError(err, "failed to sign issue transaction for "+tx.ID())
```
adecaro, line 118: this is like Printf, use coma instead of +

token/validator.go
```diff
@@ -43,8 +43,8 @@ func (c *Validator) UnmarshallAndVerify(ctx context.Context, ledger Ledger, anch
 
 // UnmarshallAndVerifyWithMetadata behaves as UnmarshallAndVerify. In addition, it returns the metadata extracts from the token request
 // in the form of map.
-func (c *Validator) UnmarshallAndVerifyWithMetadata(ctx context.Context, ledger Ledger, anchor string, raw []byte) ([]interface{}, map[string][]byte, error) {
-	actions, meta, err := c.backend.VerifyTokenRequestFromRaw(ctx, ledger.GetState, anchor, raw)
+func (c *Validator) UnmarshallAndVerifyWithMetadata(ctx context.Context, getState driver.GetStateFnc, anchor string, raw []byte) ([]interface{}, map[string][]byte, error) {
```
adecaro, line 46: I would prefer as it was because it is easier in the future to extend it with additional functions, if needed. Wdyt?

adecaro (changes requested):
Please, have a look at the comments.

PR #747, opened 2024-08-30
dbs must be agnostic to any backend implementation

adecaro:
This PR does the following: It removes from the db implementation any reference to the generation of keys used to store tokens in the rwset. The dbs must be agnostic to any backend implementation

### DIFF
--- a/token/services/network/common/rws/translator/translator.go
+++ b/token/services/network/common/rws/translator/translator.go
@@ -202,12 +202,20 @@ func (w *Translator) checkIssue(issue IssueAction) error {
 }
 
 func (w *Translator) checkTransfer(t TransferAction) error {
-	keys, err := t.GetInputs()
+	inputs, err := t.GetInputs()
 	if err != nil {
 		return errors.Wrapf(err, "invalid transfer: failed getting input IDs")
 	}
 	if !t.IsGraphHiding() {
-		for _, key := range keys {
+		rwsetKeys := make([]string, len(inputs))
+		for i, input := range inputs {
+			rwsetKeys[i], err = keys.CreateTokenKey(input.TxId, input.Index)
+			if err != nil {
+				return errors.Wrapf(err, "invalid transfer: failed creating output ID [%v]", input)
+			}
+		}
+
+		for _, key := range rwsetKeys {
 			bytes, err := w.RWSet.GetState(w.namespace, key)
 			if err != nil {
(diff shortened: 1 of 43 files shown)


token/services/network/common/rws/translator/translator.go
```diff
@@ -202,12 +202,20 @@ func (w *Translator) checkIssue(issue IssueAction) error {
...
 
 func (w *Translator) checkTransfer(t TransferAction) error {
-	keys, err := t.GetInputs()
+	inputs, err := t.GetInputs()
 	if err != nil {
 		return errors.Wrapf(err, "invalid transfer: failed getting input IDs")
 	}
 	if !t.IsGraphHiding() {
-		for _, key := range keys {
+		rwsetKeys := make([]string, len(inputs))
```
alexandrosfilios, line 210: do we need this slice? we could merge the two for loops right?

alexandrosfilios: but a better solution might be
```
var keys []string
if t.IsGraphHiding() {
 keys = t.GetSerialNumbers()
} else {
 keys = make([]string, len(inputs))
 ...
}

for _, key := range keys {
 bytes, err := w.RWSet.GetState(w.namespace, key)
 if err != nil {
    return errors.Wrapf(err, "invalid transfer: failed getting state [%s]", key)
 }
 if len(bytes) != 0 {
  return errors.Errorf("invalid transfer: input is already spent [%s:%v]", key, bytes)
 }
}
```

token/services/network/common/rws/translator/translator.go
```diff
@@ -414,9 +418,21 @@ func (w *Translator) commitTransferAction(transferAction TransferAction) error {
...
-	if !graphHiding {
-		for _, id := range ids {
+func (w *Translator) spendTokens(transferAction TransferAction) error {
+	if !transferAction.IsGraphHiding() {
+		ids, err := transferAction.GetInputs()
+		if err != nil {
+			return err
+		}
+		rwsetKeys := make([]string, len(ids))
+		for i, input := range ids {
```
alexandrosfilios, line 428: We could re-use the code in line 211.

alexandrosfilios (approved):
LGTM modulo comments

token/services/network/common/rws/translator/translator.go
adecaro: okay

PR #753, opened 2024-09-06
Fix CI

alexandrosfilios:


integration/token/fungible/tests.go
```diff
@@ -286,7 +287,7 @@ func TestAll(network *integration.Infrastructure, auditorId string, onAuditorRes
 	Eventually(DoesWalletExist).WithArguments(network, issuer, "pineapple", views.IssuerWallet).WithTimeout(1 * time.Minute).WithPolling(15 * time.Second).Should(Equal(false))
 	Eventually(DoesWalletExist).WithArguments(network, alice, "", views.OwnerWallet).WithTimeout(1 * time.Minute).WithPolling(15 * time.Second).Should(Equal(true))
 	Eventually(DoesWalletExist).WithArguments(network, alice, "mango", views.OwnerWallet).WithTimeout(1 * time.Minute).WithPolling(15 * time.Second).Should(Equal(false))
-	IssueCash(network, "", "USD", 110, alice, auditor, true, issuer)
+	IssueSuccessfulCash(network, "", "USD", 110, alice, auditor, true, issuer, endorsers...)
```
adecaro, line 290: Why only here?

alexandrosfilios: Also the other ones are successful, but only here there was a problem. We could do it everywhere

PR #760, opened 2024-09-19
add query parameters to RestoreTMS so it only tries to update status …

arner:
…of not-finalized transactions.


alexandrosfilios (approved):
LGTM

adecaro:
shouldn't deleted recoreds be excluded from the restore?

Issue #684, opened 2024-07-08
upgradability: ensure that the issuer in the metadata is correct


adecaro:
Currently, both the issuer metadata and action carry the identity of the issuer. Though, the code does not check that they match. Therefore, this Github Issue asks for the following:
- `driver.IssueMetadata`, make the `Issuer` field an auditable identity. Adjust serialization as needed; (Addressed by #911 )
- `token.Request#IsValid`, should match the issuer identity in the action with corresponding metadata;
- `audit.Auditor`, should check the issuer of the Issue actions too;
- `ttxdb.TransactionRecords` should extract the enrollment ID of the issuer and add it to the transaction records for an issue.

PR #765, opened 2024-09-25
FSC from #684

adecaro:


mbrandenburger:
closed as #767 upgraded to newer FSC dep

PR #768, opened 2024-09-30
Extracted default PP fetcher to injectable dependency

alexandrosfilios:


mbrandenburger (approved):
LGTM

Injecting of custom fetcher works nicely.

PR #777, opened 2024-11-11
Extract services to allow for re-use

alexandrosfilios:


token/provider.go
```diff
@@ -234,3 +251,11 @@ func GetManagementServiceProvider(sp ServiceProvider) *ManagementServiceProvider
 	}
 	return s.(*ManagementServiceProvider)
 }
+
+func GetTMSNormalizer(sp ServiceProvider) TMSNormalizer {
```
adecaro, line 255: this function is never used

adecaro (approved):
LGTM, modulo comment

PR #779, opened 2024-11-18
spendable tokens

adecaro:
In the context of #785 , this PR tackles the following:
- [x] Extend the tokens service to provide the supported token types
- [x] Introduce a `spendable` flag in the tokensdb to mark the tokens that cannot be spent with the current driver.
- [x] Introduce a `token type` field in the tokensdb to track the type of a token.

### DIFF
--- a/token/core/fabtoken/tokens.go
+++ b/token/core/fabtoken/tokens.go
@@ -19,19 +19,19 @@ import (
 
 type TokensService struct {
 	*common.TokensService
-	OutputTokenType token2.TokenType
+	OutputTokenFormat token2.TokenFormat
 }
 
 func NewTokensService(pp *PublicParams) (*TokensService, error) {
-	supportedTokens, err := supportedTokenTypes(pp.QuantityPrecision)
+	supportedTokens, err := supportedTokenFormat(pp.QuantityPrecision)
 	if err != nil {
 		return nil, errors.WithMessagef(err, "failed getting supported token types")
 	}
-	return &TokensService{TokensService: common.NewTokensService(), OutputTokenType: supportedTokens}, nil
+	return &TokensService{TokensService: common.NewTokensService(), OutputTokenFormat: supportedTokens}, nil
 }
 
 // Deobfuscate returns a deserialized token and the identity of its issuer
-func (s *TokensService) Deobfuscate(output []byte, outputMetadata []byte) (*token2.Token, driver.Identity, token2.TokenType, error) {
+func (s *TokensService) Deobfuscate(output []byte, outputMetadata []byte) (*token2.Token, driver.Identity, token2.TokenFormat, error) {
 	tok := &Output{}
 	if err := tok.Deserialize(output); err != nil {
 		return nil, nil, "", errors.Wrap(err, "failed unmarshalling token")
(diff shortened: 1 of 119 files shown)


alexandrosfilios (commented):
Comments to address

token/core/fabtoken/tokens.go
```diff
@@ -7,33 +7,63 @@ SPDX-License-Identifier: Apache-2.0
...
+		Quantity: tok.Quantity,
+	}, metadata.Issuer, s.TokenTypes[0], nil
+}
 
-	return tok, tokInfo.Issuer, nil
+func (s *TokensService) SupportedTokenTypes() []string {
+	return s.TokenTypes
+}
+
+func SupportedTokenTypes(precisions ...uint64) ([]string, error) {
```
alexandrosfilios, line 55: It would be more clear if we had a type
```
type TokenType struct {
 TokenType
 IdentityType
 Precision
}

func (t *TokenType) Hash() (string, error) {...}
```

token/core/zkatdlog/nogh/transfer.go
```diff
@@ -39,40 +50,46 @@ func NewTransferService(
...
 			LabelNames: []tracing.LabelName{},
 		})),
+		TokenDeserializer: tokenDeserializer,
 	}
 }
 
 // Transfer returns a TransferActionMetadata as a function of the passed arguments
 // It also returns the corresponding TransferMetadata
-func (s *TransferService) Transfer(ctx context.Context, txID string, wallet driver.OwnerWallet, tokenIDs []*token3.ID, outputTokens []*token3.Token, opts *driver.TransferOptions) (driver.TransferAction, *driver.TransferMetadata, error) {
+func (s *TransferService) Transfer(ctx context.Context, txID string, wallet driver.OwnerWallet, tokenIDs []*token2.ID, outputTokens []*token2.Token, opts *driver.TransferOptions) (driver.TransferAction, *driver.TransferMetadata, error) {
```
alexandrosfilios, line 75: We can replace txID with _ to make it clear it is not needed for the transfer.

token/sdk/tms/tms.go
```diff
@@ -41,13 +39,26 @@ func (p *PostInitializer) PostInit(tms driver.TokenManagerService, networkID, ch
...
 	if err := p.auditorManager.RestoreTMS(tmsID); err != nil {
 		return errors.WithMessagef(err, "failed to restore auditor dbs for [%s]", tmsID)
 	}
+
+	// set supported tokens
+	tokens, err := p.tokensProvider.Tokens(tmsID)
+	if err != nil {
+		return errors.WithMessagef(err, "failed to get tokens for [%s]", tmsID)
+	}
+	supportedTokens := tms.TokensService().SupportedTokenTypes()
```
alexandrosfilios, line 58: Would it be maybe better to remove `SetSupportedTokenTypes` and instead get the supported token types on the fly in the token service? This way we can update them only in one place.

token/services/db/sql/common/tcondition.go
```diff
@@ -59,6 +67,15 @@ func (c *tokenInterpreter) HasTokenDetails(params driver.QueryTokenDetailsParams
 	if !params.IncludeDeleted {
 		conds = append(conds, common.ConstCondition("is_deleted = false"))
 	}
+	if params.OnlyNonSpendable {
```
alexandrosfilios, line 70: This allows us to set both properties with unexpected results. We can have an enum property `Spendable` with `TRUE`, `FALSE`, `ANY`.

token/services/db/sql/common/tokens.go
```diff
@@ -1010,17 +1044,18 @@ func (t *TokenTransaction) Delete(ctx context.Context, txID string, index uint64
 	return nil
 }
 
-func (t *TokenTransaction) StoreToken(ctx context.Context, tr driver.TokenRecord, owners []string) error {
```
alexandrosfilios, line 1013: This will not capture anymore the events under the same span, when we create a new context for the token transaction.

```

 tx := NewTokenTransaction(ctx)
 ...
func MyFunc1(tx *TokenTransaction) {
 span, newCtx := tracer.NewSpanFromCtx(ctx, ...)
span.AddEvent("one")

tx.Delete(...)
}

func MyFunc2(tx *TokenTransaction) {
 span, newCtx := tracer.NewSpanFromCtx(ctx, ...)
span.AddEvent("two")

tx.Add(...)
}
```

token/services/db/sql/common/tokens.go
```diff
@@ -1080,6 +1117,55 @@ func (t *TokenTransaction) StoreToken(ctx context.Context, tr driver.TokenRecord
 	return nil
 }
 
+func (t *TokenTransaction) SetSpendable(txID string, index uint64, spendable bool) error {
```
alexandrosfilios, line 1120: We can use the tokenID instead of txID and index maybe. And use TxNum instead of uint64

alexandrosfilios (approved):
LGTM

PR #780, opened 2024-11-20
selector: support for additional configuration

adecaro:
This PR adds support for the following configuration properties (`evictionInterval`, and `cleanupTickPeriod`) :

```yaml
# token selector configuration allows to use different implementations of the token selector
  # the "sherdlock" driver is the default implementation, other possible configurations are: "simple"
  # if empty, the default selector is used
  selector:
    driver: sherdlock
    # tokens might be locked because of an ongoing transaction from the same wallet. Instead of failing immediately, the selector can retry.
    # The interval is the exact amount of seconds (for simple selector) or the max amount of seconds (sherdlock) the selector waits before retrying.
    # default: 5s
    retryInterval: 5s
    # retry to gain a lock on tokens this amount of times before failing the transaction
    numRetries: 3
    # evictionInterval is the period a token can be locked, after which it is forcefully unlocked
    # if evictionInterval is zero, the eviction algorithm is never executed
    evictionInterval: 3m
    # cleanupTickPeriod defines how often the eviction algorithm must be executed
    # if cleanupTickPeriod is zero, the eviction algorithm is never executed
    cleanupTickPeriod: 1m
```

### DIFF
--- a/docs/core-token.md
+++ b/docs/core-token.md
@@ -22,8 +22,12 @@ token:
     retryInterval: 5s
     # retry to gain a lock on tokens this amount of times before failing the transaction
     numRetries: 3
-    evictionInterval: 1m
-    cleanupPeriod: 1m
+    # evictionInterval is the period a token can be locked, after which it is forcefully unlocked
+    # if evictionInterval is zero, the eviction algorithm is never executed
+    evictionInterval: 3m
+    # cleanupTickPeriod defines how often the eviction algorithm must be executed
+    # if cleanupTickPeriod is zero, the eviction algorithm is never executed
+    cleanupTickPeriod: 1m
 
   tms:
     mytms: # unique name of this token management system
--- a/docs/core-token.md
+++ b/docs/core-token.md
@@ -22,6 +22,8 @@ token:
     retryInterval: 5s
     # retry to gain a lock on tokens this amount of times before failing the transaction
     numRetries: 3
+    evictionInterval: 1m
+    cleanupPeriod: 1m
 
   tms:
     mytms: # unique name of this token management system
--- a/token/services/selector/config/driver.go
+++ b/token/services/selector/config/driver.go
@@ -1,0 +1,79 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package config
+
+import (
+	"time"
+
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/selector/driver"
+	"github.com/pkg/errors"
+)
+
+const (
+	defaultDriver            = driver.Sherdlock
+	defaultEvictionInterval  = 3 * time.Minute
+	defaultCleanupTickPeriod = 1 * time.Minute
+	defaultNumRetries        = 3
+	defaultRetryInterval     = 5 * time.Second
+)
+
+type configService interface {
+	UnmarshalKey(key string, rawVal interface{}) error
+}
+
+type Config struct {
+	Driver            driver.Driver `yaml:"driver,omitempty"`
+	RetryInterval     time.Duration `yaml:"retryInterval,omitempty"`
+	NumRetries        int           `yaml:"numRetries,omitempty"`
+	EvictionInterval  time.Duration `yaml:"evictionInterval,omitempty"`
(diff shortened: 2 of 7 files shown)


token/services/selector/config/driver.go
```diff
@@ -0,0 +1,79 @@
...
+import (
+	"time"
+
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/selector/driver"
+	"github.com/pkg/errors"
+)
+
+const (
+	defaultDriver            = driver.Sherdlock
+	defaultEvictionInterval  = 3 * time.Minute
```
arner, line 18: The defaults were:
- cleanupPeriod (evictionInterval): 30 seconds
- cleanupTickPeriod : 1 minute.

I don't know what good defaults would be, maybe something inbetween? We want to give some space for exceptionally slow commits, but not lock tokens for too long if something went wrong... Tick every 90 seconds, clean every lock older than a minute?

token/services/selector/sherdlock/manager.go
```diff
@@ -82,12 +83,12 @@ func (m *manager) Close(id transaction.ID) error {
...
 
 func (m *manager) cleaner() {
-	ticker := time.NewTicker(5 * time.Second) // Change the duration as needed
+	ticker := time.NewTicker(m.cleanupTickPeriod) // Change the duration as needed
 	defer ticker.Stop()
 
 	for range ticker.C {
-		logger.Debugf("cleanup locked tokens with eviction delay of [%s]", m.evictionDelay)
-		if err := m.locker.Cleanup(m.evictionDelay); err != nil {
+		logger.Debugf("cleanup locked tokens with eviction delay of [%s]", m.evictionInterval)
```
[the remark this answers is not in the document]
adecaro: what about something like this

logger.Debugf("release token locks older than [%s]", m.evictionInterval)

? Unless I misunderstood your point.

token/services/selector/config/driver.go
adecaro: yeah, I also don't know. I would leave evictionInternal to 3 minutes and the tick to 90 seconds.

do you think that evictionInternal is a good name? What about lockAge?

token/services/selector/sherdlock/manager.go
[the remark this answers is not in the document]
arner: even better 😄

token/services/selector/config/driver.go
arner: leaseExpiry?

adecaro: perfect

token/services/selector/sherdlock/manager.go
```diff
@@ -82,12 +83,12 @@ func (m *manager) Close(id transaction.ID) error {
 }
 
 func (m *manager) cleaner() {
-	ticker := time.NewTicker(5 * time.Second) // Change the duration as needed
+	ticker := time.NewTicker(m.cleanupTickPeriod) // Change the duration as needed
```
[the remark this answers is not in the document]
adecaro: indeed

token/services/selector/sherdlock/manager.go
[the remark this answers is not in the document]
adecaro: perfect

adecaro:
@arner , please, have a look at the changes. Thanks :)

docs/core-token.md
```diff
@@ -22,6 +22,12 @@ token:
     retryInterval: 5s
     # retry to gain a lock on tokens this amount of times before failing the transaction
     numRetries: 3
+    # leaseExpiry is the period a token can be locked, after which it is forcefully unlocked
+    # if leaseExpiry is zero, the eviction algorithm is never executed
+    leaseExpiry: 3m
+    # leaseCleanupTickPeriod defines how often the eviction algorithm must be executed
+    # if leaseCleanupTickPeriod is zero, the eviction algorithm is never executed
+    leaseCleanupTickPeriod: 90s
```
arner, line 30: Shall we make the example the same as the default value?

arner (approved):
Nice, thanks!

docs/core-token.md
adecaro: yes

PR #783, opened 2024-11-25
optimised sql transaction management 

adecaro:
The goal of this PR is to optimise the use of sql transactions. This means the following:
1. Keep the transaction open for the shortest time possible
2. Lazy load the sql transaction


adecaro:
I'll close this now cause it needs a deeper discussion first

PR #789, opened 2024-12-03
make the network handler for fabric generic

adecaro:
This allows the developer to plugin different ways to manage the fabric namespaces.


integration/nwo/token/fabric/cc/tcc.go
```diff
@@ -17,20 +17,139 @@ import (
...
+	var cc *topology.ChannelChaincode
+	for _, chaincode := range p.Fabric(tms).Topology().Chaincodes {
+		if chaincode.Chaincode.Name == tms.Namespace {
+			cc = chaincode
+			break
+		}
+	}
+	Expect(cc).NotTo(BeNil(), "failed to find chaincode [%s]", tms.Namespace)
+
+	packageDir := filepath.Join(
```
alexandrosfilios, line 114: Maybe we can extract 114-145 to a method

```
func (p *GenericBackend) packageChaincode(tmsID TMSID, cc *topology.ChannelChaincode, chaincodeNameSuffix string) (filename, error)
```

integration/nwo/token/fabric/cc/tcc.go
```diff
@@ -17,20 +17,139 @@ import (
...
+		DefaultTokenChaincodeParamsReplaceSuffix,
+		tokenPlatform,
+	)
+}
+
+func NewGenericBackend(tokenChaincodePath string, tokenChaincodeParamsReplaceSuffix string, tokenPlatform common2.TokenPlatform) *GenericBackend {
+	return &GenericBackend{TokenChaincodePath: tokenChaincodePath, TokenChaincodeParamsReplaceSuffix: tokenChaincodeParamsReplaceSuffix, TokenPlatform: tokenPlatform}
+}
+
+func (p *GenericBackend) PrepareNamespace(entry *fabric.Entry, tms *topology3.TMS) {
```
alexandrosfilios, line 64: Do we need the entry as a parameter? If another backend comes that needs the entry as input, we can add it later to the interface.

alexandrosfilios (commented):
Some comments to address

integration/nwo/token/fabric/cc/tcc.go
adecaro: okay

integration/nwo/token/fabric/cc/tcc.go
adecaro: let's address this later

PR #790, opened 2024-12-03
Translator: Switch from Any to Latest

adecaro:
With these add-ons:
- pass to the approval responder view a translator provider. This will simplify the possibility to change the transaltor.

### DIFF
--- a/token/services/network/fabric/endorsement/approval.go
+++ b/token/services/network/fabric/endorsement/approval.go
@@ -4,7 +4,7 @@ Copyright IBM Corp. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
-package fts
+package endorsement
 
 import (
 	"time"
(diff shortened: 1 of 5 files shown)


token/services/network/fabric/endorsement/approval.go
```diff
@@ -101,8 +93,21 @@ func (r *RequestApprovalView) Call(context view.Context) (interface{}, error) {
 	return tx.Envelope()
 }
 
+type Translator interface {
```
alexandrosfilios, line 96: I don't know how intuitive it is not to pass the RWSet as a param and include it in the fields of the translator. I expect the translator to be a service and act on different RWSets.

token/services/network/fabric/endorsement/approval.go
```diff
@@ -198,13 +203,16 @@ func (r *RequestApprovalResponderView) translate(
 ) error {
 	// prepare the rws as usual
 	txID := tx.ID()
-	w := translator.New(txID, translator.NewRWSetWrapper(&rwsWrapper{stub: rws}, tms.Namespace(), txID), &keys.Translator{})
+	w, err := r.getTranslator(txID, tms.Namespace(), rws)
```
alexandrosfilios, line 206: Can we inject the translator instead of passing it during instantiation?

alexandrosfilios (approved):
LGTM

PR #791, opened 2024-12-04
Extracted functionality for finality listener manager for modularity

alexandrosfilios:


adecaro (approved):
LGTM.
I have just noticed tthat the use of Key Translator is not consistent. 
The driver should decide it for all the other components.
I'll submit a PR after this to fix it.

PR #794, opened 2024-12-09
benchmark for sig service and improvements

adecaro:
This PR does the following:
- Resolve a lock-related bug in RegisterVerifier
- It moves out from the critical zone the storage of the identity

### DIFF
--- a/token/services/identity/sig/sigservice.go
+++ b/token/services/identity/sig/sigservice.go
@@ -66,6 +66,7 @@ func (o *Service) RegisterSigner(identity driver.Identity, signer driver.Signer,
 	if logger.IsEnabledFor(zapcore.DebugLevel) {
 		logger.Debugf("register signer and verifier [%s]:[%s][%s]", idHash, GetIdentifier(signer), GetIdentifier(verifier))
 	}
+	// First check with read lock
 	o.sync.RLock()
 	s, ok := o.signers[idHash]
 	o.sync.RUnlock()
@@ -74,6 +75,7 @@ func (o *Service) RegisterSigner(identity driver.Identity, signer driver.Signer,
 		return nil
 	}
 
+	// write lock
 	o.sync.Lock()
 
 	// check again the cache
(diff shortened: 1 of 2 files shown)


token/services/identity/sig/sigservice.go
```diff
@@ -181,17 +193,17 @@ func (o *Service) IsMe(identity driver.Identity) bool {
 	}
 
 	// last chance, deserialize
```
arner, line 195: This is commented out, maybe it's better to remove it?

token/services/identity/sig/sigservice_test.go
```diff
@@ -0,0 +1,70 @@
...
+	backend, err := kvs.NewWithConfig(d, "", cp)
+	assert.NoError(b, err)
+	err = registry.RegisterService(backend)
+	assert.NoError(b, err)
+	sigService := NewService(NewMultiplexDeserializer(), kvs2.NewIdentityDB(backend, token.TMSID{Network: "pineapple"}))
+	b.StartTimer()
+
+	for i := 0; i < b.N; i++ {
+		// Code to be benchmarked
+		b.StopTimer()
```
arner, line 61: is there a reason that we Stop first and then Start? Why not StopTimer after line 68?

arner (commented):
Nice, thanks! Two small comments

token/services/identity/sig/sigservice_test.go
Grimnir777: To mesure each loop I guess ? Because we start the timer again at the end of the loop

token/services/identity/sig/sigservice.go
adecaro: yes

arner (approved):

token/services/identity/sig/sigservice_test.go
adecaro: okay

PR #795, opened 2024-12-09
Alternatvie implementation for the finality listeners

alexandrosfilios:
This implementation simulates a notification service. It scans the chain from the beginning and informs any listeners have subscribed so far.


adecaro (approved):
LGTM

PR #796, opened 2024-12-09
remove github.com/lib/pq

adecaro:


arner (approved):
Shall we merge it now (if tests succeed) and then make sure the indirect dependency to pq is gone when we upgrade FSC?

PR #799, opened 2024-12-10
log error but don't fail hard if there are no configured identities

arner:
Currently, the identity setup returns an error if there is an issue with loading identities. While that makes sense, it seems that we also need TMS to be able to register a new identity. And TMS doesn't load if the identity setup fails, so it can lead to a catch 22 situation. It's possible that there's a nicer solution (do we really need the larger TMS component to register identities?) but for now just logging the error seems a valid approach too.

### DIFF
--- a/token/services/identity/common/lm.go
+++ b/token/services/identity/common/lm.go
@@ -251,7 +251,8 @@ func (l *LocalMembership) registerIdentityConfiguration(identity *driver.Identit
 		l.logger.Warnf("failed to load local identity at [%s]:[%s]", identity.URL, err)
 		// Does path correspond to a folder containing multiple identities?
 		if err := l.registerLocalIdentities(identity); err != nil {
-			return errors.WithMessage(err, "failed to register local identity")
+			// we don't return the error so that the token manager can still load
+			l.logger.Errorf("failed to register local identity from folder: %s", err.Error())
 		}
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


token/services/identity/common/lm.go
```diff
@@ -251,7 +251,8 @@ func (l *LocalMembership) registerIdentityConfiguration(identity *driver.Identit
 		l.logger.Warnf("failed to load local identity at [%s]:[%s]", identity.URL, err)
 		// Does path correspond to a folder containing multiple identities?
 		if err := l.registerLocalIdentities(identity); err != nil {
-			return errors.WithMessage(err, "failed to register local identity")
+			// we don't return the error so that the token manager can still load
```
adecaro, line 254: Should we allow the caller to decide if the error is fatal or not? For example, this function is called by `RegisterIdentity` which can be called at runtime to register an identity. I would expect to get the error if something wrong happens. No?
I agree that when this function is called at `load` time, then we can just print the errors but continue with the processing. What do you think?

adecaro (changes requested):

token/services/identity/common/lm.go
arner: Good point, changed it now so that when loading all the identities at startup (from database or disk), it won't fail the whole setup if any individual identity fails to load.

adecaro (approved):
LGTM

PR #804, opened 2024-12-12
Update FSC

adecaro:
Testing FSC versions


tools/tools.go
```diff
@@ -15,6 +15,7 @@ import (
 	_ "github.com/gordonklaus/ineffassign"
 	_ "github.com/maxbrunsfeld/counterfeiter/v6"
 	_ "github.com/onsi/ginkgo/v2/ginkgo"
+	_ "github.com/securego/gosec/v2/cmd/gosec"
```
mbrandenburger, line 18: It seems this should be in a separate PR to enable its use via the makefile or even GH action.

mbrandenburger (commented):
Did you forget to add the FSC update in `go.mod` as suggested by the title of this PR.

tools/tools.go
adecaro: good catch, will remove

alexandrosfilios (approved):
LGTM

PR #809, opened 2024-12-16
introduce sql query support

adecaro:


token/services/db/sql/common/query.go
```diff
@@ -0,0 +1,190 @@
...
+const (
+	SelectStatement         = `SELECT`
+	SelectDistinctStatement = `SELECT DISTINCT`
+	InsertStatement         = `INSERT INTO`
+	UpdateStatement         = `UPDATE`
+)
+
+type Select struct {
+	stmt    string
+	columns []string
```
alexandrosfilios, line 23: I would add optionally the table name for each column

token/services/db/sql/common/query.go
```diff
@@ -0,0 +1,190 @@
...
+	from    []string
+	where   string
```
alexandrosfilios, line 25: Where can be a *common.Condition

token/services/db/sql/common/query.go
```diff
@@ -0,0 +1,190 @@
...
+	orderBy string
```
alexandrosfilios, line 26: order by could be a slice of Column

token/services/db/sql/common/query.go
```diff
@@ -0,0 +1,190 @@
...
+	s.from = tables
+	return s
+}
+
+func (s *Select) Where(where string) *Select {
+	s.where = where
+	return s
+}
+
+func (s *Select) Compile() (string, error) {
```
alexandrosfilios, line 53: I would leave compile to an interpreter. Different SQL implementations have different results (e.g. check the common Interpreter and Postgres Interpreter). So each db implementation could be injected its own interpreter like is already done for the conditions.
```
type Interpreter interface {
 Select(s Select) (string, error)
 Insert(I Insert) (string, error)
}
```

token/services/db/sql/common/query.go
```diff
@@ -0,0 +1,190 @@
...
+}
+
+func NewInsertInto(table string) *Insert {
+	return &Insert{
+		stmt:  InsertStatement,
+		table: table,
+	}
+}
+
+func (i *Insert) Rows(rows string) *Insert {
```
alexandrosfilios, line 99: This refers to fields/columns I think. Also we could pass a slice to make it clear.

alexandrosfilios (commented):
Some suggestions

adecaro:
@alexandrosfilios , thanks for the comment. Let's address them in another PR when time will come :)

alexandrosfilios (approved):
We can address these comments at a later stage. LGTM

PR #811, opened 2024-12-17
use uint64 where possible

adecaro:
new mathlib

### DIFF
--- a/token/core/zkatdlog/crypto/rp/bulletproof_test.go
+++ b/token/core/zkatdlog/crypto/rp/bulletproof_test.go
@@ -22,8 +22,8 @@ var _ = Describe("Range Proof", func() {
 		Context("If IPA is generated correctly", func() {
 			It("Succeeds", func() {
 				curve := math.Curves[1]
-				nr := 3
-				l := int(math2.Pow(2, float64(nr)))
+				nr := uint64(3)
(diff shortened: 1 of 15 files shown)


token/core/zkatdlog/crypto/rp/bulletproof_test.go
```diff
@@ -22,8 +22,8 @@ var _ = Describe("Range Proof", func() {
 		Context("If IPA is generated correctly", func() {
 			It("Succeeds", func() {
 				curve := math.Curves[1]
-				nr := 3
-				l := int(math2.Pow(2, float64(nr)))
+				nr := uint64(3)
+				l := uint64(math2.Pow(2, float64(nr)))
```
arner, line 26: This reminds me of https://github.com/hyperledger-labs/fabric-token-sdk/pull/592; float64 is not the same across arm64 and amd64 architectures. Not sure if it matters in this test (and the other one), but we also have it in setup.go. Maybe a good moment to take a look at those to make sure that we're safe?

adecaro: very good catch

adecaro: I have removed the use of math.Pow everywhere.

alexandrosfilios (approved):
LGTM

PR #817, opened 2024-12-27
token conversion via issue

adecaro:
When updating the public parameters, some of the existing tokens might not be convertible on the fly. This PR introduces the possibility for the issuer to redeem these tokens and issue a token for the same amount.


integration/nwo/token/platform.go
```diff
@@ -197,12 +199,16 @@ func (p *Platform) PublicParametersDir() string {
 }
 
 func (p *Platform) PublicParametersFile(tms *topology2.TMS) string {
+	filename := fmt.Sprintf("%s_%s_%s_%s", tms.Network, tms.Channel, tms.Namespace, tms.Driver)
+	if len(tms.Alias) != 0 {
```
alexandrosfilios, line 203: What does this alias stand for?

integration/token/fungible/support.go
```diff
@@ -884,6 +885,20 @@ func GetTMS(network *integration.Infrastructure, networkName string) *topology.T
 	return tms
 }
 
+func GetTMSByAlias(network *integration.Infrastructure, alias string) *topology.TMS {
```
alexandrosfilios, line 888: Maybe it would be better to use a type

```
type Alias string
```
to avoid confusion with the network

integration/token/fungible/views/conversion.go
```diff
@@ -0,0 +1,209 @@
...
+		// have been issued already including the current request.
+		// No check is performed for other types.
+		wallet := token.GetManagementService(context, token.WithTMSID(conversionRequest.TMSID)).WalletManager().IssuerWallet("")
+		assert.NotNil(wallet, "issuer wallet not found")
+
+		// At this point, the issuer is ready to prepare the token transaction.
+		// The issuer creates a new token transaction and specifies the auditor that must be contacted to approve the operation.
+		var tx *ttx.Transaction
+		var auditorID string
+		if len(p.Auditor) == 0 {
```
alexandrosfilios, line 170: We don't need to check whether the tx is unspendable here, right?

adecaro: you mean the tokens? It is up to the issuer to perform all the checks that are relevant

integration/nwo/token/platform.go
adecaro: it is a way to distinguish two TMSs for the same triplet `network, channel, namespace`. 
In the update integration test, I need to generate different public parameters for the same triplet in order to perform the test. The alias allows me to distinguish them

integration/token/fungible/support.go
adecaro: okay

integration/token/fungible/support.go
```diff
@@ -1233,3 +1259,32 @@ func CheckPrometheusMetrics(ii *integration.Infrastructure, viewName string) {
 		Expect(v.Value).NotTo(Equal(model.SampleValue(0)))
 	}
 }
+
+func Conversion(network *integration.Infrastructure, wpm *WalletManagerProvider, user *token3.NodeReference, wallet string, typ token.Type, auditor *token3.NodeReference, issuer *token3.NodeReference, expectedErrorMsgs ...string) string {
```
alexandrosfilios, line 1263: Maybe convert is very generic. it might create confusion and it might also clash with other possible conversions in the future (e.g. between currencies).

maybe we could call it `upgrade` or `promote`.

alexandrosfilios, line 1263: not very important but instead of passing expectedErrorMsgs I think it makes more sense to pass `errorMatchers ...GomegaMatcher`. This way you can have more complex cases that are needed already in some tests like

```
Equal("..."), Contains("..."), Not(Contains("..."))
```

integration/token/fungible/topology/topology.go
```diff
@@ -360,6 +367,22 @@ func Topology(opts common.Opts) []api.Topology {
 	for _, sdk := range opts.SDKs {
 		fscTopology.AddSDK(sdk)
 	}
+
+	// any extra TMS
+	for _, tmsOpts := range opts.ExtraTMSs {
+		tms := tokenTopology.AddTMS(nodeList, backendNetwork, backendChannel, tmsOpts.TokenSDKDriver)
+		tms.Alias = tmsOpts.Alias
+		tms.Namespace = "token-chaincode"
+		tms.Transient = true
```
alexandrosfilios, line 376: What does transient stand for?

token/core/zkatdlog/nogh/tokens.go
```diff
@@ -128,3 +140,79 @@ func supportedTokenFormat(pp *crypto.PublicParams) (token.Format, error) {
 	}
 	return token.Format(hasher.HexDigest()), nil
 }
+
+func (s *TokensService) ProcessTokenConversionRequest(utp *driver.TokenConversionRequest) ([]token.Type, []uint64, error) {
+	if utp == nil {
+		return nil, nil, errors.New("nil token conversion request")
+	}
+
+	ok, err := s.CheckConversionProof(utp.Proof, utp.Proof, utp.Tokens)
```
alexandrosfilios, line 149: `utp.Challenge`

token/driver/issue.go
```diff
@@ -12,10 +12,24 @@ import (
...
+	// Proof is a proof that the prover has solved the challenge
+	Proof []byte
+}
+
 // IssueOptions models the options that can be passed to the issue command
 type IssueOptions struct {
 	// Attributes is a container of generic options that might be driver specific
 	Attributes map[interface{}]interface{}
+	// TokenConversionRequest is a request to convert tokens
+	TokenConversionRequest *TokenConversionRequest
```
alexandrosfilios, line 30: Maybe we don't have to specify the reason that these tokens are there but the objective. That is, we could call them `TokenRedeemRequest`. The same for the interface `CheckRedeemProof` `GenRedeemProof` etc. This could make this field more versatile for other cases where we redeem some tokens to issue others transactionally (e.g. currency conversion).

token/core/zkatdlog/crypto/issue/issue.go
```diff
@@ -14,59 +14,80 @@ import (
...
-type IssueAction struct {
+// Action specifies an issue of one or more tokens
+type Action struct {
 	// Issuer is the identity of issuer
-	Issuer []byte
-	// OutputTokens are the newly issued tokens
-	OutputTokens []*token.Token `protobuf:"bytes,1,rep,name=outputs,proto3" json:"outputs,omitempty"`
+	Issuer driver.Identity
+	// Inputs specify the identifiers in of the tokens to redeem
+	Inputs []*token2.ID
```
alexandrosfilios, line 26: From an interface perspective it would be maybe more clear to have `Inputs` as an aggregation of both id and token
```
Inputs []struct {
 ID *token2.ID
 Token []byte
}
```

token/core/zkatdlog/crypto/issue/issue.go
```diff
@@ -80,34 +101,60 @@ func (i *IssueAction) GetSerializedOutputs() ([][]byte, error) {
...
+		if i.Outputs[j] == nil {
 			return nil, errors.New("invalid issue: there is a nil output")
 		}
-		com[j] = i.OutputTokens[j].Data
+		com[j] = i.Outputs[j].Data
 	}
 	return com, nil
 }
 
 // IsGraphHiding returns false, indicating that fabtoken does not hide the transaction graph
```
alexandrosfilios, line 125: Not relevant to this pr but maybe update the token and say "this implementation of zkat" instead of "fabtoken"

token/core/zkatdlog/crypto/setup.go
```diff
@@ -272,12 +276,19 @@ func (pp *PublicParams) Validate() error {
 	if pp.RangeProofParams == nil {
 		return errors.New("invalid public parameters: nil range proof parameters")
 	}
+	bitLength := pp.RangeProofParams.BitLength
+	if bitLength > 64 {
```
alexandrosfilios, line 280: Do we need this condition? We make the check in the constructor right?

token/core/zkatdlog/nogh/issue.go
```diff
@@ -98,8 +139,24 @@ func (s *IssueService) Issue(ctx context.Context, issuerIdentity driver.Identity
 		return nil, nil, err
 	}
 
+	// prepare inputs to redeem, if any
+	if opts != nil && opts.TokenConversionRequest != nil {
+		tokens := opts.TokenConversionRequest.Tokens
+		if len(tokens) != 0 {
```
alexandrosfilios, line 145: to avoid nesting, we could add this condition to the previous clause
```
if opts != nil && opts.TokenConversionRequest != nil && len(opts.TokenConversionRequest.Tokens) > 0
```

token/core/zkatdlog/nogh/tokens.go
```diff
@@ -128,3 +140,79 @@ func supportedTokenFormat(pp *crypto.PublicParams) (token.Format, error) {
...
+	}
+	if !ok {
+		return nil, nil, errors.New("invalid conversion proof")
+	}
+
+	var tokenTypes []token.Type
+	var tokenValue []uint64
+
+	// which types do we recognize? Any other type is not convertable
+	fabtoken16Type, err := fabtoken2.SupportedTokenFormat(16)
```
alexandrosfilios, line 161: These can also be defined as constants outside of this method
```
const fabtoken16Type = utils.MustGet(fabtoken2.SupportedTokenFormat(16))
```

alexandrosfilios: Is this supposed to be fabtoken? I think the format for zkat was comm.

token/driver/request.go
```diff
@@ -154,8 +160,28 @@ func (m *TokenRequestMetadata) Bytes() ([]byte, error) {
...
 		}
 	}
+
+	// marshal issues
+	issues := make([]IssueMetadataSer, len(m.Issues))
+	for i, issue := range m.Issues {
+		TokenIDs := make([]TokenIDSer, len(issue.TokenIDs))
+		for j, tokenID := range issue.TokenIDs {
+			TokenIDs[j].TxId = tokenID.TxId
+			TokenIDs[j].Index = int(tokenID.Index)
```
alexandrosfilios, line 170: Is there a problem with the serialization of uint64?

token/services/db/sql/common/tokens.go
```diff
@@ -186,6 +186,37 @@ func (db *TokenDB) SpendableTokensIteratorBy(ctx context.Context, walletID strin
...
+		return nil, errors.Wrapf(err, "failed to get unspendable token formats")
+	}
+	logger.Debugf("after filtering we have [%v]", includeFormats)
+
+	span := trace.SpanFromContext(ctx)
+	// now, select the tokens with the list of ledger tokens
+	where, args := common.Where(db.ci.HasTokenDetails(driver.QueryTokenDetailsParams{
+		WalletID:           walletID,
+		TokenType:          tokenType,
+		Spendable:          driver.Any,
```
alexandrosfilios, line 203: For the sake of completeness if we're querying all unspendable tokens with this walletID and tokenType the condition should be

```
WHERE wallet_id = ... AND token_type = ... AND (spendable == true OR format IN (...))
```
Otherwise maybe we could rename this method to `UnspendableFormatTokensIteratorBy`

token/services/db/sql/common/tokens.go
```diff
@@ -949,13 +980,68 @@ func (db *TokenDB) SetSupportedTokenFormats(formats []token.Format) error {
...
+		if err := rows.Scan(&tmp); err != nil {
+			return nil, errors.Wrapf(err, "failed to scan row")
+		}
+
+		logger.Debugf("format from db [%s]", tmp)
+
+		// include tmp only if it is not in supportedFormats
+		format := token.Format(tmp)
+		found := false
+		for _, t := range supportedFormats {
```
alexandrosfilios, line 1027: We can either use `slices.Contains(supportedFormats, token.Format(tmp))` or (in case we expected there to be many supported formats) use
```
supportedFormats collection.Set[token.Format]
```
and then `supportedFormats.Contains(token.Format(tmp))`

token/services/identity/common/lm.go
```diff
@@ -288,8 +292,29 @@ func (l *LocalMembership) registerLocalIdentities(configuration *driver.Identity
 }
 
 func (l *LocalMembership) addLocalIdentity(config *driver.IdentityConfiguration, keyManager KeyManager, defaultID bool) error {
-	eID := keyManager.EnrollmentID()
+	// check for duplicates
 	name := config.ID
+	_, ok := l.localIdentitiesByName[name]
+	if ok && !keyManager.Anonymous() && len(l.targetIdentities) > 0 {
+		// is the identity in the target identities? If no, ignore it.
+		identity, _, err := keyManager.Identity(nil)
```
alexandrosfilios, line 300:
```
if identity, _, err := keyManager.Identity(nil); err != nil {
 return err
} else if found := slices.ContainsFunc(l.targetIdentities, identity.Equal); !found {
 return nil
}
```

alexandrosfilios: A bit more concise and less nesting:
```
if _, ok := l.localIdentitiesByName[config.ID]; !ok || keyManager.Anonymous() || len(l.targetIdentities) == 0 {
 logger.Debugf("no check")
} else if identity, _, err := keyManager.Identity(nil); err != nil {
 return err
} else if found := slices.ContainsFunc(l.targetIdentities, identity.Equal); !found {
 return nil
}
```

token/services/ttx/convert.go
```diff
@@ -0,0 +1,271 @@
...
+	Tokens        []token2.LedgerToken
+	Proof         []byte
+	NotAnonymous  bool
+}
+
+// ConversionInitiatorView is the initiator view to request an issuer the conversion of tokens.
+// The view prepares an instance of ConversionRequest and send it to the issuer.
+type ConversionInitiatorView struct {
+	Issuer       view.Identity
+	TokenType    string
```
alexandrosfilios, line 40: `token.Type`

token/services/ttx/convert.go
```diff
@@ -0,0 +1,271 @@
...
+	if err != nil {
+		logger.Errorf("failed to send recipient data: [%s]", err)
+		return nil, errors.Wrapf(err, "failed to send recipient data")
+	}
+
+	return []interface{}{wr, session.Session()}, nil
+}
+
+// WithWallet sets the wallet to use to retrieve a recipient identity if it has not been passed already
+func (r *ConversionInitiatorView) WithWallet(wallet string) *ConversionInitiatorView {
```
alexandrosfilios, line 145: It is a bit confusing to have two strategies to pass arguments: `With...` and passing them to the constructor. We could either have `opts` or pass them directly. In any case I would force to pass them on instantiation to enforce immutability. Otherwise one might think that WithWallet is meant to be called also after instantiation.

token/services/ttx/convert.go
```diff
@@ -0,0 +1,271 @@
...
+	return r
+}
+
+// WithRecipientData sets the recipient data to use
+func (r *ConversionInitiatorView) WithRecipientData(data *RecipientData) *ConversionInitiatorView {
+	r.RecipientData = data
+	return r
+}
+
+func (r *ConversionInitiatorView) getRecipientIdentity(context view.Context) (*token.TMSID, view.Identity, []byte, []byte, error) {
```
alexandrosfilios, line 162: Maybe rename this method. Or otherwise inline it. Its logic is pretty straightforward and it has too many outputs.

alexandrosfilios (changes requested):
Comments to address

integration/token/fungible/support.go
```diff
@@ -884,6 +885,20 @@ func GetTMS(network *integration.Infrastructure, networkName string) *topology.T
 	return tms
 }
 
+func GetTMSByAlias(network *integration.Infrastructure, alias topology.TMSAlias) *topology.TMS {
+	tp := tplatform.GetPlatform(network.Ctx, "token")
+	Expect(tp).NotTo(BeNil())
+	var tms *topology.TMS
+	for _, TMS := range tp.GetTopology().TMSs {
+		if TMS.Alias == alias {
+			tms = TMS
```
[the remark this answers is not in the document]
adecaro: okay

integration/token/fungible/topology/topology.go
adecaro: I'll add a comment, but it means that it is not an actual TMS, it is a TMS for which we need to generate only the public parameters

token/core/zkatdlog/nogh/tokens.go
adecaro: I'll introduce types to avoid this issue

token/driver/issue.go
adecaro: in this case, I think it is better to clarify the reason to avoid confusion. Different operations might have different requirements. Better to explicitly differentiate.

token/services/ttx/convert.go
adecaro: okay

token/services/ttx/convert.go
adecaro: indeed, they can also after initialization

token/services/ttx/convert.go
adecaro: but inlining makes the calling method bigger, no?

token/services/db/sql/common/tokens.go
```diff
@@ -949,13 +980,68 @@ func (db *TokenDB) SetSupportedTokenFormats(formats []token.Format) error {
...
+	if err := rows.Err(); err != nil {
+		return nil, errors.Wrapf(err, "error querying db")
+	}
+	defer rows.Close()
+	// read the types from the query result and remove discard those in db.getSupportedTokenFormats()
+	supportedFormats := db.getSupportedTokenFormats()
+	logger.Debugf("supported token formats are [%v]", supportedFormats)
+	includeFormats := make([]token.Format, 0)
+	for rows.Next() {
+		var tmp string
```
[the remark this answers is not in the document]
adecaro: there is already a var with that name after...

token/services/db/sql/common/tokens.go
adecaro: okay

token/driver/request.go
adecaro: yes, asn1 does not support it

token/core/zkatdlog/nogh/tokens.go
adecaro: the dlog driver supports the conversion from fabtoken, it might support more. To arrive later.

token/core/zkatdlog/nogh/issue.go
adecaro: okay

token/core/zkatdlog/crypto/issue/issue.go
adecaro: okay

token/services/ttx/convert.go
alexandrosfilios: It's a matter of taste. They are simple calls in this method without any data calculations. If they had more calculations that all produce one result, I would definitely extract them. But still whatever is better for you. In this case we can just rename this method to reflect what it gets.

token/core/zkatdlog/crypto/issue/issue.go
adecaro: okay

integration/token/fungible/support.go
adecaro: went for `upgrade`

token/core/zkatdlog/nogh/issue.go
```diff
@@ -98,8 +139,25 @@ func (s *IssueService) Issue(ctx context.Context, issuerIdentity driver.Identity
 		return nil, nil, err
 	}
 
+	// prepare inputs to redeem, if any
+	var tokenIDs []*token.ID
+	if opts != nil && opts.TokenConversionRequest != nil && len(opts.TokenConversionRequest.Tokens) > 0 {
+		tokens := opts.TokenConversionRequest.Tokens
```
alexandrosfilios, line 145: TokenConversionRequest we can rename as well

token/core/zkatdlog/nogh/tokens.go
```diff
@@ -128,3 +140,79 @@ func supportedTokenFormat(pp *crypto.PublicParams) (token.Format, error) {
 	}
 	return token.Format(hasher.HexDigest()), nil
 }
+
+func (s *TokensService) ProcessTokenConversionRequest(utp *driver.TokenConversionRequest) ([]token.Type, []uint64, error) {
```
alexandrosfilios, line 144: Upgrade

token/request.go
```diff
@@ -380,6 +380,57 @@ func (r *Request) Redeem(ctx context.Context, wallet *OwnerWallet, typ token.Typ
 	return nil
 }
 
+func (r *Request) Convert(
```
alexandrosfilios, line 383: Upgrade

alexandrosfilios (approved):
LGTM

PR #821, opened 2025-01-09
Make tx type for delivery mapper configurable

alexandrosfilios:


adecaro (changes requested):
Please, move the configuration section under
```
     services:
        # This section contains network specific configuration
        network:
          # Configuration related to the Fabric network
          fabric:
```
It belongs there.

PR #823, opened 2025-01-09
metadata cleanup

adecaro:
In this PR, we get rid of the `Outputs` fields in the metadata. It was just redundant. In addition to this, we add checks to make sure the metadata fields are verifiable. Notice that, some of these fields can only be verified by an auditor.


alexandrosfilios (approved):
LGTM

PR #824, opened 2025-01-13
local tokens upgrade

adecaro:
This PR brings in the ability to upgrade a token in place when possible. It is the token driver implementation that decides when this is possible. The validator provided by the same token driver must validate the upgrade witness.


alexandrosfilios (approved):
LGTM modulo comments

PR #826, opened 2025-01-15
rotate idemix issuer pk

adecaro:
This PR introduces the possibility to rotate the idemix issuer pk. This means  that multiple public keys for the same idemix issuer can co-exist. The dlog driver does not support the possibility to upgrade from a non-supported idemix issuer public key.


cmd/tokengen/main_test.go
```diff
@@ -41,6 +40,7 @@ func TestGenFullSuccess(t *testing.T) {
 
 	tempOutput, err := os.MkdirTemp("", "tokengen-test")
 	gt.Expect(err).NotTo(HaveOccurred())
+	// tempOutput := "./testdata"
```
alexandrosfilios, line 43: remember to remove

token/core/zkatdlog/nogh/driver/base.go
```diff
@@ -81,12 +84,37 @@ func (d *base) newWalletService(
 		deserializerManager,
 		ignoreRemote,
 	)
+	// owner role
+	// we have one key manager for fabtoken and one for each idemix issuer public key
+	var kmps []common2.KeyManagerProvider
+	for _, key := range pp.IdemixIssuerPublicKeys {
```
alexandrosfilios, line 90: Is there a callback when we change pp, in order to create a new key manager and remove the ones that are not relevant anymore?

token/core/zkatdlog/nogh/driver/base.go
```diff
@@ -81,12 +84,37 @@ func (d *base) newWalletService(
 		deserializerManager,
 		ignoreRemote,
 	)
+	// owner role
+	// we have one key manager for fabtoken and one for each idemix issuer public key
+	var kmps []common2.KeyManagerProvider
```
alexandrosfilios, line 89: since you know the exact size its better to initialize it with the length of pp.IdemixIssuerPublicKeys. This makes it a bit more clear that they are 1-1 mapped.

token/services/auditor/auditor.go
```diff
@@ -146,3 +147,35 @@ func (a *Auditor) GetTokenRequest(txID string) ([]byte, error) {
 func (a *Auditor) Check(context context.Context) ([]string, error) {
 	return a.checkService.Check(context)
 }
+
+func (a *Auditor) completeInputsWithEmptyEID(record *token.AuditRecord) error {
+	filter := record.Inputs.ByEnrollmentID("")
```
alexandrosfilios, line 152: Why would an input have an empty enrollment id?

token/services/identity/msp/rolefactory.go
```diff
@@ -85,27 +85,34 @@ func NewRoleFactory(
...
+			issuerPublicKey.PublicKey,
+			issuerPublicKey.Curve,
+			RoleToMSPID[role],
+			keyStore,
+			f.SignerService,
+			f.Config,
+			cacheSize,
+			f.ignoreRemote,
+		)
+		kmps = append([]common.KeyManagerProvider{kmp}, additionalKMPs...)
```
alexandrosfilios, line 114: The reason you didn't do `kmps = append(kmps, kmp)` is because you would modify the additionalKMPs slice reference?

alexandrosfilios (commented):
Some comments to address

token/services/identity/msp/rolefactory.go
adecaro: no, it is the order. The order gives priority to the different kmps

token/services/auditor/auditor.go
adecaro: When redeeming tokens during an issue. These tokens might not be parsable by the current token driver.

token/services/db/sql/common/identity.go
```diff
@@ -345,10 +345,10 @@ func (db *IdentityDB) GetSchema() string {
 			url TEXT NOT NULL,
 			conf BYTEA,
 			raw BYTEA,
-			PRIMARY KEY(id, type)
+			PRIMARY KEY(id, type, url)
```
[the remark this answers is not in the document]
adecaro: yes, one id can have multiple configurations. Each configuration might work with a different token driver.

token/core/zkatdlog/nogh/driver/base.go
adecaro: okay

token/core/zkatdlog/nogh/driver/base.go
adecaro: yes, when the public params change a new TMS is loaded and together with it a new wallet service that reloads all the keys.

cmd/tokengen/main_test.go
adecaro: okay

token/services/identity/common/lm.go
```diff
@@ -159,47 +164,79 @@ func (l *LocalMembership) Load(identities []*config.Identity, targets []view.Ide
...
+	}
+	storedIdentityConfigurations, err := l.storedIdentityConfigurations()
+	if err != nil {
+		return errors.Wrap(err, "failed to load stored identity configurations")
+	}
+
+	// merge identityConfigurations and storedIdentityConfigurations
+	// filter out stored configuration that are already in identityConfigurations
+	var filtered []driver.IdentityConfiguration
+	if len(storedIdentityConfigurations) != 0 {
```
alexandrosfilios, line 187: This if is not needed

token/services/identity/common/lm.go
```diff
@@ -159,47 +164,79 @@ func (l *LocalMembership) Load(identities []*config.Identity, targets []view.Ide
...
+		for _, stored := range storedIdentityConfigurations {
+			found := false
+			// if stored is in identityConfigurations, skip it
+			for _, identityConfiguration := range identityConfigurations {
+				if stored.ID == identityConfiguration.ID && stored.URL == identityConfiguration.URL {
+					// we don't need this configuration
+					found = true
```
alexandrosfilios, line 194: Here we can break the loop.

token/services/identity/common/lm.go
adecaro: okay

token/services/identity/common/lm.go
adecaro: okay

alexandrosfilios (approved):
LGTM

PR #831, opened 2025-01-19
OnwerWallet GetRecipientData

adecaro:
introduce GetRecipientData for owner wallet to avoid to retrieve audt info and token related metadata in seperated calls


alexandrosfilios (approved):
LGTM

PR #835, opened 2025-01-26
driver cleanup

adecaro:
This PR does the following:
- It remove an interface from the driver package that was only used in fabric network service


alexandrosfilios (approved):
LGTM

PR #838, opened 2025-01-29
proto messages

adecaro:
This PR is the first in train of PRs whose goal is to introduce protobuf message to marshal core data structure as `TokenRequest`, `TokenRequestMetadata`, `TokenRequestWithMetadata`, and `PublicParameters`. The code has been updated to make sure signatures are generated on fixed bytes and not on regenerated bytes. 

In the coming PRs, we will further structure `TokenRequestMetadata` to group input and output related fields. We will rework the `ttx` package to get rid of the TODOs and introduce protobuf messages where needed.


alexandrosfilios (approved):
LGTM

token/driver/protos/request.proto
```diff
@@ -0,0 +1,93 @@
...
+  repeated Action actions = 2; // List of token actions to perform
+  repeated Signature signatures = 3; // Signatures for the actions
+  repeated Signature auditor_signatures = 4; // Additional signatures from auditors
+}
+
+message TokenRequestWithMetadata {
+  uint32 version = 1; // Version number
+  string anchor = 2; // Request anchor
+  TokenRequest request = 3; // the request
+  TokenRequestMetadata metadata = 4; // the corresponding metadata
```
alexandrosfilios, line 92: Isn't there one metadata per action? If yes, it's better to add it under the `Action` instead of having two slices as separate fields.

token/driver/request.go
```diff
@@ -108,145 +192,194 @@ type TokenRequestMetadata struct {
...
-		TokenIDs := make([]TokenIDSer, len(transfer.TokenIDs))
+	for _, transfer := range m.Transfers {
+		tokenIDs := make([]*request.TokenID, len(transfer.TokenIDs))
 		for j, tokenID := range transfer.TokenIDs {
-			TokenIDs[j].TxId = tokenID.TxId
-			TokenIDs[j].Index = int(tokenID.Index)
+			if tokenID == nil {
+				return nil, errors.Errorf("nil token ID at index [%d]", j)
+			}
+			tokenIDs[j] = &request.TokenID{}
```
alexandrosfilios, line 216: You can define the method `ToProtos' and `FromProtos` for each subtype:
```
func (t *TokenID) ToProtos() (*request.TokenID, error) {
 return &request.TokenID{...}
}
func (I *Identity) ToProtos() (*request.Identity, error) {
 return &request.Identity{...}
}
type ProtoSource[T any] interface {
 ToProtos() (T, error)
}
func ToProtosSlice[T any, S ProtoSource[T]](s []S) ([]T, error) {

}
```
This way you write more modular code that can be re-used (e.g. tokenID, slices, identities)

```
func (m *TokenRequestMetadata) ToProtos() (*requestTokenRequestMetadata, error) {
 transfers, err := ToProtosSlice(m.Transfers)
 metadata, err := ToProtosSlice(m.Metadata)
 issues, err := ToProtosSlice(m.Issues)
 ...
 return &request.TokenRequestMetadata{
  Version: Version,
  Metadata: metadata,
  Application: m.Application,
  Transfers: transfers,
  Issues: issues,
}
}
```

alexandrosfilios: The same for `FromProtos`

token/driver/protos/request.proto
adecaro: they cannot go together because metadata needs to be stripped out from the TokenRequest, so better to keep them seperated.

token/driver/request.go
adecaro: Not sure, I don't understand why the TokenID struct should know anything about protos.

PR #839, opened 2025-01-30
tcc: query states function

adecaro:
This PR does the following: It introduces a new function to the token chaincode that allows the developer to query any state in the token chaincode's namespace.


alexandrosfilios (approved):
LGTM

PR #840, opened 2025-01-30
lookup listener manager

adecaro:
This PR does the following: Similarly to the listener manager, the fabric network driver uses a lookup listener manager to implement `LookupTransferMetadataKey`. This way one can choose which strategy to use for the looking up of the key.

### DIFF
--- a/integration/token/common/topology.go
+++ b/integration/token/common/topology.go
@@ -42,6 +42,7 @@ type Opts struct {
 	OnlyUnity           bool
 	FSCBasedEndorsement bool
 	ExtraTMSs           []TMSOpts
+	FinalityType        string
 }
 
 func SetDefaultParams(tms *topology.TMS, opts TMSOpts) {
--- a/integration/token/common/topology.go
+++ b/integration/token/common/topology.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/api"
 	"github.com/hyperledger-labs/fabric-token-sdk/integration/nwo/token/generators/dlog"
 	"github.com/hyperledger-labs/fabric-token-sdk/integration/nwo/token/topology"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/network/fabric/config"
 	. "github.com/onsi/gomega"
 )
 
@@ -42,7 +43,7 @@ type Opts struct {
 	OnlyUnity           bool
 	FSCBasedEndorsement bool
 	ExtraTMSs           []TMSOpts
-	FinalityType        string
+	FinalityType        config.ManagerType
 }
 
 func SetDefaultParams(tms *topology.TMS, opts TMSOpts) {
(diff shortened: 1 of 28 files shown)


integration/token/common/topology.go
```diff
@@ -42,6 +42,7 @@ type Opts struct {
 	OnlyUnity           bool
 	FSCBasedEndorsement bool
 	ExtraTMSs           []TMSOpts
+	FinalityType        string
```
alexandrosfilios, line 45: We can use the type that exists already `finality.ManagerType` and the two constants `finality.Delivery` `finality.Committer`

token/services/network/fabric/lookup/channelflm.go
```diff
@@ -0,0 +1,185 @@
...
+
+		logger.Debugf("scanning [%s]...", tx.TxID())
+
+		rws, err := v.InspectRWSet(tx.Results())
+		if err != nil {
+			return false, err
+		}
+
+		found := false
+		for _, ns := range rws.Namespaces() {
```
alexandrosfilios, line 138:
```
if !slices.Contains(rws.Namespaces(), s.namespace) {
 logger.Debugf(...)
 return false, nil
}
```

token/services/network/fabric/lookup/channelflm.go
```diff
@@ -0,0 +1,185 @@
...
+		delete(c.listeners, id)
+	}
+	return nil
+}
+
+type Scanner struct {
+	context      context.Context
+	channel      *fabric.Channel
+	namespace    string
+	key          string
```
alexandrosfilios, line 100: it might make sense to accept multiple keys here so that we can batch the scans and not launch one scan per key.

token/services/network/fabric/lookup/deliveryllm.go
```diff
@@ -0,0 +1,185 @@
...
+	return other != nil && other.(*listenerEntry).listener == e.listener
+}
+
+type TxInfo struct {
+	Namespace driver2.Namespace
+	Key       driver2.TxID
+	Value     []byte
+}
+
+func (i TxInfo) TxID() driver2.TxID {
```
alexandrosfilios, line 61: Maybe we should rename this interface method to LookupID

adecaro: we cannot because we must `TxID` to match the expected interface

token/services/network/fabric/lookup/channelflm.go
adecaro: okay

token/services/network/fabric/lookup/channelflm.go
```diff
@@ -0,0 +1,185 @@
...
+		s.Stop()
+		delete(c.listeners, id)
+	}
+	return nil
+}
+
+type Scanner struct {
+	context      context.Context
+	channel      *fabric.Channel
+	namespace    string
```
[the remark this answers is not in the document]
adecaro: okay

token/services/network/fabric/lookup/channelflm.go
adecaro: okay

integration/token/common/topology.go
adecaro: okay

PR #851, opened 2025-02-10
inputs, outsputs, and metadata rework

adecaro:
and output with different recipients should generate an request outptts with the same index and the specified recipients


token/core/fabtoken/issue.go
```diff
@@ -62,34 +58,41 @@ func (s *IssueService) Issue(ctx context.Context, issuerIdentity driver.Identity
 			Quantity: q.Hex(),
 		})
 
-		meta := &OutputMetadata{
+		outputMetadata := &OutputMetadata{
```
alexandrosfilios, line 61: `outputMetadataRaw` can be calculated outside of the loop if I see it correctly.
Then we can also create a method to fetch multiple owner audit info at the same time in order to
* allow batching (because each of them is a db request)
* allow re-use with the transfer logic
```
func (s *IssueService) GetAuditableIdentities(owners []view.Identity) ([]*driver.AuditableIdentity, error)
```
Also it's better to define the length of the outputsMetadata from the beginning to make it more obvious that it's a 1-1 mapping. Then the loop would look something like this
```
outputsMetadata := make([]*driver.IssueOutputMetadata, len(values))
for i, amount := range values {
 quantity, err := ...
 if err != nil ...
 outs[i] = &Output{...}
 outputsMetadata[i] = &driver.IssueOutputMetadata{outputMetadataRaw, auditableIdentities[I]}
}
```

token/core/fabtoken/transfer.go
```diff
@@ -50,108 +50,109 @@ func (s *TransferService) Transfer(ctx context.Context, _ string, _ driver.Owner
...
+				return nil, nil, errors.Wrap(err, "failed getting recipients")
+			}
+			receivers = append(receivers, recipients...)
+			audiInfo, err := s.Deserializer.GetOwnerAuditInfo(output.Owner, ws)
+			if err != nil {
+				return nil, nil, errors.Wrapf(err, "failed getting audit info for sender identity [%s]", driver.Identity(output.Owner).String())
+			}
+			receiversAuditInfo = append(receiversAuditInfo, audiInfo)
+		}
+		for i, receiver := range receivers {
```
alexandrosfilios, line 125: this would become
```
receivers, err := s.GetOwnerAuditableIdentities(recipients)
transferOutputMetadata := &driver.TransferOutputMetadata{
 OutputMetadata: outputMetadataRaw,
 Receivers: receivers,
}
```
It's also a bit more clear if we instantiate the struct with all its values instead of modifying it later. The more immutable, the easier to understand.

token/core/fabtoken/issue.go
```diff
@@ -62,34 +58,41 @@ func (s *IssueService) Issue(ctx context.Context, issuerIdentity driver.Identity
...
-	}
 
 	meta := &driver.IssueMetadata{
-		Issuer:              issuerIdentity,
-		OutputsMetadata:     outputsMetadata,
-		Receivers:           []driver.Identity{driver.Identity(owners[0])},
-		ReceiversAuditInfos: auditInfo,
-		ExtraSigners:        nil,
+		Issuer:       issuerIdentity,
+		Inputs:       nil,
```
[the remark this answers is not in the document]
adecaro: yes

token/core/fabtoken/issue.go
adecaro: regarding `GetAuditableIdentities`, in practice, you are not batching much there. Most of the time you have a single recipient per output and only very few outputs. It doesn't look as much of a gain.
Essentially, we have many scattered invocations of `GetAuditInfo` due to the constructions of multiple transactions. What if the backend batches transparently the requests? We could gain when the system is under high load.
What do you think?
For other comments, it is okay.

token/core/fabtoken/transfer.go
adecaro: okay

token/metadata.go
```diff
@@ -43,124 +48,135 @@ func (m *Metadata) SpentTokenID() []*token.ID {
...
+			found := false
+			for _, receiver := range output.Receivers {
+				// If the receiver has the given enrollment ID, add it
+				recipientEID, err := m.WalletService.GetEnrollmentID(receiver.Identity, receiver.AuditInfo)
+				if err != nil {
+					return nil, errors.Wrap(err, "failed getting enrollment ID")
+				}
+				if eIDSet.Contains(recipientEID) {
+					logger.Debugf("eid [%s] found in list [%v]", recipientEID, eIDs)
+					skip = false
```
alexandrosfilios, line 119: The variable `skip` can be replaced with `counter == 0`

token/metadata.go
```diff
@@ -43,124 +48,135 @@ func (m *Metadata) SpentTokenID() []*token.ID {
...
 		// if the receiver has the given enrollment ID, add it. Otherwise, add empty entries
 		skip := true
-		for i, auditInfo := range transfer.ReceiverAuditInfos {
-			recipientEID, err := m.WalletService.GetEnrollmentID(transfer.Receivers[i], auditInfo)
-			if err != nil {
-				return nil, errors.Wrap(err, "failed getting enrollment ID")
+		counter := 0
+		for _, output := range transfer.Outputs {
+			found := false
+			for _, receiver := range output.Receivers {
```
alexandrosfilios, line 111: We can reuse the logic for the filtering (in issues and transfers):

```
			if found, err := m.contains(output.Receivers, eIDSet); err != nil {
				return nil, err
			} else if found {
				clone.Outputs = append(clone.Outputs, output)
				counter++
			} else {
				clone.Outputs = append(clone.Outputs, nil)
			}


```

where
```
func (m *Metadata) contains(receivers []*driver.AuditableIdentity, eIDSet collections.Set[string]) (bool, error) {
	for _, receiver := range receivers {
		// If the receiver has the given enrollment ID, add it
		recipientEID, err := m.WalletService.GetEnrollmentID(receiver.Identity, receiver.AuditInfo)
		if err != nil {
			return false, errors.Wrap(err, "failed getting enrollment ID")
		}
		if eIDSet.Contains(recipientEID) {
			logger.Debugf("eid [%s] found in list [%v]", recipientEID, eIDSet)
			return true, nil
		} else {
			logger.Debugf("eid [%s] not found in list [%v]", recipientEID, eIDSet)
		}
	}
	return false, nil
}
```

token/metadata.go
```diff
@@ -43,124 +48,135 @@ func (m *Metadata) SpentTokenID() []*token.ID {
...
-
-			transferRes.Receivers = append(transferRes.Receivers, receiver)
-			transferRes.ReceiverAuditInfos = append(transferRes.ReceiverAuditInfos, receiverAuditInfos)
 		}
 
 		// if skip = true, it means that this transfer does not contain any output for the given enrollment IDs.
 		// Therefore, no metadata should be given to the passed enrollment IDs.
 		// if skip = false, it means that this transfer contains at least one output for the given enrollment IDs.
 		// Append the senders to the transfer metadata.
+		for range transfer.Inputs {
```
alexandrosfilios, line 138: We can skip this loop
```
inputs := make([]any, length(transfer.Inputs))
if counter > 0 {
 for I, input := range transfer.Inputs {
  clone.Inputs[I] = &TransferInputMetadata{Senders: input.Senders}
 }
}
```

token/metadata.go
```diff
@@ -43,124 +48,135 @@ func (m *Metadata) SpentTokenID() []*token.ID {
...
+	}
+
+	clonedMetadata := &Metadata{
 		TokenService:         m.TokenService,
 		WalletService:        m.WalletService,
 		TokenRequestMetadata: &driver.TokenRequestMetadata{},
 		Logger:               m.Logger,
 	}
-
+	eIDSet := collections.NewSet(eIDs...)
```
alexandrosfilios, line 61: We can split this method to smaller pieces
```
issues, err := m.filterIssues(m.TokenRequestMetadata.Issues, eIDSet)
transfers, err := m.filterTransfers(m.TokenRequestMetadata.Transfers, eIDSet)
return &Metadata{
 TokenService: m.TokenService,
 WalletService: m.WalletService,
 Logger: m.Logger,
 TokenRequestMetadata: &driver.TokenRequestMetadata{
  Issues: issues,
  Transfers: transfers,
  Application: m.TokenRequestMetadata.Application
 }
}, nil
```

token/metadata.go
```diff
@@ -225,55 +243,42 @@ func (m *TransferMetadata) Match(action *TransferAction) error {
...
+
+	// extra signer
 	extraSigner := action.a.ExtraSigners()
 	if len(m.ExtraSigners) != len(extraSigner) {
-		return errors.Errorf("expected [%d] extra signers but got [%d]", len(extraSigner), len(m.ExtraSigners))
+		return errors.Errorf("expected [%d] extra signers but got [%d]", len(m.ExtraSigners), len(extraSigner))
 	}
 	// check that the extra signers are the same
 	for i, signer := range extraSigner {
 		if !signer.Equal(m.ExtraSigners[i]) {
```
alexandrosfilios, line 263:
```
if !slices.ContainsFunc(m.ExtraSigners, signer.Equal) {return err}
```

token/services/identity/deserializer/typed.go
```diff
@@ -23,7 +23,7 @@ var logger = logging.MustGetLogger("token-sdk.services.identity.deserializer")
 type TypedVerifierDeserializer interface {
 	DeserializeVerifier(typ string, raw []byte) (driver.Verifier, error)
 	Recipients(id driver.Identity, typ string, raw []byte) ([]driver.Identity, error)
-	GetOwnerAuditInfo(id driver.Identity, typ string, raw []byte, p driver.AuditInfoProvider) ([][]byte, error)
+	GetOwnerAuditInfo(id driver.Identity, typ string, raw []byte, p driver.AuditInfoProvider) ([]byte, error)
```
alexandrosfilios, line 26: We could use `identity.Type` for the `typ` input of these 3 methods.

alexandrosfilios (approved):
LGTM modulo comments

token/core/zkatdlog/nogh/issue.go
```diff
@@ -145,26 +127,52 @@ func (s *IssueService) Issue(ctx context.Context, issuerIdentity driver.Identity
...
 		}
 	}
 
+	var outputsMetadata []*driver.IssueOutputMetadata
+	for _, meta := range zkOutputsMetadata {
+		raw, err := meta.Serialize()
+		if err != nil {
+			return nil, nil, errors.WithMessage(err, "failed serializing token info")
+		}
+		auditInfo, err := s.Deserializer.GetOwnerAuditInfo(owners[0], s.WalletService)
```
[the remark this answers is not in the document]
adecaro: This logic was flawed. I fixed it.

token/core/zkatdlog/nogh/transfer.go
```diff
@@ -130,9 +130,6 @@ func (s *TransferService) Transfer(ctx context.Context, txID string, _ driver.Ow
 	}
 	var values []uint64
```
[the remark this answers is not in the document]
adecaro: okay

token/metadata.go
adecaro: okay

token/services/identity/deserializer/typed.go
adecaro: okay

token/metadata.go
adecaro: okay

token/metadata.go
adecaro: okay

token/metadata.go
adecaro: okay

PR #854, opened 2025-02-13
cleanup the use of the smart client's KVS

adecaro:
To goal of this PR is to avoid reliance on external database, by default.
We are still using in non-testing cases the FSC's KVS to store token-sdk related information.
In a subsequent PR, we will get rid of the use of the KVS by introducing support by the token-sdk dbs.


integration/token/common/sdk/identity/provider.go
```diff
@@ -0,0 +1,34 @@
...
+package identity
+
+import (
+	"github.com/hyperledger-labs/fabric-token-sdk/token"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/db/driver"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/identity"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/storage/kvs"
+)
+
+type KVSStorageProvider struct {
```
alexandrosfilios, line 16: To put things into perspective, we could implement this interface that exists already and is implemented by `unity`, `sql` and `memory`. Then KVS would be an alternative implementation to it (of course what is not implemented can panic, but it will still be clear what it is about).
```
type Driver interface {
	NewTokenLock(opts common.Opts) (TokenLockDB, error)

	NewWallet(opts common.Opts) (WalletDB, error)

	NewIdentity(opts common.Opts) (IdentityDB, error)

	NewToken(opts common.Opts) (TokenDB, error)

	NewTokenNotifier(opts common.Opts) (TokenNotifier, error)

	NewAuditTransaction(opts common.Opts) (AuditTransactionDB, error)

	NewOwnerTransaction(opts common.Opts) (TokenTransactionDB, error)
}
```

alexandrosfilios: Otherwise maybe we don't need the walletdb and the identitydb, since we could use the memory implementation instead.

alexandrosfilios (commented):
LGTM modulo comments

alexandrosfilios (approved):

integration/token/common/sdk/identity/provider.go
adecaro: good point, please, open a github issue for this 👍

PR #855, opened 2025-02-13
use the new FSC events listener

adecaro:
This PR does the following: For Fabric and when possible, It uses the new FSC events listener to listen to events related to the update of setup key. This allows the token-sdk stack to avoid any dependency from the FSC's commit pipeline.

### DIFF
--- a/token/services/network/fabric/lookup/deliveryllm.go
+++ b/token/services/network/fabric/lookup/deliveryllm.go
@@ -15,8 +15,8 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/collections"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/committer"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/events"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/fabricutils"
-	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/finality"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/rwset"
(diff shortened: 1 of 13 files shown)


token/services/network/fabric/lookup/deliveryllm.go
```diff
@@ -15,19 +15,27 @@ import (
...
 	"github.com/hyperledger/fabric-protos-go/common"
 	"github.com/pkg/errors"
 	"go.opentelemetry.io/otel/trace"
 )
 
-type newTxInfoMapper = func(network, channel string) finality.TxInfoMapper[TxInfo]
+type newTxInfoMapper = func(network, channel string) events.EventInfoMapper[KeyInfo]
+
+type EventsListenerManager interface {
+	AddPermanentEventListener(key string, e events.ListenerEntry[KeyInfo]) error
```
alexandrosfilios, line 35: `PKey`

token/services/network/fabric/lookup/deliveryllm.go
```diff
@@ -122,23 +136,32 @@ func (p *deliveryBasedLLMProvider) NewManager(network, channel string) (Listener
...
 
 func (m *deliveryBasedLLM) RemoveLookupListener(key string, listener Listener) error {
-	return m.lm.RemoveFinalityListener(key, &listenerEntry{"", listener})
+	return m.lm.RemoveEventListener(key, &listenerEntry{"", listener})
 }
 
 type endorserTxInfoMapper struct {
 	network string
-	prefix  string
+	prefix1 string
```
alexandrosfilios, line 160: Is there a guarantee that they'll never be more? We could use a slice

alexandrosfilios (approved):
LGTM modulo comments

token/services/network/fabric/lookup/deliveryllm.go
adecaro: moved to slice

token/services/network/fabric/lookup/deliveryllm.go
adecaro: yes

PR #856, opened 2025-02-13
Simplified db driver logic

alexandrosfilios:


token/services/drivers/holder.go
```diff
@@ -7,57 +7,43 @@ SPDX-License-Identifier: Apache-2.0
 package drivers
```
adecaro, line 7: can we move this package under `utils`?

PR #858, opened 2025-02-14
introduce v1 for both fabtoken and dlog

adecaro:


token/core/fabtoken/v1/setup.go
```diff
@@ -16,6 +16,7 @@ import (
 const (
 	// PublicParameters is the key to be used to look up fabtoken parameters
 	PublicParameters = "fabtoken"
+	Version          = "1.0.0"
```
alexandrosfilios, line 19: We could type the version

token/core/zkatdlog/nogh/tokens.go
```diff
@@ -54,7 +54,7 @@ func NewTokensService(publicParametersManager common.PublicParametersManager[*cr
 
 	supportedTokenFormatList := make([]token.Format, 0, 3*len(pp.IdemixIssuerPublicKeys))
 	for _, key := range pp.IdemixIssuerPublicKeys {
-		for _, precision := range crypto.SupportedPrecisions {
+		for _, precision := range v1.SupportedPrecisions {
```
alexandrosfilios, line 57: Is this code related to `v1` only?

alexandrosfilios: Maybe everything that is not under `v1` should not have references to `v1` but to interfaces or common structs?

alexandrosfilios (approved):
LGTM modulo comments

token/core/fabtoken/v1/setup.go
adecaro: what do you mena?

PR #891, opened 2025-03-01
identity/msp/idemix: get rid of the msp structure and functions

adecaro:


token/services/identity/idemix/crypto/deserializer.go
```diff
@@ -4,21 +4,17 @@ Copyright IBM Corp. All Rights Reserved.
...
 import (
 	bccsp "github.com/IBM/idemix/bccsp/types"
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/proto"
-	m "github.com/hyperledger/fabric-protos-go/msp"
 	"github.com/pkg/errors"
 )
 
 type DeserializedIdentity struct {
-	Identity           *Identity
-	NymPublicKey       bccsp.Key
```
AkramBitar, line 18: What happens if we have msp config? how we can support old configurations (backward compatibility?) do we need to test it (loading msp config to the new one)?

token/services/identity/x509/crypto/bccsp.go
```diff
@@ -4,50 +4,47 @@ Copyright IBM Corp. All Rights Reserved.
...
+func GetPKCS11BCCSP(conf *BCCSP, keyStore bccsp.KeyStore) (bccsp.BCCSP, error) {
 	if conf.PKCS11 == nil {
-		return nil, nil, errors.New("invalid BCCSP.PKCS11. missing configuration")
+		return nil, errors.New("invalid BCCSP.PKCS11. missing configuration")
 	}
 
 	p11Opts := conf.PKCS11
-	ks := sw.NewDummyKeyStore()
+	if keyStore == nil {
+		keyStore = sw.NewDummyKeyStore()
```
AkramBitar, line 42: When we need dummy? it test? If so, what happen if we are in production and we have null keyStore?

token/services/identity/x509/crypto/msp.go
```diff
@@ -0,0 +1,226 @@
...
+type verifyingIdentity struct {
+	bccsp               bccsp.BCCSP
+	SignatureHashFamily string
+	cert                *x509.Certificate
+	pk                  bccsp.Key
+}
+
+func (f *verifyingIdentity) Serialize() ([]byte, error) {
+	pb := &pem.Block{Bytes: f.cert.Raw, Type: "CERTIFICATE"}
+	pemBytes := pem.EncodeToMemory(pb)
```
AkramBitar, line 29: Do we need to check that f.cert.Raw is null?

token/core/zkatdlog/nogh/v1/crypto/ecdsa/ecdsa.go
```diff
@@ -17,10 +18,8 @@ import (
 	"fmt"
 	"math/big"
 
-	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/proto"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/identity"
-	msp2 "github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/msp"
-	"github.com/hyperledger/fabric-protos-go/msp"
+	x510 "github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/x509"
```
AkramBitar, line 22: Why this x510 and not x509?

token/core/fabtoken/driver/deserializer.go
```diff
@@ -11,8 +11,7 @@ import (
 	v1 "github.com/hyperledger-labs/fabric-token-sdk/token/core/fabtoken/v1"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/deserializer"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/interop/htlc"
-	"github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/msp"
-	"github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/msp/x509"
+	x510 "github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/x509"
```
AkramBitar, line 14: Why this x510 and not x509?

token/services/identity/idemix/crypto/protos/idemix_config.proto
```diff
@@ -0,0 +1,76 @@
+// Copyright the Hyperledger Fabric contributors. All rights reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+
+syntax = "proto3";
+
+option go_package = "github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/idemix/crypto/protos-go/config";
+
+package identity_idemix;
```
AkramBitar, line 9: Why we need a new proto for idemix config? am wondering if we had one in the past? 
By the way why proto? is that for grpc call to fabric?

token/services/identity/idemix/crypto/protos/generate.go
```diff
@@ -0,0 +1,9 @@
+/*
```
AkramBitar, line 1: Why we need this file? if it is for generating the proto buff of idemix then I noticed that for x509 proto file we do not have this file

AkramBitar (changes requested):

token/services/identity/x509/testdata/msp1/signcerts/auditor.org1.example.com-cert.pem
```diff
@@ -0,0 +1,14 @@
+-----BEGIN CERTIFICATE-----
```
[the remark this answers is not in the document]
adecaro: the certificate is not an issue, it contains only public information. But, if you want, please, open a Github Issue to ask for the removal of all these pieces and the their generation on the fly :)

cmd/tokengen/cobra/pp/common/common.go
```diff
@@ -32,23 +31,10 @@ type PP interface {
...
-		mspID = entries[1]
-	} else {
-		if len(entries) <= 0 || len(entries) > 2 {
-			return nil, errors.Errorf("invalid input [%s], expected <MSPConfigPath>:<MSPID> or <MSPConfigPath>", entry)
-		}
-	}
-
+// GetX509Identity returns the x509 identity from the passed entry.
+func GetX509Identity(entry string) (driver.Identity, error) {
 	// read certificate from entries[0]/signcerts
```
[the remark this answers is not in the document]
adecaro: because we got rid of the MSP_ID. It is necessary in Fabric, but not anymore here.

token/core/zkatdlog/nogh/v1/crypto/ecdsa/ecdsa.go
```diff
@@ -119,16 +118,7 @@ func (v *Verifier) Serialize() ([]byte, error) {
...
-		// Type:    msp.SerializedIdentity_PK,
-		IdBytes: pkRaw,
-	}
-	raw, err := proto.Marshal(mspSI)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed marshalling msp serialized identity")
-	}
-
-	wrap, err := identity.WrapWithType(msp2.X509Identity, raw)
+	wrap, err := identity.WrapWithType(x510.IdentityType, pkRaw)
```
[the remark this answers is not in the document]
adecaro: no, because it comes from `pkRaw, err := PemEncodeKey(v.PK)`. If there is a problem with `v.PK`, `PemEncodeKey` will fail.

token/services/identity/idemix/crypto/id.go
```diff
@@ -0,0 +1,152 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return idemixIDBytes, nil
+}
+
+func (id *Identity) verifyProof() error {
+	// Verify signature
+	var metadata *bccsp.IdemixSignerMetadata
+	if len(id.Idemix.NymEID) != 0 {
```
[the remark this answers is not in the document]
adecaro: yes, then the signature will not be verified against this metadata too. Essentially, one can say, please check the signature and the revealed attributes, when needed.

token/services/identity/idemix/crypto/id.go
```diff
@@ -0,0 +1,152 @@
...
+				{Type: bccsp.IdemixHiddenAttribute},
+			},
+			RhIndex:          RHIndex,
+			EidIndex:         EIDIndex,
+			Epoch:            id.Idemix.Epoch,
+			VerificationType: id.VerificationType,
+			Metadata:         metadata,
+		},
+	)
+	if err == nil && !valid {
```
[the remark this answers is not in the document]
adecaro: the contract of `Verify` is that when valid is true, err is nil.

token/services/identity/idemix/crypto/id.go
```diff
@@ -0,0 +1,152 @@
...
+func (id *SigningIdentity) Sign(msg []byte) ([]byte, error) {
+	sig, err := id.Idemix.Csp.Sign(
+		id.UserKey,
+		msg,
+		&bccsp.IdemixNymSignerOpts{
+			Nym:      id.NymKey,
+			IssuerPK: id.Idemix.IssuerPublicKey,
+		},
+	)
+	if err != nil {
```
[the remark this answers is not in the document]
adecaro: the contract of `Sign` is that if err is nil, then sig is also not nil.

token/services/identity/idemix/crypto/id.go
```diff
@@ -0,0 +1,152 @@
...
+func (v *NymSignatureVerifier) Verify(message, sigma []byte) error {
+	_, err := v.CSP.Verify(
+		v.NymPK,
+		sigma,
+		message,
+		&bccsp.IdemixNymSignerOpts{
+			IssuerPK: v.IPK,
+		},
+	)
+	return err
```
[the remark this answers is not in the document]
adecaro: correct.

token/services/identity/idemix/crypto/deserializer.go
adecaro: At this stage we can break backward compatibility. That's the reason also why we introduce local protobuf message to handle backward and forward compatibility.

token/services/identity/x509/crypto/bccsp.go
adecaro: in the case of PKCS11, this is the HSM. Therefore the HSM is keystore as well. We don't need an external keystore.

token/services/identity/x509/crypto/msp.go
adecaro: if the identity is valid, we should not worry about this.

token/core/zkatdlog/nogh/v1/crypto/ecdsa/ecdsa.go
adecaro: good catch :)

token/core/fabtoken/driver/deserializer.go
adecaro: yes

token/services/identity/idemix/crypto/protos/idemix_config.proto
adecaro: protobuf are very good to handle backward and forward compatibility. We are introducing them here to be independent from external dependencies.

token/services/identity/idemix/crypto/protos/generate.go
adecaro: just generation in case we change something in the protobuf. Go has this tooling integration that is very convinient.

adecaro:
@AkramBitar thanks for your comments. I have answered all of them, I think :)

PR #916, opened 2025-03-05
switch to Fabric v3: protos v2

adecaro:


adecaro:
dropping this because the Issue needs to be addressed differently.

PR #937, opened 2025-03-07
km stores user key into the keystore, not ephemeral anymore

adecaro:
This PR brings in the following: 
- SKI in the configuration (remove reference to SK once loaded) 
- Identity configuration's raw field is filled so that it gets stored in identity db 
- Use of SKI instead of reference to the key itself
- unit-tests


token/services/identity/idemix/deserializer_test.go
```diff
@@ -0,0 +1,118 @@
...
+	_, err = NewDeserializer([]byte{}, math.FP256BN_AMCL)
+	assert.Error(t, err)
+	_, err = NewDeserializer([]byte{0, 1, 2}, math.FP256BN_AMCL)
+	assert.Error(t, err)
+
+	// instantiate a deserializer and validate it
+	d, err := NewDeserializer(config.Ipk, math.FP256BN_AMCL)
+	assert.NoError(t, err)
+	assert.NotNil(t, d)
+	assert.Equal(t, "Idemix with IPK [dJZK5i5D2i5B8S9DsVWDFzdHSJE/jcTLk9VaJzFB4fo=]", d.String())
```
AkramBitar, line 57: I am wondering why d.String should be equal to "Idemix with IPK [dJZK5i5D2i5B8S9DsVWDFzdHSJE/jcTLk9VaJzFB4fo=]"?

token/services/identity/idemix/deserializer_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/IBM/idemix/bccsp/types"
+	math "github.com/IBM/mathlib"
+	"github.com/hyperledger-labs/fabric-token-sdk/token"
+	crypto2 "github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/idemix/crypto"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/sig"
+	kvs2 "github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/storage/kvs"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestNewDeserializer(t *testing.T) {
```
AkramBitar, line 22: If we have a unit test for old Deserializer then may we need to deprecate  (remove) it

token/services/identity/idemix/km.go
```diff
@@ -82,18 +85,33 @@ func NewKeyManager(conf *crypto2.Config, signerService SignerService, sigType bc
...
 	var userKey bccsp.Key
-	if len(conf.Signer.Sk) != 0 && len(conf.Signer.Cred) != 0 {
-		// A credential is present in the config, so we set up a default signer
-		logger.Debugf("the signer contains key material, load it")
+	var userKeySKI SKI
 
-		// Import User secret key
-		userKey, err = cryptoProvider.KeyImport(conf.Signer.Sk, &bccsp.IdemixUserSecretKeyImportOpts{Temporary: true})
+	// use the ski
+	if len(conf.Signer.Ski) != 0 {
```
AkramBitar, line 91: What we will have in conf.Signer.Cred in case of len(conf.Signer.Ski)  != 0? do we need to check also  len(conf.Signer.Cred) != 0  as in the if  len(conf.Signer.Sk) != 0 && len(conf.Signer.Cred)

token/services/identity/idemix/kmp_test.go
```diff
@@ -0,0 +1,105 @@
...
+func TestNewKeyManagerProvider(t *testing.T) {
+	backend, err := kvs.NewInMemory()
+	assert.NoError(t, err)
+	sigService := sig.NewService(sig.NewMultiplexDeserializer(), kvs.NewIdentityDB(backend, token.TMSID{Network: "pineapple"}))
+	config, err := crypto.NewConfig("./testdata/idemix")
+	assert.NoError(t, err)
+	keyStore, err := crypto.NewKeyStore(math.FP256BN_AMCL, backend)
+	assert.NoError(t, err)
+
+	kmp := NewKeyManagerProvider(
```
AkramBitar, line 46: What is the provider? I am wondering what is the differences between kmp and km tests?

token/services/identity/storage/kvs/tracker.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// KeyValuePair stores tracking info
+type KeyValuePair struct {
+	Key   string
+	Value interface{}
+	Error string
+}
+
+// TrackedKVS wraps a Backend and tracks operations
```
AkramBitar, line 22: What is a backend ? is that key value storage?

token/services/identity/storage/kvs/tracker.go
```diff
@@ -0,0 +1,73 @@
...
+type TrackedKVS struct {
+	Backend    Backend
+	PutCounter int
+	GetCounter int
+	PutHistory []KeyValuePair
+	GetHistory []KeyValuePair
+}
+
+func NewTrackedMemory() *TrackedKVS {
```
AkramBitar, line 31: Why we need this tracker? is it for testing purposes only?

AkramBitar (changes requested):

token/services/identity/idemix/deserializer_test.go
adecaro: the name there just signal that we are testing the function `NewDeserializer`. Just that.

token/services/identity/idemix/kmp_test.go
adecaro: Very good. We have the `KayManagerProvider` that returns instances of `KeyManager`, just a factory, if you want.. Then, a `KeyManager` handles a given secret key.

token/services/identity/storage/kvs/tracker.go
adecaro: good point, the `backend` is the underlying KVS on top of view we build a Tracker.

token/services/identity/storage/kvs/tracker.go
adecaro: just for testing purposes to make sure that the code stick to the promises of handling the keys in a certain way.

token/services/identity/idemix/deserializer_test.go
adecaro: This is because `String` is implemented as
```go
func (i *Deserializer) String() string {
	return fmt.Sprintf("Idemix with IPK [%s]", hash.Hashable(i.Ipk).String())
}
```
that value is the hash of the issuer public key, and we load this key from a file so it is always the same.

AkramBitar (approved):

PR #949, opened 2025-03-11
integration tests: views SDKs for the fungible test

adecaro:
introduce SDK for a business party, issuer, auditor, and endorser

### DIFF
--- a/integration/token/fungible/sdk/auditor/sdk.go
+++ b/integration/token/fungible/sdk/auditor/sdk.go
@@ -7,11 +7,8 @@ SPDX-License-Identifier: Apache-2.0
 package auditor
 
 import (
-	"context"
 	errors2 "errors"
-	"reflect"
 
-	"github.com/hyperledger-labs/fabric-smart-client/pkg/node"
 	dig2 "github.com/hyperledger-labs/fabric-smart-client/platform/common/sdk/dig"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/driver"
 	views1 "github.com/hyperledger-labs/fabric-token-sdk/integration/token/common/views"
@@ -22,11 +19,6 @@ import (
 
 type SDK struct {
 	dig2.SDK
-	registry node.Registry
-}
-
-func NewSDK(registry node.Registry) *SDK {
-	return &SDK{registry: registry}
 }
 
 func NewFrom(sdk dig2.SDK) *SDK {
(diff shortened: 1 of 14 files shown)


integration/token/fungible/sdk/auditor/sdk.go
```diff
@@ -0,0 +1,65 @@
...
+	return &SDK{SDK: sdk}
+}
+
+func (p *SDK) Install() error {
+	if err := p.SDK.Install(); err != nil {
+		return err
+	}
+
+	if err := p.Container().Invoke(func(in struct {
+		dig.In
```
alexandrosfilios, line 61: you don't need in. just pass directly driver.Registry

integration/token/fungible/sdk/auditor/sdk.go
```diff
@@ -0,0 +1,65 @@
...
+		Registry driver.Registry // replace this with an external interface
+	}) error {
+		return errors2.Join(
+			in.Registry.RegisterFactory("registerAuditor", &views.RegisterAuditorViewFactory{}),
```
alexandrosfilios, line 65: I think it would make sense to adopt a naming pattern for the strings

integration/token/fungible/sdk/auditor/sdk.go
```diff
@@ -0,0 +1,65 @@
...
+func NewFrom(sdk dig2.SDK) *SDK {
+	return &SDK{SDK: sdk}
+}
+
+func (p *SDK) Install() error {
+	if err := p.SDK.Install(); err != nil {
+		return err
+	}
+
+	if err := p.Container().Invoke(func(in struct {
```
alexandrosfilios: it's better to extract this to another function to keep the Install method cleaner:
```p.Container().Invoke(registerViews)```

File: integration/token/fungible/sdk/endorser/sdk.go [hunk 4]
```diff
@@ -0,0 +1,91 @@
...
+}
+
+func NewFrom(sdk dig2.SDK) *SDK {
+	return &SDK{SDK: sdk}
+}
+
+func (p *SDK) Install() error {
+	// get dig from registry, this was installed by the FTS's sdk
+	var digContainer *dig.Container
+	if p.SDK != nil {
```
alexandrosfilios, line 39: In what case should we accept that SDK is nil?

integration/token/fungible/sdk/endorser/sdk.go
```diff
@@ -0,0 +1,91 @@
...
+		if err := p.SDK.Install(); err != nil {
+			return err
+		}
+		digContainer = p.SDK.Container()
+	} else {
+		// get it from the registry
+		if p.registry == nil {
+			return errors.New("registry not set")
+		}
+		wrapper, err := p.registry.GetService(reflect.TypeOf((*dig.Container)(nil)))
```
alexandrosfilios, line 49: Registering the container as a service doesn't look right. It is available from `p.Container()`. Let's not introduce multiple ways of accessing it, because it becomes too complex to manage.

integration/token/fungible/sdk/endorser/sdk.go
```diff
@@ -0,0 +1,91 @@
...
+			in.Registry.RegisterFactory("GetPublicParams", &views.GetPublicParamsViewFactory{}),
+			in.Registry.RegisterFactory("EndorserFinality", &endorser.FinalityViewFactory{}),
+		)
+	}); err != nil {
+		return errors.WithMessage(err, "failed to install endorser's views")
+	}
+	return nil
+}
+
+func (p *SDK) Start(ctx context.Context) error {
```
alexandrosfilios, line 74: start and post start are not needed if we make sure that SDK exists. So we either create an sdk like all others, or we create another basesdk like the viewsdk, just without installing anything, but I'm not sure what we are trying to achieve with this.

adecaro:
Sorry, @alexandrosfilios , I should have marked the PR as a WIP. Most of your comments are not relevant anymore.

alexandrosfilios (approved):
LVGTM

PR #963, opened 2025-03-12
ttx service: remove the call to ProcessNamespace

adecaro:
We don't need the call to ProcessNamespace because the auditor can subscribe directly to all the transactions it has processed so far.


alexandrosfilios (approved):
LGTM

PR #974, opened 2025-03-14
upgredability: check that a token is spendable

adecaro:
This PR brings the following:
- DLog Token Format is computed excluding the issuer public key
- A new db check verifies the spendability of tokens in the tokens db. If a token is not spendable, a message is reported.

### DIFF
--- a/integration/token/fungible/support.go
+++ b/integration/token/fungible/support.go
@@ -947,24 +947,15 @@ func DoesWalletExist(network *integration.Infrastructure, id *token3.NodeReferen
 	return exists
 }
 
-func CheckOwnerDB(network *integration.Infrastructure, expectedErrors []string, ids ...*token3.NodeReference) {
+func CheckOwnerDB(network *integration.Infrastructure, errorCheck func([]string) error, ids ...*token3.NodeReference) {
 	for _, id := range ids {
 		for _, replicaName := range id.AllNames() {
 			errorMessagesBoxed, err := network.Client(replicaName).CallView("CheckTTXDB", common.JSONMarshall(&views.CheckTTXDB{}))
 			Expect(err).NotTo(HaveOccurred())
-			var errorMessages []string
-			common.JSONUnmarshal(errorMessagesBoxed.([]byte), &errorMessages)
-
-			Expect(len(errorMessages)).To(Equal(len(expectedErrors)), "expected %d error messages from [%s], got [% v]", len(expectedErrors), replicaName, errorMessages)
-			for _, expectedError := range expectedErrors {
-				found := false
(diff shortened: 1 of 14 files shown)


token/services/db/common/checks.go
```diff
@@ -204,3 +209,59 @@ func (a *DefaultCheckers) checkUnspentTokens(context context.Context) ([]string,
 	}
 	return errorMessages, nil
 }
+
+func (a *DefaultCheckers) checkTokenSpendability(context context.Context) ([]string, error) {
+	var errorMessages []string
+
```
alexandrosfilios, line 215: not really important but it would make more sense to have the TMS, network, tokenVault, queryEngine etc as fields of the checker, since we know the tmsID. That spares us the search every time.

alexandrosfilios (approved):
LGTM modulo comments

token/services/db/common/checks.go
adecaro: good point, will do.

alexandrosfilios: makes sense. Can you add a comment in the code please so that it's clearly defined why we don't make this optimization?

adecaro: will do

integration/token/fungible/support.go
```diff
@@ -947,24 +947,15 @@ func DoesWalletExist(network *integration.Infrastructure, id *token3.NodeReferen
 	return exists
 }
 
-func CheckOwnerDB(network *integration.Infrastructure, expectedErrors []string, ids ...*token3.NodeReference) {
+func CheckOwnerDB(network *integration.Infrastructure, errorCheck func([]string) error, ids ...*token3.NodeReference) {
```
alexandrosfilios, line 950: it would be more intuitive to add a matcher instead of the function
* This method would have a signature:
```
func CheckOwnerDB(network *integration.Infrastructure, matchesErrors gomega.Matcher, ids ...*token3.NodeReference)
```

* The actual check
```
if errorCheck != nil {
  var errorMessages []string
  common.JSONUnmarshal(errorMessagesBoxed.([]byte), &errorMessages)
  Expect(errorMessages).To(matchErrors, "failed to check errors")
}
```
* The invocation of the check:
```
CheckOwnerDB(network, ConsistsOf(ContainsSubstring("token format not supported ["), alice)
```
This way we can combine even different types of conditions, e.g. the first error contains this substring, the second error has a specific type etc. Also,  `ConsistsOf` does not enforce the right ordering of the elements. If we know the order (or we just have one element as we do here) we can use `HasExacatElements` instead.

PR #975, opened 2025-03-14
remove the use of the network's TokenVault

adecaro:
This PR does the following: It removes the concept of vault from the network service. The rest of the token-sdk can access directly the tokens and transactions dbs.


token/sdk/vault/pps.go
```diff
@@ -0,0 +1,23 @@
...
+
+import (
+	"github.com/pkg/errors"
+)
+
+type PublicParamsStorage struct {
+	Provider *Provider
+}
+
+func (p *PublicParamsStorage) PublicParams(networkID string, channel string, namespace string) ([]byte, error) {
```
alexandrosfilios, line 17: could we do this with the tokendb?

token/services/tokens/tokens.go
```diff
@@ -225,6 +233,96 @@ func (t *Tokens) UnsupportedTokensIteratorBy(ctx context.Context, walletID strin
...
+	// get unspent tokens
+	it, err := tms.Vault().NewQueryEngine().UnspentTokensIterator()
+	if err != nil {
+		return nil, errors.WithMessage(err, "failed to get an iterator of unspent tokens")
+	}
+	defer it.Close()
+
+	var deleted []*token2.ID
+	var buffer []*token2.UnspentToken
+	bufferSize := 50
```
alexandrosfilios, line 261: This buffer size doesn't look very correct. I guess your limitation is the chaincode? We should pass all tokens and then the chain code should split them into batches and return the response. 3 reasons:
* If we have a different implementation that supports bigger chunks, we cannot leverage it.
* All subsequent calls (e.g. SpentIds, deleteTokens) cannot delete faster.
* It is not clear what the reason, i.e. limitation is. Is it the cc?
The APIs should be agnostic to the limitations of the implementation.

token/services/ttx/multisig/wallet.go
```diff
@@ -115,20 +115,18 @@ func Wallet(sp token.ServiceProvider, wallet *token.OwnerWallet) *OwnerWallet {
...
-		logger.Errorf("failed to get vault for [%s:%s:%s]", tms.Network(), tms.Channel(), tms.Namespace())
 		return nil
 	}
-
 	return &OwnerWallet{
 		wallet:      wallet,
-		vault:       vault,
-		queryEngine: vault.QueryEngine(),
+		vault:       tokens,
+		queryEngine: tms.Vault().NewQueryEngine(),
```
alexandrosfilios, line 129: Do we need to have a unified query engine? Would it be better to split it into smaller engines? i.e. only for tokens.

alexandrosfilios (approved):
LGTM modulo comments

token/sdk/vault/pps.go
adecaro: under the hood, the tokens db is doing that, yes. But the Driver API that needs an implementation of the `PublicParametersStorage` storage interface should not depend on the tokens db.

token/services/tokens/tokens.go
adecaro: will open a Github issue

token/services/ttx/multisig/wallet.go
adecaro: will open a Github issue

token/services/tokens/tokens.go
adecaro: https://github.com/hyperledger-labs/fabric-token-sdk/issues/988

token/services/ttx/multisig/wallet.go
adecaro: https://github.com/hyperledger-labs/fabric-token-sdk/issues/989

PR #977, opened 2025-03-14
dlog v1 finalization: cleanup versioning of public params

adecaro:


alexandrosfilios:
In order to be more future proof, I think we should have a structure like this:
```
nogh/
 api/
  driver.go
  ...
 v1/
  The files that are already here
 v2/
```
Then the whole codebase only interacts with services/structs from the `api` package. `v1` and `v2` also have a dependency on the `api` package.

Only at one point we decide whether we switch to v1 or v2 (I guess the SDK). What do you think?

What we have right now is not versioned, because the code assumes in many places that the version is v1. This means:
* Switching between versions has to be done in many places.
* Switching cannot be configurable and easily testable (run integration tests for both v1 and v2)
* Support different pp versions per TMS or for any other partition (if that is required).

alexandrosfilios (approved):
LGTM

adecaro:
all good points, they will come when we will have `v2`. Thanks, @alexandrosfilios .

PR #978, opened 2025-03-14
upgradability: issuer upgrade service

adecaro:


token/core/zkatdlog/nogh/v1/crypto/upgrade/service.go
```diff
@@ -0,0 +1,249 @@
...
+	if len(witness) != 0 {
+		return nil, errors.Errorf("proof witness not expected")
+	}
+
+	proof := &Proof{
+		Challenge:  ch,
+		Tokens:     ledgerTokens,
+		Signatures: make([]Signature, 0, len(ledgerTokens)),
+	}
+	digest, err := proof.SHA256Digest()
```
alexandrosfilios, line 109: I would make a static method `SHA256Digest(challenge, tokens)` instead of defining it on the proof struct. The reason is because its normal use is based on an incomplete struct. If a struct doesn't have complete data, it should not be used and we should not call methods upon it, because this is like exposing its internal workings.

Instead, for code clarity, I would aim for as much immutability as possible: When we instantiate a struct we don't try to modify its content. Or at least we don't interact with it until it is fully instantiated.

token/core/zkatdlog/nogh/v1/crypto/upgrade/service.go
```diff
@@ -0,0 +1,249 @@
...
+
+func (s *Service) ProcessTokens(ledgerTokens []token.LedgerToken) ([]token.Token, error) {
+	// for each token, extract type and value
+	tokens := make([]token.Token, len(ledgerTokens))
+	for i, tok := range ledgerTokens {
+		precision, ok := token2.Precisions[tok.Format]
+		if !ok {
+			return nil, errors.Errorf("unsupported token format [%s]", tok.Format)
+		}
+		fabToken, _, err := token2.ParseFabtokenToken(tok.Token, precision, s.MaxPrecision)
```
alexandrosfilios, line 178: Do we mean fabtoken?

token/core/zkatdlog/nogh/v1/crypto/upgrade/service.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+//go:generate counterfeiter -o mock/ip.go -fake-name IdentityProvider . IdentityProvider
+
+type IdentityProvider interface {
+	GetSigner(id driver.Identity) (driver.Signer, error)
+}
+
+type Service struct {
+	Logger                          logging.Logger
```
alexandrosfilios, line 43: We can keep these private and when we want to extend them we can make the ones we are interested in public.

token/token/token.go
```diff
@@ -111,6 +114,22 @@ type LedgerToken struct {
 	TokenMetadata []byte
 }
 
+func (t LedgerToken) Equal(right LedgerToken) bool {
+	if !t.ID.Equal(right.ID) {
```
alexandrosfilios, line 118: Maybe a matter of taste but you could also do
```
return t.ID.Equal(right.ID) &&
 bytes.Equal(...) &&
 bytes.Equal(...) ...
```

alexandrosfilios (approved):
LGTM modulo comments

token/core/zkatdlog/nogh/v1/crypto/upgrade/service.go
adecaro: yes, `ParseFabtokenToken`.

token/core/zkatdlog/nogh/v1/crypto/upgrade/service.go
adecaro: okay

PR #983, opened 2025-03-17
hashicorp vault: integration test, run the identity db tests against the vault

adecaro:


token/services/db/dbtest/identity.go
```diff
@@ -0,0 +1,128 @@
...
+
+	it, err := db.IteratorConfigurations("core")
+	assert.NoError(t, err)
+	assert.True(t, it.HasNext())
+	c, err := it.Next()
+	assert.NoError(t, err)
+	assert.True(t, reflect.DeepEqual(expected, c))
+	assert.NoError(t, it.Close())
+
+	exists, err := db.ConfigurationExists("pineapple", "core", "look here")
```
AkramBitar, line 47: I am wondering how the code works if the URL is "look here"?

token/services/db/dbtest/identity.go
```diff
@@ -0,0 +1,128 @@
...
+	tokMeta2, tokMetaAudit2, err := db.GetTokenInfo(id)
+	assert.NoError(t, err, "failed to retrieve token info for [%s]", id)
+	assert.Equal(t, tokMeta, tokMeta2)
+	assert.Equal(t, tokMetaAudit, tokMetaAudit2)
+}
+
+func TSignerInfo(t *testing.T, db driver.IdentityDB) {
+	alice := []byte("alice")
+	bob := []byte("bob")
+	assert.NoError(t, db.StoreSignerInfo(alice, nil))
```
AkramBitar, line 89: I noticed that in the test there is no remove and check after that if exists. Do we need to test remove for each add?

token/services/db/dbtest/identity.go
```diff
@@ -0,0 +1,128 @@
...
+
+func TSignerInfoConcurrent(t *testing.T, db driver.IdentityDB) {
+	wg := sync.WaitGroup{}
+	n := 100
+	wg.Add(n)
+
+	for i := 0; i < n; i++ {
+		go func(i int) {
+			alice := []byte(fmt.Sprintf("alice_%d", i))
+			bob := []byte(fmt.Sprintf("bob_%d", i))
```
AkramBitar, line 107: This code is same as in TSignerInfo. Can we add a method that receives parameter for   "alice_%d" and call this method from the concurrency test and from TSignerInfo?

AkramBitar (changes requested):

token/services/db/dbtest/identity.go
adecaro: indeed, there was a failure, I fixed it.

token/services/db/dbtest/identity.go
```diff
@@ -0,0 +1,128 @@
...
+	assert.False(t, exists)
+
+	expected = driver.IdentityConfiguration{
+		ID:     "pineapple",
+		Type:   "no core",
+		URL:    "look here",
+		Config: []byte("config"),
+		Raw:    []byte("raw"),
+	}
+	assert.NoError(t, db.AddConfiguration(expected))
```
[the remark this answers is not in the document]
adecaro: Good point, the IdentityDB interface doesn't have a `RemoveConfiguration`, we will test when it is time to add this functionality.

token/services/db/dbtest/identity.go
adecaro: we don't have API to remove stuff. I guess this will be a new task for us.

token/services/db/dbtest/identity.go
adecaro: good point, will create a shared function.

token/services/identity/storage/kvs/identitydb.go
```diff
@@ -0,0 +1,219 @@
...
+
+func (s *IdentityDB) GetAuditInfo(identity []byte) ([]byte, error) {
+	k := kvs.CreateCompositeKeyOrPanic(
+		IdentityDBPrefix,
+		[]string{
+			IdentityDBData,
+			driver2.Identity(identity).String(),
+		},
+	)
+	if !s.kvs.Exists(k) {
```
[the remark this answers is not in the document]
adecaro: good point, it is more about the expected behaviour. Now, if the key is not there, the caller expects no error.

token/services/identity/storage/kvs/identitydb.go
```diff
@@ -0,0 +1,219 @@
...
+				IdentityDBSigner,
+				id.UniqueID(),
+			},
+		)
+		if err != nil {
+			return nil, err
+		}
+		keys[i] = k
+	}
+	return s.kvs.GetExisting(keys...), nil
```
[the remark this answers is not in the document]
adecaro: this is very good point, we need to address this in a better way. What about another PR?

token/services/identity/storage/kvs/walletdb.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func (s *WalletDB) LoadMeta(identity driver2.Identity, wID driver.WalletID, roleID int) ([]byte, error) {
+	idHash := identity.UniqueID()
+	k, err := kvs.CreateCompositeKey("walletDB", []string{s.tmsID.String(), strconv.Itoa(roleID), idHash, wID, "meta"})
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create key")
+	}
+	var meta []byte
+	if err := s.kvs.Get(k, &meta); err != nil {
```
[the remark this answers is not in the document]
adecaro: we really need to clean this, totally agree.

AkramBitar:
I noticed that the test Tests / itest (interop-dlog-t4) fails with the following postgres error. I do not see how this is related to the changes that you made in this PR. Do you know why we have these failure?
I2025-03-17 16:50:01.693 UTC [104] FATAL:  role "postgres" does not exist
I2025-03-17 16:50:02.180 UTC [150] FATAL:  role "postgres" does not exist

token/services/identity/storage/kvs/walletdb.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+	var walletIDs []string
+	for it.HasNext() {
+		var wID string
+		if _, err := it.Next(&wID); err != nil {
+			return nil, errors.Wrapf(err, "failed to get next wallets from iterator")
+		}
+		found := false
+		for _, walletID := range walletIDs {
+			if walletID == wID {
```
AkramBitar, line 92: If found == true, then why we need to continue the loop. Can we break the loop in this case and stop it since we found the walletID?

AkramBitar (changes requested):

token/services/identity/storage/kvs/hashicorp/vault_kvs_test_utils.go
```diff
@@ -9,61 +9,92 @@ package hashicorp
...
-	defer out.Close()
-	_, _ = io.Copy(os.Stdout, out)
 
+	cli, err := client.NewClientWithOpts(client.FromEnv, client.WithAPIVersionNegotiation())
+	if err != nil {
+		t.Fatal(err)
+	}
 	// Define the container configuration
+	portStr := fmt.Sprintf("%d", port)
 	token := "00000000-0000-0000-0000-000000000000"
```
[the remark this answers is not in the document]
adecaro: we need to be able to change port because the use the vault from different tests.

token/services/identity/storage/kvs/hashicorp/vault_kvs_test_utils.go
```diff
@@ -9,61 +9,92 @@ package hashicorp
...
-					HostPort: "8200",
+					HostPort: portStr,
 				},
 			},
 		},
 	}
 	// Create the container
-	resp, err := cli.ContainerCreate(ctx, containerConfig, hostConfig, nil, nil, "")
+	ctx := context.Background()
+	containerName := "dev-hashicorp-vault-container-" + portStr
```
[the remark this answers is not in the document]
adecaro: will do

token/services/identity/storage/kvs/walletdb.go
adecaro: yes, will fix using `collections.Set`

PR #987, opened 2025-03-18
hashicorp vault: documentation

adecaro:


docs/services/identity.md
```diff
@@ -135,4 +135,37 @@ type WalletService interface {
...
+An implementation for this interface can be found under [`token/sdk/identity`](./../../token/sdk/identity).
+It uses the `identitydb` service for the `IdentityDB` and the `WalletDB`, and the Fabric-Smart-Client's KVS for the `Keystore`.
+
+### HashiCorp Vault Secrets Engine Support
+
+The HashiCorp Vault Secrets Engine is a modular component of Vault designed to securely manage, store, or generate sensitive data such as API keys, passwords, certificates, and encryption keys.
+The identity service provides an implementation for both the `IdentityDB` and `Keystore` based on the `HashiCorp Vault Secrets Engine`.
+This implementation can be found under [`hashicorp`](./../../token/services/identity/storage/kvs/hashicorp).
+
+In order to use this integration, the developer must do the following:
```
AkramBitar, line 165: I am wondering if to add the following section here, since the impelemnation works with Vault Secrets Engine to run in non-versioned mode (so called KV1), and not the versioned one (KV2).

"The Fabric Token SDK requieres to configure the HashiCorp 
 Vault Secrets Engine to run in non-versioned mode (i.e., stores the most recently written value for a key). For more information about non-versioned secrets engine mode, refer to the (https://developer.hashicorp.com/vault/docs/secrets/kv/kv-v1). "

AkramBitar (changes requested):

docs/services/identity.md
adecaro: done

AkramBitar (approved):

Issue #890, opened 2025-02-28
Enhance Redeem with the signature of the issuer


KElkhiyaoui:
Redeem is a transfer action that contains an output with a `nil` `Owner`. Currently, any `Owner` can redeem tokens without the consent of the issuer. We would like to restrict this capability by requiring the issuer's signature on the  token requests that contain redeem. 
* A redeem operation can be initiated by the owner of the token. The corresponding `Redeem` view will reaches out to the relevant issuer to obtain redeem confirmation in the form of a signature from the issuer on the relevant token request. In this manner, the issuer will be aware of all redeem operations, and can withhold the confirmation when it is warranted. 
* The validation semantics of a redeem action will stay the same with the additional check that verifies whether the issuer of the tokens being redeemed signed the token request.

AkramBitar:
For the team education session, it will be great if we can address the following:
1) What is a Redeem transfer, and why we need it
2) What we have today
3) Which enhancement is needed
4) Flows of the Redeem transfer and involved components.

arner:
Nice! Does this epic make the redeem transaction show up in the Issuer transaction db? And does it make it possible for the Issuer to query a "total in circulation" per token type based on the issued tokens minus the redeemed tokens?

adecaro:
@arner , yes, the transaction will be marked as a redeem in the transactions db. We might added additional query to infer the total in circulation, yes.

aaadir:
PR #1027 soon to be submitted implements the above sub-issues, except for #905 that still needs to be completed for a correct handling of the issuer parameter of Redeem. See there.

adecaro:
@aaadir , thanks for the great work so far.
Additional points to take in consideration:
- We must ensure that the auditor can check the identity of the issuer for a redeem operation
- We must ensure that the transactions db tracks correctly the redeem operation. We might have to include the identity of the issuer in the storage as well
- We need to update the `Validate` function of the actions to ensure the Issuer identity is there 
- We need to extend the `Match` function of the `token.TransferMetadata` to make sure metadata and action match.

If you agree, let's have Github Issues for each item so that we can address them. Thanks much :)

adecaro:
done

PR #990, opened 2025-03-18
Enhanced Redeem

barvhaim:
First iteration completed using Extra Signers and Binding to pass the Issuer identity


token/core/fabtoken/v1/transfer.go
```diff
@@ -146,13 +151,26 @@ func (s *TransferService) Transfer(ctx context.Context, _ string, _ driver.Owner
...
 		ExtraSigners: nil,
 	}
 
+	// add redeem signer
+	if isRedeem {
+		issuers := s.PublicParametersManager.PublicParameters().Issuers()
+		if len(issuers) < 1 {
+			return nil, nil, errors.New("no issuer found")
+		}
+		issuer := issuers[0]
```
arner, line 168: Can it be possible that there is more than one issuer? If so, how do we know which one we have to pick?

barvhaim: We assumed that it is enough for a single issuer to sign and if there are multiple we take the first. It's about the definition of the flow https://github.com/hyperledger-labs/fabric-token-sdk/issues/890

arner: Ok I don't know about that, maybe @KElkhiyaoui or @adecaro can confirm :)

adecaro: good catch, Arne, we will need to fix this in coming PRs to let the user decide which issuer to choose.

PR #995, opened 2025-03-19
chore: make function comment match function name

murongshaozong:
make function comment match function name


adecaro:
@murongshaozong , could you write a script to check that the comments follow the go guidelines to write them? I think it would make this PR stronger. Thanks :)

adecaro:
closing this for now

PR #1010, opened 2025-03-23
dlog v1 finalization: enforce at least one issuer

adecaro:
We start by adding the check that at least one issuer identity is specified in the `Validate` function of the public parameters. From there, we make sure the rest still works.
Many of the integration tests were assuming that anyone could issue, therefore we had to change the test to update the public parameters with the identities of who needed to issue.

### DIFF
--- a/integration/token/fungible/support.go
+++ b/integration/token/fungible/support.go
@@ -942,10 +942,11 @@ func GetPublicParams(network *integration.Infrastructure, id *token3.NodeReferen
 	return pp.([]byte)
 }
 
-func FetchAndUpdatePublicParams(network *integration.Infrastructure, id *token3.NodeReference) []byte {
-	pp, err := network.Client(id.ReplicaName()).CallView("FetchAndUpdatePublicParams", common.JSONMarshall(&views.UpdatePublicParams{}))
-	Expect(err).NotTo(HaveOccurred())
-	return pp.([]byte)
+func FetchAndUpdatePublicParams(network *integration.Infrastructure, id *token3.NodeReference) {
+	for _, name := range id.AllNames() {
+		_, err := network.Client(name).CallView("FetchAndUpdatePublicParams", common.JSONMarshall(&views.UpdatePublicParams{}))
+		Expect(err).NotTo(HaveOccurred())
+	}
 }
 
 func DoesWalletExist(network *integration.Infrastructure, id *token3.NodeReference, wallet string, walletType int) bool {
--- a/integration/token/fungible/support.go
+++ b/integration/token/fungible/support.go
@@ -10,6 +10,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"os"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -27,6 +28,8 @@ import (
 	common2 "github.com/hyperledger-labs/fabric-token-sdk/integration/token/common"
 	"github.com/hyperledger-labs/fabric-token-sdk/integration/token/fungible/views"
 	token2 "github.com/hyperledger-labs/fabric-token-sdk/token"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/nogh/v1/crypto"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/driver"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/identity"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/storage/kvs"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/x509"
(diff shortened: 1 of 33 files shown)


integration/token/fungible/support.go
```diff
@@ -1346,3 +1368,94 @@ func MultiSigSpendCashForTMSID(network *integration.Infrastructure, sender *toke
 	return txID
 
 }
+
+func PrepareUpdatedPublicParams(network *integration.Infrastructure, auditor string, networkName string) []byte {
+	tms := GetTMSByNetworkName(network, networkName)
+	auditorId := GetAuditorIdentity(tms, auditor)
+	issuerId := GetIssuerIdentity(tms, "newIssuer")
```
alexandrosfilios, line 1375: Gives a bit more clarity if we pass the issuer id as a parameter too.

token/core/fabtoken/v1/driver/driver.go
```diff
@@ -100,6 +100,15 @@ func (d *Driver) NewTokenService(tmsID driver.TMSID, publicParams []byte) (drive
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to initiliaze public params manager")
 	}
+	// log information about the public params
+	pp := publicParamsManager.PublicParams()
+	logger.Infof("new token driver for tms id [%s] with label and version [%s:%s]", tmsID, pp.Identifier(), pp.Version())
+	for _, issuer := range pp.Issuers() {
```
alexandrosfilios, line 106: You can do
```
logger.Debug("issuers are [%s]", pp.Issuers())
```

adecaro: I wanted to have them on separated lines, for clarify.

integration/token/fungible/tests.go
```diff
@@ -274,21 +274,25 @@ var BobAcceptedTransactions = []TransactionRecord{
 
 type OnRestartFunc = func(*integration.Infrastructure, string)
 
-func TestAll(network *integration.Infrastructure, auditorId string, onRestart OnRestartFunc, aries bool, sel *token3.ReplicaSelector) {
+func TestAll(network *integration.Infrastructure, auditorId string, onRestart OnRestartFunc, aries bool, orion bool, sel *token3.ReplicaSelector) {
```
aaadir, line 277: consider, instead of "aries", "orion", and more possible future system-specific parameter names, maybe use a single generic parameter like network_name.

integration/token/fungible/tests.go
```diff
@@ -274,21 +274,25 @@ var BobAcceptedTransactions = []TransactionRecord{
...
 	endorsers := GetEndorsers(network, sel)
+	custodian := sel.Get("custodian")
 	RegisterAuditor(network, auditor)
 
 	// give some time to the nodes to get the public parameters
 	time.Sleep(10 * time.Second)
 
 	SetKVSEntry(network, issuer, "auditor", auditor.Id())
 	CheckPublicParams(network, issuer, auditor, alice, bob, charlie, manager)
+	if orion {
```
aaadir, line 293: consider, instead of mentioning a specific network_name in the code, maybe refer to its relevant trait. Then, either pass the trait as a parameter or maintain a structure with generic traits for every network type.

token/core/fabtoken/v1/driver/driver.go
aaadir, line 106: I see that this kind of debug-printing occurs often in the code. 
Consider having a general debug-printing method for printing slices of strings - e.g. something like logger.logSliceOf("issuer", pp.Issuers())

token/services/identity/membership/lm.go
```diff
@@ -400,9 +400,11 @@ func (l *LocalMembership) addLocalIdentity(config *driver.IdentityConfiguration,
 	if keyManager.Anonymous() || len(l.targetIdentities) == 0 {
 		l.logger.Debugf("no target identity check needed, skip it")
 	} else if found := slices.ContainsFunc(l.targetIdentities, identity.Equal); !found {
-		l.logger.Debugf("identity [%s:%s] not in target identities, ignore it", name, config.URL)
-		return nil
+		// the identity is not in the target identities, we should give it a lower priority
+		l.logger.Debugf("identity [%s:%s] not in target identities", name, config.URL)
 	} else {
+		// give it high priority
+		priority = -1
```
alexandrosfilios, line 407: We could have it as a constant or even better a type to explain whether a low or high number has higher priority

alexandrosfilios (approved):
LGTM modulo comments

token/core/fabtoken/v1/driver/driver.go
adecaro: cleaned

token/services/identity/membership/lm.go
adecaro: done

integration/token/fungible/tests.go
adecaro: good point, I think we need to rework these tests. I'll open a github issue for it.

token/core/fabtoken/v1/driver/driver.go
adecaro: removed, I'll just print the public params.

aaadir (approved):

PR #1014, opened 2025-03-25
upgradability: token drivers, introduce protobufs

adecaro:
This PR does the following:
- It introduces protobuf serialization for the fabtoken driver's actions.
- It introduces a version field to the action protobuf message with version check enforcement.


token/core/fabtoken/v1/actions/transfer.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+// IsGraphHiding returns false, indicating that fabtoken does not hide the transaction graph
+func (t *TransferAction) IsGraphHiding() bool {
+	return false
+}
+
+// SerializeOutputAt marshals the output at the specified index in TransferAction
+func (t *TransferAction) SerializeOutputAt(index int) ([]byte, error) {
+	if index >= len(t.Outputs) {
```
alexandrosfilios, line 124: is there a reason we don't want to expose only one method `GetOutputAt(index int)` and then call the methods `IsRedeem()` or `Serialize()`?

adecaro: no, good point, I can open a github issue for this because it requires a few changes, but we can do it.

adecaro: ah, yes, now I remember, it is the translator. The translator defines local interfaces for the action it can translate, `token/services/network/common/rws/translator/action.go`. If we return a complex struct, then we must have an import for that struct and I didn't want that link.

PR #1022, opened 2025-03-26
Pagination support for Transactions Query

AkramBitar:


token/services/auditdb/db.go
```diff
@@ -125,6 +126,12 @@ type QueryTransactionsParams = driver.QueryTransactionsParams
 // QueryTokenRequestsParams defines the parameters for querying token requests
 type QueryTokenRequestsParams = driver.QueryTokenRequestsParams
 
+// Pagination defines the pagination for querying movements
+type Pagination = driver2.Pagination
+
+// Pagination iterator defines the pagination iterator for movements query results
+type PageIterator = driver2.PageIterator[*TransactionRecord]
```
alexandrosfilios, line 133: This name is a bit too generic to concern only transaction records. I don't think we need it.

token/services/ttxdb/db.go
```diff
@@ -179,9 +180,15 @@ type QueryTokenRequestsParams = driver.QueryTokenRequestsParams
 // QueryValidationRecordsParams defines the parameters for querying movements
 type QueryValidationRecordsParams = driver.QueryValidationRecordsParams
 
+// Pagination defines the pagination for querying movements
+type Pagination = driver3.Pagination
+
+// Pagination iterator defines the pagination iterator for movements query results
+type PageIterator = driver3.PageIterator[*TransactionRecord]
```
alexandrosfilios, line 187: same

alexandrosfilios (commented):
Some minor comments to address

token/services/auditdb/db.go
AkramBitar: Changed to
type PageTransactionsIterator = driver3.PageIterator[*TransactionRecord]
Is that ok?
or do you want me to remove it at all?

token/services/ttxdb/db.go
AkramBitar: Changed to
type PageTransactionsIterator = driver3.PageIterator[*TransactionRecord]
Is that ok?
or do you want me to remove it at all?

AkramBitar:
A new PR opened instead of this PR: Pagination support for Transactions Query #1054

PR #1024, opened 2025-03-27
flaky tests: fix interop-t4

adecaro:
In the fast exchange test, the auditor is contacted multiple time in the same context. To do that, the CollectEndorsmentView close the connection to the auditor after completion. Sometimes this does not happen timely and the connection is closed when is still used. We avoid this issue by putting the last transaction in its own fresh context.


alexandrosfilios (approved):
LGTM

PR #1025, opened 2025-03-31
identity service: use always aries

adecaro:


adecaro:
closing this because we need to consider the fact that fabric-ca does not support aries.

PR #1026, opened 2025-03-31
upgradability: key rotation of the auditor public key

adecaro:


token/core/common/validator_auditing.go
```diff
@@ -0,0 +1,48 @@
...
+		}
+		if !found {
+			return errors.Errorf("auditor [%s] is not in auditors", auditor)
+		}
+
+		verifier, err := ctx.Deserializer.GetAuditorVerifier(auditor)
+		if err != nil {
+			return errors.Wrapf(err, "failed to deserialize auditor's public key")
+		}
+		_, err = ctx.SignatureProvider.HasBeenSignedBy(auditor, verifier)
```
alexandrosfilios, line 42: If I understand correctly, HasBeenSignedBy has to be called in the right order. Would it make sense to check in any order and keep a counter of how many have been checked already?

token/core/common/validator.go
```diff
@@ -248,13 +265,36 @@ func (v *Validator[P, T, TA, IA, DS]) verifyTransfer(tr TA, ledger driver.Ledger
 		}
```
alexandrosfilios, line 265: do you mean `if c >= 1`?

token/core/common/validator.go
```diff
@@ -205,30 +209,43 @@ func (v *Validator[P, T, TA, IA, DS]) verifyIssue(tr IA, ledger driver.Ledger, s
...
 	context := &Context[P, T, TA, IA, DS]{
 		Logger:            v.Logger,
 		PP:                v.PublicParams,
+		TokenRequest:      tokenRequest,
 		Deserializer:      v.Deserializer,
-		TransferAction:    tr,
+		TransferAction:    action,
 		Ledger:            ledger,
 		SignatureProvider: signatureProvider,
 		MetadataCounter:   map[MetadataCounterID]int{},
```
alexandrosfilios, line 251: What I don't find very neat here is that sometimes we pass a transfer action in the context, sometimes an issue action and sometimes none. I would pass none of them and then have different validator types:
```
type TransferValidator interface {
 Validate(Context, TransferAction)
}
```
Similarly for `IssueValidator` and `AuditValidator`

token/driver/request.go
```diff
@@ -22,6 +22,32 @@ const (
...
+			Raw: r.Identity,
+		},
+		Signature: &request.Signature{
+			Raw: r.Signature,
+		},
+	}, nil
+}
+
+func (r *AuditorSignature) FromProtos(tr *request.AuditorSignature) error {
+	if tr.Identity != nil {
```
alexandrosfilios, line 42: maybe also `if tr == nil` if it can happen

token/services/ttx/auditor.go
```diff
@@ -121,12 +121,13 @@ func (r *RegisterAuditorView) Call(context view.Context) (interface{}, error) {
...
-	local bool
+	tx                               *Transaction
+	local                            bool
+	skipAuditorSignatureVerification bool
 }
 
-func newAuditingViewInitiator(tx *Transaction, local bool) *AuditingViewInitiator {
-	return &AuditingViewInitiator{tx: tx, local: local}
+func newAuditingViewInitiator(tx *Transaction, local, skipAuditorSignatureVerification bool) *AuditingViewInitiator {
+	return &AuditingViewInitiator{tx: tx, local: local, skipAuditorSignatureVerification: skipAuditorSignatureVerification}
```
alexandrosfilios, line 130: Should we include a testing feature in this view?

alexandrosfilios (approved):
LGTM modulo comments

token/core/common/validator_auditing.go
adecaro: checking in any order is expensive because we would need to verify the signatures against multiple public keys.
Anyhow, a driver can use a different implementation of `SignatureProvider` as you suggest, sure.

token/services/ttx/auditor.go
adecaro: what do you mean?

token/core/common/validator.go
adecaro: no, one time is okay.

PR #1028, opened 2025-04-01
ttx service: timeout via context - part 1

sreenidhidrIBM:


token/services/ttx/auditor.go
```diff
@@ -146,11 +147,15 @@ func (a *AuditingViewInitiator) Call(context view.Context) (interface{}, error)
...
 	logger.Debugf("Receiving signature for [%s]", a.tx.ID())
 	span.AddEvent("start_receiving")
-	signature, err := ReadMessage(session, time.Minute)
+	jsonSession := session2.NewFromSession(context, session)
+	signature, err := jsonSession.ReceiveRawWithTimeout(time.Minute)
 	if err != nil {
 		span.RecordError(err)
 		return nil, errors.WithMessage(err, "failed to read audit event")
 	}
+	if len(signature) == 0 {
```
alexandrosfilios, line 156: is this check needed? it might be misleading as to what is expected from the code if this can't happen. normally I expect that if err == nil, there data is valid.

token/services/ttx/endorse.go
```diff
@@ -736,8 +738,8 @@ func (f *ReceiveTransactionView) Call(context view.Context) (interface{}, error)
 	span := trace.SpanFromContext(context.Context())
 	span.AddEvent("start_receive_transaction_view")
 	defer span.AddEvent("end_receive_transaction_view")
-
-	msg, err := ReadMessage(context.Session(), time.Minute*4)
+	jsonSession := session2.JSON(context)
+	msg, err := jsonSession.ReceiveRawWithTimeout(time.Minute * 4)
```
alexandrosfilios, line 742: we could also extract these durations to constants, depending on what we are waiting for, e.g. `signatureRcvTimeout`, `txRcvTimeout`, to make sure we are aligned and to have a better overview. ideally I would even add them as fields in the structure objects and set the default values in the constructors.

alexandrosfilios (commented):
Some minor comments to address

adecaro (approved):
LGTM

PR #1032, opened 2025-04-09
Update FSC version

alexandrosfilios:


adecaro:
does this belong to an EPIC?

adecaro (approved):
LGTM. Please, attach it to an EPIC.

adecaro:
maybe, you want to rename the PR also to better represent what you are doing here. The association with an EPIC might not be necessary then

PR #1034, opened 2025-04-12
make the transfer and issuance metadata available in the db

arner:
This PR makes the public metadata of Issuance and Transfer actions, which are stored on the ledger, available to the user via the database (on the sender, recipient, and auditor side). At the moment this includes the hash and preimage of HTLC lock and claim, which can be used to display to the end user or for reporting.

The user can already add this kind of metadata to a Transfer with the opt:`tkn.WithTransferMetadata("user.x."+tx.ID(), []byte("this is metadata"))`. At the moment this causes the validation to fail, but that can be fixed if we add a new `ValidateTransferFunc`, somewhat like the following:

```go
func TransferExtraDataValidate(ctx *validator.Context) error {
	for key, val := range ctx.TransferAction.Metadata {
		if strings.HasPrefix(key, "user.") {
                        // TBD: validation?
			ctx.CountMetadataKey(key)
		}
	}
	return nil
}
```

It would add some useful tools to the toolbox. For instance you could share metadata privately with the counterparty and optionally an auditor in the ApplicationMetadata, and a hash of that data on the public ledger, all tied to the same transaction. Or link it to external data, or create another custom validator for additional proofs. It would be the responsibility of the user to make sure that privacy is protected.

> [!IMPORTANT]
> This change is not backwards compatible due to changing the database schema. It would require a manual `ALTER TABLE ... ADD COLUMN public_metadata JSONB NOT NULL;` before upgrading. Until we have a common method for database migrations, I couldn't find a good way to handle it.

Looking forward to feedback on this PR and the idea for the followup.

### DIFF
--- a/token/request.go
+++ b/token/request.go
@@ -136,9 +136,13 @@ func WithRestRecipientIdentity(recipientData *RecipientData) TransferOption {
 // AuditRecord models the audit record returned by the audit command
 // It contains the token request's anchor, inputs (with Type and Quantity), and outputs
 type AuditRecord struct {
-	Anchor     string
-	Inputs     *InputStream
-	Outputs    *OutputStream
+	// Anchor is used to bind the Actions to a given Transaction
+	Anchor string
+	// Inputs represent the input tokens of the transaction
+	Inputs *InputStream
+	// Outputs represent the output tokens of the transaction
+	Outputs *OutputStream
+	// Attributes are metadata which are stored on the public ledger as part of the transaction Actions.
 	Attributes map[string][]byte
 }
 
--- a/token/request.go
+++ b/token/request.go
@@ -136,9 +136,10 @@ func WithRestRecipientIdentity(recipientData *RecipientData) TransferOption {
 // AuditRecord models the audit record returned by the audit command
 // It contains the token request's anchor, inputs (with Type and Quantity), and outputs
 type AuditRecord struct {
-	Anchor  string
-	Inputs  *InputStream
-	Outputs *OutputStream
+	Anchor     string
+	Inputs     *InputStream
+	Outputs    *OutputStream
+	Attributes map[string][]byte
 }
 
 // Issue contains information about an issue operation.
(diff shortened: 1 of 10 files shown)


adecaro:
Hi @arner , thanks for this PR. I will look at it ASAP.
I have a question: Do you have any suggestion on how to handle these schema changes? Should we have a schema versioning? Should the code be able to handle multiple versions? What's your take.
@alexandrosfilios , please, jump in as well :)

alexandrosfilios:
> Hi @arner , thanks for this PR. I will look at it ASAP. I have a question: Do you have any suggestion on how to handle these schema changes? Should we have a schema versioning? Should the code be able to handle multiple versions? What's your take. @alexandrosfilios , please, jump in as well :)

I have only used liquibase before, but I can compare some alternatives.

adecaro:
I can see an epic to handle all this, honestly. There are many things to streamline in the lines of what you say @arner .
@alexandrosfilios , shall we craft with Arne, an EPIC on the changes to do?

token/request.go
```diff
@@ -136,9 +136,10 @@ func WithRestRecipientIdentity(recipientData *RecipientData) TransferOption {
 // AuditRecord models the audit record returned by the audit command
 // It contains the token request's anchor, inputs (with Type and Quantity), and outputs
 type AuditRecord struct {
-	Anchor  string
-	Inputs  *InputStream
-	Outputs *OutputStream
+	Anchor     string
+	Inputs     *InputStream
+	Outputs    *OutputStream
+	Attributes map[string][]byte
```
adecaro, line 142: Please, update the comment of the struct and maybe some information about what `Attributes` is supposed to contain.

token/services/db/driver/common.go
```diff
@@ -120,6 +120,9 @@ type TransactionRecord struct {
 	// ApplicationMetadata is the metadata sent by the application in the
 	// transient field. It is not validated or recorded on the ledger.
 	ApplicationMetadata map[string][]byte
+	// PublicMetadata is the metadata that is stored on the ledger as part
+	// of an Issuance or Transfer Action (for instance the HTLC hash).
+	PublicMetadata map[string][]byte
```
adecaro, line 125: do we test that we get back what we store?

adecaro:
@arner , could you check my comments, please? I'll then merge. Thanks :)

token/services/db/driver/common.go
arner: Looking into the tests, a good place to validate that would be in the `CheckAcceptedTransactions` function in `integration/token/fungible/support.go`. At the moment that function is only used for standard transactions, and not for HTLC transactions. HTLC lock and claim are currently the only place where we use transfer attributes on the ledger. It seems like not the best place to test the data that comes back in the PublicMetadata field, because what the HTLC package stores exactly is its own implementation and not something we would have to test for.

Of course we _should_ have a test for the PublicMetadata though. I added a simple validation in the last commit, but it's not used yet. Since we decided to add a new PR with the feature to add user-supplied data that is stored on the ledger, it seems like a more pure place to add the integration test for the full feature. What do you think, shall we split it up that way?

adecaro: @arner , I just ment to have a unit-test that set `PublicMetadata` when storing a record and check if they are retrieved by a query 😅

PR #1035, opened 2025-04-14
dlog v1 finalization: unit-tests coverage above 80% - token

adecaro:


token/core/zkatdlog/nogh/v1/token/service_test.go
```diff
@@ -0,0 +1,376 @@
...
+		name          string
+		init          func() (logging.Logger, *setup.PublicParams, token2.IdentityDeserializer, error)
+		check         func(pp *setup.PublicParams, ts *token2.TokensService) error
+		wantErr       bool
+		expectedError string
+	}{
+		{
+			name: "publicParams cannot be nil",
+			init: func() (logging.Logger, *setup.PublicParams, token2.IdentityDeserializer, error) {
+				return nil, nil, nil, nil
```
alexandrosfilios: do we need to return the logger? we could use some default or nil value always, or are we testing something?

token/core/zkatdlog/nogh/v1/token/service_test.go
```diff
@@ -0,0 +1,376 @@
...
+					return nil, nil, err
+				}
+				return ts, raw, nil
+			},
+			wantErr:            false,
+			expectedIdentities: []driver.Identity{driver.Identity("alice")},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
```
alexandrosfilios, line 231: I think that using matrices here makes the tests more complex. The amount of re-usable code is negligible, so I would inline the tests, e.g.

```
func Test_Success(t *testing.T) {
				RegisterTestingT(t)
				id := &mock.IdentityDeserializer{}
				id.RecipientsReturns([]driver.Identity{driver.Identity("alice")}, nil)
				ts, err := token2.NewTokensService(nil, pp, id)
				Expect(err).ToNot(HaveOccurred())

				tok := &token2.Token{}
				raw, err := tok.Serialize()
				Expect(err).ToNot(HaveOccurred())

				identities, err := ts.Recipients(raw)
				Expect(err).ToNot(HaveOccurred())
				Expect(identities).To(ConsistOf(driver.Identity("Alice"))
}
```

alexandrosfilios: This way also your errors can become more descriptive, e.g.
```
Expect(err).To(ContainSubstring(...))
```

alexandrosfilios (approved):
LGTM modulo comments

PR #1036, opened 2025-04-14
allow custom transfer options for claim and reclaim

arner:
and merge them with the htlc opts.

Same as we do in the Lock function. Can be helpful in the case of custom validators and/or storing extra attributes on the ledger.


adecaro (approved):
LGTM

PR #1039, opened 2025-04-16
check owner to identify redeemed tokens

adecaro:
This PR fixes the parsing the token request's inputs and outputs. Namely: We clearly identify when an output is redeemed and we don't include it in the list of tokens to be created in the db. We return an error in case of a failure of any sort.

### DIFF
--- a/integration/ports.go
+++ b/integration/ports.go
@@ -60,8 +60,8 @@ var (
 
 	AllTestTypes = []*InfrastructureType{
 		WebSocketNoReplication,
-		// LibP2PNoReplication,
-		// WebSocketWithReplication,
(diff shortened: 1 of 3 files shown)


integration/ports.go
```diff
@@ -60,8 +60,8 @@ var (
 
 	AllTestTypes = []*InfrastructureType{
 		WebSocketNoReplication,
-		LibP2PNoReplication,
-		WebSocketWithReplication,
+		// LibP2PNoReplication,
```
alexandrosfilios, line 63: to reactivate

alexandrosfilios (approved):
LGTM modulo comment

PR #1040, opened 2025-04-16
ttx service: timeout via context - part 2

adecaro:


token/services/ttx/endorse.go
```diff
@@ -656,8 +656,11 @@ func NewReceiveTransactionView() *ReceiveTransactionView {
 }
 
 func (f *ReceiveTransactionView) Call(context view.Context) (interface{}, error) {
-	jsonSession := session2.JSON(context)
-	msg, err := jsonSession.ReceiveRawWithTimeout(time.Minute * 4)
+	span := trace.SpanFromContext(context.Context())
+	span.AddEvent("start_receive_transaction_view")
```
alexandrosfilios, line 660: It's better to use the method `logger.DebugfContext`

token/services/utils/json/jsession/session.go
```diff
@@ -0,0 +1,143 @@
...
+
+type JSession struct {
+	s       Session
+	context context.Context
+}
+
+// New returns a JSON-based session wrapping the passed session.
+func New(context context.Context, s Session) *JSession {
+	span := trace.SpanFromContext(context)
+	span.AddEvent(fmt.Sprintf("Open json session to [%s:%s]", string(s.Info().Caller), s.Info().CallerViewID))
```
alexandrosfilios, line 31: logger

alexandrosfilios, line 31: `s.Info()` is an expensive call (it contains a lock). I would suggest you use it only once or even better you use the `Eval` util:
```
logger.DebugfContext(context, "Open json session: %s", logging.Eval(s.Info))
```

alexandrosfilios (approved):
LGTM modulo comments

PR #1041, opened 2025-04-16
go1.23-24

adecaro:
This PR does the following:
- moves the minimum go version to 1.23, and
- uses go1.24 as toolchain


Vibhuair20:
i saw your updates in the above sdk @adecaro
using the toolchain is a great option
But you still have issues with the golanggci-lint
Can i help you fix those issues and push a pr

adecaro:
@Vibhuair20 that would be great :)

Vibhuair20:
Can you give me a basic rundown 
To work properly

adecaro:
actually, given that it might be more complicated on the token-sdk, could you apply the same process to the smart client? Updating also the linter as I did here. What do you think?

Vibhuair20:
Okay i think i can do the same for the smart client.

adecaro:
@Vibhuair20 , as soon as you have the PR on the smart client, I'll update the `go.mod` here.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/hyperledger-labs/fabric-token-sdk
 
-go 1.22.6
+go 1.23.0
```
arner, line 3: It might be best to depend on `1.23.8` (or whichever is the latest now) to have the security fixes included.

alexandrosfilios (approved):
LGTM

PR #1045, opened 2025-04-17
F db config

alexandrosfilios:
This PR aligns the db drivers and the configuration with FSC so that:
* we re-use the configuration objects and services instead of defining new logic
* we use one consistent way of configuring (that will be changed in the next step to the simplified structure, but only in one place now)
* we set a standard way for developers to develop new DBs: the method signatures, drivers, and configurations are always the same


token/services/db/sql/common/init.go
```diff
@@ -30,26 +30,23 @@ type tableNames struct {
...
-		Tokens:                 nc.MustGetTableName("tokens"),
-		Ownership:              nc.MustGetTableName("token_ownership"),
-		Certifications:         nc.MustGetTableName("token_certifications"),
-		TokenLocks:             nc.MustGetTableName("token_locks"),
-		PublicParams:           nc.MustGetTableName("public_params"),
-		Wallets:                nc.MustGetTableName("wallets"),
-		IdentityConfigurations: nc.MustGetTableName("identity_configurations"),
-		IdentityInfo:           nc.MustGetTableName("identity_information"),
-		Signers:                nc.MustGetTableName("identity_signers"),
+		Movements:              nc.MustGetTableName(prefix, "movements", params...),
```
alexandrosfilios, line 37: postgres has a limit in the name length of the tables and it's easy to get to the limit considering the tms prefix. we can find a more sustainable solution like hashing

token/services/db/sql/common/tokenlock.go
```diff
@@ -40,26 +40,18 @@ func newTokenLockDB(readDB, writeDB *sql.DB, tables tokenLockTables) *TokenLockD
...
-	if opts.CreateSchema {
-		if err = common.InitSchema(writeDB, []string{tokenLockDB.GetSchema()}...); err != nil {
-			return nil, err
-		}
-	}
-	return tokenLockDB, nil
+		}), nil
+}
+
+func (db *TokenLockDB) CreateSchema() error {
```
alexandrosfilios, line 53: extracted CreateSchema in order to be aligned with the FSC db's and re-use the configs.

token/services/db/sql/postgres/driver.go
```diff
@@ -0,0 +1,128 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package postgres
+
+import (
```
alexandrosfilios, line 9: I split the postgres driver from the sqlite driver:
* because they have different configs (although they overlap significantly). We shouldn't create configs that contain properties of both solutions and use them partially, because it is not clear what is needed/used and what not. Plus, this is more extensible if we add more db drivers.
* to flatten the config in the next steps

token/services/db/sql/postgres/test_utils.go
```diff
@@ -4,7 +4,7 @@ Copyright IBM Corp. All Rights Reserved.
 SPDX-License-Identifier: Apache-2.0
 */
 
-package common
```
alexandrosfilios, line 7: this test functionality is only relevant to postgres, hence I moved it under the postgres package and removed the postgres prefix.

token/services/db/dbtest/notifier.go
```diff
@@ -0,0 +1,112 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
```
alexandrosfilios, line 4: Moved all test cases under the same detest package. I don't expose them directly, but through the test method that is called by each driver: memory, sqlite, postgres

token/sdk/dig/sdk.go
```diff
@@ -128,28 +127,16 @@ func (p *SDK) Install() error {
...
-			return tokendb.NewManager(dh, "tokendb.persistence", "db.persistence")
-		}),
-		p.Container().Provide(func(dh *db2.DriverHolder) *tokendb.NotifierManager {
-			return tokendb.NewNotifierManager(dh, "tokendb.persistence", "db.persistence")
-		}),
+		p.Container().Provide(tokendb.NewManager),
+		p.Container().Provide(tokendb.NewNotifierManager),
 		p.Container().Provide(digutils.Identity[*tokendb.Manager](), dig.As(new(tokens.DBProvider))),
 		p.Container().Provide(NewDriverHolder),
-		p.Container().Provide(func(dh *db2.DriverHolder) *auditdb.Manager {
```
alexandrosfilios, line 143: The config prefix in now moved in the manager instead of the SDK to be aligned with the SDK

token/services/db/dbtest/identity.go
```diff
@@ -16,7 +16,20 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-var IdentityCases = []struct {
+func IdentityTest(t *testing.T, cfgProvider cfgProvider) {
```
alexandrosfilios, line 19: The test cases (IdentityCases here) are not exporter anymore. Only IdentityTest is exported and used by the 3 drivers (memory, postgres, sqlite). The db cases are agnostic to the underlying driver.

adecaro (approved):
LGTM. Please, update any relevant documentation here or in the next PR.

Issue #1064, opened 2025-04-27
Redeem: extend the Match function of the token.TransferMetadata to make sure metadata and action match


aaadir:
We need to extend the Match function of the token.TransferMetadata to make sure metadata and action match.

PR #1067, opened 2025-04-28
Redeem: match TransferMetadata issuer

aaadir:
This PR implements issue #1064:
Extend the Match function of the token.TransferMetadata to make sure metadata and action match.


adecaro:
@aaadir , thanks for the PR. Please, introduce a unit-test to verify that the check is performed. Thanks.

adecaro (changes requested):
Please, look at the comments above.

aaadir:
token/metadata_test.go:TestMetadata_TestMatchTransferAction now includes a unit test for the match function.
done.

Issue #1073, opened 2025-05-08
Redeem integration test where the issuer is also the party that asks to redeem a token


aaadir:
Add an integration test with the following scenario:
Alice transfers to the Issuer a token with 10 units of some type.
The Issuer then redeems a subset of this amount (say 5 units).

The test can be added to TestAll or put in a new separate integration test for the various Redeem tests that we have.

PR #1076, opened 2025-05-11
Redeem integration test where the issuer is also the party that asks to redeem a token

aaadir:
Implements issue #1073


adecaro (approved):
LGTM

PR #1078, opened 2025-05-13
Create a logger that automatically extracts the package and uses it a…

SaidAltury-ibm:
…s a prefix #815


alexandrosfilios (approved):
LGTM

PR #1079, opened 2025-05-13
Remove redundant default persistence from core.yaml

alexandrosfilios:


adecaro (approved):
LGTM, please make sure that the core.yaml generated doesn't contain any entry for the storage

Issue #1081, opened 2025-05-18
SonarQube maintainability issues: avoid duplicated literals


aaadir:
Define a constant instead of duplicating a literal multiple times.
- cmd/tokengen/cobra/artifactgen/gen/gen.go: **gen** function

PR #1082, opened 2025-05-18
SonarQube maintainability issues: avoid duplicated literals

aaadir:
This PR implements issue #1081 
Define a constant instead of duplicating a literal multiple times.


alexandrosfilios:
I think it makes more sense to configure sonar than applying these changes.

Issue #955, opened 2025-03-12
Add contexts to db operations


alexandrosfilios:
We need to pass the `context.Context` to as many functions as possible, so that we can enable tracing with little effort. And example of such a change would be from

```func (db *DB) GetTransactions(walletID string) ([]Transaction, error)```

to

```func (db *DB) GetTransactions(ctx context.Context, walletID string) ([]Transaction, error)```

PR #1089, opened 2025-05-21
Add context to DB operations

drucker-nir:
Required by Issue #955


token/services/db/dbtest/identity.go
```diff
@@ -16,30 +17,30 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func IdentityTest(t *testing.T, cfgProvider cfgProvider) {
+func IdentityTest(ctx context.Context, t *testing.T, cfgProvider cfgProvider) {
```
alexandrosfilios, line 20: you don't need to pass it here. use `context.Background()`

alexandrosfilios: same for all test methods

token/core/zkatdlog/nogh/v1/driver/driver.go
```diff
@@ -125,7 +127,7 @@ func (d *Driver) NewTokenService(tmsID driver.TMSID, publicParams []byte) (drive
 	ip := ws.IdentityProvider
 
 	authorization := common.NewAuthorizationMultiplexer(
-		common.NewTMSAuthorization(logger, ppm.PublicParams(), ws),
+		common.NewTMSAuthorization(ctx, logger, ppm.PublicParams(), ws),
```
alexandrosfilios, line 130: We shouldnt have a context here. Conceptually the context is bound to a request. Here we are instantiating a service, which doesn't happen with every request.

token/services/db/dbtest/notifier.go
```diff
@@ -45,7 +45,7 @@ type dbEvent struct {
 	vals map[driver2.ColumnKey]string
 }
 
-func collectDBEvents(db driver.TokenNotifier) (*[]dbEvent, error) {
+func collectDBEvents(ctx context.Context, db driver.TokenNotifier) (*[]dbEvent, error) {
```
alexandrosfilios, line 48: this method doesnt use its context. we don't need a context here.

token/services/db/dbtest/tokens.go
```diff
@@ -51,7 +51,7 @@ func TokensTest(t *testing.T, cfgProvider cfgProvider) {
 
 var tokensCases = []struct {
 	Name string
-	Fn   func(*testing.T, TestTokenDB)
+	Fn   func(context.Context, *testing.T, TestTokenDB)
```
alexandrosfilios, line 54: all these methods in the test don't need a context. as we said, when they call a method that requires a context (a method that is in an outside package), then we pass the background context.

token/services/db/sql/common/wallet.go
```diff
@@ -44,7 +45,7 @@ func (db *WalletStore) CreateSchema() error {
 	return common.InitSchema(db.writeDB, []string{db.GetSchema()}...)
 }
 
-func (db *WalletStore) GetWalletID(identity token.Identity, roleID int) (driver.WalletID, error) {
+func (db *WalletStore) GetWalletID(ctx context.Context, identity token.Identity, roleID int) (driver.WalletID, error) {
 	idHash := identity.UniqueID()
 	result, err := common.QueryUnique[driver.WalletID](db.readDB,
```
alexandrosfilios, line 50: Since you are adding the context in the stores, you can also use the context and change these methods
```
QueryUnique -> QueryUniqueContext
Query -> QueryContext
Exec -> ExecContext
```

Issue #1091, opened 2025-05-23
upgradability: add multiplexing driver


adecaro:
Any public parameters must carry two field related to versioning: Identifier and Version.
The `Identifier` is used to point to a class of drivers with certain characteristics. The `Version` is used to pick a specific instance from the class.

We currently don't have a mechanism to select the version given the class.

This Issue is about the introduction of a driver multiplexer that given the `Identifier` chooses the driver version to use looking at the `Version` field.

PR #1100, opened 2025-05-30
upgradability: add multiplexing driver #1091

adecaro:
This PR does the following: It introduces a new naming convention for the token drivers.
Before, a driver was identifier just by a label. Now, a driver is identified as <id>.v<version>. Both information can be retrieved from the interface `driver.PublicParameters`. This way, we can re-use the actual code that selects a driver to implicit select by version as well.


integration/token/fungible/fabtoken/fabtoken_test.go
```diff
@@ -45,6 +45,7 @@ func UpdatePublicParams(network *integration.Infrastructure, selector *token2.Re
 	issuerId := fungible.GetIssuerIdentity(tms, "newIssuer")
 	publicParam := fabtokenv1.PublicParams{
 		Label:             "fabtoken",
+		Ver:               1,
```
alexandrosfilios, line 48: We could have a constant under `fabtokenv1` called `Version`

alexandrosfilios: In case this `1` is used also elsewhere in the production code.

token/core/fabtoken/v1/setup/setup.go
```diff
@@ -167,8 +169,13 @@ func (p *PublicParams) Deserialize(raw []byte) error {
 	if err != nil {
 		return errors.Wrapf(err, "failed to deserialize public parameters")
 	}
-	if container.Identifier != p.Label {
-		return errors.Errorf("invalid identifier, expecting 'fabtoken', got [%s]", container.Identifier)
+	expectedID := string(core.TokenDriverName(FabtokenIdentifier, ProtocolV1))
```
alexandrosfilios, line 172: for disambiguation I would call them `TokenDriverIdentifier` (instead of name), `TokenDriveryType` (instead of identifier), and `TokenDriverVersion`

alexandrosfilios (approved):
LGTM

PR #1106, opened 2025-06-09
Test TokenLockStore queries

aaadir:


token/services/db/sql/postgres/tokenlock.go
```diff
@@ -56,16 +57,16 @@ func (db *TokenLockStore) Cleanup(ctx context.Context, leaseExpiry time.Duration
...
-	_, err := db.WriteDB.ExecContext(ctx, query, args...)
+	_, err = db.WriteDB.ExecContext(ctx, query, args...)
 	if err != nil {
 		db.Logger.Errorf("query failed: %s", query)
 	}
 	return err
 }
 
-func (db *TokenLockStore) logStaleLocks(ctx context.Context, leaseExpiry time.Duration) error {
+func (db *TokenLockStore) LogStaleLocks(ctx context.Context, leaseExpiry time.Duration) ([]LockEntry, error) {
```
alexandrosfilios: you can keep this private as it was. test the cleanup method instead and make sure it runs two queries.

alexandrosfilios (changes requested):

token/services/db/sql/postgres/tokenlock.go
aaadir: done.

token/services/db/sql/postgres/tokenlock.go
```diff
@@ -40,7 +40,8 @@ func NewTokenLockStore(dbs *common2.RWDB, tableNames common.TableNames) (*TokenL
 }
 
 func (db *TokenLockStore) Cleanup(ctx context.Context, leaseExpiry time.Duration) error {
-	if err := db.logStaleLocks(ctx, leaseExpiry); err != nil {
+	_, err := db.logStaleLocks(ctx, leaseExpiry)
```
alexandrosfilios: I think we don't need to modify this method at all

token/services/db/sql/postgres/tokenlock_test.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	store, _ := NewTokenLockStore(&dbs, common.TableNames{
+		TokenLocks: "TOKEN_LOCKS",
+		Requests:   "REQUESTS",
+	})
+	return store.TokenLockStore
+}
+
+func TestLogStaleLocks(t *testing.T) {
```
alexandrosfilios: TestCleanup

token/services/db/sql/postgres/tokenlock.go
```diff
@@ -80,23 +81,23 @@ func (db *TokenLockStore) logStaleLocks(ctx context.Context, leaseExpiry time.Du
...
+		return nil, err
 	}
 
 	db.Logger.Infof("Found following entries ready for deletion: [%v]", lockEntries)
-	return nil
+	return lockEntries, nil
 }
 
-type lockEntry struct {
+type LockEntry struct {
```
alexandrosfilios: this can be private again

token/services/db/sql/postgres/tokenlock_test.go
```diff
@@ -0,0 +1,90 @@
...
+	gomega.RegisterTestingT(t)
+	db, mockDB, err := sqlmock.New()
+	gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+	input := driver.Deleted
+
+	record := LockEntry{
```
alexandrosfilios: you don't need to pack it in a `LockEntry` struct. And then you can convert lock entry again into a private struct.

alexandrosfilios (changes requested):

token/services/db/sql/postgres/tokenlock.go
aaadir: I now changed it to 
`	if _, err := db.logStaleLocks(ctx, leaseExpiry); err != nil {`
I had to make this change because now logStaleLocks now returns two outputs (see below)
done

token/services/db/sql/postgres/tokenlock_test.go
aaadir: done

token/services/db/sql/postgres/tokenlock_test.go
aaadir: done

token/services/db/sql/postgres/tokenlock.go
aaadir: done

token/services/db/sql/postgres/tokenlock.go
```diff
@@ -40,7 +40,7 @@ func NewTokenLockStore(dbs *common2.RWDB, tableNames common.TableNames) (*TokenL
 }
 
 func (db *TokenLockStore) Cleanup(ctx context.Context, leaseExpiry time.Duration) error {
-	if err := db.logStaleLocks(ctx, leaseExpiry); err != nil {
+	if _, err := db.logStaleLocks(ctx, leaseExpiry); err != nil {
```
alexandrosfilios: I don't think this file should be modified at all. Why do we need to return the lock entries?

token/services/db/sql/common/tokenlock_test_utils.go
```diff
@@ -0,0 +1,57 @@
...
+	"context"
+	"database/sql"
+	"testing"
+
+	"github.com/DATA-DOG/go-sqlmock"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/token"
+	"github.com/onsi/gomega"
+)
+
+type storeConstructor func(*sql.DB) *TokenLockStore
```
alexandrosfilios: rename

token/services/db/sql/sqlite/tokenlock_test.go
```diff
@@ -7,28 +7,51 @@ SPDX-License-Identifier: Apache-2.0
...
+	store, _ := NewTokenLockStore(&dbs, common.TableNames{
+		TokenLocks: "TOKEN_LOCKS",
+		Requests:   "REQUESTS",
+	})
+	return store.TokenLockStore
+}
+
 func TestIsStale(t *testing.T) {
-	RegisterTestingT(t)
+	gomega.RegisterTestingT(t)
```
alexandrosfilios: import gomega with `.` wherever possible (e.g. here)

token/services/db/sql/postgres/tokenlock.go
aaadir: done

token/services/db/sql/common/tokenlock_test_utils.go
aaadir: done

token/services/db/sql/sqlite/tokenlock_test.go
aaadir: done

token/services/db/sql/postgres/tokenlock.go
aaadir: yes - you're right. I did that when I directly tested the records returned by select inside logStaleLocks, but now logStaleLocks is tested as part of the Cleanup, and the returned values are not really tested.
done.

Issue #1112, opened 2025-06-11
use AllTestTypes only for DLog


adecaro:
Currently we run all integration tests against all test types. This makes the CI running for too long and doesn't bring any additional value.

This Github Issues asks to keep AllTestTypes only for fungible/dlog.

adecaro:
might be reopened in the future

PR #1113, opened 2025-06-11
use AllTestTypes only for DLog #1112 Open Feature

adecaro:
Implement #1112


adecaro:
dropping this to favor a different strategy

PR #1125, opened 2025-06-30
remove key that was renamed to skipCreateTable, as it already default…

arner:
…s to false


adecaro:
@arner , thanks for submitting this. 
Please, next time always open first a Github Issue.
Thanks.

PR #1126, opened 2025-06-30
add support for issue metadata

adecaro:
This PR makes it possible for the developer to specify `issue action metadata`. These are keys that will be stored on the ledger with uniqueness guarantees. By using `WithIssueMetadata`, the developer can specify a metadata to be stored on the ledger. This is similar to the already existing use of `WithTransferMetadata`.

### DIFF
--- a/token/request.go
+++ b/token/request.go
@@ -22,6 +22,7 @@ import (
 
 const (
 	TransferMetadataPrefix = meta.TransferMetadataPrefix
+	IssueMetadataPrefix    = meta.IssueMetadataPrefix
 )
 
 type Binder interface {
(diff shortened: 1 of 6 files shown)


KElkhiyaoui (approved):

token/request.go
```diff
@@ -108,6 +109,11 @@ func WithTransferMetadata(key string, value []byte) TransferOption {
 	return WithTransferAttribute(TransferMetadataPrefix+key, value)
 }
 
+// WithIssueMetadata sets issue action metadata
+func WithIssueMetadata(key string, value []byte) TransferOption {
```
KElkhiyaoui: It should be an `IssueOption`

KElkhiyaoui (approved):

Issue #1135, opened 2025-07-11
expand the reach of context.Context


adecaro:
There are still many parts of the code in the critical path that does not receive a context. This means that code cannot leave events in the traces. For example, the package `token/services/identity/membership` contains many debug logs without context. Therefore, a way to identify this code is to search for logging without context (Debugf, Infof, Warnf, and so on).

In addition, we need to double check the use of `context.Background()` and `context.TODO()` to see if that code can be reached by an existing context or if we need to create a trace to track those events.

Testing files can be excluded.

adecaro:
Here is the list of places where we still use `context.Background()`:

```text
Targets
    Occurrences of 'context.Background()' in Directory /Users/adc/go/src/github.com/hyperledger-labs/fabric-token-sdk
Found occurrences in Directory /Users/adc/go/src/github.com/hyperledger-labs/fabric-token-sdk  (236 usages found)
    Production  (163 usages found)
        Unclassified  (163 usages found)
            fabric-token-sdk  (163 usages found)
                docs/core/extension/zkatdlog/nogh/v2/driver  (5 usages found)
                    driver.go  (5 usages found)
                        110 pp := ppm.PublicParams(context.Background())
                        123 ppm.PublicParams(context.Background()),
                        134 common.NewTMSAuthorization(logger, ppm.PublicParams(context.Background()), ws),
                        144 tokensUpgradeService, err := upgrade.NewService(logger, ppm.PublicParams(context.Background()).QuantityPrecision, deserializer, ip)
                        150 ppm.PublicParams(context.Background()),
                integration/nwo/runner  (1 usage found)
                    view_user.go  (1 usage found)
                        160 ctx, span := u.tracer.Start(context.Background(), operationType)
                integration/nwo/token/common  (4 usages found)
                    ppmgen.go  (4 usages found)
                        66 id, _, err := km.Identity(context.Background(), nil)
                        91 id, _, err := km.Identity(context.Background(), nil)
                        170 id, _, err := km.Identity(context.Background(), nil)
                        195 id, _, err := km.Identity(context.Background(), nil)
                integration/token/fungible  (7 usages found)
                    support.go  (3 usages found)
                        1144 newIdentity, _, err := kmp.Identity(context.Background(), nil)
                        1363 val, warnings, err := cli.Query(context.Background(), metric.String(), time.Now())
                        1520 newIdentity, _, err := kmp.Identity(context.Background(), nil)
[...]

PR #1137, opened 2025-07-14
expand the reach of context.Context #1135

adecaro:


integration/token/fungible/dlogstress/dlog_test.go
```diff
@@ -44,9 +44,9 @@ func newTestSuite(backend string) (*token.TestSuite, *token.ReplicaSelector) {
 			DefaultTMSOpts:  common.TMSOpts{TokenSDKDriver: "dlog", Aries: true},
 			ReplicationOpts: opts,
 			CommType:        fsc.LibP2P,
-			// FSCLogSpec:      "token-sdk=debug:info",
-			SDKs:       []nodepkg.SDK{sdks[backend]},
-			Monitoring: true,
+			FSCLogSpec:      "token-sdk=debug:debug",
```
arner, line 47: is this an intentional change? Might lead to a lot of logs for a stress test.

token/core/zkatdlog/nogh/v1/crypto/upgrade/service.go
```diff
@@ -113,7 +114,7 @@ func (s *Service) GenUpgradeProof(ch driver.TokensUpgradeChallenge, ledgerTokens
 	signatures := make([]Signature, len(tokens))
 	for i, token := range tokens {
 		// get a signer for each token
-		signer, err := s.IdentityProvider.GetSigner(token.Owner)
+		signer, err := s.IdentityProvider.GetSigner(context.Background(), token.Owner)
```
arner, line 117: If we can't pass a context into the GenUpgradeProof function, we could maybe create a context object outside of the loop and use it for all of the tokens?

token/services/ttx/recipients.go
```diff
@@ -159,7 +159,7 @@ func (f *RequestRecipientIdentityView) Call(context view.Context) (interface{},
 			continue
 		}
 
-		results[i], err = w.GetRecipientIdentity()
+		results[i], err = w.GetRecipientIdentity(context.Context())
```
arner, line 162: It's a bit confusing that we use "context" as the name of the variable... In other places I've started calling it vctx for view context... Not for this PR but maybe we can rename them at some point.

token/services/ttx/withdrawal.go
```diff
@@ -65,10 +65,10 @@ func RequestWithdrawalForRecipient(context view.Context, issuer view.Identity, w
 	return ir.RecipientData.Identity, result[1].(view.Session), nil
 }
 
-func (r *RequestWithdrawalView) Call(context view.Context) (interface{}, error) {
-	logger.Debugf("Respond request recipient identity using wallet [%s]", r.Wallet)
+func (r *RequestWithdrawalView) Call(ctx view.Context) (interface{}, error) {
```
arner, line 68: Ah you've changed it here, it's probably better than how it was, but conventionally ctx is often already used for a context.Context, what do you think of going to vctx?

adecaro:
close this because will restore the other branch

PR #1142, opened 2025-07-17
don't panic if the wallet is not found

arner:
we try to invoke a function on `w` which is nil if the owner wallet is not found on the node


adecaro:
@arner , why switching to fmt?

adecaro:
@arner , what's the status of this? Please, can you update the metadata on the left hand side?

PR #1149, opened 2025-07-24
nwo: do not create folders that are not used

adecaro:
This PR cleans up some of the folders created by NWO and not used anymore.

### DIFF
--- a/integration/nwo/token/common/ne.go
+++ b/integration/nwo/token/common/ne.go
@@ -8,7 +8,6 @@ package common
 
 import (
 	"fmt"
-	"os"
 	"path/filepath"
 	"time"
 
@@ -17,7 +16,6 @@ import (
 	sfcnode "github.com/hyperledger-labs/fabric-smart-client/integration/nwo/fsc/node"
 	"github.com/hyperledger-labs/fabric-token-sdk/integration/nwo/token/generators"
 	topology2 "github.com/hyperledger-labs/fabric-token-sdk/integration/nwo/token/topology"
-	"github.com/onsi/gomega"
 	"github.com/onsi/gomega/gexec"
 )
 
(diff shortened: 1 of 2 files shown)


integration/nwo/token/common/ne.go
```diff
@@ -79,25 +57,9 @@ func (p *NetworkHandler) FSCNodeKVSDir(uniqueName string) string {
 }
 
 func (p *NetworkHandler) DeleteDBs(node *sfcnode.Node) {
-	for _, uniqueName := range node.ReplicaUniqueNames() {
-		for _, path := range []string{p.TokensDBSQLDataSourceDir(uniqueName)} {
-			logger.Infof("remove all [%s]", path)
-			gomega.Expect(os.RemoveAll(path)).ToNot(gomega.HaveOccurred())
```
AkramBitar, line 85: Can we add ""failed to delete [%s]", path)" if we have error?

integration/nwo/token/common/ne.go
```diff
@@ -79,25 +57,9 @@ func (p *NetworkHandler) FSCNodeKVSDir(uniqueName string) string {
 }
 
 func (p *NetworkHandler) DeleteDBs(node *sfcnode.Node) {
```
AkramBitar, line 59: When this method called? when we run make clean? or before we create the dbs in init of regression?

AkramBitar (changes requested):

alexandrosfilios (approved):

integration/nwo/token/common/ne.go
adecaro: this code has been removed.

integration/nwo/token/common/ne.go
adecaro: it is called during the integration tests, if needed.

PR #1155, opened 2025-07-24
upgradability: fixes to allow more easily to extend an existing driver #1154

adecaro:
This PR does the following:
- It extends the use of driver identifier to tokengen and the integration tests.
- It adds the possibility to customize the validator

### DIFF
--- a/docs/core/extension/zkatdlog/nogh/v2/driver/base.go
+++ b/docs/core/extension/zkatdlog/nogh/v2/driver/base.go
@@ -30,7 +30,7 @@ import (
 type base struct{}
 
 func (d *base) PublicParametersFromBytes(params []byte) (driver.PublicParameters, error) {
-	pp, err := v2.NewPublicParamsFromBytes(params, v2.DLogIdentifier, v2.ProtocolV2)
+	pp, err := v2.NewPublicParamsFromBytes(params, v2.DLogNoGHDriverName, v2.ProtocolV2)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to unmarshal public parameters")
 	}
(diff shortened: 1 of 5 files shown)


integration/nwo/token/generators/pp/fabtokenv1/gen.go
```diff
@@ -0,0 +1,107 @@
...
+				return nil, errors.WithMessage(err, "failed to create x509 km")
+			}
+			id, _, err := km.Identity(context.Background(), nil)
+			if err != nil {
+				return nil, errors.WithMessage(err, "failed to get identity")
+			}
+			if issuersSet.Contains(issuer.ID) {
+				wrap, err := identity.WrapWithType(x509.IdentityType, id)
+				if err != nil {
+					return nil, errors.WithMessagef(err, "failed to create x509 identity for auditor [%v]", issuer)
```
arner, line 92: for issuer*

token/core/zkatdlog/nogh/v1/setup/setup.go
```diff
@@ -472,9 +472,9 @@ func (p *PublicParams) String() string {
 }
 
 func (p *PublicParams) Validate() error {
-	if p.DriverVersion != ProtocolV1 {
-		return errors.Errorf("invalid version [%d], expected [%d]", p.DriverVersion, ProtocolV1)
-	}
+	// if p.DriverVersion != ProtocolV1 {
```
arner, line 475: Maybe remove this instead of commenting out? Or if the version is now embedded in the protocol name, do we want to verify the full protocol name?

arner (approved):
Small typo, approved otherwise. Do we need a label or a naming convention for PRs to signify breaking changes? That would make it easier when we create a new tag to compile a change log and instructions for developers on how to upgrade.

In this case, it looks like you could update the params on a running system with the new driver name and continue using it (?) -- but the change in name of the driver will break scripts at least so we have to document it. 

Another option might be to retain backwards compatibility by having `dlog` as an alias for `zkatdlognogh.v1`?

Issue #1169, opened 2025-07-30
introduce stricter linter configuration


adecaro:
Currently, golangci-lint is executed with the default configuration. This enables only few linters.
This Issue asks to introduce stricter linting rules.

PR #1170, opened 2025-07-30
introduce stricter linter configuration #1169

adecaro:
This PR introduces an explicit configuration for golangci-lint enabling the first set of linters.
We applied also all the autofix linters.
More linters will be enabled in upcoming PRs.


adecaro:
The following linters will be enabled in other PRs.
```
* containedctx: 3
* contextcheck: 11
* copyloopvar: 1
* depguard: 3
* dupl: 8
* err113: 19
* errcheck: 3
* errname: 5
* fatcontext: 1
* forcetypeassert: 1
* gocognit: 50
* goconst: 5
* gocritic: 38
* gosec: 22
* intrange: 3
* maintidx: 1
* prealloc: 31
* promlinter: 7
* rowserrcheck: 1
* sqlclosecheck: 3
* staticcheck: 3
* testifylint: 4
* unparam: 23
* ```

File: token/core/zkatdlog/nogh/v1/issue/prover.go [hunk 1]
```diff
@@ -59,7 +59,7 @@ func NewProver(tw []*token.Metadata, tokens []*math.G1, pp *v1.PublicParams) (*P
 
 	values := make([]uint64, len(tw))
 	blindingFactors := make([]*math.Zr, len(tw))
-	for i := 0; i < len(tw); i++ {
+	for i := range tw {
```
aaadir, line 62: I think in this case you can also write the following to have less referencing (though maybe the compiler avoid them anyway):
```
for _, t:= range tw {
   if t == nil || t.BlindingFactor == nil {
			return nil, errors.New("invalid token witness")
```
  }

token/core/zkatdlog/nogh/v1/setup/setup.go
```diff
@@ -417,7 +417,7 @@ func (p *PublicParams) GenerateRangeProofParameters(bitLength uint64) error {
 	p.RangeProofParams.LeftGenerators = make([]*mathlib.G1, bitLength)
 	p.RangeProofParams.RightGenerators = make([]*mathlib.G1, bitLength)
 
-	for i := uint64(0); i < bitLength; i++ {
+	for i := range bitLength {
```
aaadir, line 420: i is of type int, not uint64. 
the old code seemed to make an effort of ensuring that i is uint64. 
is the new version equivalent to the old here?

token/services/selector/testutils/test_cases.go
```diff
@@ -83,7 +83,7 @@ func TestInsufficientTokensOneReplica(t *testing.T, replica EnhancedManager) {
 	err := storeTokens(replica, unspentTokens)
 	assert.NoError(t, err)
 
-	// The replica asks for CHF1, CHF1, CHF1 (total CHF3)
+	// The replica asks for CHF1, CHF1 (total CHF3)
```
aaadir, line 86: only with the deleted CHF we'll have CHF 3 times - no?

adecaro:
@aaadir thanks for your comments. The changes made were only to satisfy the linters and produce equivalent code.

PR #1174, opened 2025-07-30
introduce nocache and ristretto-based cache #1173

adecaro:
This PR does the following:
- introduce a NoCache instance
- Remove the extra synchronization in `token/services/db/sql/common/identity.go`
- Replace `isMeCache` with a cache interface.
- When caching TokenRequest, in the cache it is added also the message to sign that is later compared during the commit phase.

### DIFF
--- a/token/services/identity/provider.go
+++ b/token/services/identity/provider.go
@@ -100,7 +100,7 @@ func (p *Provider) RegisterRecipientData(ctx context.Context, data *driver.Recip
 
 func (p *Provider) RegisterSigner(ctx context.Context, identity driver.Identity, signer driver.Signer, verifier driver.Verifier, signerInfo []byte) error {
 	defer func() {
-		p.isMeCache.Add(identity.String(), true)
+		p.isMeCache.Add(identity.UniqueID(), true)
 	}()
 	return p.SigService.RegisterSigner(ctx, identity, signer, verifier, signerInfo)
 }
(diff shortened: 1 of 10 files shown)


token/services/identity/provider.go
```diff
@@ -142,18 +135,14 @@ func (p *Provider) IsMe(ctx context.Context, identity driver.Identity) bool {
 
 func (p *Provider) RegisterRecipientIdentity(id driver.Identity) error {
 	p.Logger.Debugf("Registering identity [%s]", id)
-	p.isMeCacheLock.Lock()
-	p.isMeCache[id.String()] = false
-	p.isMeCacheLock.Unlock()
+	p.isMeCache.Add(id.String(), false)
```
alexandrosfilios: I would use String only for logging. UniqueID would fit better here.

token/services/utils/cache/nocache.go
```diff
@@ -0,0 +1,34 @@
...
+	return v, false, err
+}
+
+func (n *NoCache[T]) Add(key string, value T) {
+}
+
+func (n *NoCache[T]) Delete(key string) {
+}
+
+func Zero[T any]() T {
```
alexandrosfilios: You can use `utils.Zero[T]`

alexandrosfilios (approved):
LGTM modulo comments

token/services/identity/provider.go
```diff
@@ -74,7 +78,7 @@ func NewProvider(
 		SigService:              sigService,
 		Binder:                  binder,
 		enrollmentIDUnmarshaler: enrollmentIDUnmarshaler,
-		isMeCache:               make(map[string]bool),
+		isMeCache:               secondcache.NewTyped[bool](1000),
```
AkramBitar, line 81: I think the size of cache should be tunable and customers can change it based on the stress of the system. Maybe we need a config that specify the size of caches and we will have defaults if not defined.

PR #1180, opened 2025-07-31
expand the reach of context.Context #1135

adecaro:


SaidAltury-ibm (approved):
LGTM

token/services/identity/role/role.go
```diff
@@ -21,8 +21,8 @@ import (
 type localMembership interface {
 	DefaultNetworkIdentity() driver.Identity
 	IsMe(ctx context.Context, id driver.Identity) bool
-	GetIdentityInfo(label string, auditInfo []byte) (idriver.IdentityInfo, error)
-	GetIdentifier(id driver.Identity) (string, error)
+	GetIdentityInfo(ctx context.Context, label string, auditInfo []byte) (idriver.IdentityInfo, error)
+	GetIdentifier(ctx context.Context, id driver.Identity) (string, error)
 	GetDefaultIdentifier() string
```
mbrandenburger, line 26: should `GetDefaultIdentifier()` also receive a `ctx` to be consistent with `GetIdentifyInfo()`?

token/services/identity/role/role.go
```diff
@@ -101,48 +101,48 @@ func (r *Role) mapStringToID(ctx context.Context, v string) (driver.Identity, st
...
-		r.logger.Debugf("passed empty label")
+		r.logger.DebugfContext(ctx, "passed empty label")
 		return nil, defaultIdentifier, nil
 	case label == defaultIdentifier:
-		r.logger.Debugf("passed default identifier")
+		r.logger.DebugfContext(ctx, "passed default identifier")
 		return nil, defaultIdentifier, nil
 	case label == defaultNetworkIdentity.UniqueID():
-		r.logger.Debugf("passed default identity")
+		r.logger.DebugfContext(ctx, "passed default identity")
```
mbrandenburger, line 116: I am wondering if these logs really must be traces. 

Do you have a strategy in mind when something is logged vs traced+logged?

I have the fear that adding more and more traces will make it difficult to read.

mbrandenburger (commented):
While I think adding a `ctx` makes sense for many functions the SDK provides, I am also wondering if we are over using it. In particular, if the context is only used for logging/tracing rather than other functional things like aborting a request.

adecaro:
good point, @mbrandenburger. It makes sense to have trace events for sure when something is in the critical path. 
These events give us also information about the timing.

PR #1182, opened 2025-07-31
optimize membership/TypedIdentityInfo #1181

adecaro:


token/services/db/sql/common/identity.go
```diff
@@ -317,6 +331,51 @@ func (db *IdentityStore) GetSignerInfo(ctx context.Context, identity []byte) ([]
...
+func (db *IdentityStore) RegisterIdentityDescriptor(ctx context.Context, descriptor *idriver.IdentityDescriptor, alias tdriver.Identity) (err error) {
+	if descriptor == nil {
+		return errors.New("identity descriptor is nil")
+	}
+	tx, err := db.writeDB.BeginTx(ctx, nil)
+	if err != nil {
+		return err
+	}
+	defer func() {
+		if err != nil {
```
arner, line 343: Shouldn't we rollback on errors? Now we don't finish the transaction if storeIdentityData or storeSignerInfo fails.

token/services/db/sql/common/common.go
```diff
@@ -18,3 +18,12 @@ func Close(closer Closer) {
 		logger.Errorf("failed closing connection: %s", err)
 	}
 }
+
+func CloseFunc(closer func() error) {
+	if closer == nil {
```
arner, line 23: can it be nil?

token/services/db/sql/common/identity.go
```diff
@@ -317,6 +331,51 @@ func (db *IdentityStore) GetSignerInfo(ctx context.Context, identity []byte) ([]
 	return common.QueryUniqueContext[[]byte](ctx, db.readDB, query, args...)
 }
 
+func (db *IdentityStore) RegisterIdentityDescriptor(ctx context.Context, descriptor *idriver.IdentityDescriptor, alias tdriver.Identity) (err error) {
```
arner, line 334: Just wondering, do we always expect only one alias? 

Maybe we can be more explicit, and reduce our database INSERTS, by just adding a column for the alias to the identity table? Then if we query we do "where id is not empty and equals %1, or the hash equals %1". It always felt a bit weird to me that we create a `view.Identity("alice")`, where the "alice" bytes are stored in the same column as the unique id of the identity. If we need to we could add one more column even (e.g. one for the local alias, one for the official id like IBAN or public Eth address, and one for the hash). Come to think of it, how do we deal with duplicate aliases? Do or should we enforce uniqueness?

arner: We could even consider having the SignerInfo and IdentityData in the same table. I don't know our usage patterns and what the optimal solution would be from a performance point of view (we would have to query for verifiers in this same table, and they would each have an empty Signer field)... If the storage of identities is a bottleneck, it might be worth some benchmarking? It would reduce writes a lot, and considering the use of IdentityDescriptor it does make logical sense to have the verifier identity in the same row.

token/services/identity/provider.go
```diff
@@ -13,12 +13,15 @@ import (
...
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/cache/secondcache"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/collections"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/driver"
 	idriver "github.com/hyperledger-labs/fabric-token-sdk/token/services/identity/driver"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/logging"
 	"go.uber.org/zap/zapcore"
 )
 
+var logger = logging.MustGetLogger()
```
arner: shouldn't this have a name? Do we need it, since it's also part of the Provider?

token/services/identity/provider.go
```diff
@@ -167,52 +192,211 @@ func (p *Provider) GetRevocationHandler(ctx context.Context, identity driver.Ide
...
+}
+
+func (p *Provider) localRegisterVerifier(ctx context.Context, identity driver.Identity, verifier driver.Verifier) error {
+	if verifier == nil {
+		return errors.New("invalid verifier, expected a valid instance")
+	}
+
+	idHash := identity.UniqueID()
+	entry := &VerifierEntry{Verifier: verifier}
+	if logger.IsEnabledFor(zapcore.DebugLevel) {
```
arner: p.logger? also for the other logs in the provider?

token/services/identity/provider.go
```diff
@@ -167,52 +192,211 @@ func (p *Provider) GetRevocationHandler(ctx context.Context, identity driver.Ide
...
+		// Nothing to do here
+		return nil
+	}
+
+	idHash := identity.UniqueID()
+	logger.DebugfContext(ctx, "register signer and verifier [%s]:[%s][%s]", idHash, logging.Identifier(signer), logging.Identifier(verifier))
+	// First check with read lock
+	s, ok := p.signers.Get(idHash)
+	if ok {
+		logger.Warnf("another signer bound to [%s]:[%s][%s] from [%s]", identity, logging.Identifier(s), logging.Identifier(signer), string(s.DebugStack))
```
arner: we get a lot of these when we start the node. Should it be a warning? Maybe only if it's different from what we're trying to save?

token/services/identity/provider.go
```diff
@@ -167,52 +192,211 @@ func (p *Provider) GetRevocationHandler(ctx context.Context, identity driver.Ide
...
+		return nil
+	}
+
+	entry := &SignerEntry{Signer: signer}
+	if logger.IsEnabledFor(zapcore.DebugLevel) {
+		entry.DebugStack = debug.Stack()
+	}
+	p.signers.Add(idHash, entry)
+
+	// store, if a failure happens then remove the entry
```
arner: what if we turn it around, save it first in storage and then in cache? then we might not have to delete on error?

token/services/identity/provider.go
```diff
@@ -167,52 +192,211 @@ func (p *Provider) GetRevocationHandler(ctx context.Context, identity driver.Ide
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed deserializing identity for signer [%s]", identity)
+	}
+	entry = &SignerEntry{Signer: signer}
+	if logger.IsEnabledFor(zapcore.DebugLevel) {
+		entry.DebugStack = debug.Stack()
+	}
+	p.signers.Add(idHash, entry)
+	if err := p.storage.StoreSignerInfo(ctx, identity, nil); err != nil {
+		return nil, errors.Wrap(err, "failed to store entry in storage for the passed signer")
```
arner, line 274: also here maybe add it to the cache after storing it in the db?

token/services/identity/provider.go
```diff
@@ -167,52 +192,211 @@ func (p *Provider) GetRevocationHandler(ctx context.Context, identity driver.Ide
...
+	id := descriptor.Identity.UniqueID()
+	aliasID := alias.UniqueID()
+
+	// isMe?
+	{
+		p.isMeCache.Add(id, true)
+	}
+
+	// signers
+	{
```
arner: do we need these {}?

token/services/db/sql/common/identity.go
```diff
@@ -317,6 +331,51 @@ func (db *IdentityStore) GetSignerInfo(ctx context.Context, identity []byte) ([]
...
+	if err != nil {
+		return errors.Wrapf(err, "failed to store audit info for descriptor's identity")
+	}
+
+	if !descriptor.Identity.Equal(alias) {
+		h = alias.UniqueID()
+		_, err = db.storeSignerInfo(ctx, tx, h, alias, descriptor.SignerInfo)
+		if err != nil {
+			return errors.Wrapf(err, "failed to store signer info for alias")
+		}
```
AkramBitar, line 372: In case if errors, do we need to Rollback?

token/services/db/sql/common/identity.go
```diff
@@ -362,3 +421,56 @@ func (db *IdentityStore) GetSchema() string {
...
+	query, args := q.InsertInto(db.table.Signers).
+		Fields("identity_hash", "identity", "info").
+		Row(h, id, info).
+		Format()
+
+	logger.Debug(query, h, hash.Hashable(info))
+	exists := false
+	_, err := tx.ExecContext(ctx, query, args...)
+	if err != nil {
+		if err != nil && errors.Is(db.errorWrapper.WrapError(err), driver2.UniqueKeyViolation) {
```
AkramBitar: We can remove if err != nil since the outer if is if err != nil

token/services/db/sql/common/identity.go
adecaro: yes, for now it is always one. I have been refactoring the code slowly to identify exactly what was happening. The current functions should reflect exactly what we need.

adecaro: for sure we will need to see how to optimize the data model. I just wanted to reduce with this PR the number of DB connections per pseudonym generated .

token/services/identity/provider.go
adecaro: no need for names anymore, it is inferred from the package.

token/services/identity/provider.go
adecaro: good catch. I'll double check

token/services/identity/provider.go
adecaro: I'll revise the RegisterSigner function too.

token/services/identity/provider.go
adecaro: syntactic sugar, a way to reuse the variable names with different types

token/services/db/sql/common/identity.go
arner: I'm not asking for a change in this PR, in fact it was indeed this refactor that made it more clear :). Might be a nice next step though.

token/services/identity/provider.go
arner: Interesting! I had to check it out, but it's not necessary because the if block is already providing the scope: https://go.dev/play/p/Re71tNWKRso

token/services/db/sql/common/common.go
adecaro: just for consistency with the Close function

token/services/db/sql/common/identity.go
adecaro: This is what happens. CloseFunc will invoke the passed function ignoring the returned value.

token/services/db/sql/common/identity.go
adecaro: the Rollback happens in the defer above if an error occurred. Notice the named output parameter that signals that something wrong happened.
This way, we don't miss the rollback.

token/services/db/sql/common/identity.go
adecaro: good catch.

token/services/identity/provider.go
adecaro: fixed

token/services/identity/provider.go
adecaro: did you open an Issue once about this? I would address it in another PR.

token/services/identity/provider.go
adecaro: cleaned up. The code is leaner now.

token/services/identity/provider.go
adecaro: cleared

token/services/db/sql/common/identity.go
adecaro: fixed

arner (approved):

Issue #1184, opened 2025-08-04
Integrate with FSC PR 1007


HayimShaul:
Check that FTS works after integrating with FSC PR 1007

PR #1185, opened 2025-08-04
use FSC with PR 1007

HayimShaul:
solves issue #1184


adecaro:
FSC gets updates in #1188

Issue #1189, opened 2025-08-07
network service: query states


adecaro:
Currently, it is not possible to retrieve from the committer the value of a given key. This Issues asks to add a function to the Ledger interface to do that.

PR #1190, opened 2025-08-07
network service: query states #1189

adecaro:
This PR exposes a `GetStates` function to query state values directly from the ledger.


KElkhiyaoui (approved):

arner (approved):
nice! How does it work, does it use a system chaincode to query the state directly?

Issue #1209, opened 2025-08-16
config service: import a TMS configuration


adecaro:
This is issue is about the ability to load into the configuration service a new TMS configuration.

Linked to https://github.com/hyperledger-labs/fabric-smart-client/issues/1029

PR #1210, opened 2025-08-17
config service: import a TMS configuration #1209

adecaro:


token/core/tms.go
```diff
@@ -38,7 +38,7 @@ type PublicParameters struct {
 // It is responsible for creating token management services for different networks.
 type TMSProvider struct {
 	logger                  logging.Logger
-	configProvider          ConfigProvider
+	configProvider          ConfigService
```
mbrandenburger: I suggest to also rename the variable `configProvider` to `configService`

token/core/tms.go
```diff
@@ -47,10 +47,14 @@ type TMSProvider struct {
...
-func NewTMSProvider(logger logging.Logger, configProvider ConfigProvider, pps PublicParametersStorage, tokenDriverService *TokenDriverService) *TMSProvider {
+func NewTMSProvider(
+	configService ConfigService,
+	pps PublicParametersStorage,
+	tokenDriverService *TokenDriverService,
+) *TMSProvider {
 	ms := &TMSProvider{
-		logger:                  logger,
-		configProvider:          configProvider,
+		logger:                  logging.MustGetLogger(),
```
mbrandenburger: should this logger get a specific name? this code now attaches the package logger to the TMS, what is the actual idea to give to the TMS provider a special instance rather than accessing the logger through the package logger?

Maybe this is is a separate discussion - we used to use package based loggers - only very special situation, we were creating a specific logger and attached to a struct.

token/sdk/tms/config.go
```diff
@@ -0,0 +1,36 @@
...
+
+type ConfigServiceWrapper struct {
+	*config.Service
+}
+
+func NewConfigServiceWrapper(service *config.Service) *ConfigServiceWrapper {
+	return &ConfigServiceWrapper{Service: service}
+}
+
+func (c *ConfigServiceWrapper) Configurations() ([]driver.Configuration, error) {
```
mbrandenburger, line 22: @adecaro it is not so clear to me why we add the `ConfigServiceWrapper`?

token/services/config/service_test.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/config"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// TestConfigurations tests the Configurations function
+func TestConfigurations(t *testing.T) {
+	// create a new config service by loading the config file
+	cp, err := config.NewProvider("./testdata/token0")
+	assert.NoError(t, err)
```
mbrandenburger: Is there a particular reason why using `assert` and `required` in this test? I personally would always prefer `required` as it calls `t.FailNow()` and thus stops the tests immediately.

token/core/tms.go
adecaro: okay

token/core/tms.go
adecaro: agreed

token/sdk/tms/config.go
adecaro: Yes, to be able to implement these interfaces dig.As(new(ftscore.ConfigService), new(db2.ConfigService), new(token.ConfigService)),

*config.Service returns structs as Go prescribes. But a local interface uses also interfaces for the returned values so to be able to mock them for the unit-tests.

This is reason why we endup with these adaptors.

token/services/config/service_test.go
adecaro: agreed

adecaro:
@mbrandenburger , please, have another look. Thanks.

Issue #1009, opened 2025-03-23
dlog v1 finalzation: enforce at list one issuer


adecaro:
Currently, the dlog token driver supports public parameters with no issuer specified. This is interpreted as anyone is able to issue. This Github Issue asks to enforce that at least one issuer is specified in the public parameters.

PR #1214, opened 2025-08-27
FSC Update: endpoint.Service: Bind multiple ephemeral identities at the same time #1009

adecaro:


HayimShaul:
@adecaro @arner 
Can you please explain you comment? is there any changes I should make?

adecaro:
@arner , good point. Yesterday, I was discussing with @ale-linux about reworking the FSC's identity provider and endpoint service. Your comments will be used during that rework.

PR #1216, opened 2025-08-29
enhanced redeem: make issuer signature optional #1134

adecaro:
This PR allows the developer to define public parameters without issuers for the two available token drivers.
In this case, the redeem will not require an additional signature from the issuer.


arner (approved):
Considering the potential impact of having no issuers specified in the params, should we add a short explanation to the help text of the tokengen cli about why it's not mandatory (or alternatively a warning log when generating)?

Besides that, LGTM.

adecaro:
@arner , good point, I'll add warnings

PR #1229, opened 2025-10-08
fabricx integration

adecaro:


integration/token/fungible/views/fabricx/tmsdeploy/view.go
```diff
@@ -0,0 +1,51 @@
...
+		f.PublicParamsRaw,
+	)
+}
+
+type ViewFactory struct{}
+
+func (p *ViewFactory) NewView(in []byte) (view.View, error) {
+	f := &View{Deploy: &Deploy{}}
+	err := json.Unmarshal(in, f.Deploy)
+	assert.NoError(err, "failed unmarshalling input")
```
mbrandenburger: As a general comment here, I think we should not use the assert package within a view anymore even though this view is part of testing code.

token/services/network/common/rws/translator/rwset.go
```diff
@@ -8,8 +8,8 @@ package translator
 
 import (
 	"crypto/sha256"
+	"encoding/hex"
 
-	"github.com/gobuffalo/packr/v2/file/resolver/encoding/hex"
```
mbrandenburger, line 12: Beautiful catch

integration/token/fungible/views/fabricx/tmsdeploy/view.go
adecaro: let me update

token/services/network/common/rws/translator/rwset.go
adecaro: by chance

PR #1238, opened 2025-10-29
Proper Error Handling and Normalization

adecaro:
This PR does the following:
- It makes sure that `GetManagementService` returns an error when the request TMS cannot be returned. 
- It rework the Normalizer whose job is to fill the missing options with the best available alternatives given the current configurations. This is particular useful when there is only one TMS. In this case, the developer doesn't need which one they want.
- make sure that when marshaling and unmarshaling a transaction, network and namespace are properly set

### DIFF
--- a/integration/token/fungible/update/update_test.go
+++ b/integration/token/fungible/update/update_test.go
@@ -23,7 +23,7 @@ import (
 )
 
 var _ = Describe("EndToEnd", func() {
-	for _, t := range integration.WebSocketNoReplicationOnly {
+	for _, t := range integration.AllTestTypes {
 		Describe("Tokens Upgrade with Auditor ne Issuer", t.Label, func() {
 			ts, selector := newTestSuite(t.CommType, 64, []common.TMSOpts{
 				{
--- a/token/sdk/tms/normalizer.go
+++ b/token/sdk/tms/normalizer.go
@@ -43,12 +43,10 @@ func (p *tmsNormalizer) Normalize(opt *token.ServiceOptions) (*token.ServiceOpti
 
 	logger.Debugf("normalizing opts [%v]", opt)
 	if len(opt.Network) != 0 {
-		// filter configurations by netowkr
-		for i, c := range configs {
-			if c.ID().Network != opt.Network {
-				configs = append(configs[:i], configs[i+1:]...)
-			}
-		}
+		// filter configurations by network
+		configs = Filter(configs, func(c driver.Configuration) bool {
+			return c.ID().Network == opt.Network
+		})
 		if len(configs) == 0 {
 			return nil, errors.Errorf("no token management service config found for network [%s]", opt.Network)
 		}
@@ -56,11 +54,9 @@ func (p *tmsNormalizer) Normalize(opt *token.ServiceOptions) (*token.ServiceOpti
 
 	if len(opt.Channel) != 0 {
 		// filter configurations by channel
-		for i, c := range configs {
-			if c.ID().Channel != opt.Channel {
-				configs = append(configs[:i], configs[i+1:]...)
-			}
-		}
+		configs = Filter(configs, func(c driver.Configuration) bool {
+			return c.ID().Channel == opt.Channel
+		})
 		if len(configs) == 0 {
 			return nil, errors.Errorf("no token management service config found for network and channel [%s:%s]", opt.Network, opt.Channel)
 		}
@@ -68,11 +64,9 @@ func (p *tmsNormalizer) Normalize(opt *token.ServiceOptions) (*token.ServiceOpti
 
 	if len(opt.Namespace) != 0 {
 		// filter configurations by namespace
-		for i, c := range configs {
-			if c.ID().Namespace != opt.Namespace {
-				configs = append(configs[:i], configs[i+1:]...)
-			}
-		}
+		configs = Filter(configs, func(c driver.Configuration) bool {
+			return c.ID().Namespace == opt.Namespace
+		})
 		if len(configs) == 0 {
 			return nil, errors.Errorf("no token management service config found for network, channel, and namespace [%s:%s:%s]", opt.Network, opt.Channel, opt.Namespace)
 		}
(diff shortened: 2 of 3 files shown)


integration/ports.go
```diff
@@ -60,8 +60,8 @@ var (
 
 	AllTestTypes = []*InfrastructureType{
 		WebSocketNoReplication,
-		LibP2PNoReplication,
-		WebSocketWithReplication,
+		// LibP2PNoReplication,
```
aaadir: don't forget to uncomment

integration/token/common/views/finality.go
```diff
@@ -40,8 +40,8 @@ func (r *TxFinalityView) Call(context view.Context) (interface{}, error) {
 	errs := make(chan error, 2)
 
 	// Listen for finality from vault
-	tms := token.GetManagementService(context, token.WithTMSID(tmsID))
-	assert.NotNil(tms)
+	tms, err := token.GetManagementService(context, token.WithTMSID(tmsID))
+	assert.NoError(err)
```
aaadir, line 44: in the code below you check both err and also that tms isn't nil and give corresponding messages. May do the same everywhere for consistency.

integration/token/fungible/views/owner.go
```diff
@@ -27,7 +27,9 @@ type SetTransactionOwnerStatusView struct {
 }
 
 func (r *SetTransactionOwnerStatusView) Call(context view.Context) (interface{}, error) {
-	owner := ttx.NewOwner(context, token.GetManagementService(context))
+	tms, err := token.GetManagementService(context)
+	assert.NoError(err, "failed getting management service")
```
aaadir, line 31: again, you may also want to assert that tms isn't nil for consistency. there are some more like this below.
since this is a very common check of both err and tms, consider letting GetManagementService() itself check that the tms that it returns is not nil, and if it is nil then GetManagementService would return a corresponding error indicating a nil tms.

Then in all the places that call GetManagementService you can just assert that there's no err.
If you want to be on the safe side before using the tms you can have a single assert that err != nil && tms != nil and print the err message.

token/sdk/tms/normalizer_test.go
```diff
@@ -0,0 +1,157 @@
...
+func TestNormalize_FilterByNetwork_HappyPath(t *testing.T) {
+	configs := []driver.Configuration{
+		&fakeConfig{id: driver.TMSID{Network: "net1", Channel: "c1", Namespace: "ns1"}},
+		&fakeConfig{id: driver.TMSID{Network: "net2", Channel: "c2", Namespace: "ns2"}},
+	}
+	cs := &fakeConfigService{configs: configs}
+	norm := &fakeNormalizer{}
+	n := tms.NewTMSNormalizer(cs, norm)
+
+	opt := &token.ServiceOptions{Network: "net2"}
```
aaadir, line 87: is it useful to support some kind of regular expression here, for example Channel: "c[200-299]" for the case where the user doesn't care or remember what precise channel is used?

integration/token/fungible/update/update_test.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 var _ = Describe("EndToEnd", func() {
-	for _, t := range integration.AllTestTypes {
+	for _, t := range integration.WebSocketNoReplicationOnly {
```
AkramBitar: What about (P2P) protocol?

integration/token/common/views/finality.go
adecaro: Indeed. This was because no error was returned before. The general assumption is that if no error is returned, all the other outputs are expected to be useful.

integration/token/fungible/views/owner.go
adecaro: We can avoid the check on tms because the general assumption is that if no error occurs, the other outputs are usable.

integration/ports.go
adecaro: good catch

token/sdk/tms/normalizer_test.go
adecaro: Too complex to handle. Simplicity first unless there is a requirement for.

token/sdk/tms/normalizer.go
```diff
@@ -0,0 +1,90 @@
...
+			}
+		}
+		if len(configs) == 0 {
+			return nil, errors.Errorf("no token management service config found for network [%s]", opt.Network)
+		}
+	}
+
+	if len(opt.Channel) != 0 {
+		// filter configurations by channel
+		for i, c := range configs {
```
AkramBitar: Iterating on configs and appending to it in configs = append(configs[:i], configs[i+1:]...)

You’re modifying (appending to) the same slice while iterating over it, which causes skipped elements or unpredictable behavior.

Suggestion: 

for _, c := range configs {
	if c.ID().Channel == opt.Channel {
		filtered = append(filtered, c)
	}
}
configs = filtered

integration/token/fungible/views/owner.go
aaadir: but if that is the assumption, then why do you check that tms is not nil after asserting that there's no error? e.g. in the above change to integration/token/fungible/views/info.go and in many other places here.

adecaro: The check was already there, I simply didn't remove it. 
I can remove those checks on `tms` if they are confusing.
Shall I remove them? @aaadir

integration/token/fungible/update/update_test.go
adecaro: good catch.

token/sdk/tms/normalizer.go
adecaro: great catch, I'll update.

integration/token/fungible/views/owner.go
aaadir: hmm.. maybe just leave the tms checks then. 
it's strange because they're all dead code, but the reader doesn't know that without knowing the assumption, so will not be confused. It's also perhaps safe to leave these checks just in case the assumption stops holding sometime in the future.

Issue #1213, opened 2025-08-27
Provide FSC/FTS node a resync process with the ledger after a restart or crash


mffrench:
Hello,

after some stop on an inopportune time, it seems that our FSC/FTS node has lost token synchronisation (token was locked) with the actual HLF tx (which indeed was committed). It seems to us it would be important for FSC/FTS node resiliency to provide a startup process to insure all played transactions within FSC/FTS node are in sync with HLF finality.

**Summary of issue:** 

it seems to us that there was transfer in progress at the moment of the restart and so FTS node didn't reflected the HLF committed state within the database after the restart. That result into locked tokens we're not able reuse (lack of funds error). 

**Details of issue :** 

We try to redeem 10 tokens but we have this error:

```
[1;32m[ info	| 2025-08-18 12:42:52.980 (UTC) ]  [0m[1;37m[ AMS.ffhlflcidemix@BDF - settlement_engine_ff_zkat.go:85 ]  [0mredeem ttr : &fabflow.TransferTokenRequest{
	SenderWalletEPName: WFREURPBGSGGXXDLT-PBG_WALLET_001
	Type: EUR
	Quantity: 695195.24
	
...
[31m2025-08-18 12:42:59.785 UTC [token-sdk.tokens] parse -> ERRO 23f[0m could not unmarshal identity when storing token: failed to unmarshal to TypedIdentity: asn1: syntax error: sequence truncated 
```

On the 19th august  (1 day after a reboot):

```
[...]

PR #1239, opened 2025-10-29
fix: delete token loken even if it doesn't exist on request table

anisherbadji1:
provide a fix to lock issue that happened after a VM was stopped while still processing a transaction

Fix: use sort of left join instead of USING that does inner join and implies tx to be present in both tables

related ticket https://github.com/hyperledger-labs/fabric-token-sdk/issues/1213#issuecomment-3443890632


adecaro (approved):
LGTM

PR #1240, opened 2025-10-29
Bump FSC

AkramBitar:
This PR updates FSC dep [8bde666](https://github.com/hyperledger-labs/fabric-smart-client/commit/8bde666534da00f2e8644147e330896f6990c0c1) and integrates corresponding changes.


AkramBitar:
> @AkramBitar @angelo the tests are most probably failing due to a bug in [hyperledger-labs/fabric-smart-client@ea5aa20#diff-65733d2b46f196197a0740de95a8e1ba1ff8cc35fba42e7d967e8b52a92d8b3dR217](https://github.com/hyperledger-labs/fabric-smart-client/commit/ea5aa2095c1dab12b0130b08f6db325803df013c#diff-65733d2b46f196197a0740de95a8e1ba1ff8cc35fba42e7d967e8b52a92d8b3dR217) I have introduced. Blame on me 😩
> 
> In detail, my commit introduced a function to derive a lock key as follows
> 
> ```
> func createLockTag(m string) uint64 {
> 	h := sha256.Sum256([]byte(m))
> 	return binary.BigEndian.Uint64(h[:])
> }
> ```
> 
> However, this method must return `int64` instead of `uint64` as `pg_advisory_xact_lock` requires a `bigint` (https://www.postgresql.org/docs/9.1/functions-admin.html). The integration tests fail when `createLockTag` produces a number that is too large to fit into a `int64`.
> 
> For example `14913332284672489711` is too large to fit in a signed 64-bit integer (bigint), which is the valid input type for pg_advisory_xact_lock(bigint).
> 
> While we can easily fix this in FSC, I am more concerned about the fact that this bug was not detected with the current integration tests and unit tests in FSC.
> 
> It seems that the tokenSDK is the main consumer of `platform/view/services/storage/driver/sql/postgres/notifier.go`. What about moving it into the tokenSDK?

@mbrandenburger 
Thanks a lot for your help with this.
I’m fine with moving it to tokenSDK; however, I noticed that it’s currently part of the Postgres package in FSC, which provides a generic infrastructure for database access. The notifier.go file includes a test for DB event notifications, which could serve as a generic implementation for other components if needed.
I agree that, for now, tokenSDK is the only consumer, but this might be useful for other consumers in the future.
What do you think?

token/services/network/fabric/finality/deliveryqs.go
```diff
@@ -80,7 +80,7 @@ func (q *DeliveryScanQueryByID) queryByID(ctx context.Context, keys []driver.TxI
 		return
 	}
 
-	startingBlock := finality.MaxUint64(FirstBlock, lastBlock-NumberPastBlocks)
+	startingBlock := max(FirstBlock, lastBlock-NumberPastBlocks)
```
HayimShaul, line 83: I suppose there was a reason they used MaxUint64 and not plain max.
Maybe max doesn't always work on the hashes? (e.g., on 32bit computers?)

token/services/network/fabric/lookup/deliveryqs.go
```diff
@@ -105,7 +104,7 @@ func (q *DeliveryScanQueryByID) queryByID(ctx context.Context, keys []driver.PKe
 		return
 	}
 
-	startingBlock := finality.MaxUint64(FirstBlock, lastBlock-NumberPastBlocks)
+	startingBlock := max(FirstBlock, lastBlock-NumberPastBlocks)
```
HayimShaul, line 107: I suppose there was a reason they used MaxUint64 and not plain max.
Maybe max doesn't always work on the hashes? (e.g., on 32bit computers?)

HayimShaul (approved):
looks good, but please verify my comments about the max function

token/services/network/fabric/finality/deliveryqs.go
mbrandenburger: Go's built-in `max()` and `min()` functions where introduced just with go1.21 https://go.dev/ref/spec#Min_and_max. We had to have those helper functions in the past :)

Issue #940, opened 2025-03-07
Check the `IssueAction` against the passed metadata.


KElkhiyaoui:
`VerifyIssue` in `zkatdlog/nogh/issue.go` should check that the passed metatdata matches the outputs of the `IssueAction`

PR #1242, opened 2025-10-29
dlog v1 finalization: check the IssueAction against the passed metadata. #940

adecaro:
This PR addresses #940

### DIFF
--- a/token/core/zkatdlog/nogh/v1/issue.go
+++ b/token/core/zkatdlog/nogh/v1/issue.go
@@ -224,7 +224,7 @@ func (s *IssueService) VerifyIssue(ctx context.Context, ia driver.IssueAction, o
 		if err := metadata.Deserialize(outputMetadata[i].OutputMetadata); err != nil {
 			return errors.Wrap(err, "failed unmarshalling metadata")
 		}
-		if err := metadata.Validate(); err != nil {
+		if err := metadata.Validate(true); err != nil {
 			return errors.Wrap(err, "invalid metadata")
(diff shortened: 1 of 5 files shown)


token/core/fabtoken/v1/issue.go
```diff
@@ -108,7 +108,7 @@ func (s *IssueService) Issue(ctx context.Context, issuerIdentity driver.Identity
 }
 
 // VerifyIssue checks if the outputs of an IssueAction match the passed tokenInfos
-func (s *IssueService) VerifyIssue(tr driver.IssueAction, metadata []*driver.IssueOutputMetadata) error {
+func (s *IssueService) VerifyIssue(ctx context.Context, ia driver.IssueAction, metadata []*driver.IssueOutputMetadata) error {
 	// TODO:
```
AkramBitar, line 112: I am wondering if to add debug message that says "this method is not implemented".

adecaro: it definitely needs to be implemented and it can be a good task for the team.

adecaro: @HayimShaul , please, take the task on. Thanks.

adecaro: your concern was about the implementation of the function, no?

adecaro: but good point, please, identify the TODOs in the code, classify them in category so that we can assign priorities and address. Thanks much.

PR #1249, opened 2025-11-02
Document education session 1

AkramBitar:


docs/education/sessions.md
```diff
@@ -0,0 +1,111 @@
+## Fabric Token SDK - Session 1 - 29.10.2025
```
adecaro, line 1: I think you can remove both date and `Presenter`.
It is very nice to have this in form of sessions or lessons.

docs/education/README.md
```diff
@@ -0,0 +1,7 @@
+ # Education Resources
```
adecaro: Please, add a link to this page form the main README.md

docs/education/sessions.md
aaadir: Can we add a link to the box that keeps the corresponding video recording?

docs/education/README.md
AkramBitar: @adecaro Which page? the record in box or a link were the education session docs. found?

adecaro: to `docs/education/README.md`, yes.

Issue #925, opened 2025-03-06
identity service


adecaro:
In the `x509` package, we don't check the header of the loaded PEM file.
This Github Issue asks to address the following TODO:
```
// TODO: also check that the type is what we expect (cert vs key..)
```
We can pass to the function a string to compare the PEM's header to.

PR #1252, opened 2025-11-04
check the type of PEM block

HayimShaul:
Address issue #925 and check the type of PEM block

### DIFF
--- a/token/services/ttx/collectactions.go
+++ b/token/services/ttx/collectactions.go
@@ -19,7 +19,7 @@ type Actions struct {
 	Transfers []*ActionTransfer
 }
 
-// ActionTransfer describe a transfer operation.
+// ActionTransfer describe a transfer operation
 type ActionTransfer struct {
 	// From is the sender
 	From view.Identity


token/services/identity/x509/crypto/io.go
```diff
@@ -31,10 +31,23 @@ func readPemFile(file string) ([]byte, error) {
 	}
 
 	b, _ := pem.Decode(bytes)
-	if b == nil { // TODO: also check that the type is what we expect (cert vs key..)
+	if b == nil {
 		return nil, errors.Errorf("no pem content for file %s", file)
 	}
 
+	// Also check that the type is what we expect (cert vs key..)
+	expectedTypes := map[string]bool{
```
adecaro: a slice of strings should suffice then you can use `slices.Contains` . You can declare this slice as global variable.

Please, add also a unit-test for this function. Thanks

adecaro (changes requested):
Changes request

token/services/identity/x509/crypto/io.go
```diff
@@ -31,10 +32,24 @@ func readPemFile(file string) ([]byte, error) {
...
 
 	b, _ := pem.Decode(bytes)
-	if b == nil { // TODO: also check that the type is what we expect (cert vs key..)
+	if b == nil {
 		return nil, errors.Errorf("no pem content for file %s", file)
 	}
 
+	// Also check that the type is what we expect (cert vs key..)
+	// AllowedPEMTypes defines the valid PEM block types we accept
+	var AllowedPEMTypes = []string{
```
adecaro: make this global, please.

Issue #1233, opened 2025-10-09
`AcceptView` may expose a signing oracle


ale-linux:
`AcceptView` responds to two signature requests without any specific validation of what payload is being signed. As such, the node may be tricked into signing something else.

PR #1254, opened 2025-11-05
AcceptView may expose a signing oracle #1233

adecaro:
This PR reworks `AcceptView`. This view must be used only when no signature is required by the acceptor of tokens.

In another PR we will rework `EndorseView` that is supposed to generate signatures and @ale-linux 's comments are relevant.

### DIFF
--- a/token/services/ttx/accept.go
+++ b/token/services/ttx/accept.go
@@ -17,25 +17,27 @@ import (
 
 // AcceptView is used to accept tokens without the need to generate any signature.
 // This is a view executed by a responder.
+// This view is to be used in conjunction with CollectEndorsementsView.
+// Usually, AcceptView is preceded by an invocation of `tx.ReceiveTransaction(context)`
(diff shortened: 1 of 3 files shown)


token/services/ttx/accept.go
```diff
@@ -95,84 +66,36 @@ func (s *AcceptView) Call(context view.Context) (interface{}, error) {
 	return s.tx, nil
 }
 
-func (s *AcceptView) respondToSignatureRequests(context view.Context) error {
-	requestsToBeSigned, err := requestsToBeSigned(context.Context(), s.tx.TokenRequest)
+func (s *AcceptView) ack(context view.Context) error {
+	// Send back an acknowledgement
+	txRaw := s.tx.FromRaw
```
AkramBitar, line 69: Can this be nil? do we need to validate it before continue?

adecaro: done

AkramBitar (approved):

Issue #1255, opened 2025-11-05
EndorseView may expose a signing oracle


adecaro:
`EndorseView` responds to two signature requests without any specific validation of what payload is being signed. As such, the node may be tricked into signing something else.

PR #1257, opened 2025-11-05
EndorseView may expose a signing oracle #1255

adecaro:
This PR reworks the `EndorseView` to make sure it does check what is received.
The PR introduces also code for unit-testing of the view that will be useful also for other views.

A correlated bug discovered during the development of this PR: https://github.com/hyperledger-labs/fabric-token-sdk/issues/1266

### DIFF
--- a/integration/ports.go
+++ b/integration/ports.go
@@ -60,8 +60,8 @@ var (
 
 	AllTestTypes = []*InfrastructureType{
 		WebSocketNoReplication,
-		// LibP2PNoReplication,
-		// WebSocketWithReplication,
+		LibP2PNoReplication,
+		WebSocketWithReplication,
 	}
 )
 
--- a/integration/token/fungible/dlog/dlog_test.go
+++ b/integration/token/fungible/dlog/dlog_test.go
@@ -149,8 +149,8 @@ func newTestSuite(commType fsc.P2PCommunicationType, mask int, factor int, token
 			Monitoring:          false,
 			ReplicationOpts:     opts,
 			FSCBasedEndorsement: mask&WithEndorsers > 0,
-			FSCLogSpec:          "fts=debug:fsc=debug:info",
-			TokenSelector:       tokenSelector,
+			// FSCLogSpec:          "token-sdk=debug:fabric-sdk=debug:info",
+			TokenSelector: tokenSelector,
 		},
 	))
 	return ts, selector


token/driver/mock/tms.go
```diff
@@ -0,0 +1,952 @@
...
+	defer fake.tokensUpgradeServiceMutex.RUnlock()
+	fake.transferServiceMutex.RLock()
+	defer fake.transferServiceMutex.RUnlock()
+	fake.validatorMutex.RLock()
+	defer fake.validatorMutex.RUnlock()
+	fake.walletServiceMutex.RLock()
+	defer fake.walletServiceMutex.RUnlock()
+	copiedInvocations := map[string][][]interface{}{}
+	for key, value := range fake.invocations {
+		copiedInvocations[key] = value
```
AkramBitar, line 935: If we update this array only why we need to lock all the above?

token/services/ttx/dep/wrapper/tmsp.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+	req.SetTokenService(t.ManagementService)
+	return nil
+}
+
+type TokenManagementServiceProvider struct {
+	tmsProvider *token.ManagementServiceProvider
+}
+
+func NewTokenManagementServiceProvider(tmsProvider *token.ManagementServiceProvider) *TokenManagementServiceProvider {
```
AkramBitar, line 32: We assume that tmsProvider could not be nil. Right?

token/services/ttx/receivetx.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (f *ReceiveTransactionView) Call(context view.Context) (interface{}, error) {
+	// options
+	options, err := CompileOpts(f.opts...)
+	if err != nil {
+		return nil, errors.Join(err, ErrFailedCompilingOptions)
+	}
+	if options.Timeout == 0 {
+		options.Timeout = time.Minute * 4
```
AkramBitar, line 46: Default timeout is 4 min? (seems to me too large)

token/services/ttx/receivetx.go
```diff
@@ -0,0 +1,85 @@
...
+type ReceiveTransactionView struct {
+	opts []TxOption
+}
+
+func NewReceiveTransactionView(opts ...TxOption) *ReceiveTransactionView {
+	return &ReceiveTransactionView{
+		opts: opts,
+	}
+}
+
```
AkramBitar, line 31: It will help if we can have description of this view (purpose etc. like you did for other)

token/services/ttx/collectendorsements.go
```diff
@@ -0,0 +1,667 @@
...
+
+	sigmas := make(map[string][]byte)
+	for i, signerIdentity := range signers {
+		// we have the following possibilities:
+		// - there is a signer locally bound to the party, use it to generate the signature
+		// - there is a wallet bound to the party but the signer is not local, the signature is generated externally
+		// - the identity is a multi-sig identity
+		// - the signature must be generated by a remote party
+
+		signatureRequest := &SignatureRequest{
```
AkramBitar, line 194: signatureRequest := &SignatureRequest .. This allocated a new object of signatureRequest in side the for. Doe we need that? If not, we can define it outside the for and have one object that overwritten in each integration of the for.

token/services/ttx/dep/mock/nis.go
```diff
@@ -0,0 +1,121 @@
...
+	fake.signArgsForCall = append(fake.signArgsForCall, struct {
+		arg1 []byte
+	}{arg1Copy})
+	stub := fake.SignStub
+	fakeReturns := fake.signReturns
+	fake.recordInvocation("Sign", []interface{}{arg1Copy})
+	fake.signMutex.Unlock()
+	if stub != nil {
+		return stub(arg1)
+	}
```
AkramBitar, line 45: What if stub is nil?

token/services/ttx/dep/mock/nis.go
```diff
@@ -0,0 +1,121 @@
...
+	fake.signMutex.Lock()
+	ret, specificReturn := fake.signReturnsOnCall[len(fake.signArgsForCall)]
+	fake.signArgsForCall = append(fake.signArgsForCall, struct {
+		arg1 []byte
+	}{arg1Copy})
+	stub := fake.SignStub
+	fakeReturns := fake.signReturns
+	fake.recordInvocation("Sign", []interface{}{arg1Copy})
+	fake.signMutex.Unlock()
```
AkramBitar, line 42: Maybe add differ will be better.

AkramBitar (changes requested):

token/driver/mock/tms.go
adecaro: this is auto generated code used only for testing. We don't have to worry about it :)

token/services/ttx/dep/wrapper/tmsp.go
adecaro: right, because it is DIG filling it and DIG will fail if the dependency cannot be resolved. If you want a field optional (meaning nil), then we have to tell DIG explicitly

token/services/ttx/receivetx.go
adecaro: at this stage it is hardcoded. Would could we use?

token/services/ttx/receivetx.go
adecaro: will do

token/services/ttx/collectendorsements.go
adecaro: good catch. Let me investigate it.

token/services/ttx/dep/mock/nis.go
adecaro: this is used only for testing, we are good here.

token/services/ttx/dep/mock/nis.go
adecaro: everything under `mock` is autogenerated using counterfeiter. We use it for testing and we are good there.

token/services/ttx/collectendorsements.go
adecaro: okay, the signer changes. Now, to avoid confusion about when fields are initialized, I would propose to keep as it is so the code is more readable.

PR #1258, opened 2025-11-05
make generators versioned, add input interface for validator context

arner:
adds some flexibility for generic validators and reusing generators

### DIFF
--- a/integration/nwo/token/platform.go
+++ b/integration/nwo/token/platform.go
@@ -25,8 +25,8 @@ import (
 	"github.com/hyperledger-labs/fabric-token-sdk/integration/nwo/token/generators"
 	"github.com/hyperledger-labs/fabric-token-sdk/integration/nwo/token/generators/crypto/fabtokenv1"
 	"github.com/hyperledger-labs/fabric-token-sdk/integration/nwo/token/generators/crypto/zkatdlognoghv1"
-	fabtokenv2 "github.com/hyperledger-labs/fabric-token-sdk/integration/nwo/token/generators/pp/fabtokenv1"
-	common2 "github.com/hyperledger-labs/fabric-token-sdk/integration/nwo/token/generators/pp/zkatdlognoghv1"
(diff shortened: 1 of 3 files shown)


token/core/common/validator.go
```diff
@@ -22,7 +22,7 @@ const (
 	TokenRequestSignatures driver.ValidationAttributeID = "sigs"
 )
 
-type Context[P driver.PublicParameters, T any, TA driver.TransferAction, IA driver.IssueAction, DS driver.Deserializer] struct {
+type Context[P driver.PublicParameters, T driver.Input, TA driver.TransferAction, IA driver.IssueAction, DS driver.Deserializer] struct {
```
adecaro, line 25: okay, I need an argument for adding this restriction.

integration/nwo/token/platform.go
```diff
@@ -84,8 +84,8 @@ func NewPlatform(ctx api2.Context, t api2.Topology, builder api2.Builder) *Platf
 		TokenGenPath:           DefaultTokenGenPath,
 		NetworkHandlers:        map[string]NetworkHandler{},
 	}
-	p.PublicParamsGenerators[fabtokenv1.DriverIdentifier] = fabtokenv2.NewFabTokenPublicParamsGenerator()
-	p.PublicParamsGenerators[zkatdlognoghv1.DriverIdentifier] = common2.NewDLogPublicParamsGenerator(math3.BN254)
+	p.PublicParamsGenerators[fabtokenv1.DriverIdentifier] = fabtokenv2.NewFabTokenPublicParamsGenerator(0)
```
adecaro: what about introducing a constant to signal the use of predefined version instead of having the magic number 0?

token/core/common/validator.go
arner: It's so we can use the Owner field of input tokens in validators (via GetOwner) without having to cast it to an interface. I haven't done it in this PR, but we can move the HTLC validators to the common package and reduce duplicate code, for instance.

adecaro (approved):
LGTM

PR #1267, opened 2025-11-18
ttx service: create TxAuditor from the TMSID

adecaro:
Currently, a TxAuditor can be instantiated only from an auditor wallet. 
This PR allows the developer to obtain an instance of `TxAuditor` from the TMS ID alone.

### DIFF
--- a/token/sdk/dig/sdk.go
+++ b/token/sdk/dig/sdk.go
@@ -227,7 +227,7 @@ func (p *SDK) Install() error {
 		digutils.Register[dep.NetworkProvider](p.Container()),
 		digutils.Register[*wrapper2.StorageProvider](p.Container()),
 		digutils.Register[*wrapper2.NetworkIdentityProvider](p.Container()),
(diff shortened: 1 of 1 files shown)


token/services/ttx/auditor.go
```diff
@@ -26,12 +26,37 @@ import (
...
 
 type TxAuditor struct {
-	w                       *token.AuditorWallet
 	auditor                 *auditor.Service
 	auditDB                 *auditdb.StoreService
 	transactionInfoProvider *TransactionInfoProvider
 }
 
+// NewAuditorFromTMSID returns a new TxAuditor for the given TMS ID
+func NewAuditorFromTMSID(sp token.ServiceProvider, tmsID token.TMSID) (*TxAuditor, error) {
```
AkramBitar: Looks good do we need to add a test?

adecaro: good point, added.

PR #1272, opened 2025-11-24
dlog token driver: better buckets for metrics

adecaro:
This PR brings the following: It defines buckets for the zk issue and zk transfer histograms so that we get more precise data for grafana.


token/core/zkatdlog/nogh/v1/metrics.go
```diff
@@ -15,11 +18,13 @@ var (
 		Name:       "issue_duration",
 		Help:       "Duration of zk issue token",
 		LabelNames: []string{"network", "channel", "namespace"},
+		Buckets:    utils.ExponentialBucketTimeRange(0, 1*time.Second, 10),
```
AkramBitar, line 21: I see this pattern (0, 1*time.Second, 10) used a lot in the code, especially in the smart client. Now I’m wondering:

1. Can we move this into a config file so we can tune it at runtime without needing a new deployment each time? Another option would be to have a default defined in one place, and only override it for metrics that need more buckets.

2. What happens if buckets are not set? I think the default is a large number of buckets. I think this value should either be set by the user or the default should be small.”

AkramBitar (approved):
Looks ok for me. Put some comments.

token/core/zkatdlog/nogh/v1/metrics.go
adecaro: good points: 
1. having them in the configuration requires a certain amount of work, we can plan for it, if it becomes compelling
2. If not set, the default one is used. I have seen that jaeger supports self-adjusting buckets so that we don't need to worry about this anymore.

adecaro: I created this issue to track this: https://github.com/hyperledger-labs/fabric-token-sdk/issues/1274

PR #1279, opened 2025-11-26
dlog transfer service: benchmark tests #1278

adecaro:
This PR improves testing and benchmark of the `token/core/zkatdlog/nogh/v1/transfer` package.
Documentation has been produced that explain how to run the benchmark.
I have reported results produced on my local machine, an Apple M1.


docs/benchmark/tools.md
```diff
@@ -0,0 +1,106 @@
...
+
+Run the current code 10 times to gather statistically significant data.
+```bash
+# Save current performance to old.txt
+go test -bench=. -benchmem -count=10 -run=^$ . > old.txt
+```
+
+### Step 3: Capture Experiment (New)
+
+Apply your code changes (e.g., the allocation optimizations you researched) and run the same benchmark command.
```
AkramBitar, line 62: This section is not clear to me. 
Could you please elaborate what the objectives of it ? 
What you mean by "Apply your code changes"?

docs/benchmark/tools.md
```diff
@@ -0,0 +1,106 @@
...
+
+### Resource & Profiling Flags
+
+These flags are critical for performance tuning and memory analysis, which aligns with your recent work on allocation optimization.
+
+| Flag | Description | Example |
+| :--- | :--- | :--- |
+| `-benchmem` | Print memory allocation statistics (allocations per op, bytes per op). | `-benchmem` |
+| `-cpu=<n,m...>` | Run benchmarks with specific `GOMAXPROCS` values. | `-cpu=1,2,4,8` |
+| `-cpuprofile=<file>` | Write a CPU profile to the specified file. | `-cpuprofile=cpu.out` |
```
AkramBitar, line 28: I am wondering if we can connect to a port and see the Pprof while the benchmark is running?

docs/benchmark/tools.md
```diff
@@ -0,0 +1,106 @@
...
+| `-failfast` | Stop execution immediately after the first failure. | `-failfast` |
+
+### Resource & Profiling Flags
+
+These flags are critical for performance tuning and memory analysis, which aligns with your recent work on allocation optimization.
+
+| Flag | Description | Example |
+| :--- | :--- | :--- |
+| `-benchmem` | Print memory allocation statistics (allocations per op, bytes per op). | `-benchmem` |
+| `-cpu=<n,m...>` | Run benchmarks with specific `GOMAXPROCS` values. | `-cpu=1,2,4,8` |
```
AkramBitar, line 27: What is the default?  Is it the number of core of the VM where we will run?

When runtime.GOMAXPROCS(0) then it will take the number of logical CPU cores on your machine.

docs/benchmark/tools.md
```diff
@@ -0,0 +1,106 @@
...
+To run benchmarks, use the `go test` command with specific flags. The most essential flag is `-bench`, which accepts a regular expression to select which benchmark functions to execute.
+
+### Execution Control Flags
+
+| Flag | Description | Example |
+| :--- | :--- | :--- |
+| `-bench=<regexp>` | Run benchmarks matching the regex. Use `.` to run all. | `-bench=.` |
+| `-run=<regexp>` | Run unit tests matching the regex. Use `^$` to skip all unit tests and only run benchmarks. | `-run=^$` |
+| `-benchtime=<t>` | Duration to run each benchmark (default `1s`). Can also specify iterations (suffix `x`). | `-benchtime=5s` or `-benchtime=1000x` |
+| `-count=<n>` | Run each benchmark `n` times. Essential for statistical analysis with `benchstat`. | `-count=10` |
```
AkramBitar, line 16: You wrote "Run each benchmark `n` times"
It is not clear for me does it runs the same benchmark "n" times?
or do we have different benchmarks?


If I would like to run it over night (e.g., to check mem leaks) or ... other ... can I say how much time I want it to run?

docs/benchmark/tools.md
```diff
@@ -0,0 +1,106 @@
...
+| `-memprofile=<file>` | Write a memory profile to the specified file. | `-memprofile=mem.out` |
+| `-blockprofile=<file>`| Write a goroutine blocking profile (contention analysis). | `-blockprofile=block.out` |
+| `-mutexprofile=<file>`| Write a mutex contention profile. | `-mutexprofile=mutex.out` |
+| `-trace=<file>` | Write an execution trace for the `go tool trace` viewer. | `-trace=trace.out` |
+
+> **Note:** When using profiling flags like `-cpuprofile`, the resulting binary is preserved in the current directory for analysis with `go tool pprof`.
+
+## 2. The Analysis Workflow (A/B Testing)
+
+Reliable performance optimization requires measuring the "before" and "after" states. The standard workflow involves saving benchmark output to files and comparing them.
```
AkramBitar, line 38: Does it print total time of run?

docs/benchmark/dlognogh.md
```diff
@@ -0,0 +1,589 @@
...
+
+```sh
+go test ./token/core/zkatdlog/nogh/v1/transfer -bench=BenchmarkSender -benchmem -count=10 -cpu=1 | tee bench.txt
+```
+
+You can then aggregate/parse the output (e.g., with simple shell tools or a small script) to compute averages across `-count` repetitions.
+
+### Results
+
+```shell
```
KElkhiyaoui, line 99: The performance of the zero-knowledge proof verification is dominated by range proof verification. Therefore, we see that performance does not vary when we change the number of inputs but it does vary greatly when we increase the number of outputs. 
Indeed, the number of range proof verification is equal to the number of the outputs. The exception is (single input, single output). 
This corresponds to a pure transfer of ownership, during which there is no range proof verification. The assumption is that the input is already in the correct range and we just check equality of input and output. 
Note that this assumption is valid since tokens are created as a result of an issue or a transfer (during which the output is verified to be in the correct range).
The performance does not vary drastically when we change the curve. This could be due to the fact that while the field over `BLS_381` is larger than that of `BN_254` , the subgroups on which the cryptographic operations are performed are of comparable sizes.

docs/benchmark/tools.md
```diff
@@ -0,0 +1,106 @@
...
+### Step 1: Install Benchstat
+
+`benchstat` is the standard tool for computing statistical summaries and comparing Go benchmarks.
+```bash
+go install golang.org/x/perf/cmd/benchstat@latest
+```
+or
+
+```shell
+makefile install-tools
```
AkramBitar: Change "makefile" to "make".
make install-tools

docs/benchmark/tools.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+### Step 2: Capture Baseline (Old)
+
+Run the current code 10 times to gather statistically significant data.
+```bash
+# Save current performance to old.txt
+go test -bench=. -benchmem -count=10 -run=^$ . > old.txt
```
AkramBitar, line 57: tried to follow the instructions and run from root folder of token sdk and failed. Could you add to the instruction to go to the folder "./hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/nogh/v1/transfer " and run from that folder.

docs/benchmark/tools.md
```diff
@@ -0,0 +1,106 @@
...
+go test -bench=. -benchmem -count=10 -run=^$ . > new.txt
+```
+
+## 3. Analyzing Results with Benchstat
+
+Run `benchstat` against your two captured files to see the delta.
+
+```bash
+benchstat old.txt new.txt
+```
```
AkramBitar, line 74: I ran the following:

transfer-service-benchmark-tests) $ go test -bench=. -benchmem -count=10 -run=^$ . > old.txt 
transfer-service-benchmark-tests) $ go test -bench=. -benchmem -count=10 -run=^$ . > new.txt 
tatransfer-service-benchmark-tests) $ benchstat old.txt new.txt 


I did not receive any output?

token/core/zkatdlog/nogh/v1/transfer/sender_test.go
```diff
@@ -6,7 +6,9 @@ SPDX-License-Identifier: Apache-2.0
 package transfer_test
```
AkramBitar, line 7: Do we have only ZKP benchmark for the transfer? 
Is there one for Issuance, Funds, Redeem operations, etc.?
“Are different ZKP methods or protocols used for issuance, transfer, and funds operations?”

docs/benchmark/tools.md
AkramBitar: > You wrote "Run each benchmark `n` times" It is not clear for me does it runs the same benchmark "n" times? or do we have different benchmarks?
> 
> If I would like to run it over night (e.g., to check mem leaks) or ... other ... can I say how much time I want it to run?

Based on go benchmark.
-count=<n> Runs each benchmark n times. Helpful for getting stable numbers, especially if you want to compare results with tools like benchstat.

-benchtime=<duration>
Sets how long each benchmark runs. Can be a duration (e.g. 1s) or number of iterations (e.g. 100x).

docs/benchmark/tools.md
AkramBitar: > What is the default? Is it the number of core of the VM where we will run?
> 
> When runtime.GOMAXPROCS(0) then it will take the number of logical CPU cores on your machine.


GOMAXPROCS typically defaults to the number of logical CPUs available on your machine — so by default Go will try to use all CPUs/cores

docs/benchmark/tools.md
AkramBitar: > I am wondering if we can connect to a port and see the Pprof while the benchmark is running?

Benchmarks (go test -bench) cannot expose a live pprof web UI

docs/benchmark/tools.md
AkramBitar: > Does it print total time of run?

Yes, in the output of the run:
"BenchmarkFoo-8   	  2000000	       600 ns/op
PASS
ok  	example	3.210s"

In this example 3.210s

docs/benchmark/tools.md
AkramBitar: > This section is not clear to me. Could you please elaborate what the objectives of it ? What you mean by "Apply your code changes"?

Ok, this the case where we have:
Old code: go test -bench=. -benchmem -count=10 -run=^$ . > old.txt
Fixed mem optimization
go test -bench=. -benchmem -count=10 -run=^$ . > new.txt

Then analyze using benchstat 
benchstat old.txt new.txt

docs/benchmark/tools.md
AkramBitar: Worked
go test ./token/core/zkatdlog/nogh/v1/transfer -bench=. -benchmem -count=10 -run=^$ . 
goos: linux
goarch: amd64
pkg: github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/nogh/v1/transfer
cpu: Intel(R) Xeon(R) CPU E5-2699 v4 @ 2.20GHz
BenchmarkActionMarshalling/With_Protos-88                  76818             16095 ns/op            3159 B/op         40 allocs/op
BenchmarkActionMarshalling/With_Protos-88                  82558             15154 ns/op            3159 B/op         40 allocs/op
BenchmarkActionMarshalling/With_Protos-88

docs/benchmark/tools.md
AkramBitar: or go test ./token/core/zkatdlog/nogh/v1/transfer -bench=. -benchmem -count=10 -run=^$ .

PR #1283, opened 2025-12-02
remove id caches

HayimShaul:
remove id caches

### DIFF
--- a/token/services/identity/provider.go
+++ b/token/services/identity/provider.go
@@ -306,60 +306,6 @@ func (p *Provider) getSignerAndCache(ctx context.Context, identity driver.Identi
 	return signer, shouldCache, nil
 }
 
-// func (p *Provider) getSigner(ctx context.Context, identity driver.Identity, idHash string) (driver.Signer, error) {
-// 	// check again the cache
-// 	entry, ok := p.signers.Get(idHash)
-// 	if ok {
-// 		p.Logger.DebugfContext(ctx, "signer for [%s] found", idHash)
-// 		return entry.Signer, nil
-// 	}
-
-// 	p.Logger.DebugfContext(ctx, "signer for [%s] not found, try to deserialize", idHash)
-// 	// ask the deserializer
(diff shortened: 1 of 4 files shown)


token/services/identity/provider.go
```diff
@@ -117,13 +116,6 @@ func (p *Provider) RegisterVerifier(ctx context.Context, identity driver.Identit
 	if v == nil {
 		return errors.New("invalid verifier, expected a valid instance")
 	}
```
adecaro: it looks like this function is not really needed anymore. I would start an investigation on the callers of this function. Maybe we can save further cycles.

token/services/identity/provider.go
```diff
@@ -274,6 +266,7 @@ func (p *Provider) getSigner(ctx context.Context, identity driver.Identity, idHa
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed deserializing identity for signer [%s]", identity)
 	}
+
```
adecaro, line 291: I think we said to cache only the identity whose type is x509, no?
If we still cache the signers and the cache is only of size 5 there will be even more contention, no?

token/services/identity/provider.go
```diff
@@ -331,15 +325,4 @@ func (p *Provider) updateCaches(descriptor *idriver.IdentityDescriptor, alias dr
 			p.signers.Add(aliasID, entry)
 		}
 	}
```
adecaro, line 327: we are still caching any signer no matter the identity type. This will generate even more contention on the signers cache given its new size (5)

token/services/identity/provider.go
HayimShaul: actually we said to have the first PR use NoCache and a 2nd PR avoid caching x509.
The system doesn't work with signers being NoCache, so I limited the cache to 5 instead.

but sure. I added to this PR avoiding to insert x509 into the cache and increased the signer cache to 50.

Issue #1287, opened 2025-12-08
dlog token driver: improve the use of mathlib


adecaro:
This Issue asks to introduce to following improvements:
- Use of Mul2 and MultiExp when possible
- Stay in Jacobian coordinates as much as possible
- Introduce pools for temporary objects

PR #1288, opened 2025-12-08
dlog token driver: improve the use of mathlib #1287

adecaro:


token/core/common/crypto/math/curves.go
```diff
@@ -22,20 +22,23 @@ var (
...
+			math.NewG2(bls12381.NewBls12_381BBS().GenG2(), BLS12_381_BBS_GURVY_FAST_RNG),
+			math.NewGt(bls12381.NewBls12_381BBS().GenGt(), BLS12_381_BBS_GURVY_FAST_RNG),
+			math.NewZr(bls12381.NewBls12_381().GroupOrder(), BLS12_381_BBS_GURVY_FAST_RNG),
+			bls12381.NewBls12_381BBS().CoordinateByteSize(),
+			bls12381.NewBls12_381BBS().G1ByteSize(),
+			bls12381.NewBls12_381BBS().CompressedG1ByteSize(),
+			bls12381.NewBls12_381BBS().G2ByteSize(),
+			bls12381.NewBls12_381BBS().CompressedG2ByteSize(),
+			bls12381.NewBls12_381BBS().ScalarByteSize(),
+			BLS12_381_BBS_GURVY_FAST_RNG,
```
AkramBitar, line 39: Is that something that we need to tune during testing? or does this fix? (does it worth to have it in config file)?

token/core/common/crypto/slice.go
```diff
@@ -0,0 +1,41 @@
...
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package crypto
+
+import "encoding/binary"
+
+// AppendFixed32 appends slices prefixed with a 4-byte Little Endian length.
```
AkramBitar, line 11: Little Endian this means that we support only x86 not power/z deployments?

token/core/zkatdlog/nogh/v1/crypto/rp/ipa_test.go
```diff
@@ -35,14 +53,81 @@ func TestIPAProofVerify(t *testing.T) {
...
+		setup.com,
+		setup.nr,
+		setup.curve,
+	)
 	err = verifier.Verify(proof)
-	assert.NoError(t, err)
+	require.NoError(t, err)
+}
+
+func BenchmarkIPAProver(b *testing.B) {
```
AkramBitar, line 102: Do we have doc. of what this function does?

token/core/common/crypto/slice_test.go
```diff
@@ -0,0 +1,120 @@
...
+	b.ResetTimer()
+
+	for i := 0; i < b.N; i++ {
+		// Use nil to strictly measure allocation of the result
+		_ = AppendFixed32(nil, data)
+	}
+}
+
+// Comparison: Naive Loop (No pre-calculation)
+func BenchmarkAppendFixed32_Naive(b *testing.B) {
```
AkramBitar, line 107: Could you please add doc to this one?
It is not clear to me what this benchmarks tests and how to connect it to the Approver performance?

token/core/zkatdlog/nogh/v1/crypto/common/array.go
```diff
@@ -28,11 +28,21 @@ func (a *G1Array) Bytes() ([]byte, error) {
...
+
+func (a *G1Array) BytesTo(b []byte) ([]byte, error) {
+	raw := make([][]byte, len([]*math.G1(*a)))
+	for i, e := range []*math.G1(*a) {
+		if e == nil {
+			return nil, errors.Errorf("failed to marshal array of G1")
+		}
+		raw[i] = e.Bytes()
+	}
+	clear(b)
```
AkramBitar: What does the clear do? Is that ok to clear the "b" here? what happen if the caller needs it? Maybe to add flag to clear and this will be clear to the caller.

AkramBitar: What if "b" is nil?

token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof_test.go
```diff
@@ -36,12 +55,85 @@ func TestBFProofVerify(t *testing.T) {
...
+		setup.Q,
+		setup.nr,
+		setup.l,
+		setup.curve,
+	)
 	err = verifier.Verify(proof)
 	assert.NoError(t, err)
 }
+
+func BenchmarkBFProver(b *testing.B) {
```
AkramBitar, line 113: Same as other comment need to document this (need to understand what this does)

token/core/common/crypto/math/curves.go
adecaro: This curve is only for quick experimentation. It is never used in the actual code or testing.

token/core/common/crypto/slice.go
adecaro: Fortunately, it does not matter the hardware architecture here. It is just that we need to choose a convention when converting from int to bytes and vice-versa. The choise is arbitrary here.

token/core/common/crypto/slice_test.go
adecaro: This benchmark is to have data about the naive way of appending data in a slice in contract to the custom function that is optimized for minimum memory allocation. `BenchshomarkAppendFixed32` should show better number than `BenchmarkAppendFixed32_Naive`.

token/core/zkatdlog/nogh/v1/crypto/common/array.go
adecaro: clear is part of the Go SDK. From the comments of the function:
// The clear built-in function clears maps and slices.
// For maps, clear deletes all entries, resulting in an empty map.
// For slices, clear sets all elements up to the length of the slice
// to the zero value of the respective element type. If the argument
// type is a type parameter, the type parameter's type set must
// contain only map or slice types, and clear performs the operation
// implied by the type argument. If t is nil, clear is a no-op.

I'll add a comment to BytesTo to clarify its behaviour.

adecaro: nothing happens.

PR #1299, opened 2026-01-18
Fix some benchmarks

aaadir:
A few benchmarks failed, mainly because the index to the array of environments went out of range during the main benchmark loop.

### DIFF
--- a/cmd/benchmarking/run_benchmarks.py
+++ b/cmd/benchmarking/run_benchmarks.py
@@ -242,7 +242,6 @@ non_parallel_tests = [
     ('BenchmarkTransferProofGeneration', "", ""), 
     ('BenchmarkIssuer', "", issuer_benchmarks_folder), 
     ('BenchmarkProofVerificationIssuer', "", issuer_benchmarks_folder), 
-    ('BenchmarkVerificationSenderProof', "", ""), 
     ('BenchmarkTransferServiceTransfer', "", v1_benchmarks_folder), 
 ]
 parallel_tests = [
--- a/docs/benchmark/core/dlognogh/dlognogh.md
+++ b/docs/benchmark/core/dlognogh/dlognogh.md
@@ -216,3 +216,35 @@ Timeline: [▇▇▇█▇▇▇▇▆▇] (Max: 131 ops/s)
 PASS
 ok      github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/nogh/v1/transfer       14.566s
 ```
+
+### Running selected benchmarks with run_benchmarks.py
+The run_benchmarks.py script runs selected benchmarks and summarizes their results in a csv file, so it could be followed as more optimizations are added. 
+To run this script goto the ../fabric-token-sdk/cmd/benchmarking folder and run
+```shell
+python run_benchmarks.py
+```
+
+This creates a subfolder that collects the logs of all the benchmarks and the csv file (**benchmark_results.csv**) that has a separate row for every invocation of the script with the selected metrics collected for all the benchmarks. 
+The folder is named **benchmark_logs_<date>** for example **benchmark_logs_2026-01-19_06-56-41**, where the date indicates the time when the script was run.
+
+The script supports the following flags:
+
+`--count`
+: the number of times to run the benchmark
+
+`--timeout`
+: the maximum time (in seconds) to for the benchmark. Thje default is 0, implying no limit.
+
+`--benchName`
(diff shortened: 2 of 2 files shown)


cmd/benchmarking/run_benchmarks.py
```diff
@@ -0,0 +1,244 @@
+import csv
+import os
+import sys
+import shutil
+import subprocess
+import re
+from datetime import datetime
+
+I=1
+TOKENSDK_ROOT = "../../"
```
AkramBitar: 1) Can we add description of that script to readme and explain how to use it?
2) Regarding TOKENSDK_ROOT = "../../" this assumes that the script should be run from a specific folder. Maybe we can ask to define this as env variable.

cmd/benchmarking/run_benchmarks.py
```diff
@@ -0,0 +1,244 @@
...
+validator_benchmarks_folder = os.path.join(TOKENSDK_ROOT, "token/core/zkatdlog/nogh/v1/validator")
+
+def run_and_parse_non_parallel_metrics(benchName, params, folder=transfer_benchmarks_folder) -> dict:
+    global I
+    global output_folder_path
+
+    if folder == "":
+        folder = transfer_benchmarks_folder
+
+    cmd = f"go test {folder} -run='^$' -bench={benchName} -v -benchmem -count=10 -cpu=1 -timeout 0 {params} | tee bench.txt; benchstat bench.txt"
```
AkramBitar: Maybe we should define default values for count and timeout, allowing the user to override them if needed.

cmd/benchmarking/run_benchmarks.py
```diff
@@ -0,0 +1,244 @@
...
+    }
+
+def run_and_parse_parallel_metrics(benchName, params, folder=transfer_benchmarks_folder) -> dict:
+    if folder == "":
+        folder = transfer_benchmarks_folder
+
+    global I
+    global output_folder_path
+
+    cmd = f"go test {folder} -test.run={benchName} -test.v -test.timeout 0 -bits='32' -num_inputs='2' -num_outputs='2' -workers='NumCPU' -duration='10s' -setup_samples=128 {params}"
```
AkramBitar: Same as above comment maybe to have defaults for the params and user can change it if needed.

cmd/benchmarking/run_benchmarks.py
```diff
@@ -0,0 +1,244 @@
...
+    results.update(run_and_parse_non_parallel_metrics(test, params, benchType)) 
+
+print("\n*******************************************************")
+print("Running parallel tests")
+for test, params, folder in parallel_tests:
+    results.update(run_and_parse_parallel_metrics(test, params, folder))
+
+# add new row to benchmark_results.csv and copy it to the output folder
+append_dict_as_row("benchmark_results.csv", results)
+src = os.path.join(".", "benchmark_results.csv")
```
AkramBitar: Can we describe this output in graphs like we did with workload_generator (generate PDF with graphs)?

AkramBitar (changes requested):

cmd/benchmarking/run_benchmarks.py
aaadir: 1. what readme?
2. done.

cmd/benchmarking/run_benchmarks.py
aaadir: for this comment and for the similar comment below, for the sake of avoiding too many features and params I added params for count and timeout, but didn't have separate count and timeout params for the non-parallel and parallel runs. 
Instead of this, I added another param --benchName for running just the specified bench, parallel or not, and the count, timeout, ... applies to it.

I also didn't add a param for the "params" because we never ended up using it. After we use this script for some time, if I see that we use it I'll add a param for it, and if we don't I'll drop the feature itself.

done

cmd/benchmarking/run_benchmarks.py
aaadir: see reply to this above. 
done.

cmd/benchmarking/run_benchmarks.py
aaadir: maybe we'll wait until we accumulate a few separate lines in the csv file and then see what figues make sense. 
For now, we can start focusing on the optimizations.

cmd/benchmarking/run_benchmarks.py
aaadir: 1. done (fixed dlognogh.md)

PR #1307, opened 2026-01-31
performance improvements

adecaro:
This PR introduces the following improvements:
- use of MultiScalarMul for the computation of the commitment
- pre-computation of powers of 2, sums of powers of 2, and Zr elements from 0 to 64
- better use of mathlib where possible.

This PR depends on https://github.com/IBM/mathlib/pull/42


token/core/zkatdlog/nogh/v1/crypto/math/curves.go
```diff
@@ -0,0 +1,160 @@
...
+// cached value when present.
+func One(c *math.Curve) *math.Zr {
+	return NewCachedZrFromInt(c, 1)
+}
+
+// Two returns the curve element representing 2 in Zr. It prefers the
+// cached value when present.
+func Two(c *math.Curve) *math.Zr {
+	return NewCachedZrFromInt(c, 2)
+}
```
HayimShaul, line 62: Doesn't a small number X simply correspond to (0,0,...,X) in math.Zr? How much does caching these values save?

adecaro: zero, one, and two were instantiated multiple times in the code. Having them pre-generated once and for all saves a lot of allocations in the long run.

token/core/zkatdlog/nogh/v1/crypto/math/curves.go
```diff
@@ -0,0 +1,160 @@
...
+	if !ok {
+		logger.Warnf("no hit for [%d:%d]", c.ID(), i)
+		two := c.NewZrFromUint64(2)
+		return two.PowMod(c.NewZrFromUint64(i))
+	}
+	v, ok := cc[i]
+	if !ok {
+		logger.Warnf("no hit for [%d:%d]", c.ID(), i)
+		two := c.NewZrFromUint64(2)
+		return two.PowMod(c.NewZrFromUint64(i))
```
HayimShaul, line 114: why should this happen after you have populated the cache?

HayimShaul: If this happens a lot, maybe find a high power of 2 that is in the cache and multiply it by the factor that it is missing?

HayimShaul: Also, there's Exp(Element, Big.Int) I think this will be faster than converting to Zr

AkramBitar (approved):
Looks ok for me

Issue #1310, opened 2026-02-03
cachedFetcher loads all tokens into memory causing excessive memory allocation


AkramBitar:
The cachedFetcher (used by Mixed strategy) loads ALL tokens from the database into memory, regardless of which user's tokens are actually needed. This causes significant memory allocation visible in pprof.

**pprof Output**
```
56.50MB   100% |   github.com/hyperledger-labs/fabric-token-sdk/token/services/selector/sherdlock.(*cachedFetcher).update
30MB 13.90% 13.90%    56.50MB 26.19%                | github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/sql/common.(*rowIterator[go.shape.struct { Id github.com/hyperledger-labs/fabric-token-sdk/token/token.ID; WalletID string; Type github.com/hyperledger-labs/fabric-token-sdk/token/token.Type; Quantity string }]).Next
26.50MB 46.90% |   database/sql.(*Rows).Next
```

<img width="2502" height="1241" alt="Image" src="https://github.com/user-attachments/assets/df7dae56-f6ee-4812-b10c-a03a8a4afa19" />

**Root Cause**

token/services/selector/sherdlock/fetcher.go:184

it, err := f.tokenDB.SpendableTokensIteratorBy(ctx, "", "")

Empty string parameters result in SQL query without wallet/type filters:

SELECT * FROM tokens WHERE spendable = true
-- Returns ALL tokens from ALL users

**Impact**

100K tokens: 56.50MB allocated per refresh
1M tokens: 565MB allocated per refresh
Refreshes every 1 second OR every 5 queries
Memory grows unbounded with database size
High GC pressure and system slowdown

**Configuration**

Set in token/sdk/dig/sdk.go:147:
sherdlock.NewFetcherProvider(..., sherdlock.Mixed)

Mixed Strategy: Combines cachedFetcher (eager) + lazyFetcher (fallback)

- Tries cachedFetcher first (loads ALL tokens)
- Falls back to lazyFetcher if cache empty

The current strategy of fetching spendable tokens using cachedFetcher cache  is fast however it is not scalable since the cache has no size limits or eviction policy, and loads all tokens even though only one user's tokens are needed per transfer request.

One solution is to replace the current cache unlimited implementation (map) to LRU cache.

AkramBitar:
# Solution Proposal - Adaptive Cache Strategy Summary

## Overview
[...]

PR #1312, opened 2026-02-05
Optimized Cache Fetcher

AkramBitar:
The cachedFetcher (used by Mixed strategy) loads ALL tokens from the database into memory, regardless of which user's tokens are actually needed. 
See #1310


AkramBitar:
> @AkramBitar , `token/services/selector/benchmark_test.go` contains benchmarks for the available selectors. Please, can you update it and verify the performance of the new strategy are good enough? Please, update the README.md with the result on a target machine. Thanks.

Thanks a lot @adecaro. Good idea, will do that.

AkramBitar:
I proposed the following solution in the meantime:

we replaced the unbounded map with a Ristretto cache.

1) Cache size is limited (configurable) to prevent memory issues
2) Old entries are automatically removed (LRU)
3) Refresh interval and max queries are configurable
4) Existing behavior stays the same

This prevents memory growth, keeps performance stable, and allows us to tune cache size and refresh settings based on workload, memory limits, DB load, and the cost of cache misses.

PR #1313, opened 2026-02-06
fabricx: notification service integration

adecaro:
This PR modifies the network service's fabricx driver to support the fabricx notification service.
This PR uses https://github.com/hyperledger-labs/fabric-smart-client/pull/1125.

There will be more changes to the fabricx network driver after which I'll update the documentation.


docs/core-token.md
```diff
@@ -42,20 +42,11 @@ token:
...
-    #   For this reason, there is a periodic polling period (1s) that queries all txIDs for which there is a pending listener and invokes the listeners for the ones found.
-    #   Once we get notified about finality, we try (repeatedly) to fetch the additional tx information needed from the vault (e.g., transactionRequest).
-    #   If we work without replicas, there is no need to try more than once to fetch the additional tx information, because the finality notification and vault update happen in sync.
-    #   It is only needed in case we have replicas, where another replica may have updated the vault shortly before.
+    # notification: The manager is notified about finality events via a notification service (e.g. for FabricX).
+    #   When a new notification arrives, an event is added to a queue for asynchronous processing.
+    #   When a client subscribes to the manager for a specific transaction, we perform an immediate query to check its status.
+    # The field can also be left empty. In that case, the default option will be used depending on the network type each TMS refers to.
```
AkramBitar, line 48: What if we will not have the field at all, then I assume default option will be used depending on the network type each TMS refers to as well.

token/services/network/fabricx/finality/queue/queue.go
```diff
@@ -0,0 +1,224 @@
...
+	}
+
+	// Start worker pool
+	eq.start()
+	return eq, nil
+}
+
+// start initializes all worker goroutines
+func (eq *EventQueue) start() {
+	for i := 0; i < eq.workers; i++ {
```
AkramBitar: Do need to alert if eq.workers = 0?

token/services/network/fabricx/finality/queue/queue.go
```diff
@@ -0,0 +1,224 @@
...
+				// Channel closed, worker exits
+				logger.Debugf("Worker %d shutting down", id)
+				return
+			}
+
+			// Process the event with context
+			if err := event.Process(eq.ctx); err != nil {
+				logger.Debugf("Worker %d: error processing event: %v", id, err)
+			}
+
```
AkramBitar, line 124: I am wondering if to do return here? 
See the behavior  on case <-eq.ctx.Done()>

token/services/network/fabricx/finality/queue/queue.go
```diff
@@ -0,0 +1,224 @@
...
+		case <-eq.ctx.Done():
+			// Context canceled, exit gracefully
+			logger.Debugf("Worker %d received shutdown signal", id)
+			return
```
AkramBitar, line 129: Do we need default here?

token/services/network/fabricx/finality/queue/queue.go
```diff
@@ -0,0 +1,224 @@
...
+
+	select {
+	case eq.events <- event:
+		logger.Debugf("EnqueueBlocking event: [%v]", event)
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-eq.ctx.Done():
+		return ErrQueueClosed
+	}
```
AkramBitar, line 169: Do we need default?

token/services/network/fabricx/finality/queue/queue.go
```diff
@@ -0,0 +1,224 @@
...
+
+	select {
+	case eq.events <- event:
+		return nil
+	default:
+		return ErrQueueFull
+	}
+}
+
+// EnqueueBlocking adds an event to the queue (blocks until space available or timeout)
```
AkramBitar, line 151: I did not understand why this one is blocking and the Enqueue is blocking?
Please consider adding more documentation here.

token/services/network/fabricx/config/config_test.go
```diff
@@ -0,0 +1,31 @@
...
+
+	// Case 1: Configuration returns a valid manager type
+	fakeConfig.GetStringReturns("custom-type")
+	c := config.NewListenerManagerConfig(fakeConfig)
+	assert.Equal(t, config.ManagerType("custom-type"), c.Type())
+	assert.Equal(t, 1, fakeConfig.GetStringCallCount())
+	assert.Equal(t, config.Type, fakeConfig.GetStringArgsForCall(0))
+
+	// Case 2: Configuration returns an empty string, should default to Notification
+	fakeConfig.GetStringReturns("")
```
AkramBitar, line 28: Does this returned in case the value is "" or also in case the fields is not defined at all?

token/services/network/fabricx/finality/queue/config.go
```diff
@@ -0,0 +1,58 @@
...
+
+package queue
+
+const (
+	// Workers is the configuration key for the number of worker goroutines
+	Workers = "token.finality.notification.workers"
+	// QueueSize is the configuration key for the size of the event buffer
+	QueueSize = "token.finality.notification.queueSize"
+	// DefaultWorkers is the default number of worker goroutines
+	DefaultWorkers = 10
```
AkramBitar, line 15: Maybe in the future we need to add a benchmark and check what is the best value.
I am wondering if Bob can give a quick one?

token/services/network/fabricx/finality/queue/config_test.go
```diff
@@ -0,0 +1,77 @@
+/*
```
AkramBitar, line 1: Sine this is a new code I recommend to have 80%+ coverage 
However, we can go with PR and then open another one for converge.

token/services/network/fabricx/finality/queue/queue.go
```diff
@@ -0,0 +1,230 @@
...
+		}
+	})
+
+	return shutdownErr
+}
+
+// Stats returns queue statistics
+func (eq *EventQueue) Stats() Stats {
+	return Stats{
+		Workers:   eq.workers,
```
AkramBitar, line 218: Is there a since to add this to Prometheus and Grafana tutorial?

docs/core-token.md
adecaro: indeed

token/services/network/fabricx/config/config_test.go
adecaro: in both cases

token/services/network/fabricx/finality/queue/config.go
adecaro: I think it depends on the machine and other parameters. The fact that is configurable, should not have us worry about this value

token/services/network/fabricx/finality/queue/config_test.go
adecaro: for the `queue` package, we are well above 80%

token/services/network/fabricx/finality/queue/queue.go
adecaro: no default here because we have only two channels we are listen on

token/services/network/fabricx/finality/queue/queue.go
adecaro: when the context is done, that is the signal to stop the operations. In this case, instead we just continue waiting for the next event.

token/services/network/fabricx/finality/queue/queue.go
adecaro: eventually, we can extend this to produce metrics, indeed. 
Do you want it in this PR?

token/services/network/fabricx/finality/queue/queue.go
adecaro: nothing else, I would say.

token/services/network/fabricx/finality/queue/queue.go
adecaro: when calling Enqueue, if the channel is full, it returns immediately (see the default condition). EnqueueBlocking instead will block until something is available.

token/services/network/fabricx/finality/queue/queue.go
AkramBitar: I will leave it to you.

Issue #1164, opened 2025-07-27
Build on REHL.


drucker-nir:
When executing `make` on REHL the script ` ./ci/scripts/install_softhsm.sh ` fails.

The reason is that it uses `apt-get` which is only relevant for debian.

Suggested fix:
```
#!/usr/bin/env bash
set -euo pipefail
# Install SoftHSM2 based on OS
if [[ "$OSTYPE" == "darwin"* ]]; then
    echo "Detected macOS"
    brew install softhsm
elif [[ -f /etc/redhat-release ]]; then
    echo "Detected RHEL-based system"
    sudo dnf install -y softhsm
elif [[ -f /etc/debian_version ]]; then
    echo "Detected Debian-based system"
    sudo apt-get update
    sudo apt-get install -y softhsm2
else
    echo "Unsupported OS: $OSTYPE"
    exit 1
fi
```

aaadir:
When I run `make checks` on he4cloud I have problems related to the above. Even when I fix it temporarily, other problems turn up.
Need to make sure `make checks` passes on he4cloud.

adecaro:
@drucker-nir , thanks for opening this. Please, fill the metadata fields on the right-hand of this page. Please, assign this Issue to an epic as well. Thanks

PR #1315, opened 2026-02-07
Build on REHL #1164

adecaro:
This PR addresses #1164 as requested.

### DIFF
--- a/ci/scripts/install_softhsm.sh
+++ b/ci/scripts/install_softhsm.sh
@@ -5,8 +5,18 @@
 set -euo pipefail
 
 #Install softhsm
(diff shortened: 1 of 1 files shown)


ci/scripts/install_softhsm.sh
```diff
@@ -5,8 +5,18 @@
 set -euo pipefail
 
 #Install softhsm
+# Install SoftHSM2 based on OS
```
aaadir, line 8: was/is this code tested on all types of OS-types?

aaadir (approved):
LGTM (left a Q)

PR #1321, opened 2026-02-10
Fix potential goroutine leak in certifier registration

### LINKED ISSUE #1318: Goroutine leak in RegisterView causes memory exhaustion in long-running systems
The RegisterView launches a goroutine that never terminates if public parameters are unavailable or context is cancelled, causing memory leaks in production.

```go
// token/services/certifier/views/register.go:46-60
go func() {
    for {
        pp := tms.PublicParametersManager().PublicParameters()
        if pp == nil {
            time.Sleep(500 * time.Millisecond)
            continue  // <----------------------------------- **Never checks context.Context().Done()**
        }
        // Start service
        break
    }
}()
```

**Problems:**
1. No timeout - runs forever if parameters never arrive
2. No context cancellation check - ignores `context.Context().Done()`
3. No cleanup mechanism - cannot be stopped once started

AkramBitar:
See #1318 for more information about the leak.

### DIFF
--- a/token/services/certifier/views/register.go
+++ b/token/services/certifier/views/register.go
@@ -43,21 +43,33 @@ func (r *RegisterView) Call(context view.Context) (interface{}, error) {
 	pp := tms.PublicParametersManager().PublicParameters()
 	if pp == nil {
 		logger.Debugf("public parameters not yet available, start a background task...")
+
+		// Use the view's context for cancellation.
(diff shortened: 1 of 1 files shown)


adecaro (changes requested):
it would be great to have a unit-test to identify the issue in the previous code and make sure it is not there anymore. Thanks

adecaro (approved):
LGTM

Issue #1331, opened 2026-02-15
unit-tests: token/core/zkatdlog/nogh/v1/transfer


adecaro:
This issue is about increasing the test coverage of the package `https://github.com/hyperledger-labs/fabric-token-sdk/issues/1331` to above 80%.

PR #1333, opened 2026-02-15
unit-tests: token/core/zkatdlog/nogh/v1/transfer #1331

adecaro:
This PR brings the following:
- improved unit-test coverage
- improved error messages

### DIFF
--- a/token/core/zkatdlog/nogh/v1/transfer/action_test.go
+++ b/token/core/zkatdlog/nogh/v1/transfer/action_test.go
@@ -490,14 +490,14 @@ func TestNewTransfer(t *testing.T) {
 	assert.NotNil(t, action.InputTokens())
 
 	serializedInputs, err := action.GetSerializedInputs()
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	assert.Len(t, serializedInputs, 2)
 
 	outputs := action.GetOutputs()
 	assert.Len(t, outputs, 2)
 
 	serializedOutputs, err := action.GetSerializedOutputs()
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	assert.Len(t, serializedOutputs, 2)
 }
 
@@ -508,20 +508,20 @@ func TestNewActionFromProtos(t *testing.T) {
 
 	// serialize it
 	raw, err := action.Serialize()
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	assert.NotNil(t, raw)
 
 	// deserialize it
 	action2 := &transfer.Action{}
 	err = action2.Deserialize(raw)
-	assert.NoError(t, err)
+	require.NoError(t, err)
 
 	// check that the deserialized action is equal to the original one
 	assert.Equal(t, action, action2)
 
 	// create a new action from protos
 	action3, err := transfer.NewActionFromProtos(raw)
-	assert.NoError(t, err)
+	require.NoError(t, err)
 	assert.NotNil(t, action3)
 	assert.Equal(t, action2, action3)
 }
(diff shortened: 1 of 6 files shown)


token/core/zkatdlog/nogh/v1/transfer/action_test.go
```diff
@@ -533,3 +640,109 @@ func randomAction(curve *math.Curve, rand io.Reader, b assert.TestingT) *transfe
...
+	action := &transfer.Action{
+		Outputs: []*token2.Token{
+			{
+				Owner: []byte("owner"),
+				Data:  c.GenG1.Copy(),
+			},
+		},
+	}
+	raw, err := action.SerializeOutputAt(0)
+	assert.NoError(t, err)
```
aaadir: to test that the serialization is valid and not just garbage you may also want to deserialize it to verify that the serialize correctly saved ther information that was in the source data. Same for the following test,

token/core/zkatdlog/nogh/v1/transfer/transfer.go
```diff
@@ -17,31 +17,34 @@ import (
...
 	p.TypeAndSum = &TypeAndSumProof{}
 	p.RangeCorrectness = &rp.RangeCorrectness{}
 	return asn1.Unmarshal[asn1.Serializer](bytes, p.TypeAndSum, p.RangeCorrectness)
 }
 
+// Validate ensures the proof components are present and well-formed.
 func (p *Proof) Validate(curve math.CurveID) error {
 	if p.TypeAndSum == nil {
-		return errors.New("invalid transfer proof")
+		return errors.New("invalid transfer proof: missing type-and-sum proof")
```
aaadir: why isn't there a similar error/message for the case below where the range proof is missing?

token/core/zkatdlog/nogh/v1/transfer/transfer_test.go
```diff
@@ -26,6 +28,40 @@ var (
...
+		Type:                 c.NewZrFromInt(1),
+		TypeBlindingFactor:   c.NewZrFromInt(1),
+		EqualityOfSum:        c.NewZrFromInt(1),
+		Challenge:            c.NewZrFromInt(1),
+	}
+	err = proof.Validate(TestCurve)
+	require.NoError(t, err)
+
+	// invalid range correctness
+	proof.RangeCorrectness = &rp.RangeCorrectness{
```
aaadir, line 59: - why is the rangeproof not validated similarly?
- is there a test that creates typeAndSum proofs for different valid cases and makes sure that the proof is correctly verified? similarly for different invalid cases (sum/type mismatches) ... similarly for different valid/invalid rangeproof cases (like border cases, negatives, ...).

token/core/zkatdlog/nogh/v1/transfer/action_test.go
adecaro: will do

token/core/zkatdlog/nogh/v1/transfer/transfer_test.go
adecaro: - The test is just checking some conditions that generate errors. In order to reach the point where an error related to range proof is returned we need the type and sum proof to pass.
- There are more tests in typeandsum_test.go. Please, double check that your points are covered there.
Thanks

aaadir: yes - typeandsum_test.go had tests of the type I was suggesting for typeAndSum. But I didn't find a similar test file for the bulletproofs (with tests like border cases, negatives,value in range but the "bits" are with non-0/1 values, ...)

adecaro: anyhow, in this test we check the working of the function Proof#Validate. Nothing else.

Issue #1077, opened 2025-05-13
configuration rework


adecaro:
This Github Issue calls for the following changes:

- `token.finality` must be moved under the `services.network` section of each TMS. They must be per network.
- `token.selector` must be moved under the TMS definition
- If not specified, values take the defaults. The node should not fail.
- if no persistence is defined is must default to the default configuration under `fsc.persistences`.

Additional tasks are:
- `core-token.md` must be updated to reflect the above changes
- NWO must generate the leanest possible configuration

sid200727:
Hi,
I’d like to help with this configuration rework.
From my understanding, the required changes are:
	•	Move token.finality under services.network (per network)
	•	Move token.selector under the TMS definition
	•	Ensure sensible defaults when values are not specified
	•	Default persistence to fsc.persistences if not defined
	•	Update core-token.md accordingly
	•	Adjust NWO to generate a lean configuration
To proceed safely, I propose the following approach:
	1.	Review the current configuration structure and how TMS + network configs are parsed.
	2.	Refactor the configuration schema to relocate the fields.
	3.	Implement backward-compatible defaults to avoid breaking existing setups.
	4.	Update documentation (core-token.md).
	5.	Adjust NWO config generation to match the new structure.

Before starting implementation:
Is backward compatibility required for existing configurations, or can we assume migration to the new layout?

Looking forward to your guidance.

sid200727:
I've opened PR #1336 addressing:

- Default values for delivery.blockProcessParallelism
- Identification of optional vs mandatory sections
- Documentation clarification
- Minimal configuration example

Please let me know if anything else should be refined to fully resolve this issue.

PR #1336, opened 2026-02-15
config: add default for token.finality.delivery.blockProcessParallelism

sid200727:
This PR addresses part of issue #1077.

Changes:
- Added a sensible default for token.finality.delivery.blockProcessParallelism
  to prevent node failure when the value is not specified.
- Clarified optional configuration sections and default values in core-token.md.
- Added documentation for minimal valid configuration.

This ensures:
- The node does not fail when optional values are missing.
- Developers clearly understand which sections are mandatory vs optional.
- The SDK can start with a minimal configuration.

Please let me know if further refinements are needed.

Related to #1077

### DIFF
--- a/token/services/network/fabric/config/config.go
+++ b/token/services/network/fabric/config/config.go
@@ -90,7 +90,7 @@ func (c *serviceListenerManagerConfig) DeliveryMapperParallelism() int {
 }
 
 func (c *serviceListenerManagerConfig) DeliveryBlockProcessParallelism() int {
-	if v := c.c.GetInt(DeliveryBlockProcessParallelism); v > 0 {
+	if v := c.c.GetInt(DeliveryBlockProcessParallelism); v >= 0 {
 		return v
 	}
 


docs/core-token.md
```diff
@@ -185,3 +185,52 @@ token:
...
+- type: delivery
+- committer.maxRetries: 3
+- committer.retryWaitDuration: 5s
+- delivery.mapperParallelism: 10
+- delivery.blockProcessParallelism: 10
+- delivery.lruSize: 30
+- delivery.lruBuffer: 15
+- delivery.listenerTimeout: 10s
+
+# Minimal Valid Token SDK Configuration
```
adecaro: I would say: Either we have a unit-test to confirm this or it is better to remove this Section.
This could become another PR where both the unit-test and this Section is introduced.
What do you think?

adecaro:
@sid200727 , thanks a lot for your effort. Left a comment. Thanks.

sid200727:
Hi @adecaro,

I've removed the "Minimal Valid Token SDK Configuration" section as suggested.

Please let me know if everything looks good now.

Thanks!

Issue #1232, opened 2025-10-09
Responders to `CollectEndorsementsView` may expose signing oracle


ale-linux:
The `CollectEndorsementsView` is responsible for the retrieval of signatures from 3rd parties. This for example happens when the node requests signatures over issues, transfers and endorsement acks. The responders to these views retrieve the message to be signed from the network session and sign it. Lack of verification of what message is signed may cause the signer to sign an arbitrary message.

Note that by virtue of the fact that the signer is dynamically determined, e.g.

```go
TokenService().SigService().GetSigner(signingIdentity)
```

a node transfering a token may be tricked into approving the transfer of a different token.

AkramBitar:
@adecaro you are right, because the issue was left open we assumed that the issue still exist. We verified and indeed the solution found its way to the code. Thanks.

PR #1352, opened 2026-02-17
verify received identity equals local identity

HayimShaul:
Verify the identity we get from the network equals the identity we have stored locally


HayimShaul:
Addresses issue #1232

PR #1353, opened 2026-02-17
Performance changes to the range proof verifier:

nitsatiisc:
- During the inner product argument (ipa.go) verification, we save group-exponentiations by delaying verifier's computation of folded generators. Instead of folding in each round which incurs 4n group-exp, we do a final MSM of size 2n + 2\log n.
- Auxiliary functions to help efficiency: Batch modular inversion and computation of scalar vector for final MSM (computeSVector).
- Speeds up verifier by around a factor of 2x


token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof_test.go
```diff
@@ -91,8 +93,10 @@ func TestBFProofVerify(t *testing.T) {
 		setup.l,
 		setup.curve,
 	)
+	start := time.Now()
```
adecaro: no need to add this, we have benchmarks to measure this

PR #1358, opened 2026-02-17
unit-test: token/core/common #1357

adecaro:
This PR improves the unit-test coverage of the package `token/core/common`.

### DIFF
--- a/token/core/common/authorization.go
+++ b/token/core/common/authorization.go
@@ -17,6 +17,7 @@ import (
 	token2 "github.com/hyperledger-labs/fabric-token-sdk/token/token"
 )
 
+// Authorization defines an interface for checking the ownership, issuance, and auditor status of tokens.
 type Authorization interface {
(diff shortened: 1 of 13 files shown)


token/core/common/crypto/slice_test.go
```diff
@@ -62,6 +64,27 @@ func TestAppendFixed32(t *testing.T) {
 	}
 }
```
AkramBitar, line 66: General comment for new tests, can we add description of what the test check?

token/core/common/encoding/asn1/asn1_test.go
```diff
@@ -236,3 +236,159 @@ func TestArray(t *testing.T) {
 	require.NoError(t, a2.Deserialize(raw))
 	assert.Equal(t, a1.Values, a2.Values)
 }
+
```
AkramBitar, line 239: General comment for new tests, can we add description of what the test check?

PR #1362, opened 2026-02-18
validate proposal

HayimShaul:
The new validateProposal() method verifies:

1. The Fabric proposal structure (header, payload)
2. That the creator (BankA's FSC node X.509 certificate) is a known MSP member: mspManager.IsValid(creator)
3. That the proposal signature was made by that creator: verifier.Verify(proposalBytes, signature)

The fix closes a gap where the FSC endorser was not explicitly verifying that the Fabric proposal came from a legitimate, authenticated MSP member before endorsing it.

### DIFF
--- a/token/services/network/fabric/endorsement/fsc/responder.go
+++ b/token/services/network/fabric/endorsement/fsc/responder.go
@@ -243,6 +243,7 @@ func (r *RequestApprovalResponderView) validateProposal(ctx view.Context, reques
 	}
 
 	logger.DebugfContext(ctx.Context(), "Proposal validated successfully for TX [%s]", request.Anchor)
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


token/services/network/fabric/endorsement/fsc/responder.go
```diff
@@ -180,6 +185,80 @@ func (r *RequestApprovalResponderView) receive(ctx view.Context) (*Request, erro
...
+	// Verify the signature
+	err = verifier.Verify(proposalBytes, signature)
+	if err != nil {
+		return errors.Wrapf(err, "failed to verify proposal signature for tx [%s]", request.Anchor)
+	}
+
+	// Validate that the token actions in the request are consistent with the proposal
+	// The token request should match what's in the transient data
+	// This ensures the relationship between the action, read-write set, and token actions
+	if len(request.RequestRaw) == 0 {
```
adecaro: this is already checked in line 138

token/services/network/fabric/endorsement/fsc/responder.go
```diff
@@ -180,6 +188,65 @@ func (r *RequestApprovalResponderView) receive(ctx view.Context) (*Request, erro
 	}, nil
 }
 
+func (r *RequestApprovalResponderView) validateProposal(ctx view.Context, request *Request) error {
```
mbrandenburger, line 191: This is create functionality. We should come back to this function make is reusuable and provide it via FSC to any view.

Issue #1300, opened 2026-01-19
Add ZKAT DLog No Graph Hiding Benchmarks to Testing Pipeline


AkramBitar:
Need to add benchmarks test under https://github.com/hyperledger-labs/fabric-token-sdk/blob/a3e86c396e956a5c081ec53fe9cc1edf75749fff/docs/benchmark/core/dlognogh/dlognogh.md to Testing Pipelines

adecaro:
Hi @sid200727 , thanks for jumping on this.

For this task, we don't really want to run the benchmark to obtain performance results on the CI. That would not make sense.
Instead, we want to make sure that they compile and don't fail on the default values.

So, an approach would be the following:
- create a new entry in the makefile similar to `unit-tests`, something liek `bench-check`.  Instead of running `go test`, we `go bench` and we list all the files that contain `func Benchmark` in it (using something like `grep -rl "func Benchmark" --include="*.go"`)
- add `make bench-check` in the CI. The target file is `.github/workflows/tests.yml`. There, we can add `bench-check` in the test matrix here 
``` yaml
utest:
    needs: checks
    runs-on: ubuntu-latest
    strategy:
      matrix:
        tests: [unit-tests-race, unit-tests-regression]
````

Let me know if this sounds about right for you.

Thanks again :)

sid200727:
Hi @adecero 👋

Just a gentle follow-up on this PR in case it slipped through.

All checks are passing and I'm happy to make adjustments if needed.

Thanks!

PR #1375, opened 2026-02-19
ci: add bench-check job to validate benchmark compilation

sid200727:
## Description

This PR adds a `bench-check` job to the CI pipeline to ensure that benchmark files compile successfully without executing them.

As discussed in #1300, we do not want to run benchmarks in CI to measure performance. Instead, we only validate that:

- All benchmark functions compile correctly
- No build errors are introduced in benchmark code

## Changes

- Added a new `bench-check` job in `.github/workflows/tests.yml`
- The job runs:

  ```bash
  go test ./... -run=^$ -bench=^$

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -169,7 +169,7 @@ jobs:
           make fxconfig configtxgen fabricx-docker-images
 
       - name: Run ${{ matrix.tests }}
-      
+
         run: |
           mkdir -p covdata
           GOCOVERDIR=$(realpath covdata) make integration-tests-${{ matrix.tests }}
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -169,6 +169,7 @@ jobs:
           make fxconfig configtxgen fabricx-docker-images
 
       - name: Run ${{ matrix.tests }}
+HEAD
```
adecaro: some mistake here?

.github/workflows/tests.yml
```diff
@@ -190,3 +191,24 @@ jobs:
...
+        with:
+          go-version-file: "go.mod"
+          cache-dependency-path: "**/*.sum"
+
+      - name: Download dependencies
+        run: go mod download
+
+      - name: Compile benchmarks (do not run)
+        run: go test ./... -run=^$ -bench=^$
+ecb1d261 (ci: add bench-check job to compile benchmarks in CI without executing them)
```
adecaro: also here?

Issue #1386, opened 2026-02-20
network service, fabricx: query-service based key lookup


adecaro:
Currently, the key lookup is performed using the delivery service which requires the parsing of each and every transaction.

The goal of this Issue is to replace the existing implementation under `token/services/network/fabricx/lookup` with a new implementation.

PR #1387, opened 2026-02-20
network service, fabricx: query-service based key lookup #1386

adecaro:
This PR implements the lookup listener manager on top of the fabricx query service using cron jobs.


docs/services/network.md
```diff
@@ -7,33 +7,132 @@ The network service architecture is depicted below:
...
+- **Transaction Lifecycle Management**: Handles the submission (broadcasting) of transactions and provides mechanisms to wait for their finality (commitment and validity).
+- **Ledger Querying (QE)**: Allows services to retrieve the current state of tokens and other ledger entries through a specialized Query Engine.
+- **Public Parameters (PP) Management**: Monitors the ledger for updates to the system's public parameters and ensures the SDK is always using the latest version to maintain cryptographic integrity.
+- **Identity & Membership**: Interfaces with the local Membership Service Provider (MSP) to provide identities for signing and transaction creation.
+
+The service uses a **Driver-based architecture**, allowing for different implementations to be plugged in based on the specific requirements and features of the target network.
+
 ## Fabric 
 
-The Fabric-based network implementation utilizes the Fabric Smart Client for configuration and operations, including chaincode queries and transaction broadcasting.
```
AkramBitar, line 12: I reviewed the document:
- Looks great
- Improvements: some terminology (such as TMS, namespace, setup key, Network Service, driver, Query Engine (QE), public parameters, token vault, finality, bootstrap process, read-write set, lookup listener, etc.) appears before it is defined. It would be helpful to include a short terminology section at the beginning of the document or provide references to documentation that explain these concepts.
- For someone new to the system, having these terms explained would make the document easier to understand and follow.

Maybe we can add this in other PR if you want.

integration/token/fungible/support.go
```diff
@@ -1046,7 +1046,7 @@ func UpdatePublicParamsAndWait(network *integration.Infrastructure, publicParams
 		if node.Id() == "custodian" {
 			continue
 		}
-		gomega.Eventually(GetPublicParams).WithArguments(network, node).WithTimeout(30 * time.Second).WithPolling(15 * time.Second).Should(gomega.Equal(publicParams))
+		gomega.Eventually(GetPublicParams).WithArguments(network, node).WithTimeout(60 * time.Second).WithPolling(5 * time.Second).Should(gomega.Equal(publicParams))
```
AkramBitar, line 1049: Is the 60 and 5 need to be tuned? if so, maybe to have them is config as well with default?

token/services/network/fabricx/lookup/config.go
```diff
@@ -0,0 +1,86 @@
...
+const (
+	// PermanentInterval is the configuration key for the polling interval of permanent lookups
+	PermanentInterval = "token.fabricx.lookup.permanent.interval"
+	// OnceDeadline is the configuration key for the deadline of one-time lookups
+	OnceDeadline = "token.fabricx.lookup.once.deadline"
+	// OnceInterval is the configuration key for the polling interval of one-time lookups
+	OnceInterval = "token.fabricx.lookup.once.interval"
+
+	// DefaultPermanentInterval is the default polling interval for permanent lookups
+	DefaultPermanentInterval = 1 * time.Minute
```
AkramBitar, line 20: Do we need to have this in config as well?
I am wondering if this can effect the overall performance and can be tuned as well.

token/services/network/fabricx/lookup/gocron.go
```diff
@@ -0,0 +1,228 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
```
AkramBitar, line 3: I suppose this can work on other OS (e.g., Windows) in the future (since corn job is Linux term)

token/services/network/fabricx/driver.go
```diff
@@ -94,12 +92,14 @@ func NewDriver(
 		defaultPublicParamsFetcher: ppFetcher,
 		queryExecutorProvider:      queryExecutorProvider,
 		keyTranslator:              kt,
-		flmProvider:                flmProvider,
-		llmProvider: lookup2.NewListenerManagerProvider(
-			fnsProvider,
-			tracerProvider,
-			kt,
-			config3.NewListenerManagerConfig(configService),
+		flmProvider: finality2.NewNotificationServiceBased(
```
AkramBitar, line 95: Uses cron jobs to periodically poll the FabricX Query Engine. My question: is there a way to do this in event-driven mode, by registering for notifications directly from Fabric instead of polling?

token/services/network/fabricx/lookup/gocron.go
```diff
@@ -0,0 +1,228 @@
...
+// state checks are implemented by this manager.
+func (n *CronListenerManager) PermanentLookupListenerSupported() bool {
+	return true
+}
+
+// AddPermanentLookupListener registers a listener that is notified whenever
+// the value of a key changes. It schedules a recurring job that polls
+// the query service at the configured permanent interval.
+func (n *CronListenerManager) AddPermanentLookupListener(namespace string, key string, listener Listener) error {
+	logger.Infof("AddPermanentLookupListener [%s:%s]", namespace, key)
```
AkramBitar, line 78: I do not see remove method for PermanentJob. Do we need a cleanup approach to remove PermanentJob from scheduled job?

token/services/network/fabricx/lookup/gocron.go
```diff
@@ -0,0 +1,228 @@
...
+
+// Run executes a single poll of the ledger key. It compares the current value
+// hash with the last seen hash and notifies the listener only if a change
+// is detected.
+func (j *PermanentJob) Run() {
+	logger.Infof("[PermanentKeyCheck] check for key [%s:%s]", j.namespace, j.key)
+	v, err := j.queryService.GetState(j.namespace, j.key)
+	if err == nil && v != nil && len(v.Raw) != 0 {
+		newHash := token.Hashable(v.Raw).Raw()
+		if !bytes.Equal(j.lastValue, newHash) {
```
AkramBitar, line 222: I am wondering if two goroutine can call this run. If so, then we need to protect the j.lastValue

docs/services/network.md
adecaro: sure, we can improve on that. There is the main documentation that is already introducing most of this terminology.

integration/token/fungible/support.go
adecaro: this is just for the test, it mostly depends on the power of the machine the test is run on.

token/services/network/fabricx/lookup/config.go
adecaro: all the parameters are configurable. These are just defaults in case something is missing. I believe I have updated also the configuration example for that.

token/services/network/fabricx/lookup/gocron.go
adecaro: it is called cron because of the library gocron. It is fully written in Go. It works everywhere.

token/services/network/fabricx/driver.go
adecaro: FabricX does not support that yet.

token/services/network/fabricx/lookup/gocron.go
adecaro: they are never removed by design. Yes.

token/services/network/fabricx/lookup/gocron.go
adecaro: gocron is configured such that, it will not call this in parallel. We should be safe, but good point.

```	j, err := n.scheduler.NewJob(
		gocron.DurationJob(n.config.PermanentInterval()),
		gocron.NewTask(job.Run),
		gocron.WithSingletonMode(gocron.LimitModeReschedule),
	)
```

PR #1394, opened 2026-02-23
Increase test coverage for idemix identity package

aaadir:
write new unit tests to improve coverage of the token/services/identity/idemix package


aaadir:
coverage of idemix package now raised with the latest commits to 91.1% of statements

PR #1399, opened 2026-02-24
split token driver in two #864

### LINKED ISSUE #864: split token driver in two
This Issue requests to split `driver.Driver` in two interfaces:
- One to obtain an instance of `TokenManagerService`;
- Another one to obtain an instance `Validator`.
This is to avoid that when only the validator is needed, it carries over dependencies to the TMS as well with possible generated conflicts.

### LINKED ISSUE #1191: make it easier to extend an existing driver validation logic
If one wants to add just new validation functions, extending a driver should require minimal new code to be added.

adecaro:
This PR should make it easier to extends the validator side of a token driver.


token/core/service.go
```diff
@@ -157,11 +156,24 @@ func (s *TokenDriverService) NewTokenService(tmsID driver.TMSID, publicParams []
 	return nil, errors.Errorf("no token driver named '%s' found", DriverIdentifierFromPP(pp))
 }
 
-// NewDefaultValidator returns a new instance of driver.Validator for the passed public parameters.
+// TokenDriverService is an alias for TMSDriverService for backward compatibility.
```
arner, line 146: If you start the comment with `// DEPRECATED: `, it will be clear to users (even with warnings in the editor) that the new name is preferred.

docs/driverapi.md
```diff
@@ -2,10 +2,10 @@
 
 The **Driver API** serves as the interface bridging the generic Token API with specific token implementations. It defines the protocols for token creation, transfer, and management within a given system.
 
-Each driver must implement the `driver.Driver` interface, fulfilling three primary objectives:
-1.  **Public Parameters Discovery**: Decodes raw bytes into driver-specific public parameters.
-2.  **TMS Instantiation**: Provides a mechanism to instantiate a new `TokenManagementService` (TMS) tailored to the driver and its public parameters.
-3.  **Default Validation**: Returns a default `Validator` instance configured with the driver's public parameters.
+Each driver must implement the `driver.Driver` and `driver.ValidatorDriver` interfaces, fulfilling three primary objectives:
```
arner, line 5: I've reviewed the code, indeed it's splitting the driver, but to be honest it's not completely clear to me how this solves https://github.com/hyperledger-labs/fabric-token-sdk/issues/1191!

Maybe it does, but it requires a lot of knowledge of the inner workings. If this change means you can more easily add a new validator without rewriting an entire driver, maybe the documentation could give an example of how to do that? For instance using the zkatdlog driver and adding a new check.

adecaro: That's the direction, the ability to create a new validator more easily without having to instantiate things that are not validation-related. 

Good point, I'll add an example and ask again your review.
Thanks @arner 🙏

arner: Now it's clear! Thanks for the detailed guide.

arner (approved):
LGTM! Thanks @adecaro

PR #1404, opened 2026-02-26
docs: clarify 1-of-N auditor signature semantics

sid200727:
## Summary

This PR clarifies the semantics of auditor signature validation in the token SDK.

## Background

The current implementation follows a 1-of-N auditor signature policy:

- If auditors are configured in the public parameters, at least one valid auditor signature must be present.
- Multiple auditor public keys may be configured (e.g., during key rotation).
- Each provided auditor signature is independently verified.
- N-of-N semantics are not enforced.
- All configured auditor public keys are treated as belonging to a single logical auditor entity.

The existing behavior was correct but not explicitly documented, which could lead to ambiguity when reviewing or extending the validator logic.

## Changes

- Added detailed documentation to `validator_auditing.go` explaining the current signature model.
- Clarified auditor semantics in `core-token.md`.

## Impact

This PR does **not** change runtime behaviour.
It only documents the existing semantics to make the intended policy explicit.

### DIFF
--- a/token/core/common/validator_auditing.go
+++ b/token/core/common/validator_auditing.go
@@ -25,18 +25,17 @@ var (
 //
 // The current implementation follows a 1-of-N auditor signature policy:
 //
-// - If auditors are configured in the public parameters, at least one
-//   valid auditor signature must be present in the token request.
-// - Multiple auditor public keys may be configured (e.g., during key rotation).
-// - The validator verifies that each provided auditor signature corresponds
-//   to a configured auditor and that the signature is valid.
-// - The validator does NOT enforce N-of-N semantics.
-// - The validator does NOT enforce per-entity auditor checks.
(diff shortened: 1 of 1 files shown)


token/core/common/validator_auditing.go
```diff
@@ -20,6 +20,23 @@ var (
...
+// - Multiple auditor public keys may be configured (e.g., during key rotation).
+// - The validator verifies that each provided auditor signature corresponds
+//   to a configured auditor and that the signature is valid.
+// - The validator does NOT enforce N-of-N semantics.
+// - The validator does NOT enforce per-entity auditor checks.
+// - All configured auditor public keys are treated as belonging to a
+//   single logical auditor entity.
+//
+// This behavior matches the semantics implemented by current token drivers.
+
```
adecaro: please, remove this line so that the comments are bound to the function

PR #1406, opened 2026-02-27
Ttx/replace random nonce

### LINKED ISSUE #919: ttx service: replace `GetRandomBytes`
This is related to the following TODO: 
```
// TODO: rand could fill less bytes then len
```
`GetRandomBytes` is used to get a unique identifier. We can either replace this with some `uuid` or make sure we sample enough bytes, therefore addressing the todo.

sid200727:
Fixes #919

This PR addresses the TODO in `GetRandomBytes` within `token/services/ttx/random.go`.

Previously, `rand.Read` was used to generate random bytes. However, `rand.Read` does not guarantee that the requested number of bytes will be fully populated in a single call. This could potentially result in partially filled nonces without being detected.

This change replaces `rand.Read` with `io.ReadFull(rand.Reader, ...)`, ensuring that either:
- The exact number of requested bytes is generated, or
- An error is returned.

This guarantees correct and complete nonce generation in `GetRandomNonce`, improving reliability and aligning with Go best practices for secure randomness.


adecaro (approved):
LGTM

PR #1412, opened 2026-03-08
unit tests: identity deserializer

aaadir:
Improve coverage of token/services/identity/deserializer/deserializer_test.go


token/services/identity/deserializer/deserializer_test.go
```diff
@@ -0,0 +1,985 @@
...
+	}
+
+	return nil
+}
+
+type mockVerifier struct {
+	verifyFunc func(message, sigma []byte) error
+}
+
+func (m *mockVerifier) Verify(message, sigma []byte) error {
```
AkramBitar: Did you check if we can use Counterfeiter to generate these mocks?

token/services/identity/deserializer/deserializer_test.go
```diff
@@ -0,0 +1,985 @@
...
+
+func (m *mockSigner) Sign(message []byte) ([]byte, error) {
+	if m.signFunc != nil {
+		return m.signFunc(message)
+	}
+
+	return []byte("signature"), nil
+}
+
+type mockTypedVerifierDeserializer struct {
```
AkramBitar: Did you check if we can use Counterfeiter to generate these mocks?

Please check all of them

aaadir: Done.
- Counterfeiters for the following classes are now used instead of the previous manually defined mocks:
```
drivermock.Matcher
drivermock.Signer
drivermock.Verifier
drivermock.AuditInfoProvider
identitydrivermock.AuditInfo
identitydrivermock.AuditInfoDeserializer
identitydrivermock.TypedSignerDeserializer
identitydrivermock.TypedVerifierDeserializer
```

- Note that for the above I had to move the definition of the interface _TypedVerifierDeserializer_ 
to be in `token/services/identity/driver/deserialization.go` next to the defnition of the interface _TypedSignerDeserializer_.
It used to be in the file `token/services/identity/deserializer/verifier.go ` which includes implementations of the TypedVerifierDeserializer interface.
This caused dependency cycles for the new `token/services/identity/deserializer/deserializer_test.go`
which was resolved when I moved the interface defintion to its proper place.

PR #1416, opened 2026-03-10
feat(identity_service): check identity DB if a lookup fails #1415

### LINKED ISSUE #1415: feat(identity_service): check identity DB if a lookup fails
If a replica register a new identity configuration, the other replicas are not able to use it. 
This issue asks to investigate a possible solution to this issue.

adecaro:
This PR fixes #1415 by introducing a notification service for the identity table. The notification service is provided by the smart client. In this PR we instantiate it to track the additions performed to the identity table. Upon an event, the membership service registers the new identity.

### DIFF
--- a/token/services/storage/db/kvs/identitydb.go
+++ b/token/services/storage/db/kvs/identitydb.go
@@ -123,7 +123,7 @@ func (s *IdentityStore) ConfigurationExists(ctx context.Context, id, configurati
 }
 
 func (s *IdentityStore) Notifier() (idriver.IdentityConfigurationNotifier, error) {
-	return nil, nil
+	return nil, storage.ErrNotSupported
 }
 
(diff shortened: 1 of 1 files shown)


token/services/storage/db/sql/postgres/driver.go
```diff
@@ -88,6 +142,16 @@ func (d *Driver) NewIdentity(name driver2.PersistenceName, params ...string) (dr
...
 }
 
+// NewIdentityNotifier returns a new IdentityNotifier for the given persistence name and params.
+func (d *Driver) NewIdentityNotifier(name driver2.PersistenceName, params ...string) (idriver.IdentityConfigurationNotifier, error) {
+	opts, err := d.cp.GetOpts(name, params...)
+	if err != nil {
+		return nil, err
+	}
+
+	return d.IdentityNotifier.Get(*opts)
```
AkramBitar: I am wondering if we need to add unit test for this file also the same comments for other files go file in driver, kvs, memory, postgres, sqlite, etc.

adecaro: good point, will add.

PR #1419, opened 2026-03-11
Fix: Stop silently dropping failed finality events

atharrva01:
### What I Hit

I was debugging why some token transfers never completed even though the ledger showed them committed. Turns out the FabricX finality queue worker just... drops events when they fail. No retry, no warning in production (Debug logs), gone forever.

When `GetState()` fails during a peer restart or network blip, the whole downstream pipeline never fires. Tokens stay locked. Receivers never get notified. And you'd never know unless you had Debug logging on.

### The Pain

- **Silent data loss** - committed transactions vanish from wallets
- **Permanently locked funds** - senders can't re-spend their tokens  
- **No visibility** - production systems see nothing in logs

The old Fabric path handles this safely. Only the FabricX queue has this bug.

### What I Changed

Added retry logic with exponential backoff to `queue.go`. Now failed events retry 3 times (1s, 2s, 4s) before giving up, and we log properly at Warn/Error levels so you can actually see problems.

Also fixed worker panics from killing the pool permanently, and made retry params configurable (with defaults so nothing breaks).

Key change:
```go
// old - silent drop
logger.Debugf("error processing event: %v", err)

// new - retry with backoff
logger.Warnf("retrying event after error: %v", err)
// ...exponential backoff, then Error if exhausted
```

### Tests Pass

image

All retry behavior verified. Transient failures now recover instead of corrupting state.

### DIFF
--- a/token/services/network/fabricx/finality/finality_test.go
+++ b/token/services/network/fabricx/finality/finality_test.go
@@ -225,6 +225,7 @@ func TestListenerEvent_Process_RetriesOnTransientError(t *testing.T) {
 		if attempts < 3 {
 			return nil, errors.New("transient peer error")
 		}
+
 		return &cdriver.VaultValue{Raw: tokenRequestHash}, nil
 	}
 	mockKT.CreateTokenRequestKeyReturns(key, nil)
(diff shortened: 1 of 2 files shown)


adecaro:
okay, let's do the following. Please, move the retry logic in the event. We need only ListenerEvent to retry, given the current logic. Then, would you like to try to add to the FinalityListener interface a new function `OnError` that the backend can be used to signal an issuer to the frontend?
What do you think?

atharrva01:
Agreed @adecaro ! I'll go ahead and:

  1. Move the retry logic out of the queue and into ListenerEvent.Process() (keeping the queue as a simple dispatcher)
  2. Add OnError to the FinalityListener interface so the business layer gets notified when all retries are exhausted

Working on it now.

adecaro:
@atharrva01 many thanks for your effort 🙇

PR #1421, opened 2026-03-12
docs: update

adecaro:
This PR introduces multiple improvements to the documentation.

### DIFF
--- a/docs/upgradability.md
+++ b/docs/upgradability.md
@@ -113,6 +113,78 @@ func SupportedTokenFormat(precision uint64) (token.Format, error) {
 
 For more information on how token formats are used in the SDK's token service, see the [Tokens Service documentation](./services/tokens.md).
 
+### Public Parameters Upgrade Process
+
+The Fabric Token SDK provides a structured approach for upgrading public parameters across the network. 
+This process ensures that all participants can smoothly transition to new cryptographic parameters while maintaining compatibility.
+
+```mermaid
+sequenceDiagram
+    autonumber
+
+    actor Admin as Administrator<br/>(or Issuer)
+    participant Fabric as Ledger<br/>(Fabric/FabricX/etc)
+
+    box darkgreen Token SDK Stack
+        participant Network as Network Service
+        participant Provider as TMS Provider
+        participant TMS as Token Management<br/>Service
+        participant Driver as Driver API
+    end
+
+    Admin->>+Fabric: Generate & Publish<br/>New Public Parameters
+    Fabric->>+Network: Ledger Update<br/>Notification
+    deactivate Fabric
+
+    Network->>+Provider: GetManagementServiceProvider
+    Provider->>+TMS: Update/Create TMS<br/>(with new PP)
+    TMS->>+Driver: Set Public Parameters
+
+    Driver-->>-TMS: Ready for Token Operations
+    TMS-->>-Provider: TMS Instance
+    Provider-->>-Network: TMS Response
+    Network-->>-Admin: Update Complete<br/>(Optional Notification)
+```
+
+#### Process Overview
+
+1. **Generation**: New public parameters are created using tools like [tokengen](./development/tokengen.md) or custom processes
+2. **Publishing**: Parameters are distributed to the network backend (for Fabric: via chaincode transactions or configuration updates)
+3. **Detection**: The Network Service monitors the ledger for public parameter updates
+4. **Fetching**: The SDK retrieves new parameters through the `PublicParamsFetcher` interface
+5. **Update**: The `TokenManagerServiceProvider` compares new vs existing parameters and updates the TMS if changed
+6. **Propagation**: All subsequent token operations use the upgraded public parameters
+
+#### Verification
+
+After the upgrade process completes, administrators can verify that all nodes are synchronized by retrieving the public parameters using the Token API's PublicParametersManager. This ensures that the new parameters have been successfully propagated throughout the network.
+
+```go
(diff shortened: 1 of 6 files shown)


docs/upgradability.md
```diff
@@ -0,0 +1,197 @@
...
+1.  **Format Support**: The token's format must be included in the driver's `SupportedTokenFormats()`.
+2.  **Precision Compatibility**: For Fabtoken to DLog upgrades, the original token's precision must be less than or equal to the current driver's maximum supported precision (e.g., 64-bit).
+3.  **Automatic Commitment**: When the driver encounters a compatible legacy token (like Fabtoken), it automatically generates a Pedersen commitment and an **Upgrade Witness**. This witness allows the new driver to prove the validity of the original token while treating it as a zero-knowledge commitment in the new transaction.
+
+### The "Burn and Re-issue" Mechanism
+
+When in-place upgrade is not possible (e.g., moving to a completely incompatible cryptographic curve or increasing precision beyond limits), the SDK implements an atomic "Burn and Re-issue" protocol. 
+
+#### Step-by-Step Flow:
+1.  **Identification**: The owner identifies tokens that are no longer supported.
```
AkramBitar, line 31: Does the customer need to do the identification on each new version?
I am wondering if he can know that based on release notes and then he can run the identification

docs/upgradability.md
```diff
@@ -0,0 +1,197 @@
...
+The SDK manages upgradability at three distinct layers:
+1.  **Ledger Layer**: Upgrading existing tokens to new formats.
+2.  **Driver Layer**: Managing compatibility between the SDK and underlying token implementations.
+3.  **Storage Layer**: Handling local database schema evolutions.
+
+---
+
+## Token Upgradability (Ledger)
+
+The SDK manages token upgrades using two distinct mechanisms: the **Atomic "Burn and Re-issue" protocol** for across-format migrations and **In-place Upgrades** for backward-compatible transitions.
```
AkramBitar, line 14: Can you give a short description of these two mechanisms?

In the burn-and-reissue process, is there a linkage between burned tokens and newly issued tokens, and can they be correlated?

Is the burn-and-reissue process executed in the background during system operation? If so, how should we handle transactions involving tokens that are pending upgrade but have not yet been processed?

docs/upgradability.md
```diff
@@ -0,0 +1,197 @@
...
+- [**DLog (NOGH) Protobuf Messages**](drivers/dlogwogh.md#protobuf-messages)
+
+3.  **Extensible Metadata**: 
+    Most core messages (like `IssueMetadata` or `PublicParameters`) include a `map<string, bytes> extra_data` or `application` field. This allows developers to attach arbitrary information to transactions or configurations without modifying the underlying `.proto` definitions, avoiding the need for a full protocol migration for application-specific changes.
+
+### Recommendations for Protocol Changes
+*   **Never Reuse Field Numbers**: Once a field number is assigned in a `.proto` file, it must never be reassigned to a different field, even if the original field is deprecated.
+*   **Prefer Optional Fields**: Use `proto3` defaults or explicitly check for presence to ensure that missing fields from older clients don't cause crashes.
+*   **Versioned Packages**: For major, breaking changes in a driver's internal logic, create a new protobuf package (e.g., `package zkatdlognogh.v2;`). This allows both the old and new unmarshallers to coexist in the same codebase.
+
```
AkramBitar, line 260: Could we have a diagram that explains the upgradability flow? It would help to visualize the overall process.

docs/upgradability.md
```diff
@@ -0,0 +1,197 @@
...
+---
+
+## Summary of Upgradability Responsibilities
+
+| Component | Responsibility | Mechanism |
+| :--- | :--- | :--- |
+| **Tokens** | Owner / Issuer | `ttx.Transaction.Upgrade` (Burn & Re-issue) |
+| **Driver** | Admin / SDK | `PostInit` (Automatic Spendability Toggle) |
```
AkramBitar, line 268: Is there a way that the customer can validate that the Upgradability finished successfully?

docs/imgs/storage_db.puml
```diff
@@ -0,0 +1,173 @@
+@startuml
```
AkramBitar, line 1: If a DDL schema is available, it would be clearer to use it rather than redefining the tables. This would also make future changes more transparent (something to consider going forward).

AkramBitar (changes requested):

docs/upgradability.md
adecaro: the example is right after under `### Code Example: Identifying Unsupported Tokens`

docs/upgradability.md
adecaro: Description is below. Regarding your questions: (1) it depends on the token driver but an adversary can see an issue transaction burning tokens and that is sign of the conversion process. (2) No, the application needs to orchestrate this. The token-sdk just gives the mechanisms. At least, for now.

docs/upgradability.md
AkramBitar: Yes I saw that. My question was does the customer needs to run ### Code Example: Identifying Unsupported Tokens on each release, and then decide. Other, option to check the release not and he can know that he need to upgrade (maybe both)

docs/upgradability.md
adecaro: Good point, I'll prepare one.

docs/upgradability.md
adecaro: Good point, will clarify this point.

PR #1424, opened 2026-03-15
fix: nil dereference and panic in setupListener callbacks crash peer on transient errors

atharrva01:
**What's this about**

While digging through the TMS setup flow, I found two bugs in `setupListener` that can crash the peer under completely normal distributed system conditions.

**The Problem**

`setupListener` is registered permanently, it watches the setup key for public parameter updates for the entire lifetime of the peer. Both its callbacks were broken:

**1. Nil dereference in `OnStatus()`**
When `GetTokens()` fails (a transient DB hiccup is enough), the code logs a warning but doesn't return, so it walks straight into a nil pointer dereference:
```go
tokens, err := s.GetTokens()
if err != nil {
    logger.Warnf("failed to get tokens db [%v]", err)
    // no return - falls through with tokens == nil
}
tokens.StorePublicParams(ctx, value) // 💥 panic
```

**2. Unconditional panic in `OnError()`**
Any lookup error reconnect, timeout, peer restart hits this:
```go
func (s *setupListener) OnError(...) {
    panic("implement me") // crashes every time, no exceptions
}
```

**Why it matters**

Since this listener lives for the peer's entire lifetime, both crash paths are always active. Worse, if the condition persists across restarts, the peer enters a crash loop.

**Fix**

Added a `return` after the `GetTokens()` error check, and replaced the `panic` in `OnError()` with a warning log. The permanent listener retries naturally on the next delivery cycle.

### DIFF
--- a/token/services/network/fabric/network.go
+++ b/token/services/network/fabric/network.go
@@ -417,6 +417,7 @@ func (s *setupListener) OnStatus(ctx context.Context, key string, value []byte) 
 	tokens, err := s.GetTokens()
 	if err != nil {
 		logger.Warnf("failed to get tokens db [%v]", err)
+
 		return
 	}
 	if err := tokens.StorePublicParams(ctx, value); err != nil {


adecaro (approved):
LGTM.

PR #1425, opened 2026-03-15
fix: propagate error in ReceiveSpendRequestView

atharrva01:
**What's the problem?**

While digging through the multisig spend flow, I noticed `ReceiveSpendRequestView` was silently swallowing receive errors. If the session timed out or dropped mid-receive, the error got logged but never returned, so the caller kept moving forward with an empty `SpendRequest`.

**Why it matters**

In a multisig spend flow, Party B would receive a nil-Token request, send an approval anyway, and blindly endorse whatever transaction came next. In a network partition or peer-restart scenario, that's B endorsing something it never actually agreed to.

**The fix**

```go
// Before
err := jsonSession.ReceiveWithTimeout(tx, time.Minute*4)
if err != nil {
    logger.ErrorfContext(...)
}
return tx, nil  // always nil, even on failure

// After
if err != nil {
    logger.ErrorfContext(...)
    return nil, err  // abort instead of proceeding blind
}
return tx, nil
```

One line added. Happy path is unchanged, I only touched error cases, which is exactly where the behavior was wrong.

The `TODO: check tx matches request` in `EndorseSpendView` is something I'll leave as a separate follow-up.

### DIFF
--- a/token/services/ttx/multisig/spend.go
+++ b/token/services/ttx/multisig/spend.go
@@ -64,6 +64,7 @@ func (f *ReceiveSpendRequestView) Call(context view.Context) (interface{}, error
 	err := jsonSession.ReceiveWithTimeout(tx, time.Minute*4)
 	if err != nil {
 		logger.ErrorfContext(context.Context(), "failed receiving request: %s", err)
+
 		return nil, err
 	}
 


adecaro:
@atharrva01 , great catch. Merging this.

PR #1426, opened 2026-03-16
fix: prevent batch-wide finality event loss in queryByID

### LINKED ISSUE #1546: bug: batch-wide finality event loss in queryByID when any single transaction fails
### Summary

The `queryByID` function in `DeliveryScanQueryByID` exits the entire batch goroutine whenever a single transaction fails. This causes all remaining transactions in the batch to be silently dropped without emitting finality events , a serious correctness issue. 

---

### Root Cause

Two error paths incorrectly use `return` instead of continuing:

* **Mapping failure (`MapProcessedTx`)** → exits the whole batch
* **Transient ledger error (`GetTransactionByID`)** → also exits the whole batch

Both cases stop processing remaining transactions prematurely. 

---

### Impact

* Some transactions in the batch may be skipped if an earlier one fails
* Finality events might not be emitted for those skipped transactions
* Token state updates could be incomplete in certain cases
* Transactions may remain in an `Unknown` state longer than expected
* In some scenarios, this could contribute to timeouts or inconsistent behavior

---

### Expected Behavior

Each transaction should be handled independently:

* Log errors and continue processing the rest
* Trigger fallback mechanisms for transient failures
* Never let one failure break the entire batch

---

### Fix

* Replace `return` with `continue` in both error paths
* Trigger fallback scanning for transient errors
* Improve logging and add proper unit tests

### Trigger Path 1 - Transient Ledger Error (common case)

**Setup:**

* Use default LRU size (30)
* Submit >30 concurrent transactions to trigger eviction (which calls `queryByID`)

**What happens:**

* Restart a peer (or simulate a transient ledger error) while `queryByID` is running
* One `GetTransactionByID` call fails → function exits early

**Result:**

* Remaining transactions in the batch may never get finality events
* Callers can end up waiting until timeout
* Some transactions may stay in `Unknown` state

---

### Trigger Path 2 - Mapping Failure

**Setup:** Same as above (force LRU eviction)

**What happens:**

* One transaction fails in `MapProcessedTx` (e.g., non-token/system tx)
* Function exits early again

**Result:**

* Later transactions in the batch may be skipped
* Similar symptoms: missing finality, delays, or inconsistent state

---

### Minimal Repro (Unit Test)

* Mock a ledger that fails for one tx in a batch
* Call `queryByID` with multiple txIDs
* Observe: txs after the failure may never be processed

 v0.10.0

### 🐹 Go Version

 go version go1.25.7 linux/amd64

### 💻 Operating System

Linux (Ubuntu/Debian)

atharrva01:
Closes #1546 

**What's the problem?**

While digging into the finality delivery code, I found a bug in `queryByID` where a single failure — transient ledger error or failed tx mapping, causes the goroutine to exit early via `return`. Since the channel closes through `defer close(ch)`, every other transaction in the same evicted batch silently loses its finality event.

```go
// before - one bad txID kills the whole batch
infos, err := q.Mapper.MapProcessedTx(pt)
if err != nil {
    logger.Errorf("failed to map tx [%s]: [%s]", txID, err)
    return  // ← drops all remaining txIDs
}
```

**Why it matters**

This is the *only* recovery path for transactions evicted from the LRU cache. When it silently bails: `tokens.Append` never runs, `ttxDB.SetStatus` never runs (tx stays `Unknown` forever), and the finality view times out after 5–10 minutes. Input tokens stay spendable locally even though they're spent on-chain, next spend attempt gets endorsed but rejected at ordering as a double-spend. User is stuck.

Easy to hit during routine maintenance or peer restarts.

**Fix**

Instead of exiting on a per-item failure, I made each txID fail independently:

```go
// MapProcessedTx failure → skip this txID, continue the batch
if err != nil {
    logger.Errorf("failed to map tx [%s]: [%s], skipping", txID, err)
    keySet.Remove(txID)
    continue
}

// Transient ledger error → fall back to ScanFromBlock
logger.Errorf("scan for tx [%s] failed, falling back to block scan", txID, err)
startDelivery = true
// txID stays in keySet for ScanFromBlock recovery
```

Mapping failures get removed from `keySet` so block scan doesn't re-encounter them. Transient ledger errors stay in `keySet` for a second chance via `ScanFromBlock` , same as existing "not found" handling.

**Result**

A failure for one evicted transaction now has a blast radius of one. All co-evicted transactions proceed normally and get their finality events.

### DIFF
--- a/token/services/network/fabric/finality/deliveryqs_test.go
+++ b/token/services/network/fabric/finality/deliveryqs_test.go
@@ -81,6 +81,7 @@ func evicted(txIDs ...string) map[cdriver.TxID][]events2.ListenerEntry[finality.
 	for _, id := range txIDs {
 		m[id] = nil
 	}
+
 	return m
 }
 
@@ -89,6 +90,7 @@ func drain(ch <-chan []finality.TxInfo) []finality.TxInfo {
 	for batch := range ch {
 		all = append(all, batch...)
 	}
+
 	return all
 }
 


AkramBitar:
Hello @atharrva01, could you please open a Git issue for this and link it to this PR?

atharrva01:
hey @AkramBitar , i have opened the issue and linked this pr

PR #1431, opened 2026-03-17
fix(storage): support large token amounts using NUMERIC and big.Int

sid200727:
## Problem

Currently, token quantities are stored using BIGINT in the storage layer.  
Since BIGINT is a signed 64-bit integer, it cannot represent the full uint64 range, leading to overflow issues for large values.

## Solution

This PR updates the storage handling of token amounts to support large values safely:

- Changed database column type from BIGINT to NUMERIC
- Updated read logic to parse amounts using big.Int
- Updated write logic to store amounts using string representation
- Ensured compatibility with existing query flows

## Implementation Details
- QueryMovements:
  - Read amount as string and convert using big.Int

- QueryTransactions:
  - Read amount as generic type (interface{}) and safely convert to string → big.Int

- AddTransaction / AddMovement:
  - Store amounts using `Amount.String()`

- Schema:
  - Updated `amount` column to NUMERIC

## Notes
- NUMERIC is chosen for compatibility across different SQL backends
- This avoids overflow issues while maintaining precision

## Future Work (optional)
- Migration strategy for existing databases with BIGINT columns
- Performance evaluation of NUMERIC vs BIGINT for large datasets
Note:
While the schema and read paths now support large (256-bit) values via NUMERIC and big.Int,
write paths still use int64 for compatibility with existing tests and drivers.

This can be extended in a follow-up to fully support large values end-to-end.

### DIFF
--- a/token/services/storage/db/sql/common/transactions.go
+++ b/token/services/storage/db/sql/common/transactions.go
@@ -125,7 +125,7 @@ func (db *TransactionStore) QueryMovements(ctx context.Context, params driver4.Q
 			return err
 		}
 		r.Amount = amount.Int
-		logger.DebugfContext(ctx, "movement [%s:%s:%s]", r.TxID, r.Status, r.Amount.Int64())
+		logger.DebugfContext(ctx, "movement [%s:%s:%s]", r.TxID, r.Status, r.Amount)
 
 		return nil
 	})
@@ -468,7 +468,7 @@ func (w *AtomicWrite) AddTransaction(ctx context.Context, rs ...driver4.Transact
 			return errors.Wrapf(err, "error generating uuid")
 		}
 		if r.Amount.BitLen() > maxAmountBits {
-			return errors.Errorf("amount [%s] exceeds maximum supported size of %d bits", r.Amount.String(), maxAmountBits)
+			return errors.Errorf("amount [%s] exceeds maximum supported size of %d bits", r.Amount, maxAmountBits)
 		}
 		rows[i] = common3.Tuple{id, r.TxID, int(r.ActionType), r.SenderEID, r.RecipientEID, r.TokenType, r.Amount.String(), r.Timestamp.UTC()}
 	}
@@ -527,7 +527,7 @@ func (w *AtomicWrite) AddMovement(ctx context.Context, rs ...driver4.MovementRec
 			return errors.Wrapf(err, "error generating uuid")
 		}
 		if r.Amount.BitLen() > maxAmountBits {
-			return errors.Errorf("amount [%s] exceeds maximum supported size of %d bits", r.Amount.String(), maxAmountBits)
+			return errors.Errorf("amount [%s] exceeds maximum supported size of %d bits", r.Amount, maxAmountBits)
 		}
 		rows[i] = common3.Tuple{id, r.TxID, r.EnrollmentID, r.TokenType, r.Amount.String(), now}
 	}


token/services/storage/db/sql/common/transactions.go
```diff
@@ -459,14 +470,11 @@ func (w *AtomicWrite) AddTransaction(ctx context.Context, rs ...driver4.Transact
 	rows := make([]common3.Tuple, len(rs))
 	for i, r := range rs {
 		logger.DebugfContext(ctx, "adding transaction record [%s:%d,%s:%s:%s:%s]", r.TxID, r.ActionType, r.TokenType, r.SenderEID, r.RecipientEID, r.Amount)
-		if !r.Amount.IsInt64() {
```
adecaro, line 462: let's not remove this but check that the number of bits used by amount is less or equal to than 255

token/services/storage/db/sql/common/transactions.go
```diff
@@ -153,13 +158,19 @@ func (db *TransactionStore) QueryTransactions(ctx context.Context, params driver
 	}
 
 	results := common.NewIterator(rows, func(r *driver4.TransactionRecord) error {
-		var amount int64
+		var amountRaw interface{}
```
adecaro: I would rather introduce something like:

// BigInt is a custom wrapper around math/big.Int
type BigInt struct {
big.Int
}

// Scan implements the sql.Scanner interface for reading from the DB
func (b *BigInt) Scan(value interface{}) error {
if value == nil {
b.SetInt64(0) // Handle NULLs, or return an error if you strictly enforce NOT NULL
return nil
}

	var str string
	switch v := value.(type) {
	case []byte:
		str = string(v) // PostgreSQL NUMERIC generally comes over the wire as []byte
	case string:
		str = v
	case int64:
		b.SetInt64(v)
		return nil
	default:
		return fmt.Errorf("cannot scan type %T into BigInt", value)
	}

	// Parse the string directly into the embedded big.Int
	if _, ok := b.SetString(str, 10); !ok {
		return fmt.Errorf("failed to parse PostgreSQL NUMERIC %q into big.Int", str)
	}

	return nil
}

so that we can unit-test it and be more uniform when we use it. What do you think?

token/services/storage/db/sql/common/transactions.go
```diff
@@ -459,14 +464,14 @@ func (w *AtomicWrite) AddTransaction(ctx context.Context, rs ...driver4.Transact
...
 		logger.DebugfContext(ctx, "adding transaction record [%s:%d,%s:%s:%s:%s]", r.TxID, r.ActionType, r.TokenType, r.SenderEID, r.RecipientEID, r.Amount)
-		if !r.Amount.IsInt64() {
-			return errors.New("the database driver does not support larger values than int64")
-		}
 		id, err := uuid.GenerateUUID()
 		if err != nil {
 			return errors.Wrapf(err, "error generating uuid")
 		}
-		rows[i] = common3.Tuple{id, r.TxID, int(r.ActionType), r.SenderEID, r.RecipientEID, r.TokenType, r.Amount.Int64(), r.Timestamp.UTC()}
+		if r.Amount.BitLen() > 255 {
```
adecaro: please, introduce a constant for 255 with an explanation connected to `NUMERIC(78, 0)`. Double check that 255 is right number

token/services/storage/db/sql/common/transactions.go
```diff
@@ -518,15 +523,11 @@ func (w *AtomicWrite) AddMovement(ctx context.Context, rs ...driver4.MovementRec
 	rows := make([]common3.Tuple, len(rs))
 	for i, r := range rs {
 		logger.DebugfContext(ctx, "adding movement record [%s]", r)
-
-		if !r.Amount.IsInt64() {
```
adecaro, line 522: check the bit size here as well.

token/services/storage/db/sql/common/bigint.go
```diff
@@ -0,0 +1,52 @@
...
+package common
+
+import (
+	"fmt"
+	"math/big"
+)
+
+// BigInt is a custom wrapper around math/big.Int that implements
+// sql.Scanner for reading NUMERIC values from the DB
+type BigInt struct {
```
adecaro, line 16: please, add unit-tests for this

token/services/storage/db/sql/common/transactions.go
```diff
@@ -116,12 +116,17 @@ func (db *TransactionStore) QueryMovements(ctx context.Context, params driver4.Q
 	}
 
 	it := common.NewIterator(rows, func(r *driver4.MovementRecord) error {
-		var amount int64
-		if err := rows.Scan(&r.TxID, &r.EnrollmentID, &r.TokenType, &amount, &r.Status); err != nil {
+		var amountRaw interface{}
```
adecaro: please, use the new BigInt

adecaro (changes requested):
A few more changes and we are good. Thanks a lot :)

sid200727:
Hi @adecaro, addressed all your latest review comments:
1. Added maxAmountBits constant with explanation connected to NUMERIC(78, 0)
2. Added BitLen check in AddMovement
3. Replaced amountRaw with BigInt in MovementRecord scan
4. Added 6 unit tests for BigInt
5. Fixed gofmt formatting
Please take a look when you get a chance!

token/services/storage/db/sql/common/transactions.go
```diff
@@ -116,12 +120,12 @@ func (db *TransactionStore) QueryMovements(ctx context.Context, params driver4.Q
...
 	it := common.NewIterator(rows, func(r *driver4.MovementRecord) error {
-		var amount int64
+		var amount BigInt
 		if err := rows.Scan(&r.TxID, &r.EnrollmentID, &r.TokenType, &amount, &r.Status); err != nil {
 			return err
 		}
-		r.Amount = big.NewInt(amount)
-		logger.DebugfContext(ctx, "movement [%s:%s:%d]", r.TxID, r.Status, r.Amount)
+		r.Amount = amount.Int
+		logger.DebugfContext(ctx, "movement [%s:%s:%s]", r.TxID, r.Status, r.Amount.Int64())
```
adecaro: pass just r.Amount so we avoid the computation of Int64 if the logger is set up to not log debug

token/services/storage/db/sql/common/transactions.go
```diff
@@ -459,14 +463,14 @@ func (w *AtomicWrite) AddTransaction(ctx context.Context, rs ...driver4.Transact
...
-		if !r.Amount.IsInt64() {
-			return errors.New("the database driver does not support larger values than int64")
-		}
 		id, err := uuid.GenerateUUID()
 		if err != nil {
 			return errors.Wrapf(err, "error generating uuid")
 		}
-		rows[i] = common3.Tuple{id, r.TxID, int(r.ActionType), r.SenderEID, r.RecipientEID, r.TokenType, r.Amount.Int64(), r.Timestamp.UTC()}
+		if r.Amount.BitLen() > maxAmountBits {
+			return errors.Errorf("amount [%s] exceeds maximum supported size of %d bits", r.Amount.String(), maxAmountBits)
```
adecaro: no need to add .String(), it is invoked automatically.

sid200727:
Hi @adecaro, fixed both:
1. Passing r.Amount directly in DebugfContext instead of r.Amount.Int64()
2. Removed .String() from Errorf calls — invoked automatically
Ready for final review!

PR #1440, opened 2026-03-19
 unit test: token/driver package

AkramBitar:


token/driver/request_test.go
```diff
@@ -151,3 +801,762 @@ func TestTokenRequestMetadataSerialization(t *testing.T) {
...
+						},
+					},
+				},
+			},
+		},
+	}
+
+	reqMeta := &TokenRequestMetadata{}
+	err := reqMeta.FromProtos(proto)
+	require.NoError(t, err)
```
aaadir, line 1210: can you also check that the identity "issuer1" came out fine in the conversion, in spite of the nil? (assuming that it's supposed to)

token/driver/request_test.go
```diff
@@ -151,3 +801,762 @@ func TestTokenRequestMetadataSerialization(t *testing.T) {
...
+		Senders: []*AuditableIdentity{
+			{Identity: Identity("sender1"), AuditInfo: []byte("audit1")},
+			{Identity: Identity("sender2"), AuditInfo: []byte("audit2")},
+		},
+	}
+
+	proto, err := tim.ToProtos()
+	require.NoError(t, err)
+	assert.NotNil(t, proto)
+	assert.Len(t, proto.Senders, 2)
```
aaadir, line 1272: can you also check that the "sender1/2" and "audit1/2" were correctly converted (assuming they should be).
similarly in the following tests.

you may ask Bob that in tests where no error is expected, beyond checking no-error, not nil, len, etc. he also adds a chect to verify that the result value is as expected because the specific expected value is easily known or available in the test (in cases where it is known or available).

PR #1448, opened 2026-03-25
Compressed Sigma Protocol-Based Range Proof

adecaro:
This PR introduces a new type of range proof based on Compressed Sigma Protocols.
It should be faster on validation and have similar performance for proof generation.

### DIFF
--- a/token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof/ipa.go
+++ b/token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof/ipa.go
@@ -12,6 +12,7 @@ import (
 	"github.com/hyperledger-labs/fabric-token-sdk/token/core/common/encoding/asn1"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/nogh/v1/crypto/common"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/nogh/v1/crypto/math"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/nogh/v1/crypto/rp"
 )
 
 // IPA contains the proof for the inner product argument.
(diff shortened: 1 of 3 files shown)


docs/drivers/dlogwogh.md
nitsatiisc, line 1: Perhaps mention entire group operations in form of batched MSMs for faster verification as the primary reason.

docs/drivers/dlogwogh.md
```diff
@@ -256,6 +283,23 @@ type RangeProofParams struct {
...
 
+#### 4.2.2 CSP Range Proof Parameters
+
+```go
+type CSPRangeProofParams struct {
+    LeftGenerators  []*mathlib.G1  // Length = BitLength + 1
+    RightGenerators []*mathlib.G1  // Length = BitLength + 1
+    BitLength       uint64         // 16, 32, or 64
+}
+```
```
nitsatiisc, line 294: Clarification: Ideally the bitLength of 30/62 should be chosen for CSP, which makes 2n+4 a power of 2, i.e, 64 and 128 respectively.  Since CSP inner product argument is over 2n+4 sized vector, this avoids performance hit from overflowing to next power of two.  Sacrificing 2 bits from the range is perhaps worth the performance.

token/core/zkatdlog/nogh/v1/crypto/rp/csp/rp.go
```diff
@@ -0,0 +1,793 @@
...
+	if err != nil {
+		return nil, errors.New("Unable to obtain challenge gamma")
+	}
+
+	// Extended commitment: pCommExt = pComm + eta * VCommitment.
+	pCommExt := pComm.Copy()
+	pCommExt.Add(rp.VCommitment.Mul(eta))
+
+	// Extended witness pExt = aCoeffs || bCoeffs || v || r
+	// over generators  gExt = AGenerators || BGenerators || VGenerators.
```
nitsatiisc: The comment is slightly wrong: it should be 
gExt = AGenerators || BGenerators || \eta * VGenerators.

docs/drivers/dlogwogh.md
adecaro: Very relevant point, I'll update comment and documentation about this. 
One has to find the right trade off between requirements and performance depending on the use case, I guess.

docs/drivers/dlogwogh.md
adecaro: yes, we have a section for that.

PR #1452, opened 2026-03-28
FSC Update #1185

adecaro:
Integrate https://github.com/hyperledger-labs/fabric-smart-client/pull/1185 that has been merged.

### DIFF
--- a/integration/token/interop/views/htlc/claim.go
+++ b/integration/token/interop/views/htlc/claim.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package htlc
 
 import (
+	"bytes"
 	"encoding/json"
 	"time"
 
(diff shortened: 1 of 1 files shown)


integration/token/interop/views/htlc/claim.go
```diff
@@ -40,7 +40,7 @@ type ClaimView struct {
 	*Claim
 }
 
-func (r *ClaimView) Call(context view.Context) (res interface{}, err error) {
+func (r *ClaimView) Call(ctx view.Context) (res interface{}, err error) {
```
mbrandenburger, line 44: based on previous discussions we had - I would recommend calling this variable `viewCtx` to not get confused with `ctx context.Context`

integration/token/interop/views/htlc/exchange.go
```diff
@@ -160,25 +166,25 @@ func (f *FastExchangeInitiatorViewFactory) NewView(in []byte) (view.View, error)
 
 type FastExchangeResponderView struct{}
 
-func (v *FastExchangeResponderView) Call(context view.Context) (interface{}, error) {
+func (v *FastExchangeResponderView) Call(ctx view.Context) (interface{}, error) {
```
mbrandenburger, line 169: Same comment as above

integration/token/interop/views/htlc/claim.go
adecaro: yes, I had to change this to avoid the name clash

PR #1459, opened 2026-03-31
feat(metrics): add duration and outcome metrics to sherdlock selector

atharrva01:
**Add selection duration, outcome, and retry metrics to the sherdlock token selector**

Token selection sits on the critical path of every transaction, but when I looked at the sherdlock selector, it had almost no observability, just a single invocation counter. Operators had no way to tell how long selection was taking, whether failures were due to locked funds or insufficient funds, or how often retries were happening.

I filled that gap by adding three new metrics:

- **`selection_duration_seconds`** : how long each `Select()` call takes end-to-end
- **`selection_outcome_total`** : counts outcomes by label: `success`, `insufficient_funds`, `locked_funds`, `error`
- **`selection_immediate_retries`** : distribution of retry counts per call

To avoid double-counting when `stubbornSelector` wraps `selector`, I split `Select()` into `selectInternal()` and `selectWithoutMetrics()`. The stubborn selector records its own duration (which includes backoff sleep), giving operators two views: per-attempt latency and total wall-clock latency.

I threaded the `*Metrics` object through `NewManager` → `NewSherdSelector` → the selectors themselves. On the test side, I updated all `NewManager` call sites to pass `newMetrics(&disabled.Provider{})`, consistent with how fetcher metrics are already handled.

This follows the same pattern I used in #1453 and #1451 for ttx and the finality queue, just never got around to extending it to the selector until now.

### DIFF
--- a/token/services/selector/sherdlock/selector.go
+++ b/token/services/selector/sherdlock/selector.go
@@ -65,7 +65,7 @@ type stubbornSelector struct {
 func (m *stubbornSelector) Select(ctx context.Context, ownerFilter token.OwnerFilter, q string, tokenType token2.Type) ([]*token2.ID, token2.Quantity, error) {
 	start := time.Now()
 	for retriesAfterBackoff := 0; retriesAfterBackoff <= m.maxRetriesAfterBackoff; retriesAfterBackoff++ {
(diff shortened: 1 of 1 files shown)


atharrva01:
hi @adecaro I added duration, outcome, and retry metrics to the sherdlock selector to improve observability on the critical transaction path and make latency and contention easier to diagnose.

PR #1463, opened 2026-04-02
network service: acl support during endorsement #1462

adecaro:
This PR introduces an additional check for the RequestApprovalResponderView to make sure who is proposing satisfies the channel ACL.

This PR depends on https://github.com/hyperledger-labs/fabric-smart-client/pull/1221

### DIFF
--- a/token/token/quantity.go
+++ b/token/token/quantity.go
@@ -15,15 +15,16 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/collections/iterators"
 )
 
-// Quantity models an immutable token quantity and its basic operations.
+// Quantity models a token quantity and its basic operations.
+// Note: Operations mutate the receiver and are NOT thread-safe.
+// Do not share Quantity instances across goroutines without synchronization.
 type Quantity interface {
-
-	// Add returns this + b modifying this.
-	// If an overflow occurs, it returns an error.
+	// Add returns this + b, modifying this in place.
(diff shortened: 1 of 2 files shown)


token/services/network/fabric/endorsement/provider.go
```diff
@@ -141,6 +142,25 @@ func (c *ChannelProvider) GetMSPManager(network, channel string) (fsc.MSPManager
...
+	ch, err := fns.Channel(channel)
+	if err != nil {
+		return nil, errors.WithMessagef(err, "failed to get channel [%s] for network [%s]", channel, network)
+	}
+
+	if fns.ConfigService().DriverName() == fabricx.DriverName {
+		return ch.ACLProvider(), nil
+	}
+
+	// fabric does not support yet ACL check
```
AkramBitar, line 160: Minor issue. It is worth to have a log message that "fabric does not support yet ACL check"

token/token/quantity.go
```diff
@@ -78,34 +89,37 @@ func ToQuantitySum(precision uint64) iterators.Reducer[*UnspentToken, Quantity]
 		if err != nil {
 			return nil, err
 		}
-
-		return sum.Add(q), nil
+		// Recover from panic and convert to error
+		defer func() {
+			if r := recover(); r != nil {
+				err = errors.Errorf("overflow in quantity sum: %v", r)
```
AkramBitar, line 95: Maybe we need a log message here as well.

PR #1465, opened 2026-04-02
network service documentation update

adecaro:
This PR improves the documentation of the network service highlighting also a path for integration of EVM-based blockchains.

### DIFF
--- a/docs/public_parameters.md
+++ b/docs/public_parameters.md
@@ -37,7 +37,7 @@ In a standard Hyperledger Fabric network, public parameters are managed by the *
 
 ### FabricX
 In FabricX, public parameters are managed through a dedicated **Public Parameters Service** and the **Vault**.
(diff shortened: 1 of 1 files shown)


docs/services/network-ethereum.md
```diff
@@ -0,0 +1,488 @@
...
+| **Flexibility** | Limited by EVM constraints | High (validation in Go) |
+| **Endorsement** | Not required | Required from FSC endorsers |
+| **Best For** | Simple deployments, public tokens | Complex validation, privacy needs |
+
+## Driver Interface Implementation
+
+Both approaches must implement the [`driver.Network`](../../token/services/network/driver/network.go) interface:
+
+```go
+type EthereumNetwork struct {
```
AkramBitar, line 132: LGTM. One minor issue.

One concern, if the interface is changed then we need to remember to change it here. Maybe if we have this interface in the code, then I think it is better to add a link to it.

Same concern for other code in the docs that we might replace with a link.

PR #1468, opened 2026-04-03
feat(storage): Introduce a recovery service

adecaro:
This PR introduces a robust transaction recovery mechanism for the Fabric Token SDK that automatically re-registers finality listeners for pending transactions that may have lost their listeners due to node restarts, network interruptions, or other failures.

Problem Solved
Previously, when a node restarted or experienced network interruptions, pending transactions could lose their finality listeners, leaving them in an indefinite pending state without any mechanism to detect their final status (confirmed/invalid). This created operational issues where transactions would appear stuck, requiring manual intervention to resolve.

### DIFF
--- a/docs/services/recovery.md
+++ b/docs/services/recovery.md
@@ -1,0 +1,157 @@
+# Transaction Recovery Service
+
+The **Transaction Recovery Service** provides automatic re-registration of finality listeners for pending transactions that may have lost their listeners due to node restarts, network interruptions, or other failures. This ensures that transactions eventually reach finality even after system disruptions.
+
+## Architecture
+
+The recovery system consists of three main components:
+
+1. **Manager**: Orchestrates the recovery process with periodic scanning and distributed coordination
+2. **Handler**: Implements the actual recovery logic for individual transactions
+3. **Storage**: Provides database operations for claiming and tracking recovery state
+
+## Components
+
+### Recovery Manager
+
+The Manager runs in the background and periodically scans for pending transactions that are eligible for recovery. It uses distributed locking (PostgreSQL advisory locks) to ensure only one replica in a multi-instance deployment performs recovery at a time.
+
+**Key features:**
+- Configurable scan intervals and batch sizes
+- Worker pool for parallel transaction processing
+- Lease-based claim mechanism to prevent duplicate work
+- Graceful shutdown with proper cleanup
+
+### Recovery Handler
+
+The Handler interface defines how individual transactions are recovered. The TTX service provides a concrete implementation (`TTXRecoveryHandler`) that:
+- Queries transaction status from the network
+- Applies finality logic (Valid/Invalid/Busy)
+- Updates local database state
+- Handles hash verification and token request processing
+
+### Storage Interface
+
+The Storage interface abstracts database operations needed for recovery:
+- `AcquireRecoveryLeadership`: Obtains distributed lock for leader election
+- `ClaimPendingTransactions`: Atomically claims a batch of transactions
+- `ReleaseRecoveryClaim`: Releases claim after processing
+
+## Database Support
+
+### PostgreSQL (Recommended for Production)
+
+PostgreSQL is the recommended database for production multi-instance deployments:
+- Advisory locks provide distributed coordination
+- Atomic `UPDATE...RETURNING` ensures no duplicate claims
+- Supports horizontal scaling with multiple replicas
+- Leader election prevents conflicting recovery attempts
+
+### SQLite (Development and Single-Node)
+
+SQLite is supported for single-node deployments and development:
+- Handles node restarts gracefully
+- Simpler setup for development environments
+- Not designed for multi-replica scenarios
+- No distributed locking mechanism
+
+## Configuration
+
+Recovery behavior is controlled via configuration (see [Configuration](../configuration.md)):
+
+```yaml
+recovery:
+  enabled: true              # Enable/disable recovery
+  ttl: 30s                   # Minimum age before recovery
+  scanInterval: 5s           # How often to scan
+  batchSize: 100             # Max transactions per scan
+  workerCount: 4             # Parallel workers
+  leaseDuration: 30s         # Claim lease duration
+  advisoryLockID: 8389...    # PostgreSQL lock ID
+  instanceID: ""             # Instance identifier
+```
+
+## Usage Example
+
+Creating a recovery manager:
+
+```go
+config := recovery.Config{
+    Enabled:        true,
+    TTL:            30 * time.Second,
+    ScanInterval:   5 * time.Second,
+    BatchSize:      100,
+    WorkerCount:    4,
+    LeaseDuration:  30 * time.Second,
+    AdvisoryLockID: 8389190333894887286,
+}
+
+manager := recovery.NewManager(
+    logger,
+    storage,  // Implements Storage interface
+    handler,  // Implements Handler interface
+    config,
+)
+
+// Start recovery
+if err := manager.Start(); err != nil {
+    return err
+}
+defer manager.Stop()
+```
+
+## Implementing a Custom Handler
+
+To implement a custom recovery handler:
+
+```go
+type MyHandler struct {
+    // your dependencies
+}
+
+func (h *MyHandler) Recover(ctx context.Context, txID string) error {
+    // 1. Query transaction status from your backend
+    // 2. Apply finality logic based on status
+    // 3. Update local database state
+    // 4. Return nil on success, error on failure
+    return nil
+}
+```
+
+## Recovery Process Flow
+
+1. Manager acquires leadership (PostgreSQL advisory lock)
+2. Manager queries for pending transactions older than TTL
+3. Manager atomically claims a batch of transactions
+4. Manager distributes claimed transactions to worker pool
+5. Each worker calls `Handler.Recover()` for its transactions
+6. Handler queries network and applies finality logic
+7. Manager releases claims with success/failure message
+8. Process repeats on next scan interval
+
+## Error Handling
+
+- **Transient errors** (Busy status): Released gracefully, retried on next scan
+- **Permanent errors** (Invalid tx): Marked as Deleted in database
+- **Handler errors**: Logged individually, claim released with error message
+- **Network errors**: Propagated to caller, claim released for retry
+
+## Performance Tuning
+
+### For High-Throughput Environments
+- Increase `batchSize` (200-500)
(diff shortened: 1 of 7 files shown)


docs/services/recovery.md
```diff
@@ -0,0 +1,157 @@
...
+    // 3. Update local database state
+    // 4. Return nil on success, error on failure
+    return nil
+}
+```
+
+## Recovery Process Flow
+
+1. Manager acquires leadership (PostgreSQL advisory lock)
+2. Manager queries for pending transactions older than TTL
```
AkramBitar, line 124: If the node crashed and returned less than the TTL then do these transaction still in pending until the will be transactions older than TTL?

docs/services/recovery.md
```diff
@@ -0,0 +1,157 @@
...
+
+1. **Manager**: Orchestrates the recovery process with periodic scanning and distributed coordination
+2. **Handler**: Implements the actual recovery logic for individual transactions
+3. **Storage**: Provides database operations for claiming and tracking recovery state
+
+## Components
+
+### Recovery Manager
+
+The Manager runs in the background and periodically scans for pending transactions that are eligible for recovery. It uses distributed locking (PostgreSQL advisory locks) to ensure only one replica in a multi-instance deployment performs recovery at a time.
```
AkramBitar, line 17: Does this lock will be taken until the recovery finished of the chosen replica will be finished ? Does this lock prevent also new transactions to be inserted? or obtains distributed lock for leader election only?

docs/services/recovery.md
```diff
@@ -0,0 +1,157 @@
...
+The Handler interface defines how individual transactions are recovered. The TTX service provides a concrete implementation (`TTXRecoveryHandler`) that:
+- Queries transaction status from the network
+- Applies finality logic (Valid/Invalid/Busy)
+- Updates local database state
+- Handles hash verification and token request processing
+
+### Storage Interface
+
+The Storage interface abstracts database operations needed for recovery:
+- `AcquireRecoveryLeadership`: Obtains distributed lock for leader election
```
AkramBitar, line 36: What happen if the lock is taken and before it freed the node that take it crashes? How this will be freed?

docs/services/storage.md
```diff
@@ -70,4 +70,76 @@ The `walletdb` and associated stores (IdentityDB, KeyStore) manage the cryptogra
...
+- **Configurable Behavior**: Recovery parameters can be tuned via configuration (see [Configuration](../configuration.md), Section `Optional: token.tms.<name>.services.network.fabric.recovery`)
+
+### Recovery Process
+
+The recovery service follows this workflow:
+
+1. **Scan Phase**: The recovery manager periodically scans the transaction database for pending transactions older than the configured TTL
+2. **Claim Phase**: Eligible transactions are claimed by the current instance using a lease mechanism
+3. **Recovery Phase**: For each claimed transaction, the recovery handler:
+   - Queries the network for the transaction's current status
```
AkramBitar, line 108: that is older then TTL

docs/services/storage.md
```diff
@@ -70,4 +70,76 @@ The `walletdb` and associated stores (IdentityDB, KeyStore) manage the cryptogra
...
+### Integration with Network Service
+
+The Network Service (Fabric and FabricX implementations) instantiates the recovery service during initialization:
+- Creates a recovery manager with the appropriate transaction database (TTXDB or AuditDB)
+- Configures the recovery handler with network-specific transaction status query capabilities
+- Starts the background recovery process
+
+This design ensures that recovery is handled consistently across different network backends while maintaining the separation of concerns between storage and network layers.
+
+### Database Backend Considerations
```
AkramBitar, line 124: by the way what happens for transactions that still not reached the ordering service? I suppose all under DB transaction that is not committed, and then in case of failure the database will be clean.

token/services/storage/db/sql/postgres/recovery.go
```diff
@@ -0,0 +1,116 @@
...
+
+// Close releases the advisory lock and closes the connection.
+// It is safe to call Close multiple times.
+func (l *AdvisoryLock) Close() error {
+	if l.conn == nil {
+		return nil
+	}
+
+	// Release the lock explicitly before closing the connection
+	// This is not strictly necessary as the lock auto-releases on connection close,
```
AkramBitar, line 86: What happens if we have a crash before releasing the lock?

token/services/storage/db/sql/postgres/transactions.go
```diff
@@ -7,24 +7,280 @@ SPDX-License-Identifier: Apache-2.0
...
+// This ensures only one recovery instance can claim a specific transaction.
+func (db *TransactionStore) ClaimPendingTransactions(ctx context.Context, params tokensdriver.RecoveryClaimParams) ([]*tokensdriver.TransactionRecord, error) {
+	logger.Debugf("Claiming pending transactions: owner=%s, olderThan=%s, limit=%d, lease=%s",
+		params.Owner, params.OlderThan, params.Limit, params.LeaseDuration)
+
+	// Build the atomic claim query using UPDATE...RETURNING with CTE
+	// This query:
+	// 1. Finds eligible pending transactions (old enough, not claimed or expired or owned by us)
+	// 2. Atomically updates them with our claim
+	// 3. Returns the full transaction details
```
AkramBitar, line 85: What happens if we have a lot of records? do we want to get them in chanks?

token/services/storage/ttxdb/store.go
```diff
@@ -350,6 +354,47 @@ func (d *StoreService) AppendValidationRecord(ctx context.Context, txID string,
...
+	logger.DebugfContext(ctx, "claimed %d pending transactions older than %s for owner [%s]", len(records), olderThan, owner)
+
+	return records, nil
+}
+
+// ReleaseRecoveryClaim clears the recovery claim for the passed transaction if owned by owner.
+// The message parameter is stored for audit/debugging purposes.
+func (d *StoreService) ReleaseRecoveryClaim(ctx context.Context, txID string, owner string, message string) error {
+	if err := d.db.ReleaseRecoveryClaim(ctx, txID, owner, message); err != nil {
+		return errors.Wrapf(err, "failed to release recovery claim for tx [%s]", txID)
```
AkramBitar, line 392: What happens if we have crash before this lock was released?

token/services/ttx/finality/recovery.go
```diff
@@ -0,0 +1,180 @@
...
+				return err
+			}
+		}
+
+	case network.Invalid:
+		txStatus = storage.Deleted
+		h.logger.DebugfContext(ctx, "transaction [%s] is invalid", txID)
+
+	default:
+		// Transaction is not yet finalized (Busy or Unknown status)
```
AkramBitar, line 137: This supposed to be old transaction older than TTL. 
I think in case if still Busy or Unknown status this apparently to be a problem.

docs/services/recovery.md
```diff
@@ -0,0 +1,157 @@
...
+
+## Database Support
+
+### PostgreSQL (Recommended for Production)
+
+PostgreSQL is the recommended database for production multi-instance deployments:
+- Advisory locks provide distributed coordination
+- Atomic `UPDATE...RETURNING` ensures no duplicate claims
+- Supports horizontal scaling with multiple replicas
+- Leader election prevents conflicting recovery attempts
```
AkramBitar, line 48: Do we have a test for leader election? 
Simulate multiple nodes (replicas)?

docs/services/storage.md
```diff
@@ -70,4 +70,76 @@ The `walletdb` and associated stores (IdentityDB, KeyStore) manage the cryptogra
...
+- **Multi-Database Support**:
+  - **PostgreSQL**: Recommended for production multi-instance deployments. Uses advisory locks for distributed coordination and leader election
+  - **SQLite**: Supported for single-node deployments and development. Handles node restarts gracefully but is not designed for multi-replica scenarios
+- **Configurable Behavior**: Recovery parameters can be tuned via configuration (see [Configuration](../configuration.md), Section `Optional: token.tms.<name>.services.network.fabric.recovery`)
+
+### Recovery Process
+
+The recovery service follows this workflow:
+
+1. **Scan Phase**: The recovery manager periodically scans the transaction database for pending transactions older than the configured TTL
```
AkramBitar, line 105: I’m wondering if we can simulate raising a finality event for these pending transactions, since our code already listens for such events and would then handle their recovery. Why do we need to process them separately instead of using our regular flow

docs/services/recovery.md
adecaro: yes. TTL is there to avoid situations where the transaction is still in processing.

docs/services/recovery.md
adecaro: The `PostgreSQL advisory locks` is used solely for leader election, it does not block anything else.

docs/services/recovery.md
adecaro: it is automatically freed by postgres, it is part of the semantic of the `PostgreSQL advisory locks`.

docs/services/storage.md
adecaro: what do you mean? The sentence is already grammatically sound.

docs/services/storage.md
adecaro: the transaction will never reach finality then. In these cases, one need an additional business process to decide to discard a transaction. The application must take care of that.

token/services/storage/db/sql/postgres/transactions.go
adecaro: only a batch is selected. The size of the batch is configurable.

token/services/storage/ttxdb/store.go
adecaro: if the transaction got finalized, this failure is not harming and doesn't not need to recovered.
If the transaction didn't get finalized, it will processed again.
We should be good there.

token/services/ttx/finality/recovery.go
adecaro: sure, but the listener cannot do much here. The recovery manager needs to retry. No?

docs/services/recovery.md
adecaro: the integration tests are running also in replication mode, yes.

docs/services/storage.md
adecaro: in the integration test there is such a situation. A node is killed and then we check if finality of a pending transaction is reached.

docs/services/storage.md
AkramBitar: - Queries the network for the transaction's OLDER THEN TTL current status

Issue #1432, opened 2026-03-18
Optimize ZKP Computation Using CPU SIMD and Specialized Instructions


AkramBitar:
Investigate and implement optimizations for zero-knowledge proof computations by leveraging CPU-specific instructions. Goal is to improve performance of finite field operations, FFTs, and elliptic curve calculations.

SurbhiAgarwal1:
Hi @AkramBitar ,
Regarding the ZKP optimization, here’s the plan we can follow:

1. First, we need to benchmark the current `zkatdlog` and `idemix` drivers to find the exact hotspots. Most likely, it's modular multiplication in the field arithmetic.

2. We should introduce assembly backends (`.s` files) for the core field ops. Leveraging ADX instructions (MULX, ADCX/ADOX) on amd64 will give us a massive boost over pure Go or math/big.

3. For MSM, if we aren't already using Pippenger’s algorithm, that should be the priority. We can vectorize the bucket accumulation using AVX2 to handle coordinates in parallel.

4. Instead of reinventing everything, we could probably look into how `gnark-crypto` or `blst` handles these curves and either port their assembly logic or pull them in as dependencies where it makes sense.

I can start with the benchmarking and then look into the ADX/SIMD implementation for the field multiplication. What do you think?

AkramBitar:
> Hi [@AkramBitar](https://github.com/AkramBitar) , Regarding the ZKP optimization, here’s the plan we can follow:
> 
> 1. First, we need to benchmark the current `zkatdlog` and `idemix` drivers to find the exact hotspots. Most likely, it's modular multiplication in the field arithmetic.
> 2. We should introduce assembly backends (`.s` files) for the core field ops. Leveraging ADX instructions (MULX, ADCX/ADOX) on amd64 will give us a massive boost over pure Go or math/big.
> 3. For MSM, if we aren't already using Pippenger’s algorithm, that should be the priority. We can vectorize the bucket accumulation using AVX2 to handle coordinates in parallel.
> 4. Instead of reinventing everything, we could probably look into how `gnark-crypto` or `blst` handles these curves and either port their assembly logic or pull them in as dependencies where it makes sense.
> 
> I can start with the benchmarking and then look into the ADX/SIMD implementation for the field multiplication. What do you think?

@SurbhiAgarwal1 I will be very happy if you can help with this one. Thanks a million.

Please let me know if you need any additional clarification or support.

[...]

PR #1470, opened 2026-04-04
perf(zkatdlog): parallelise IPA generator reduction and proof generation, batch verifyIPA MSM

neetance:
Addresses #1432.

### 1. `reduceGenerators()` -> parallel goroutines (`ipa.go`)

The serial loop computing `l` output generator pairs per IPA round is replaced with concurrent goroutines, one per index. For a 64-bit range proof this eliminates 252 sequential scalar multiplications across 6 rounds (32+16+8+4+2+1 iterations × 2 `Mul2InPlace` calls each) and computes all outputs in parallel. There are no data races: each goroutine writes to a distinct slice index.

### 2. `verifyIPA()` -> single `MultiScalarMul` + `BatchInverse` (`bulletproof.go`)

The serial loop that reconstructed the commitment basis called `.Mul()` 3×64 = 192 times sequentially and called `InvModOrder()` 64 times individually. This is now:  One `BatchInverse` call for the 64 `yInv` values, 1 inversion + O(n) multiplications instead of 64.
One `MultiScalarMul` call over all 2n+3 terms instead of 192 serial `.Mul()` calls

### 3. `RangeCorrectnessProver.Prove()` + `RangeCorrectnessVerifier.Verify()` -> parallel (`rangecorrectness.go`)

Range proofs for independent output tokens are generated and verified concurrently using `sync.WaitGroup`. Shared parameters are read-only.

### IPA Prover

    goos: linux
    goarch: amd64
    pkg: github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/nogh/v1/crypto/rp
    cpu: 13th Gen Intel(R) Core(TM) i5-13450HX
                       │ /tmp/before_ipa.txt │         /tmp/after_ipa.txt          │
                       │       sec/op        │   sec/op     vs base                │
    IPAProver/bench-16          33.70m ± 26%   18.41m ± 7%  -45.38% (p=0.000 n=20)

                       │ /tmp/before_ipa.txt │         /tmp/after_ipa.txt          │
                       │        B/op         │     B/op      vs base               │
    IPAProver/bench-16          538.5Ki ± 0%   544.9Ki ± 0%  +1.17% (p=0.000 n=20)

                       │ /tmp/before_ipa.txt │         /tmp/after_ipa.txt         │
                       │      allocs/op      │  allocs/op   vs base               │
    IPAProver/bench-16           2.751k ± 0%   2.833k ± 0%  +2.98% (p=0.000 n=20)

- ~45% reduction in execution time
- Major variance reduction (±26% → ±7%)

### BF Prover

    goos: linux
    goarch: amd64
    pkg: github.com/hyperledger-labs/fabric-token-sdk/token/core/zkatdlog/nogh/v1/crypto/rp
    cpu: 13th Gen Intel(R) Core(TM) i5-13450HX
                      │ /tmp/before_bf.txt │          /tmp/after_bf.txt          │
                      │       sec/op       │   sec/op     vs base                │
    BFProver/bench-16         46.90m ± 29%   32.81m ± 4%  -30.05% (p=0.000 n=20)

                      │ /tmp/before_bf.txt │          /tmp/after_bf.txt          │
                      │        B/op        │     B/op      vs base               │
    BFProver/bench-16         902.0Ki ± 0%   908.5Ki ± 0%  +0.72% (p=0.000 n=20)

                      │ /tmp/before_bf.txt │         /tmp/after_bf.txt          │
                      │     allocs/op      │  allocs/op   vs base               │
    BFProver/bench-16          6.263k ± 0%   6.346k ± 0%  +1.33% (p=0.000 n=20)

- ~30% reduction in execution time
- Major variance reduction (±29% → ±4%)

## Testing

-   All existing tests pass: `go test -race ./token/core/zkatdlog/...`
-   Zero data races detected with `-race`
-   No changes to proof format, wire protocol, or public API
-   `TestBFProofVerify`, `TestIPAProofVerify`, `TestRangeCorrectness`
    all pass, confirming cryptographic correctness is preserved

### DIFF
--- a/token/core/zkatdlog/nogh/v1/crypto/rp/executor.go
+++ b/token/core/zkatdlog/nogh/v1/crypto/rp/executor.go
@@ -6,94 +6,34 @@ SPDX-License-Identifier: Apache-2.0
 
 package rp
 
-import (
-	"runtime"
-	"sync"
-)
-
 // Executor defines a minimal interface for submitting tasks and waiting for their completion.
 type Executor interface {
 	Submit(func())
 	Wait()
 }
 
-// SerialExecutor executes tasks immediately in the caller goroutine.
-// This provides the lowest latency and zero scheduling overhead.
-type SerialExecutor struct{}
+// SerialExecutor executes tasks sequentially when Wait is invoked.
+type SerialExecutor struct {
+	tasks []func()
+}
 
+// NewSerialExecutor creates a new serial executor.
+func NewSerialExecutor() *SerialExecutor {
+	return &SerialExecutor{
+		tasks: make([]func(), 0),
+	}
+}
+
+// Submit stores the task for later execution.
 func (s *SerialExecutor) Submit(task func()) {
-	task()
+	s.tasks = append(s.tasks, task)
 }
 
-func (s *SerialExecutor) Wait() {}
-
-// WorkerPoolExecutor executes tasks using a fixed-size worker pool.
-// It bounds concurrency and avoids spawning unbounded goroutines.
-type WorkerPoolExecutor struct {
-	tasks chan func()
-	wg    sync.WaitGroup
-}
-
-// NewWorkerPoolExecutor creates a worker pool with numWorkers workers.
-func NewWorkerPoolExecutor(numWorkers int) *WorkerPoolExecutor {
-	e := &WorkerPoolExecutor{
-		tasks: make(chan func()),
+// Wait executes all submitted tasks sequentially.
+func (s *SerialExecutor) Wait() {
+	for _, task := range s.tasks {
+		task()
 	}
-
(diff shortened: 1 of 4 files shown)


adecaro:
Hi @neetance , last comment to get this PR done.

I would propose the following:
- Let's introduce in this PR only the executor that serially executes the tasks (make sure no locks are used here, run the tasks directly when `Wait` is invoked). This way, we guarantee the same behavior as before. Please, add also unit-tests for the executors.
- In another PR, we will make sure the executor can be injected when the token-driver is instantiated so that the executor can be customized depending on the specific use-case. 

What do you think?

Many many thanks for this great effort 🙏

Issue #1467, opened 2026-04-03
chore(token/token/quantity.go): improvements


adecaro:
Here is a list of improvements to `token/token/quantity.go`.

1. Consider returning errors instead of panicking for better error handling
2. Consider making operations truly immutable (return new instances)
3. Add benchmarks to track performance over time
4. Consider adding a `Clone()` method for safe copying

Daniel-1600:
Hey @adecaro can I work on this , could you assign me

PR #1475, opened 2026-04-04
Return errors instead of panicking and make make operations immutable

theycallmeaabie:
## Summary                                                                                                                                    
                                                                                                                                              
Changes `Add` and `Sub` on the `Quantity` interface to return `(Quantity, error)` instead of panicking, and makes them return new instances instead of mutating the receiver.                                                                                                             
                                                                                                                                              
This is part of the improvements listed in #1467.                                                                                          

## What changed
                                                                                                                                              
- `Add(b Quantity) Quantity` → `Add(b Quantity) (Quantity, error)`, same for `Sub`                                                            
- Both methods now return a new instance, leaving the original unchanged                                                                      
- `Cmp` stays as-is — type mismatch there is a programming error, not a runtime one                                                           
- Updated all callers to handle the new signatures                                                                                            
                                                                                                                                              
## Testing                                                                                                                                    
                                                                                                                                              
- Ran `go vet` and `go test` (all 31 tests passed)                                                                                            
- Added immutability tests to make sure the receiver isn't mutated after `Add`/`Sub`                                                          
- Converted the old panic-based tests to check for errors instead

### DIFF
--- a/token/services/selector/simple/selector.go
+++ b/token/services/selector/simple/selector.go
@@ -120,6 +120,7 @@ func (s *selector) selectByID(ctx context.Context, ownerFilter token.OwnerFilter
 				if addErr != nil {
 					s.locker.UnlockIDs(ctx, toBeSpent...)
 					s.locker.UnlockIDs(ctx, toBeCertified...)
+
 					return nil, nil, errors.Wrap(addErr, "failed to add locked quantity")
 				}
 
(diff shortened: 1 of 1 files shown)


theycallmeaabie:
Thanks for the review @adecaro!  I've resolved the conflicts and updated the branch. Let me know if anything else is needed.

PR #1483, opened 2026-04-06
added checks local lock state before DB insert in shared token selector

### LINKED ISSUE #886: shared token selector: first check locally
When a replica tries to lock a token, it might try to lock a token that it has already locked. Instead of trying to insert optimistically in the DB and getting an error, it could first make a local check.

Horiodino:
This PR fixes #886 by adding a local lock check before attempting DB lock insertion in the shared token selector.
The selector now tracks tokens already locked by the current replica and skips duplicate `TryLock` calls for the same token within the same selection cycle. The local lock state is cleared during `UnlockAll()`.
This avoids unnecessary DB round trips and optimistic insert failures when duplicate tokens are returned by the iterator.

### DIFF
--- a/token/services/selector/sherdlock/manager.go
+++ b/token/services/selector/sherdlock/manager.go
@@ -37,6 +37,7 @@ type tokenSelectorUnlocker interface {
 type manager struct {
 	selectorCache          lazy2.Provider[transaction.ID, tokenSelectorUnlocker]
 	locker                 Locker
+	localLocks             *localTokenLockTracker
 	leaseExpiry            time.Duration
 	leaseCleanupTickPeriod time.Duration
 	metrics                *Metrics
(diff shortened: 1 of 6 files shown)


adecaro:
Hi @Horiodino , did you have time to look at my comment?
Thanks 🙏

Horiodino:
Hey , i have moved the  implementation from a per-selector cache to a shared local lock tracker,different selectors are aware of tokens already locked within the same replica, could you review? , i would fix the conflicts asap if it looks good.

token/services/selector/sherdlock/manager.go
adecaro, line 39: why don't we pass an instance of locker that embeds the behavior of your localTokenLockTracker directly? This would reduce the code impact.

adecaro:
Hi @Horiodino , I left an initial comment. But, I also need you to run the benchmarks in `token/services/selector/benchmark_test.go` to prove that this feature does not make it worse the performance of the selector which sits on the critical path of a transaction execution. 
Please, make sure we can have a very efficient implementation of this local check.

Many thanks for your effort 🙏

token/services/selector/sherdlock/manager.go
Horiodino: good point,i  kept them separate because the Locker represents the shared/external locking mechanism, while the localTokenLockTracker is strictly in-memory .

keeeping that split avoids coupling the Locker to local state and makes the behavior explicit, since not all Locker implementations are necessarily local-aware.

adecaro: fair enough 👍

Horiodino:
from the benchmark comparison I ran (current branch vs baseline):

sherdlock slowdown (time)
SelectorSingle/sherdlock: +56.24%
SelectorParallel/sherdlock: +881.99%
SelectorSingle/sherdlock+lock: +46.17%
SelectorParallel/sherdlock+lock: +22.32%
SelectorParallel/sherdlock+lock+parallelism: +47.79%
Memory impact (sherdlock paths)
B/op roughly +24% to +44% (depending on case)
allocs/op about +0.6% to +1.1%


it checks a local in-memory lock map,then also checks shared local lock state,this happens in a  loop.
currently im trying to optimize it .

adecaro:
Hi @Horiodino , any progress on this PR? 
Thanks for the effort so far 🙏

adecaro:
Hi @Horiodino , any update on this? Thanks much 🙏

adecaro:
@atharrva01 , given that you are also working on the token-selector, could you check this PR for performance? Thanks much 🙏

Horiodino:
Hey, sorry for the delay. I’m currently busy with my university exams, but I’ll share an update this weekend.

AkramBitar:
> Hey, sorry for the delay. I’m currently busy with my university exams, but I’ll share an update this weekend.

Hello @Horiodino,

Did you get a chance to look it over?

Regards,
Akram

adecaro:
I think we lost momentum here. I'll close it, for now, to be resurrected later. @Horiodino , thanks a lot for your effort 🙏

Issue #1479, opened 2026-04-05
unit-test: token/services/interop/htlc


theycallmeaabie:
Add unit tests for the htlc package under token/services/interop.                                                                             
                                                                                                                                             
This package handles hash-time-lock contracts for atomic swaps but currently has no tests. Starting with script.go since it has well-isolated logic (hash validation, image computation, script validation, ownership checks) that can be tested without heavy mocking.

PR #1519, opened 2026-04-09
test(interop/encoding): add unit tests for encoding package

sid200727:
Part of #1491

Adds unit tests for `token/services/interop/encoding/encoding.go` covering previously untested paths:

- `Encoding.String()` — None, Base64, Hex and unknown values
- `Encoding.Available()` — valid and invalid encodings  
- `RegisterEncoding()` — error case for unknown encoding
- `Encoding.EncodingFunc()` — return value verification
- `Encoding.New()` — returns nil for unavailable encoding

Note: `token/services/interop/htlc` is being covered separately by @theycallmeaabie in #1479.

All tests pass locally


adecaro (approved):
LGTM. Thanks @sid200727 🙏

PR #1522, opened 2026-04-10
fix: plug five silent asset-safety bugs across selector, storage, and htlc

Aman-Cool:
Been going through some edge cases in the token pipeline and found a few things worth fixing.

**Selector lock leak**: `selectInternal` was returning `SelectorInsufficientFunds` on the genuine no-funds path without calling `UnlockAll` first. Any tokens locked during that selection attempt just stay locked until something else cleans them up, which it won't.

**HTLC wall-clock non-determinism**: `Verifier.Verify` was calling `time.Now()` directly to evaluate the deadline. Endorsing peers with any clock skew can disagree on whether the deadline has passed, causing endorsement failures near the boundary. Made the clock injectable via a `ClockFunc` field so callers with access to the block timestamp can pass it in, defaulting to `time.Now()` when not set.

**IdentityDB namespace collision**: `StoreIdentityData`, `GetAuditInfo`, `GetTokenInfo`, `StoreSignerInfo`, `GetExistingSignerInfo`, and `GetSignerInfo` were all building KVS composite keys without including the `TMSID`. In any deployment with multiple TMS instances sharing a single KVS backend, identity and signer data silently collides across TMS instances. Added the `TMSID` segment to every key.

All three are reachable through normal SDK usage, not just edge case theory.

### DIFF
--- a/token/services/interop/htlc/script_test.go
+++ b/token/services/interop/htlc/script_test.go
@@ -210,7 +210,7 @@ func TestScriptAuthAmIAnAuditor(t *testing.T) {
 
 func TestScriptAuthIssued(t *testing.T) {
 	sa := htlc.NewScriptAuth(nil)
-	require.False(t, sa.Issued(context.Background(), nil, &token3.Token{}))
+	require.False(t, sa.Issued(t.Context(), nil, &token3.Token{}))
 }
 
 func TestScriptAuthOwnerType(t *testing.T) {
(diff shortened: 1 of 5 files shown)


Aman-Cool:
Hey @adecaro, Found these while tracing the commit pipeline for the first PR. Figured I'd fix them in one shot since they're all in the same area.

The auditor `RestoreTMS` one is probably the most urgent; it's the same class of crash-recovery bug as the previous PR but on the audit side, just nobody noticed because the ttx side gets more attention. Silent permanent `Pending` record, no alert, no recovery.

The `ON CONFLICT DO NOTHING` is embarrassingly simple but without it any transient error that triggers a view retry just permanently kills the operation. Surprised it hasn't surfaced more in testing.

Rest are smaller but real; the lock leak compounds over time, the HTLC clock thing is a latent endorsement bomb near deadlines, and the identity key scoping only bites in multi-TMS setups which are increasingly common in production.

Would appreciate a second pair of eyes especially on the auditor recovery path and the SQL idempotency change.

adecaro:
Hi @Aman-Cool , great effort indeed. 

A few comments, I would remove from this PR the changes to manager.go and transactions.go. 
I think that issue needs to be addressed more globally. For instance, `RestoreTMS` is only invoked at bootstrap but we might have to recover transactions as the system is running to due a replica crash. Regarding the split-brain issue. I think we must make that transactional to be more consistent. 

The rest is totally fine. If it is okay for you, I'll merge the rest once you remove the changes I pointed to.
What do you think?

Aman-Cool:
@adecaro, Thanks a lot for the thorough review, really appreciate it; you're absolutely right on both points and I should've caught that myself before opening this.

Pulled out the `auditor/manager.go` and `transactions.go` changes. The three remaining fixes (selector lock leak, HTLC clock injection, `identitydb` namespace scoping) are clean and standalone so this should be ready to go now.

The auditor recovery and the split-brain atomicity fix are tracked separately in #1507 ; would love your eyes on that when you get a chance, especially since you clearly have good context on the commit pipeline.

adecaro (approved):
LGTM. Great effort. Thanks.

adecaro:
Hi @Aman-Cool , please, update the description of the PR to reflect the actual changes. Thanks a lot 🙏

adecaro:
great, thanks a lot @Aman-Cool 🙏

PR #1523, opened 2026-04-10
 fix(certifier): check wire message size before JSON deserialisation         

### LINKED ISSUE #1557: bug(audit): zkatdlog AuditorCheck validates transfer inputs against action-embedded tokens, not ledger commitments
### 👾 Description

The `AuditorCheck` in `token/core/zkatdlog/nogh/v1/auditor.go` builds `inputTokens` directly from `transfer.Inputs()` , basically from the transfer action itself , instead of fetching the actual Pedersen commitments from the ledger.

There *was* a `TokenCommitmentLoader` wired into `AuditorService`, but the call to `GetTokenOutputs` is commented out. Because of that, the validation ends up being self-referential.

What this means in practice is: someone could craft a transfer with arbitrary input commitments, and the auditor would still accept it, since it never checks against the real on-chain state.

There’s also a second issue in `VaultLedgerTokenLoader.GetTokenOutputs` (`token/core/common/loaders.go`). The map is keyed using `id.TxId` instead of `id.String()` (which includes the output index). So if a transaction has multiple outputs, they overwrite each other silently — and only the last one remains.

---

### Root cause

In `auditor.go`, the ledger lookup is effectively disabled:

```go
// tokenMap, err := s.TokenCommitmentLoader.GetTokenOutputs(ctx, tokenIDs)
// if err != nil {
//     return errors.Wrapf(err, "failed getting token outputs to perform auditor check")
// }
```

So instead of verifying against ledger data, it just trusts what’s inside the action.

And in `loaders.go`, the key issue is this:

```go
tokens[id.TxId] = ti   // should be id.String()
```

---

### What should actually happen

For every transfer input, the auditor should:

1. Fetch the real token commitment from the ledger
2. Deserialize it into a `*token.Token`
3. Compare it with the commitment provided in the action

If there’s any mismatch, `AuditorCheck` should fail.

---

### Fix direction

* Reintroduce `GetTokenOutputs` in `AuditorCheck`, using a proper key (`id.String()`)
* Add a per-input commitment comparison (action vs ledger)
* Add a test case specifically for commitment mismatch

### Bug 1 , No input commitment validation

**Where:** `auditor.go`

Auditor builds inputs from `transfer.Inputs()` (user-provided) instead of the ledger:

```go
inputTokens[i][j] = transfer.Inputs[j].Token
```

No cross-check with on-chain commitments → attacker can inject arbitrary commitments and still pass.

---

### Bug 2 , Loader key collision

**Where:** `loaders.go`

```go
tokens[id.TxId] = ti // ignores index
```

Multiple outputs from same tx overwrite each other → missing tokens, no error.

---

### Impact

Bug 1 ignores the ledger, Bug 2 breaks it anyway.
➡️ Input validation effectively becomes useless.

### 📋 Logs and Screenshots

_No response_

### 🌐 Network Environment

 v0.10.0-41-ge85b9f06a

### 🐹 Go Version

go1.25.7 linux/amd64

### 💻 Operating System

Linux (Ubuntu/Debian)

### ➕ Additional context

_No response_

atharrva01:
Closes #1557 

Follow-on to #1498 based on your @adecaro's review feedback.

The `MaxRequestBytes` check I added in #1498 ran *after* full JSON deserialization , so a 500 MiB payload would still spike the heap before getting rejected. The fix needs to happen earlier.

I can't touch the transport layer directly, but `ReceiveRaw()` gets us close enough: now we check `len(raw) > MaxWireMessageBytes` before ever calling `json.Unmarshal`, so an oversized message gets dropped before any allocations for the IDs slice or `Request` field.

To keep this unit-testable without a real FSC session, I added a `sessionFactory` field (defaults to `session.JSON`) so tests can inject a fake that returns preset raw bytes.

**Changes:**
- `config.go` , adds `MaxWireMessageBytes = MaxRequestBytes * 2`
- `service.go` , switches to `ReceiveRaw` + size guard + manual unmarshal; adds `sessionFactory`
- `service_test.go` ,  4 new tests: oversized reject, at-limit pass-through, transport errors, constant relationship

### DIFF
--- a/token/services/certifier/interactive/config.go
+++ b/token/services/certifier/interactive/config.go
@@ -37,4 +37,12 @@ const (
 	// in a CertificationRequest. Requests exceeding this limit are rejected to prevent
 	// memory exhaustion on the certifier node.
 	MaxRequestBytes = 1 << 20 // 1 MiB
+
+	// MaxWireMessageBytes is the maximum byte-length of the entire JSON-encoded
+	// certification request as received from the wire. This guard fires before
+	// JSON deserialisation so that an oversized message is dropped without ever
+	// allocating the decoded struct — preventing memory exhaustion from large
+	// payloads. It is set to 2 MiB to accommodate the base64 overhead of
+	// MaxRequestBytes plus the JSON-encoded IDs and header fields.
+	MaxWireMessageBytes = MaxRequestBytes * 2 // 2 MiB
 )
--- a/token/services/certifier/interactive/export_test.go
+++ b/token/services/certifier/interactive/export_test.go
@@ -10,6 +10,7 @@ package interactive
 
 import (
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/utils/json/session"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/token"
 )
 
@@ -42,3 +43,9 @@ func CRVCertifier(v *CertificationRequestView) view.Identity { return v.certifie
 
 // CRVIDs returns the ids field of a CertificationRequestView.
 func CRVIDs(v *CertificationRequestView) []*token.ID { return v.ids }
+
+// SetSessionFactory replaces the sessionFactory of s with f.
+// For testing only — lets external test packages inject a fake session.
+func SetSessionFactory(s *CertificationService, f func(view.Context) session.JsonSession) {
+	s.sessionFactory = f
+}
--- a/token/services/certifier/interactive/service.go
+++ b/token/services/certifier/interactive/service.go
@@ -7,7 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 package interactive
 
 import (
-	"encoding/json"
 	"fmt"
 	"sync"
 	"time"
@@ -47,7 +46,9 @@ func NewCertificationService(responderRegistry ResponderRegistry, mp metrics.Pro
 		metrics:           NewMetrics(mp),
 		backend:           backend,
 		ResponderRegistry: responderRegistry,
-		sessionFactory:    session.JSON,
+		sessionFactory: func(ctx view.Context) session.JsonSession {
+			return session.JSONWithLimit(ctx, MaxWireMessageBytes)
+		},
 	}
 }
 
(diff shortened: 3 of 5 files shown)


atharrva01:
caught oversized payloads before deserialization using `ReceiveRaw` + size guard, cc @adecaro

AkramBitar:
@atharrva01 thanks a lot for opening the PR. 
Could you please open an issue and link it to this PR? 

Also, could you please have a look at the failing tests?

adecaro:
Hi @atharrva01 , thanks for the effort. 
I'm not convinced by this one. I need we need a more systematic way to address these kind of issues without having to modify each view. For instance, a way to inject this limit in the comm stack of the smart client could be very helpful.
What do you think?

atharrva01:
hi @adecaro you are right, per-view patching doesn't scale.

In d5feb20 I've extracted the size guard into a reusable `SizeLimitedJsonSession` wrapper at `token/services/utils/json/session/limited.go`. It decorates any `JsonSession` and enforces a configurable byte limit before deserialization on every `Receive` / `ReceiveRaw` / `ReceiveWithTimeout` call. Two constructors: `session.NewSizeLimitedSession(inner, maxBytes)` for wrapping an existing session, and `session.JSONWithLimit(ctx, maxBytes)` as a drop-in for `session.JSON(ctx)`.

`CertificationService` now uses `JSONWithLimit` as its default `sessionFactory`, so `Call()` goes back to a plain `s.Receive()` the guard is fully transparent to view logic. Any other service that wants the same protection just swaps `session.JSON(ctx)` → `session.JSONWithLimit(ctx, limit)`, no per-view boilerplate needed.

This stops short of injecting into the FSC comm stack directly (that would need upstream changes), but it gives a single, reusable session-layer abstraction any view in this repo can opt into. Happy to discuss if a deeper comm-stack hook makes more sense.

adecaro:
Hi @atharrva01 , I think this check needs to be handled as earlier as possible. If we check at the application level, we risk resource exhaustion anyway.

So, I would close this PR, but I would like you to ask to check the fabric smart client (`platform/view/services/comm/host`) to double check that:
- they have a maximum message size limit, and
- this limit is configurable.

What do you think? Thanks for your effort in any case 🙏

atharrva01:
Hi @adecaro  investigated `platform/view/services/comm/host` in FSC (at the version pinned in this repo: `v0.10.2-0.20260428094934-a70a13e26c74`). Here's what I found:

### Does FSC have a max message size limit?

**Yes  both backends enforce a 10 MiB cap at the transport layer, before any application-level deserialization:**

| Backend | File | Constant | Enforcement point |
|---|---|---|---|
| libp2p | `platform/view/services/comm/io/reader.go:46` | `maxMessageSize = 10 * 1024 * 1024` | `varintReader.ReadData()`  reads the varint length header, rejects before allocating the body buffer |
| WebSocket | `platform/view/services/comm/host/websocket/ws/streamreader.go:17` | `maxDelimitedPayloadSize = 10 * 1024 * 1024` | `delimitedReader.Read()` rejects before accumulating the payload |

Both checks happen as early as possible: the varint/length header is read first, and if it exceeds 10 MiB the message is rejected immediately  no heap spike from body allocation.

### Is the limit configurable?

**No.** Both constants are hardcoded. The comm `config.go` only exposes `fsc.p2p.incomingMessagesBufferSize` and `fsc.p2p.streamReaderBufferSize`  there is no `fsc.p2p.maxMessageSize` or equivalent config key.

### Conclusion

The protection the maintainer was looking for already exists at the FSC comm layer. The only missing piece is **configurability**  that would require an upstream FSC change to expose the limit via `fsc.p2p.maxMessageSize` (or similar). This is out of scope for this repo.

Closing this PR as requested. Thanks for the guidance 🙏

PR #1525, opened 2026-04-10
bug(transfer_action): issuer identity check

### LINKED ISSUE #1524: bug(transfer_action): issuer identity check
### 👾 Description

The validity check of the transfer action of both the fabtoken and the dlog driver enforces that the issuer identity is not empty, if the transfer is a redeem. Now, if the public parameters do not contain issuers, the validator does not require any signature for a redeem. Therefore this check on the action is not needed. 

### 🛠️ Steps to reproduce

This was identified via code analysis.

### 📋 Logs and Screenshots

_No response_

### 🌐 Network Environment

main

### 🐹 Go Version

go1.25

### 💻 Operating System

macOS

### ➕ Additional context

_No response_

adecaro:
This PR removes the validity check of the issuer identity from the transfer action. This check is done by the validator only if no issuers are defined in the public parameters.

### DIFF
--- a/token/core/zkatdlog/nogh/v1/transfer/action.go
+++ b/token/core/zkatdlog/nogh/v1/transfer/action.go
@@ -338,9 +338,11 @@ func (t *Action) Validate() error {
 			return errors.Wrapf(err, "invalid output at index [%d]", i)
 		}
 	}
-	if t.IsRedeem() && (t.Issuer == nil) {
-		return ErrMissingIssuer
-	}
+	// The following check must happen only if the public parameters contain issuers.
+	// The validator enforces the signature of an issuer only in that case.
+	// if t.IsRedeem() && (t.Issuer == nil) {
+	// 	return ErrMissingIssuer
+	// }
 	if t.ProofType != rp.RangeProofType && t.ProofType != rp.CSPRangeProofType {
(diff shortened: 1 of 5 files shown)


token/core/zkatdlog/nogh/v1/transfer/action.go
```diff
@@ -338,9 +338,11 @@ func (t *Action) Validate() error {
 			return errors.Wrapf(err, "invalid output at index [%d]", i)
 		}
 	}
-	if t.IsRedeem() && (t.Issuer == nil) {
-		return ErrMissingIssuer
-	}
+	// The following check must happen only if the public parameters contain issuers.
+	// The validator enforces the signature of an issuer only in that case.
```
AkramBitar, line 342: What happens in the case the (t.Issuer == nil)?
// The validator enforces the signature of an issuer only in that case.
Do we check in this case that the issuer is not null?

AkramBitar (approved):
LGTM.

token/core/zkatdlog/nogh/v1/transfer/action.go
adecaro: yes, the validator checks that and the exiting unit-tests of the validator confirm that. Good sanity check 🙏

PR #1526, opened 2026-04-10
perf(csp): thread ExecutorProvider through csp.RangeCorrectnessProver and csp.RangeCorrectnessVerifier

neetance:
## Summary 

Extends the Executor abstraction from the bulletproof package (PR #1497) to the CSP-based range proof system in the csp package, as requested by @adecaro .

---
### Modified `rangecorrectness.go`

`RangeCorrectnessProver` and `RangeCorrectnessVerifier` now hold an `ExecutorProvider` field. Their constructors accept `provider ExecutorProvider` as a final parameter; `nil` maps to `DefaultProvider`. `Prove()` and `Verify()` call `provider.New()` to get a fresh `Executor` per call, so concurrent callers never share an `Executor` instance.

## Testing

```
go test -race ./token/core/zkatdlog/...
```
passes with  zero failures, zero races

---

Let me know if this looks good 🙏


adecaro:
Hi @neetance , great job. 
What remains now is the ability to select the runner from the benchmark tests so that we can run the full battery of test with this additional dimension. Would you like to take it on?

adecaro (approved):
LGTM

neetance:
Hi @adecaro, thanks

Yes, I’d be happy to take this on. I’ll extend the benchmark tests to allow selecting the executor (serial, unbounded, worker pool) so we can run the full test suite across different execution strategies.

I’ll open a follow-up PR with these changes soon 🙏

Issue #1160, opened 2025-07-25
feat(storage): Introduce a recovery service


adecaro:
This Issues asks to  introduce a robust transaction recovery mechanism for the Fabric Token SDK that automatically re-registers finality listeners for pending transactions that may have lost their listeners due to node restarts, network interruptions, or other failures.

Currently, when a node restarted or experienced network interruptions, pending transactions could lose their finality listeners, leaving them in an indefinite pending state without any mechanism to detect their final status (confirmed/invalid). This created operational issues where transactions would appear stuck, requiring manual intervention to resolve.

AkramBitar:
@adecaro is that will be addressed when we will integrate with FSC notification service: 
https://github.com/hyperledger-labs/fabric-smart-client/pull/1075

atharrva01:
Hi @adecaro, I'd like to work on this!

I checked out the upstream branch the Postgres LISTEN/NOTIFY infrastructure is mostly there but `TransactionNotifier` is never wired into `StatusSupport.Notify()`, so the poll in `dbFinality` is never actually eliminated. That's the missing piece I'll tackle first, then raise/remove the 1s fallback, fix the `panic("not implemented")` stubs in `network/fabric/network.go`, and add unit tests.

Could you assign this to me?

adecaro:
Hi @atharrva01 , sorry, I have just noticed your message. In working on this I came up with a better definition of the goal and achieved in here: https://github.com/hyperledger-labs/fabric-token-sdk/pull/1468.

Thanks for volunteering 🙏

PR #1528, opened 2026-04-11
fix(recovery): make audit recovery global and token requests idempotent

Aman-Cool:
Following up on the review feedback from #1522, here's the more global approach.

**Audit crash recovery**

Same split-brain scenario as the ttx side (PR #1507): node crashes after `auditDB.Append` writes the token record but before `SetStatus(Confirmed)` runs. On restart, `RestoreTMS` re-registers a finality listener for the stuck `Pending` record; but if the finality event was already delivered before the crash, it won't come again and the record stays `Pending` forever.

The fix: before adding the listener, check whether the tokens are already in `tokenDB`. If they are, set `Confirmed` directly and move on. Extracted into a standalone `recoverAuditCommittedPending` function so it's easy to call from other recovery paths as the codebase evolves.

**Idempotent token request writes**

`AddTokenRequest` was doing a plain `INSERT` with a `PRIMARY KEY` on `tx_id`. Any view retry after a transient failure would hit a `UNIQUE VIOLATION` and fail permanently instead of recovering. Added `ON CONFLICT DO NOTHING` so replaying the write is safe; the record is already there, nothing to do.

Together these close the two gaps pointed out in the review. Happy to adjust either if the direction isn't quite right.

### DIFF
--- a/token/services/auditor/manager.go
+++ b/token/services/auditor/manager.go
@@ -29,6 +29,42 @@ type TokenManagementServiceProvider interface {
 	GetManagementService(opts ...token.ServiceOption) (*token.ManagementService, error)
 }
 
+// auditTokenExistenceChecker can tell whether a given txID was already committed to tokenDB.
+type auditTokenExistenceChecker interface {
+	TransactionExists(ctx context.Context, id string) (bool, error)
+}
+
+// auditStatusSetter can promote an auditDB record from Pending to Confirmed.
+type auditStatusSetter interface {
+	SetStatus(ctx context.Context, txID string, status TxStatus, message string) error
+}
+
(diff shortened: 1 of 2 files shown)


Aman-Cool:
@adecaro, Opened this as a follow-up to the review feedback on #1522. Kept the audit recovery logic in its own function so it's not locked to the bootstrap path; easier to call from other places if needed down the line. 
The `ON CONFLICT DO NOTHING` on the token request insert handles the retry case cleanly without any extra plumbing.

Let me know what you think.

adecaro:
Hi @Aman-Cool , I think everything is addressed by #1160 .
If it is okay for you, I'll close this one.
What do you think?

PR #1530, opened 2026-04-11
fix: sherdlock selector leaks token lock on bad quantity

Aman-Cool:
Found a spot in the sherdlock selector where we lock a token but then hit a parse error and just... return. Never unlock. The token sits locked in `tokenlockdb` until the lease cleanup goroutine gets around to it (or forever if cleanup is off).

Two lines in `selectInternal`; after `TryLock` succeeds, if `ToQuantity` fails or `sum.Add` fails, we were bailing without calling `UnlockAll`. Every other error path in the same loop does the unlock correctly, these two just got missed.

Added the `UnlockAll` call before each of those returns. That's it.

### DIFF
--- a/token/services/selector/sherdlock/selector.go
+++ b/token/services/selector/sherdlock/selector.go
@@ -205,19 +205,17 @@ func (s *selector) selectInternal(ctx context.Context, owner token.OwnerFilter, 
 			s.logger.DebugfContext(ctx, "Got the lock on token [%v]", t)
 			q, err := token2.ToQuantity(t.Quantity, s.precision)
 			if err != nil {
-				if unlockErr := s.locker.UnlockAll(ctx); unlockErr != nil {
-					s.logger.Errorf("failed to unlock tokens after quantity parse error: %s", unlockErr)
-				}
-				return nil, nil, immediateRetries, errors.Wrapf(err, "invalid token [%s] found", t.Id)
+				err2 := s.locker.UnlockAll(ctx)
+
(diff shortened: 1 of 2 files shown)


Aman-Cool:
@adecaro, Found that `selectInternal` was calling `TryLock` successfully but returning early without unlocking when `ToQuantity` or `sum.Add` failed; tokens would stay locked until the background cleanup ran. Fixed both paths to call `UnlockAll` and surface the unlock result in the error, same as every other error path in the function. Also aligned the error wrapping style to match existing patterns and added a regression test that exercises the lock-then-bad-quantity path directly.

token/services/selector/sherdlock/selector.go
```diff
@@ -205,13 +205,17 @@ func (s *selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
 			s.logger.DebugfContext(ctx, "Got the lock on token [%v]", t)
 			q, err := token2.ToQuantity(t.Quantity, s.precision)
 			if err != nil {
-				return nil, nil, immediateRetries, errors.Wrapf(err, "invalid token [%s] found", t.Id)
+				err2 := s.locker.UnlockAll(ctx)
+
```
adecaro: Hi @Aman-Cool , I would add the UnlockAll directly to the caller of `selectInternal`. If `selectInternal` returns an error, the caller unlocks all. This way, we prevent any leakage.

adecaro (changes requested):
Hi @Aman-Cool , thanks for looking into this 🙏 
I left a comment. Please, address it to make the PR more robust.

adecaro (approved):
Great, thanks @Aman-Cool 🙏

PR #1531, opened 2026-04-12
cleanup(audit): wire ledger commitment check into zkatdlog AuditorCheck

### LINKED ISSUE #1544: bug(audit): zkatdlog auditor does not cross-check token commitments against ledger state
### Description

While looking into the `zkatdlog` auditor flow, I found a chain of issues that, when combined, allow a malicious transfer to pass validation undetected.

The core problem is that the auditor was validating ZK proofs against **data provided inside the transfer action**, instead of verifying against the **actual ledger state**. This effectively breaks the auditor as a security boundary.

---

### Root Cause

This wasn’t a single bug , it was a combination of three:

---

#### 1. Map key collision in token loader

**File:** `token/core/common/loaders.go`

The token loader was indexing tokens using only the transaction ID:

```go
tokens[id.TxId] = tok
```

If a transaction produced multiple outputs, they would overwrite each other in the map. So only the **last token survived**, and the rest were silently lost.

**Fix:**

```go
tokens[id.String()] = tok // "[txId:index]"
```

Now each output is uniquely tracked.

---

#### 2. Auditor wasn’t loading ledger state

**File:** `token/core/zkatdlog/nogh/v1/auditor.go`

The auditor had the correct logic to load tokens from the ledger  but it was commented out.

Instead, it was using token data directly from the transfer action:

```go
inputTokens = append(inputTokens, transfer.Inputs[j].Token)
```

This means the auditor was trusting **prover-supplied data**, not the ledger.

**Fix:**

```go
tokenMap, err := s.TokenCommitmentLoader.GetTokenOutputs(ctx, tokenIDs)
inputTokens = append(inputTokens, tokenMap[input.ID.String()])
```

Now validation is based on actual ledger state.

---

#### 3. Missing commitment consistency check

**File:** `token/core/zkatdlog/nogh/v1/audit/auditor.go`

Even after loading ledger tokens, there was no check to ensure that:

* the commitment inside the transfer action
* matches the commitment stored on the ledger

Without this, a malicious actor could craft a transfer that is internally consistent (so the ZK proof passes), but references **different tokens than the ones actually being spent**.

**Fix:**

```go
if !inputs[k][i].Data.Equals(ta.Inputs[i].Token.Data) {
    return nil, nil, errors.Errorf(
        "input[%d][%d]: action commitment does not match ledger commitment", k, i,
    )
}
```

---

### Security Impact

Because of these issues, an attacker could:

1. Generate a valid ZK proof for a high-value (or even fake) token
2. Attach it to a transfer referencing a different (lower-value) token
3. Pass auditor validation  since the ledger state was never checked

The auditor would then sign the transaction, giving **false confidence that the transfer is valid**.

---

### Affected Components

* Token loader → `token/core/common/loaders.go`
* Auditor entry point → `token/core/zkatdlog/nogh/v1/auditor.go`
* Commitment validation → `token/core/zkatdlog/nogh/v1/audit/auditor.go`

---

### 🛠️ Steps to reproduce

Create a transfer where the action’s commitment ≠ ledger commitment
Submit it to the auditor
Before fix: validation passes
After fix: validation correctly fails

A regression test has been added in auditor_test.go (input commitment mismatch case).

a6209c7ff 

### 🐹 Go Version

go1.25.7 linux/amd64 

### 💻 Operating System

Linux (Ubuntu/Debian)

atharrva01:
Closes #1544

Three bugs were making AuditorCheck silently self-validate, it was using the prover's own claimed commitments instead of cross-checking against the ledger.

I fixed the compound map key in `loaders.go` (was keying on `TxId` alone, which clobbers multi-output txns), uncommented and wired the dead `GetTokenOutputs` call in `auditor.go`, and added the missing Pedersen commitment equality check in `GetAuditInfoForTransfers` that closes the TODO that's been sitting there. Before this, a fraudulent transfer could pass the auditor check without ever touching the ledger.

All existing tests pass; added a multi-output collision test and a commitment-mismatch test.


token/core/common/loaders.go
```diff
@@ -55,7 +55,7 @@ func (s *VaultLedgerTokenLoader[T]) GetTokenOutputs(ctx context.Context, ids []*
 			if err != nil {
 				return errors.Wrapf(err, "failed deserializing token for id [%v][%s]", id, string(bytes))
 			}
-			tokens[id.TxId] = ti
+			tokens[id.String()] = ti
```
adecaro: great catch!

atharrva01:
hi @adecaro , i've made the changes which you mentioned!

AkramBitar:
Hello @atharrva01
Could you please, open a Github Issue related to this PR and link it to this PR?

atharrva01:
hi @AkramBitar , i have opened the issue , and linked the pr.

AkramBitar:
@adecaro could you please have a look on that one?

adecaro:
Hi @atharrva01 , you touched a very interesting point. 

Let me clarify the context here. The auditor we are talking about is, in the context of the token-sdk, an `active auditor`. If the transaction does not carry a valid signature of an auditor, it cannot be deemed valid. 

The role of the auditor is to check the content of the transaction against any policy the auditor must enforce. 
If the transaction is privacy-preserving, then the auditor is entitled to receive metadata to allow it to perform its job.

The auditor might not be able to decide if a transaction is valid with the respect to the content of the ledger because the auditor is not building the world state as a Fabric(X) committer is. Indeed, also the validator of a token transaction does not check the transaction against the ledger directly but build a read-write-set such that a conflict will arise if a token that does not exist is spent.

This is the reason why the auditor does not check the existence of the input tokens in the transfer actions. Of course, the auditor might end up signing a transaction that ultimately is deemed invalid by the Fabric committer.

So, I don't think we need to wire the ledger commitments. But, I think we need to rework the auditor service to streamline its implementation that right now is not robust enough.

If you want, I can create an Issue about that and if you want you can pick it up. What do you think?

atharrva01:
hi @adecaro Thanks for the context. I see now that the GetTokenOutputs call was intentionally commented out in commit  1b1aa7018. I'll split the map key fix into a separate PR. I'm interested in the separate issue you mentioned for rethinking the auditor's commitment validation approach, I'd like to contribute there.

atharrva01:
@adecaro  I've already cleaned up this PR directly, removed the GetTokenOutputs wiring, the ledger-sourced input tokens, and the Pedersen commitment cross-check. The branch now only contains the loaders.go map key fix (id.String() to prevent multi-output collision) and its tests. Ready for another look when you get a chance.

adecaro:
Hi @atharrva01 , I would go more radical and remove it. Indeed, `AuditorService` does not need anymore `TokenCommitmentLoader`. Let's remove that field so that we can remove also `VaultLedgerTokenLoader`. 
We don't need to maintain code that is not currently used 😄

atharrva01:
hi @adecaro  went ahead and removed TokenCommitmentLoader from AuditorService entirely along with VaultLedgerTokenLoader since nothing else uses it.  Cleaned up the constructor, the dead code block in AuditorCheck, and the tests. Let me know if you had something different in mind!

adecaro (approved):
Great. LGTM 👍

PR #1532, opened 2026-04-12
fix: ScanForPreImage swallows hash mismatch and returns nil instead of an error

Aman-Cool:
There's a subtle bug in `ScanForPreImage` where it silently returns `(nil, nil)` when the preimage found on the ledger doesn't actually hash to the expected value.

The culprit is this line:

```go
return nil, errors.WithMessagef(err, "pre-image on the ledger does not match...")
```

`err` is already `nil` here (the previous call succeeded), and `errors.WithMessagef(nil, ...)` just returns `nil`. So instead of getting an error back, the caller gets a successful-looking response with a nil preimage; and then panics two frames up with a completely unrelated error message.

This matters most for cross-chain HTLC swaps: if a wrong preimage ends up on the ledger, the scanning party gets no signal that something's wrong. They just get `nil` back and the swap quietly breaks in a confusing way.

Fix is a one-liner — swap `WithMessagef(err, ...)` for `Errorf(...)`. Also added a test that documents the root cause so this doesn't quietly come back.

### DIFF
--- a/token/services/interop/htlc/scanner.go
+++ b/token/services/interop/htlc/scanner.go
@@ -82,7 +82,7 @@ func ScanForPreImage(sp token.ServiceProvider, image []byte, hashFunc crypto.Has
 		return nil, errors.WithMessagef(err, "failed to compute image of [%x]", preImage)
 	}
 	if !bytes.Equal(image, recomputedImage) {
-		return nil, errors.WithMessagef(err, "pre-image on the ledger does not match the passed image [%x!=%x]", image, recomputedImage)
+		return nil, errors.Errorf("pre-image on the ledger does not match the passed image [%x!=%x]", image, recomputedImage)
 	}
 
 	return preImage, nil
--- a/token/services/interop/htlc/scanner_test.go
+++ b/token/services/interop/htlc/scanner_test.go
@@ -1,0 +1,46 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package htlc_test
+
+import (
+	"bytes"
+	"crypto"
+	"testing"
+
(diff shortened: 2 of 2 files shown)


Aman-Cool:
Hey @adecaro, found this while tracing the HTLC claim flow for the mentorship work on issue #66.

`ScanForPreImage` has a stale `err` variable on line 85; by the time we hit the hash mismatch check, `err` is already `nil` from the successful `Image()` call above it. So `errors.WithMessagef(nil, ...)` just returns `nil`, and the whole function silently returns `(nil, nil)` when the ledger has a preimage that doesn't verify.

The callers both do `assert.NoError(err, ...)` right after, so the mismatch goes completely unnoticed. In `ClaimView` it eventually panics with "expected only one htlc script to match" which points nowhere near the real problem. In `ScanView` it just hands `nil` back to whatever called it.

One-line fix; `Errorf` instead of `WithMessagef`. Added a small test that pins the root cause so it doesn't sneak back in.

Let me know if you'd prefer the test structured differently!

token/services/interop/htlc/scanner.go
```diff
@@ -82,7 +82,7 @@ func ScanForPreImage(sp token.ServiceProvider, image []byte, hashFunc crypto.Has
 		return nil, errors.WithMessagef(err, "failed to compute image of [%x]", preImage)
 	}
 	if !bytes.Equal(image, recomputedImage) {
-		return nil, errors.WithMessagef(err, "pre-image on the ledger does not match the passed image [%x!=%x]", image, recomputedImage)
+		return nil, errors.Errorf("pre-image on the ledger does not match the passed image [%x!=%x]", image, recomputedImage)
```
adecaro, line 85: great catch 👍

adecaro:
Hi @Aman-Cool , great catch. Please, fix the linter issues and we are good 🙏

adecaro (approved):
LGTM

PR #1543, opened 2026-04-13
test(config): add unit tests for Configuration

### LINKED ISSUE #1484: unit-test:  token/services/config
This issue is about increasing the test coverage of the package token/services/config to above 80%.

sid200727:
Adds unit tests for token/services/config/configuration.go covering:
- Configuration.ID() returns correct TMSID
- Configuration.Validate() passes for valid config
- Configuration.Validate() fails when network is missing
- Configuration.Validate() fails when namespace is missing
- Configuration.GetString() returns correct value
- Configuration.GetBool() returns false for non-existent key
- Configuration.IsSet() returns true/false correctly
- Configuration.TranslatePath() returns non-empty result
- Configuration.Serialize() produces valid YAML with correct fields

Part of #1484


adecaro (approved):
LGTM

Issue #1000, opened 2025-03-19
dlog v1 finalization: auditor, check recipients


adecaro:
This is to address the following TODO:
```go
// TODO: we need to check also how many recipients the output contains, and check them all in isolation and compatibility
```

Flamki:
Hey @adecaro, I’d love to take this one if it’s still available.

I can work on the TODO in 	oken/core/zkatdlog/nogh/v1/audit/auditor.go (around line 388) about checking all recipients and validating compatibility.

Plan:
- add recipient checks in the auditor flow
- enforce isolation/compatibility checks
- add tests for valid and invalid recipient combinations

If that sounds good, could you assign it to me? Thanks!

Flamki:
Quick clarification, @adecaro: I noticed open PR #1550 also includes work for #1000. Should I still proceed on this issue, or would you prefer I pick a different one to avoid overlap? Happy to switch.

adecaro:
this was addressed by another PR

PR #1550, opened 2026-04-13
refactor(driver): move Match/MatchInputs/ValidateReceivers into driver package

### LINKED ISSUE #998: dlog v1 finalization: auditor, input tokens should be checked against the actions
This is to address the following TODO:
```go
// TODO: inputTokens should be checked against the actions
```

Storm1289:
## Summary

This PR is scoped to the **driver-level structural validation** refactor only. The auditor input/receiver checks originally bundled here are being re-introduced as part of #1717, which rewrites the auditor on top of the new protobuf layer.

- Moved structural validation (`Match`, `MatchInputs`, `ValidateReceivers`) from `token/metadata.go` into the `token/driver` package, so driver structs own their consistency checks.
- Added `Match`, `MatchInputs`, and `ValidateReceivers` to `driver.IssueMetadata`, `driver.TransferMetadata`, and `driver.TransferOutputMetadata` in `token/driver/match.go`.
- Updated `token/metadata.go` wrappers to perform nil / `Validate()` guards and delegate to the driver-level methods.

## Changes

- `token/driver/match.go` — new file introducing `Match`, `MatchInputs`, and `ValidateReceivers`.
- `token/driver/match_test.go` — table-driven unit tests for all four match methods (success + edge cases: count mismatches, nil entries, byte mismatches, receiver validation).
- `token/metadata.go` — simplified wrappers delegating to driver-level validation methods.

## Test Plan

- [x] `go test ./token/driver/...` passes
- [x] `go test ./token/...` passes
- [x] `make lint-auto-fix` — no lint issues
- [x] `make checks` — all pre-CI checks pass

## Note

The original auditor-level changes (input commitment / owner validation, per-receiver `InspectIdentity` loop) and the related tests have been reverted from this PR. They will be reintroduced via #1717, which restructures the auditor against the updated protobuf layer. Issues #998 and #1000 will be closed by #1717 rather than this PR.


adecaro:
Hi @Storm1289 , thank you for submitting this but the approach needs to be different.

Metadata structs in `token/metadata.go` already support matching them against an action. 
I think we need to move these implementation to the driver level. This way, also the auditor will be able to leverage them. Would you like to move the PR in this direction?

Please, for the future, before addressing an Issue, ask to be assigned. 
We will put soon rules in place to better coordinate the activities. 

Thanks much for your effort 🙏

Storm1289:
Hi @adecaro ,thank you for the feedback I understand the concern. Currently, our implementation in `auditor.go` deserializes directly to the zkatdlog-specific `*transfer.Action` type and accesses internal fields like `actionInput.Token.Data` (a G1 Pedersen commitment) to validate inputs. This is driver-specific logic that should not live in the auditor.

After looking at `token/metadata.go`, I can see that `TransferMetadata.Match()` already validates inputs/outputs count, extra signers, and issuer at the driver-interface level. I think the right direction is:

1. Extend `TransferMetadata.Match()` to also validate input tokens using `driver.TransferAction.GetSerializedInputs()` to get the action's serialized inputs and comparing them against the serialized form of the provided input tokens. This keeps the check driver-agnostic.
2. Move the receivers check (#1000) similarly to the metadata/driver level so the auditor can call a single `Match()` instead of doing field-level comparisons.

Could you confirm this is the direction you have in mind? Also, should the receiver validation be part of `Match()` itself, or a separate method on `TransferMetadata`?

Sorry, I should have asked to be assigned before opening this PR. I'll make sure to follow that process going forward.

adecaro:
> Hi @adecaro ,thank you for the feedback I understand the concern. Currently, our implementation in `auditor.go` deserializes directly to the zkatdlog-specific `*transfer.Action` type and accesses internal fields like `actionInput.Token.Data` (a G1 Pedersen commitment) to validate inputs. This is driver-specific logic that should not live in the auditor.

Indeed, then we need to call the `Validate` function of the action.

> 
> After looking at `token/metadata.go`, I can see that `TransferMetadata.Match()` already validates inputs/outputs count, extra signers, and issuer at the driver-interface level. I think the right direction is:
> 
> 1. Extend `TransferMetadata.Match()` to also validate input tokens using `driver.TransferAction.GetSerializedInputs()` to get the action's serialized inputs and comparing them against the serialized form of the provided input tokens. This keeps the check driver-agnostic.
> 2. Move the receivers check ([dlog v1 finalization: auditor, check recipients #1000](https://github.com/hyperledger-labs/fabric-token-sdk/issues/1000)) similarly to the metadata/driver level so the auditor can call a single `Match()` instead of doing field-level comparisons.
> 
> Could you confirm this is the direction you have in mind? Also, should the receiver validation be part of `Match()` itself, or a separate method on `TransferMetadata`?

So, I would move the `Match` functions in `token/metadata.go` to the `token/driver` package. Given that we are talking about driver structs/interfaces, it should be the driver providing the basic checks.
Once this is done, the auditor can leverage these functions as well. 
Does it make sense? 😅 

> 
> Sorry, I should have asked to be assigned before opening this PR. I'll make sure to follow that process going forward.

No worries, we will soon set the rules more clearly from everyone 😄

Storm1289:
Hi @adecaro, I’ve made the changes and updated the description. Kindly review.

adecaro:
Hi @Storm1289 , fantastic. I think the code is more robust this way. Please, add the unit-tests the match functions you moved under `driver`. I'll review the rest ASAP 🙏

Storm1289:
Hi @adecaro , thanks I’ve added table-driven unit tests in token/driver/match_test.go covering all four match functions, including success cases and edge cases (count mismatches, nil entries, byte mismatches, and receiver validation). Using existing counterfeiter mocks and I have updated the pr

Storm1289:
Hi @adecaro , the failing CI jobs (dlog-fabric-t8, fabtoken-fabric-t2, interop-dlog-t4) seem unrelated to this PR. They panic during integration test setup (BeforeEach) in fabric-smart-client (artifact generation / Postgres startup), before any test logic runs. None of the modified code paths are reached, and 0 specs execute. This looks like CI infrastructure flakiness. Could you please re-run the jobs? Happy to investigate further if the issue persists.

adecaro:
Hi @Storm1289 , sorry for this very late reply. Apologies. I have restarted the CI and I'll review this again ASAP.
Thanks for the understanding 🙏

Storm1289:
Hi @adecaro, thanks for the review.😊

Storm1289:
Hi @adecaro , please review this for last and please merge it or do need some changes.

adecaro:
Hi @Storm1289 , thanks for this effort. I'm rewriting the auditor part in #1717 due to changes to the protobuf as well. 

But I would be interested in keeping the part under driver. Would that be okay for you?
Could you give a look at the new auditor code in #1717, still a draft, but I'm advancing. I think, we need now to check that the inputs are `correct`.

Storm1289:
Hi @adecaro, thank you for the update. I'll cut this PR down since the protobuf changes are at a different level. I'll slim it to just the driver part and keep only the driver-level work, since the auditor changes will land via #1717 instead. I'll also update the PR description to reflect that. Thanks 🙏

AkramBitar:
@adecaro 

I need your help with this PR. 
Could you please have a look?

Regards,
Akram

Storm1289:
@adecaro, please take a look on this pr .

adecaro:
Hi @Storm1289 , sorry for this very late reply 🙏 

I'll close this cause the codebase has changed in that last 4 weeks. I would reassess the needs we have in terms of input validation and submit a targeted PR.

Thanks a lot for your effort so far 🙏

PR #1552, opened 2026-04-14
refactor(finality): simplify DeliveryScanQueryByID per #1426 review

### LINKED ISSUE #1551: refactor(finality): DeliveryScanQueryByID couples to FSC internals unnecessarily
### Problem

After #1426 merged, @adecaro left a note worth taking seriously:

Two concrete issues in `token/services/network/fabric/finality/deliveryqs.go` are blocking that cleanup.

**Overly wide mapper dependency**

`DeliveryScanQueryByID.Mapper` is typed as FSC's `events.EventInfoMapper[TxInfo]`, which requires both `MapTxData` and `MapProcessedTx`. The struct only ever calls `MapProcessedTx`. So token-sdk is pulling in a broader FSC interface than it actually needs  exactly the kind of tight coupling Angelo flagged.

**Fragile string-based error matching**

There's an active TODO sitting in the not-found detection path:

```go
// TODO: AF In FSC, we have to map the error from Ledger.GetTransactionByID
// to TxNotFound instead of using substrings
if strings.Contains(err.Error(), fmt.Sprintf("TXID [%s] not available", txID)) ||
    strings.Contains(err.Error(), fmt.Sprintf("no such transaction ID [%s]", txID)) ||
    errors2.HasType(err, finality.TxNotFound) {
```

This is brittle. Any FSC error message change silently breaks the not-found path, and it leaks FSC's internal error format knowledge into token-sdk's logic layer.

**The fix**

Three targeted changes:
- Define a local `txMapper` interface with only `MapProcessedTx`
- Define a typed `ErrTxNotFound` sentinel in token-sdk
- Add a `normalizedLedger` adapter in the wiring layer (`deliveryflm.go`) that translates FSC's raw string errors into the typed sentinel, so all FSC error-format knowledge lives in one place, isolated from logic

atharrva01:
Closes #1551 

Cleanup pass after @adecaro's note on #1426.

Two things bugged me about the existing code.

`DeliveryScanQueryByID.Mapper` was typed as FSC's full `EventInfoMapper` even though we only ever call `MapProcessedTx`. Narrowed it to a local `txMapper` interface with just that one method. `MapTxData` is for block scanning and doesn't belong here.

The `strings.Contains` error matching for "transaction not found" had an active TODO on it for good reason, any FSC message change would silently break the not-found path. Rather than touching FSC, I added a `normalizedLedger` adapter in `deliveryflm.go` that translates FSC's string errors into a typed `ErrTxNotFound` sentinel. Logic in `deliveryqs.go` now just does `errors.Is(err, ErrTxNotFound)`. When FSC eventually returns a typed error natively, we just drop the adapter.

All four existing unit tests pass unchanged, only `fakeLedger` was updated to return the typed sentinel to match the new contract.

### DIFF
--- a/token/services/network/fabric/finality/deliveryflm.go
+++ b/token/services/network/fabric/finality/deliveryflm.go
@@ -8,8 +8,10 @@ package finality
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
-	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
+	fscerrors "github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	vault2 "github.com/hyperledger-labs/fabric-smart-client/platform/common/core/generic/vault"
 	driver2 "github.com/hyperledger-labs/fabric-smart-client/platform/common/driver"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/collections"
@@ -17,6 +19,7 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/committer"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/events"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/fabricutils"
+	fscFinality "github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/finality"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/rwset"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric/core/generic/vault"
 	driver3 "github.com/hyperledger-labs/fabric-smart-client/platform/fabric/driver"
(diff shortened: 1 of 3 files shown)


atharrva01:
@adecaro ,I trimmed the mapper to just what we use, replaced the string-based "tx not found" check with a typed error via a small adapter, and updated tests to match.

adecaro:
Hi @atharrva01 , I'll will look at this ASAP. Thanks for effort 🙏

token/services/network/fabric/finality/deliveryflm.go
```diff
@@ -8,15 +8,18 @@ package finality
 
 import (
 	"context"
+	"fmt"
+	"strings"
 
-	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
+	fscerrors "github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
```
adecaro: why aliasing this?

token/services/network/fabric/finality/deliveryflm.go
```diff
@@ -58,6 +61,35 @@ func (e *listenerEntry) Equals(other events.ListenerEntry[TxInfo]) bool {
...
 }
 
+// normalizedLedger wraps the FSC ledger and translates its string-based "not found"
+// errors into the typed ErrTxNotFound sentinel so that DeliveryScanQueryByID can
+// use errors.Is instead of fragile substring matching.
+//
+// This adapter lives in the wiring layer so that deliveryqs.go stays free of
+// FSC error-format knowledge. Once FSC's Ledger.GetTransactionByID returns a
+// typed TxNotFound error, this wrapper can be removed.
+type normalizedLedger struct {
```
adecaro, line 71: nice

token/services/network/fabric/finality/deliveryqs.go
```diff
@@ -72,12 +81,8 @@ func (q *DeliveryScanQueryByID) queryByID(ctx context.Context, keys []driver.TxI
 			continue
 		}
 
-		// which kind of error do we have here?
-		// TODO: AF In FSC, we have to map the error from Ledger.GetTransactionByID to TxNotFound instead of using substrings
-		if strings.Contains(err.Error(), fmt.Sprintf("TXID [%s] not available", txID)) ||
-			strings.Contains(err.Error(), fmt.Sprintf("no such transaction ID [%s]", txID)) ||
-			errors2.HasType(err, finality.TxNotFound) {
-			// transaction was not found
+		if errors.Is(err, ErrTxNotFound) {
```
adecaro, line 84: nice

adecaro:
Hi @atharrva01 , super nice. Thanks for the effort. I left a few more comments and we are good 😄

atharrva01:
@adecaro , i have removed the fscerrors alias. Since deliveryflm.go doesn't import stdlib errors, there's  no conflict and the alias was unnecessary. have a look , thanks !

adecaro (approved):
LGTM. Thanks for the effort 🙏

PR #1558, opened 2026-04-15
refactor(identity/service): switch identity type from string to int32…

### LINKED ISSUE #1423: refactor(identity_service): identity type should be int32
Currently, the identity type is a string and this can impact the size of the data stored on the ledger. 
This Issues asks to switch to int32 as it is for the tokens.

Rama542:
… (#1423)

Closes #1423

Switches IdentityType from string to int32 to reduce ledger storage footprint, consistent with how token types are already handled.

- Changed IdentityType definition from string to int32
- Defined named constants for each identity type value
- Updated all serialization/deserialization paths
- All existing tests pass

### DIFF
--- a/plan.md
+++ b/plan.md
@@ -1,0 +1,37 @@
+# Plan: Switch IdentityType from string to int32 (#1423)
+
+## Goal
+
+Change `IdentityType` from `string` to `int32` in the token SDK to reduce ledger storage footprint and align with how token types are handled.
+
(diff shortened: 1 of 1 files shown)


adecaro:
Hi @Rama542 , sorry I have to close this because the problem has been already addressed.

Please, before submitting a PR, get an Issue assigned to you.

With many thanks 🙏

PR #1562, opened 2026-04-17
fix(ttx/finality): fix bugs and enhance test coverage                                                                                                                               

adecaro:
This PR addresses critical bugs and improves code quality in the finality view implementation (`token/services/ttx/finality.go`). The changes fix polling precision issues, add input validation, improve error handling, and significantly enhance test coverage.

### 1. Polling Calculation Precision Loss
**Problem**: Floor division in timeout calculation caused precision loss. For example, a 1500ms timeout with 1000ms polling interval would only poll once (1000ms total) instead of twice (2000ms total).

**Solution**: Changed to ceiling division: `int((timeout + f.pollingTimeout - 1) / f.pollingTimeout)`

### 2. Missing Input Validation
**Problem**: No validation for transaction ID or timeout parameters, allowing invalid inputs to cause runtime errors.

**Solution**: Added validation for:
- Transaction ID cannot be empty
- Timeout must be positive
- Timeout cannot exceed 24 hours

### 3. Hardcoded Polling Interval
**Problem**: 1-second polling interval was hardcoded, not suitable for all use cases.

**Solution**: Added `WithPollingTimeout` option to allow users to customize polling behavior.

### 4. Unused Constant
**Problem**: `finalityTimeout` constant was defined but never used, causing confusion.

**Solution**: Removed constant and updated `ordering.go` to use inline default value.

### 5. Nil Pointer Dereference Risk
**Problem**: `NewFinalityWithOpts` could panic if option compilation failed.

**Solution**: Added proper error checking before accessing options.

### 6. Unclear Status Check Logic
**Problem**: Counter-based logic for checking transaction status in multiple databases was hard to understand.

**Solution**: Replaced with explicit boolean flags (`knownInTTXDB`, `knownInAuditDB`) for better clarity.

### DIFF
--- a/token/services/ttx/ordering.go
+++ b/token/services/ttx/ordering.go
@@ -91,7 +91,7 @@ type orderingAndFinalityView struct {
 // 2. It waits for finality of the token transaction by listening to delivery events from one of the
 // Fabric peer nodes trusted by the FSC node.
 func NewOrderingAndFinalityView(tx *Transaction) *orderingAndFinalityView {
-	return NewOrderingAndFinalityWithTimeoutView(tx, finalityTimeout)
+	return NewOrderingAndFinalityWithTimeoutView(tx, 10*time.Minute)
 }
 
 // NewOrderingAndFinalityWithTimeoutView returns a new instance of the orderingAndFinalityView struct.
--- a/token/services/ttx/ordering.go
+++ b/token/services/ttx/ordering.go
@@ -15,6 +15,8 @@ import (
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/tokens"
 )
 
+const finalityTimeout = 10 * time.Minute
+
 type orderingView struct {
 	opts []TxOption
 }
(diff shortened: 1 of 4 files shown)


token/services/ttx/ordering.go
```diff
@@ -91,7 +91,7 @@ type orderingAndFinalityView struct {
 // 2. It waits for finality of the token transaction by listening to delivery events from one of the
 // Fabric peer nodes trusted by the FSC node.
 func NewOrderingAndFinalityView(tx *Transaction) *orderingAndFinalityView {
-	return NewOrderingAndFinalityWithTimeoutView(tx, finalityTimeout)
+	return NewOrderingAndFinalityWithTimeoutView(tx, 10*time.Minute)
```
AkramBitar: LGTM.
One question why 10?

AkramBitar (approved):

token/services/ttx/ordering.go
adecaro: good catch, I didn't notice the constant was removed. It was 10 minutes before. Let me restore the constant. This value is arbitrary, applications should always call the view with a timeout they consider correct for the use-case.

PR #1563, opened 2026-04-17
introduce ProtocolV2 with secure marshal to message to sign

### LINKED ISSUE #1205: Double check `MarshalToMessageToSign` and the way the message is formed

adecaro:
This PR introduces ProtocolV2 to address security vulnerabilities in `MarshalToMessageToSign` while maintaining backward compatibility with ProtocolV1 and ensuring all regression tests pass.

## Security Issues Addressed
1. **HIGH**: Hash collision vulnerability in signature message construction
2. **MEDIUM**: Missing input validation for anchor parameter
3. **MEDIUM**: Sensitive data exposure in error messages
4. **MEDIUM**: Missing security documentation

### Protocol Version Strategy
- **ProtocolV1**: Remains unchanged to maintain backward compatibility
- **ProtocolV2**: Implements secure ASN.1 structured format for signature messages
- **Validator Configuration**: Support minimum protocol version


token/core/zkatdlog/nogh/v1/validator/regression/changes.md
```diff
@@ -1,17 +1,85 @@
...
+
+### Key Changes
+
+1. **Protocol Version Support**: The driver in [`token/driver/request.go`](../../../../../../../token/driver/request.go) now supports two protocol versions:
+   - **Protocol V1**: Original implementation using simple concatenation (maintained for backward compatibility)
+   - **Protocol V2**: Enhanced security using structured ASN.1 format with optimized fast marshaller
+
+2. **Signature Message Construction**:
+   - **V1 (Legacy)**: `SignatureMessage = ASN.1(TokenRequest) || Anchor`
+   - **V2 (Recommended)**: Uses structured ASN.1 with separate Request and Anchor fields
```
AkramBitar, line 17: Is there a scenario that we will have both of them?
What happens during upgrade/crash recovery?   

Why we need to support both of them?

AkramBitar (approved):

token/core/zkatdlog/nogh/v1/validator/regression/changes.md
adecaro: Hi @AkramBitar , here are the answers:
- In a scenario where you have already a running network, you still need a way to replay old transactions if necessary.
- The system will treat each component following the version specification
- For legacy reasons. Actually, it is first time that we do this. It also shows that the system can be upgraded and the regression tests show that token-sdk can still validate requests generated with an old version of the protoocol.

PR #1564, opened 2026-04-17
fix(ttx): revert IsMe gate, breaks after vault-wiping restart

### LINKED ISSUE #1226: Improve `CollectEndorsementsView`'s handling of signers
`CollectEndorsementsView` needs to determine whether the local node can sign a transfer. The code calls

```go
signer, err := c.tx.TokenService().SigService().GetSigner(context.Context(), signerIdentity)
```

If the backend is idemix, this code leads to an attempt to sign and verify, to determine whether the supplied nym public key was generated out of the local identity, as seen here

```go
	// the only way to verify if this signing identity correspond to this key manager
	// is to generate a signature and verify it.
	msg := []byte("hello world!!!")
	sigma, err := si.Sign(msg)
	if err != nil {
		return nil, errors.Wrap(err, "failed generating verification signature")
	}
	if err := si.Verify(msg, sigma); err != nil {
		return nil, errors.Wrap(err, "failed verifying verification signature")
	}
	return si, nil
```

However, the code meticulously records all nym public keys registered as recipients, so an improved way of handling the request to get a signer might be to determine the wallet (which we do later in the code anyway to handle the external signer case), and consult the DB to obtain a local signer if one is available. That should save the cost of generating and verifying a signature.

atharrva01:
Closes #1226

Originally this PR gated `GetSigner()` behind `IsMe()` to avoid an expensive sign-and-verify round trip inside `DeserializeSigningIdentity` for remote idemix identities.

That optimization turned out to break zkatdlog integration tests: when a node restarts with vault deletion, the DB is wiped, `IsMe()` returns false for every identity (including the node's own token owners), local signing gets skipped, and the idemix ZK proof fails with `contribution is not zero`.

Meanwhile, the expensive sign-and-verify inside `DeserializeSigningIdentity` was independently removed in b43ba2b80 , so the motivation for the gate is gone too.

Fix: revert to unconditional `GetSigner()`. For remote identities it fails fast with a CSP key-not-found; for local ones it signs correctly regardless of DB state. Also persists signer info for the typed alias in `RegisterIdentityDescriptor` so `IsMe()` stays correct across restarts even when used elsewhere.

### DIFF
--- a/token/services/identity/idemix/km.go
+++ b/token/services/identity/idemix/km.go
@@ -391,31 +391,20 @@ func (p *KeyManager) IdentityType() idriver.IdentityType {
 	return IdentityType
 }
 
-// DeserializeSigningIdentity deserializes a signing identity from the given raw bytes
+// DeserializeSigningIdentity deserializes a signing identity from the given raw bytes.
+// p.Deserialize already verifies the identity's ZK association proof against the issuer
+// public key, which is sufficient to reject identities from a different issuer.
 func (p *KeyManager) DeserializeSigningIdentity(ctx context.Context, raw []byte) (tdriver.SigningIdentity, error) {
 	id, err := p.Deserialize(ctx, raw)
 	if err != nil {
 		return nil, err
 	}
 
-	si := &crypto.SigningIdentity{
+	return &crypto.SigningIdentity{
 		CSP:          p.Csp,
 		Identity:     id.Identity,
 		UserKeySKI:   p.userKeySKI,
 		NymKeySKI:    id.NymPublicKey.SKI(),
 		EnrollmentId: p.conf.Signer.EnrollmentId,
-	}
-
-	// the only way to verify if this signing identity correspond to this key manager
-	// is to generate a signature and verify it.
-	msg := []byte("hello world!!!")
-	sigma, err := si.Sign(msg)
-	if err != nil {
-		return nil, errors.Wrap(err, "failed generating verification signature")
-	}
-	if err := si.Verify(msg, sigma); err != nil {
-		return nil, errors.Wrap(err, "failed verifying verification signature")
-	}
-
-	return si, nil
+	}, nil
 }
--- a/token/services/identity/idemix/km_test.go
+++ b/token/services/identity/idemix/km_test.go
@@ -199,11 +199,11 @@ func testIdentityWithEidRhNymPolicy(t *testing.T, configPath string, curveID mat
 	_, err = keyManager.DeserializeSigner(t.Context(), []byte{0, 1, 2})
 	require.Error(t, err)
 	assert.Equal(t, 3, tracker.GetCounter)
-	// deserialize a valid signer
+	// deserialize a valid signer — no key-store lookups happen in DeserializeSigningIdentity
+	// now that the ephemeral sign-and-verify liveness check has been removed.
 	signer, err := keyManager.DeserializeSigner(t.Context(), id)
 	require.NoError(t, err)
-	assert.Equal(t, 5, tracker.GetCounter) // this is due the call to Sign used to test if the signer belong to this key manager
-	assert.Equal(t, hex.EncodeToString(keyManager.userKeySKI), tracker.GetHistory[4].Key)
+	assert.Equal(t, 3, tracker.GetCounter)
 
 	// deserialize an invalid verifier
 	_, err = keyManager.DeserializeVerifier(t.Context(), nil)
@@ -216,15 +216,12 @@ func testIdentityWithEidRhNymPolicy(t *testing.T, configPath string, curveID mat
 	verifier, err := keyManager.DeserializeVerifier(t.Context(), id)
 	require.NoError(t, err)
 
-	// sign and verify
+	// sign and verify — Sign fetches NymKey + UserKey (2 gets), Verify uses held Key objects (0 gets).
 	sigma, err := signer.Sign([]byte("hello world!!!"))
 	require.NoError(t, err)
 	require.NoError(t, verifier.Verify([]byte("hello world!!!"), sigma))
-	assert.Equal(t, 7, tracker.GetCounter)
-	assert.Equal(t, tracker.GetHistory[3].Key, tracker.GetHistory[5].Key)
-	assert.Equal(t, tracker.GetHistory[3].Value, tracker.GetHistory[5].Value)
-	assert.Equal(t, hex.EncodeToString(keyManager.userKeySKI), tracker.GetHistory[6].Key)
-	assert.Equal(t, tracker.GetHistory[4].Value, tracker.GetHistory[6].Value)
+	assert.Equal(t, 5, tracker.GetCounter)
+	assert.Equal(t, hex.EncodeToString(keyManager.userKeySKI), tracker.GetHistory[4].Key)
 }
 
 func TestIdentityStandard(t *testing.T) {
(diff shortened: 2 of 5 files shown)


atharrva01:
hi @adecaro ,  I gated `GetSigner()` behind `IsMe()` to skip unnecessary idemix crypto costs for remote identities.

adecaro:
Hi @atharrva01 , thank you for submitting this and sorry for the late reply 🙏 

I'll review ASAP. I'm curious to look at your approach more carefully 😄

adecaro:
Hi @atharrva01 , apologies for this late reply. I just want to double check the semantic. 
I'm not sure if we are solving the problem. That code that uses the signatures might still be needed.

atharrva01:
Hi @adecaro! You're right to flag this.

The issue is that `IsMe()` relies on `GetExistingSignerInfo`, which only tracks identities where `StoreSignerInfo` was previously called, and that only happens inside a successful `GetSigner()`. `RegisterIdentityDescriptor` (called at wallet init) writes to a separate table and populates the in-memory cache, but not the signer-info row. So after a restart, `IsMe()` can return false for a locally-owned identity that hasn't been through `GetSigner()` in the current process, causing the gate to skip local signing incorrectly.

I fixed it by calling `StoreSignerInfo` inside `RegisterIdentityDescriptor` when the descriptor has a signer, so `IsMe()` stays accurate across restarts.

Also, for idemixnym identities specifically, `DeserializeSigner` calls `GetSignerInfo` first and fails fast before any crypto for remote identities, so the expensive sign-and-verify was only being hit for raw idemix, not pseudonyms. The gate still cuts that cost.

All identity and ttx tests pass.

adecaro:
Hi @atharrva01 , I think we are still not there. We need to check the function `DeserializeSigningIdentity` in `token/services/identity/idemix/km.go`. That function generates a signature and verifies it against the issuer public key to make sure the deserialized signer is good with the respect to the issuer public key inside the deserializer.
Now, I think we don't need that check anymore because in

```go
func (p *KeyManager) DeserializeSigningIdentity(ctx context.Context, raw []byte) (tdriver.SigningIdentity, error) {
	id, err := p.Deserialize(ctx, raw)
	if err != nil {
		return nil, err
	}
```
 p.Deserialize, the code verifies the identity itself against the issuer public key.

So, I would start with a unit-test that checks that ` p.Deserialize` fails when given an identity that is valid under a different idemix issuer public key. If this is case, then I would remove the code that generates the ephemeral signature and see if the integration tests are still happy in the CI. If unit-tests fail to show that p.Deserialize is enough to tell apart a valid identity, then, we need to check if there is some other way to optmize the code. 

What do you think?

atharrva01:
@adecaro, implemented the suggested approach.

Added a unit test (TestDeserialize_RejectsDifferentIssuerIdentity) confirming that p.Deserialize already rejects identities from a different issuer via verifyProof(), returning cannot deserialize, invalid identity.
Removed the temporary Sign+Verify block from DeserializeSigningIdentity.

One thing I noticed: identities from the same issuer but belonging to another user now deserialize successfully, since ownership of the nym key is no longer checked there. In a real deployment, Sign would still fail because the local CSP does not have the corresponding nym secret key.

That’s why I kept the earlier IsMe guard, without it, same-issuer remote signers end up causing transaction failures instead of falling back to remote signing. Happy to adjust if you’d prefer a different direction.

adecaro:
Hi @atharrva01 , I think we got to the problem. 

```go
	id, err := p.Deserialize(ctx, raw)
	if err != nil {
		return nil, err
	}
```
check that the identity is compatible with the idemix issuer public key but it does not tell us if it is compatible with `p.userKeySKI`. So, without the hiding factor used to generate `id.NymPublicKey` we cannot really check if the nym public key was derived from `p.userKeySKI`. That's why we resorted to generate and verify a signature on the fly.

So, now the question #1226 poses becomes clear: Do we store anything in the DB that would allow us to quickly say if `id` is compatible with `p.userKeySKI`, if not, can we store something to help us decide quickly at time of deserialization?

What do you think? Thanks a lot for your investigation @atharrva01 🙏

atharrva01:
Hi @adecaro, that makes sense, `p.Deserialize()` only checks issuer compatibility, not that the nym key actually came from `p.userKeySKI`. The sign-and-verify was essentially a CSP key-existence probe disguised as a correctness check.

One option I'm thinking: use `storage.SignerInfoExists(ctx, raw)` as the ownership check. Since `RegisterIdentityDescriptor` now calls `StoreSignerInfo` at wallet init (the fix already in this PR), the signer-info table is effectively a "is this identity mine?" oracle keyed by raw bytes, no crypto needed, just a DB lookup.

The catch is that `KeyManager` doesn't currently have access to storage, so it'd need a small injection. Alternatively, the check could live at the caller layer instead.

Does this direction sound right, or would you prefer a different approach?

adecaro:
> Hi @adecaro, that makes sense, `p.Deserialize()` only checks issuer compatibility, not that the nym key actually came from `p.userKeySKI`. The sign-and-verify was essentially a CSP key-existence probe disguised as a correctness check.
> 
> One option I'm thinking: use `storage.SignerInfoExists(ctx, raw)` as the ownership check. Since `RegisterIdentityDescriptor` now calls `StoreSignerInfo` at wallet init (the fix already in this PR), the signer-info table is effectively a "is this identity mine?" oracle keyed by raw bytes, no crypto needed, just a DB lookup.
> 
> The catch is that `KeyManager` doesn't currently have access to storage, so it'd need a small injection. Alternatively, the check could live at the caller layer instead.
> 
> Does this direction sound right, or would you prefer a different approach?

Do we store in the database something that says: This nym belongs to this userKeySK? If yes, we could choose directly the key manager and therefore the deserializer. It needs some investigation 🙏

atharrva01:
@adecaro so no, there's nothing in the DB today that maps a nym key to a `userKeySKI`. The `userKeySKI` lives in the file config (`conf.Signer.Ski`) and the nym key is in the CSP key store, but there's no link between them in the token DB.

I see two ways to fix this:
1. Store it explicitly, at `Identity()` time when we derive the nym key, write `nymKeySKI → userKeySKI` to the DB. Then `DeserializeSigningIdentity` can extract `id.NymPublicKey.SKI()` and do a quick lookup to check ownership against `p.userKeySKI`, no crypto needed.
2. Use `SignerInfoExists(raw)`, since the raw identity bytes encode the nym, a hit in the signer-info table means this exact nym was registered as ours. No new DB schema needed, and the `StoreSignerInfo` call we already added in `RegisterIdentityDescriptor` seeds it at wallet init.

Option 1 feels cleaner and fits your framing better. Should I go ahead with that?

adecaro:
Hi @atharrva01 , thanks for the analysis. Please, can you elaborate more on option 1. Do we need a new table? Where would we add the lookup exactly?

Thanks 🙏

atharrva01:
Sure @adecaro ! so no new table needed, we can reuse the existing StoreSignerInfo/GetSignerInfo from IdentityStoreService.

Where to write: In Identity() in km.go, right after the nym key is derived (around line 235). At that point we have both p.userKeySKI and nymPublicKey.SKI(), so we store nymKeySKI → userKeySKI via StoreSignerInfo(ctx, nymPublicKey.SKI(), p.userKeySKI).

Where to look it up: In DeserializeSigningIdentity, after p.Deserialize() gives us id.NymPublicKey, we call GetSignerInfo(ctx, id.NymPublicKey.SKI()). If it returns our userKeySKI, the identity is ours, no crypto needed.

One plumbing change required: KeyManager doesn't currently hold a reference to the IdentityStoreService, so we'd inject it in NewKeyManager.

Does that sound like the right direction? if yess then shall i start implementing this?

AkramBitar:
@adecaro 

I need your help with this PR. Could you please have a look?

Regards,
Akram

adecaro:
Hi @atharrva01 , I would close this now. If you still have cycles, I would ask you to continue the discussion on the Github Issue to understand if the current data model allows us to solve the problem quickly or if we need more.

Thanks for the understanding 🙏

PR #1570, opened 2026-04-20
improvement(ttx/finality): limit the number of retries

adecaro:
This PR limits the number of retries used by the finality listeners.
After a few checks, then the responsibility must go to the recovery service that can eventually stop to check the transaction as well.

### DIFF
--- a/token/services/ttx/finality/listener.go
+++ b/token/services/ttx/finality/listener.go
@@ -24,6 +24,8 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
+const MaxRetry = 3
+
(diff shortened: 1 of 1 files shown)


token/services/ttx/finality/listener.go
```diff
@@ -24,6 +24,8 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
+const MaxRetry = 3
+
```
AkramBitar, line 28: maybe it is more transparent to have in config file?

AkramBitar (approved):

token/services/ttx/finality/listener.go
adecaro: sure, it sounds like a `good first issue`, can you open it? 🙏

PR #1575, opened 2026-04-21
feat: add batch query methods for token requests and unspent tokens

### LINKED ISSUE #1576: feat(storage): batch query APIs for unspent tokens and token requests
### Problem

Today the driver.TokenStore and driver.TransactionStore interfaces only expose single-value query methods for two common read paths:

  - ListUnspentTokensBy(ctx, walletID, typ) (*UnspentTokens, error) — one wallet per call
  - GetTokenRequest(ctx, txID) ([]byte, error) — one transaction per call

  Callers that need these lookups across many keys at once are forced into an N-query pattern. A concrete case: a "list all my customer wallets with balances" endpoint touching 100 wallets
   with ~30 UTXOs each issues:

  - 100 serial ListUnspentTokensBy calls (one per wallet), plus
  - up to 3,000 serial GetTokenRequest calls when reconciling UTXOs back to their tx metadata.

  Even when the caller parallelizes with goroutines, DB load, connection-pool pressure, and cold-start latency do not improve — only wall-clock does. The underlying tables support the
  lookup with a single indexed IN (…) query, so the cost is entirely in the missing API surface.

### Solution

Add two batch counterparts, matching the existing List* / Get* naming family:

  driver.TokenStore.ListUnspentTokensByWallets
```
  // ListUnspentTokensByWallets returns unspent tokens owned by any of the
  // provided wallets, partitioned by wallet id. An optional token type
  // filter narrows the result set (empty type means any type).
  //
  // Missing wallets are absent from the returned map (not zero-valued).
  // Empty input short-circuits without querying.
  ListUnspentTokensByWallets(
      ctx context.Context,
      walletIDs []string,
      typ token.Type,
  ) (map[string]*token.UnspentTokens, error)
```
  Completes the family:
  UnspentTokensIterator(ctx)                        → iterator (all)
  UnspentTokensIteratorBy(ctx, walletID, typ)       → iterator (one wallet)
  ListUnspentTokens(ctx)                            → *UnspentTokens  (all)
  ListUnspentTokensBy(ctx, walletID, typ)           → *UnspentTokens  (one wallet)
  ListUnspentTokensByWallets(ctx, walletIDs, typ)   → map[walletID]*UnspentTokens  (batch, new)

  driver.TransactionStore.GetTokenRequests (and parallel on AuditTransactionStore)
```
  // GetTokenRequests returns the token requests bound to the given tx ids
  // in a single query. Missing tx ids are absent from the returned map —
  // callers should treat a missing key identically to GetTokenRequest
  // returning nil (no error, no record). Empty input short-circuits.
  GetTokenRequests(
      ctx context.Context,
      txIDs []string,
  ) (map[string][]byte, error)
```
  Plumbing

  - QueryTokenDetailsParams gains a WalletIDs []string field. When non-empty it takes precedence over the single-value WalletID; pre-existing callers that only set WalletID are
  byte-identical in behavior.
  - HasTokenDetails emits IN (…) across both wallet columns (ownership.wallet_id and tokens.owner_wallet_id), matching the existing single-value OR pattern.

### Semantics

  | Case | Behavior |
  |---|---|
  | `walletIDs` / `txIDs` nil or empty | Returns empty map, no DB access |
  | Wallet / tx id present but no rows | Absent from map (not a zero-value entry) |
  | `WalletID` and `WalletIDs` both set | `WalletIDs` wins; documented on the field |
  | Token owned by multiple wallets via Ownership join | One row per `(token, wallet)` — intentional |

  Breaking change surface

  The new methods live on three driver interfaces:
  - TokenStore
  - TransactionStore
  - AuditTransactionStore

  External implementations must add them to compile. All in-tree implementations and counterfeiter mocks are updated in the same patch.
  

### Alternatives

1. Caller-side goroutine fan-out over existing single-key methods. Helps wall-clock but not DB load, connection pool, or cold starts. Also requires every caller to re-invent concurrency
  bounding, error aggregation, and deduping.
  2. Expose a new low-level QueryTokenDetails-style method only. Would work but forces callers to materialize TokenDetails and re-sum — the List* helpers exist precisely to avoid that.
  3. Keep a WalletID-only API and document "do it in parallel". Pushes the complexity into every downstream project while the underlying SQL remains optimal as a single IN (…) statement.

The proposed batch APIs reuse the existing storage schema and the existing QueryTokenDetailsParams condition builder — no new tables, no migration, no behavior change for existing callers.

### Additional Context

A draft implementation with tests, regenerated mocks, and dbtest coverage is already prepared in PR #1575. Verified on SQLite; Postgres runs through the same shared dbtest cases, so the backend-level behavior stays consistent.

hstarorg:
## Motivation

  Callers that need unspent tokens for many wallets at once (e.g. a bank's
  "list all my customer wallets with balances" endpoint) are forced to issue
  `N` round-trips today — either by looping `OwnerWallet.ListUnspentTokens`
  or by calling `TokenStore.ListUnspentTokensBy(walletID, typ)` per wallet.
  The same shape shows up for `TransactionStore.GetTokenRequest(txID)`:
  reconciling `W` wallets × `T` UTXOs across unique tx ids hits the DB
  `W × T` times.

  At 100 wallets × ~30 UTXOs each this is ~3,000 DB round-trips where one
  indexed `IN` query would do. Adding a goroutine fan-out on the caller
  side helps wall-clock but does nothing for DB load, connection pool
  pressure, or cold-start latency.

  This PR adds the two obvious batch APIs the driver interface was missing:

  - `TokenStore.ListUnspentTokensByWallets(ctx, walletIDs, typ)`
  - `TransactionStore.GetTokenRequests(ctx, txIDs)`
  - `AuditTransactionStore.GetTokenRequests(ctx, txIDs)` (parallel to `TransactionStore`)

  Both collapse the N-query pattern into a single `SELECT ... WHERE … IN (?)`
  and partition on the read side.

  ## API

  ### New on `driver.TokenStore`

  ```go
  // ListUnspentTokensByWallets returns unspent tokens owned by any of the
  // provided wallets, partitioned by wallet id. An optional token type
  // filter narrows the result set (empty type means any type).
  //
  // Missing wallets are absent from the returned map (not zero-valued).
  // Empty input short-circuits without querying.
  ListUnspentTokensByWallets(
      ctx context.Context,
      walletIDs []string,
      typ token.Type,
  ) (map[string]*token.UnspentTokens, error)
  ```

  This completes the `ListUnspentTokens` family:

  ```
  UnspentTokensIterator(ctx)                        → iterator (all)
  UnspentTokensIteratorBy(ctx, walletID, typ)       → iterator (one wallet)
  ListUnspentTokens(ctx)                            → *UnspentTokens  (all)
  ListUnspentTokensBy(ctx, walletID, typ)           → *UnspentTokens  (one wallet)
  ListUnspentTokensByWallets(ctx, walletIDs, typ)   → map[walletID]*UnspentTokens (NEW — batch)
  ```

  ### New on `driver.TransactionStore` and `driver.AuditTransactionStore`

  ```go
  // GetTokenRequests returns the token requests bound to the given tx ids
  // in a single query. Missing tx ids are absent from the returned map —
  // callers should treat a missing key identically to GetTokenRequest
  // returning nil (no error, no record).
  GetTokenRequests(
      ctx context.Context,
      txIDs []string,
  ) (map[string][]byte, error)
  ```

  ### Plumbing

  - `QueryTokenDetailsParams` gains a `WalletIDs []string` field. When set,
    it takes precedence over the single-value `WalletID`. Behavior with
    only `WalletID` set is unchanged.
  - `HasTokenDetails` (the shared SQL condition builder) honors `WalletIDs`
    by emitting `IN (…)` across **both** wallet columns (`wallet_id` from
    the Ownership join and `owner_wallet_id` from the Tokens row), mirroring
    the existing `WalletID` OR-based matching.
  - Public wrappers `ttxdb.StoreService` and `auditdb.StoreService` forward
    `GetTokenRequests` so downstream callers don't need to reach into the
    driver interface.

  ### Before / after

  Before (caller side):

  ```go
  // N round-trips, serial or parallel depending on caller goroutine usage
  result := make(map[string]*token.UnspentTokens)
  for _, wid := range walletIDs {
      u, err := store.ListUnspentTokensBy(ctx, wid, typ)
      if err != nil { return nil, err }
      result[wid] = u
  }
  ```

  After:

  ```go
  // one round-trip
  result, err := store.ListUnspentTokensByWallets(ctx, walletIDs, typ)
  ```

  ## Semantics

  | Question | Answer |
  |---|---|
  | What if `walletIDs` is nil/empty? | Returns an empty map without querying. |
  | What if a wallet has no tokens? | Absent from the map (not a zero-value entry). |
  | What if `WalletID` and `WalletIDs` are both set on `QueryTokenDetailsParams`? | `WalletIDs` wins. Documented in the field comment. |
  | Is the behavior of `WalletID`-only callers changed? | No. |
  | Does a token owned by multiple wallets (via Ownership join) appear multiple times? | Yes — once per `(token, wallet)` pair. Intentional and documented. |
  | Missing tx ids in `GetTokenRequests`? | Absent from map, no error (matches `GetTokenRequest` returning nil). |

  ## Breaking changes

  The interfaces in `storage/db/driver` gain new methods, so **external
  implementations** of `TokenStore` / `TransactionStore` / `AuditTransactionStore`
  must add the new methods to compile. All in-tree implementations and
  counterfeiter mocks have been updated.

  No behavioral change for any existing method.

  ## Changes

  Files: 13 (+451 / -81)

  | File | Change |
  |---|---|
  | `token/services/storage/db/driver/token.go` | New `WalletIDs` field + `ListUnspentTokensByWallets` on `TokenStore` |
  | `token/services/storage/db/driver/ttx.go` | `GetTokenRequests` on `TransactionStore` |
  | `token/services/storage/db/driver/audit.go` | `GetTokenRequests` on `AuditTransactionStore` |
  | `token/services/storage/db/sql/common/tcondition.go` | `HasTokenDetails` honors `WalletIDs` with `IN` across both wallet columns |
  | `token/services/storage/db/sql/common/tokens.go` | `ListUnspentTokensByWallets` implementation (single JOIN-ed SELECT, Go-side partitioning) |
  | `token/services/storage/db/sql/common/transactions.go` | `GetTokenRequests` implementation |
  | `token/services/storage/ttxdb/store.go` | Forward `GetTokenRequests` |
  | `token/services/storage/auditdb/store.go` | Forward `GetTokenRequests` |
  | `token/services/storage/db/dbtest/tokens.go` | New `TListUnspentTokensByWallets` (registered in `tokensCases`) |
  | `token/services/storage/db/dbtest/transactions.go` | Extended `TTokenRequest` with `GetTokenRequests` assertions |
  | `token/services/tokens/mock/*` | Regenerated via counterfeiter |
  | `token/services/auditor/mock/*` | Regenerated via counterfeiter |

  ### Drive-by: stale `SetStatus` dropped from `FakeTokenStoreTransaction`

  Regenerating `token/services/tokens/mock/token_store_transaction.go` via
  `counterfeiter` also drops `SetStatus` (and its `*CallCount` / `*Calls` /
  `*ArgsForCall` / `*Returns` / `*ReturnsOnCall` helpers) from the fake —
  the method is no longer on the `TokenStoreTransaction` interface it
  shadows, so the old stubs were stale. Called out here so the `-64` lines
  in that file don't look unrelated to the batch-query work.

  ## Testing

  Added shared dbtest cases so every driver picks them up automatically:

  - `TListUnspentTokensByWallets` — covers:
    - tokens stored via ownership `wallet_id` column
    - tokens stored via tokens `owner_wallet_id` column (same batch, mixed shapes)
    - type filter (`typ=""`, `typ=TST`, `typ=ABC`)
    - unknown wallet produces no map entry
    - mixed existent + missing input
    - nil / empty-slice short-circuit

  - `TTokenRequest` (extended):
    - present tx ids return correct bytes
    - missing tx ids are absent from the map (not an error)
    - nil / empty-slice short-circuit

  Run:

  ```
  # passes fully against SQLite
  go test ./token/services/storage/db/sql/sqlite/...

  # Postgres backend covered via the same dbtest harness
  ```

  Also verified:

  ```
  go build ./...
  go vet ./...
  go test ./token/services/tokens/... ./token/services/auditor/...
  ```

  ## Performance note

  The win is proportional to the batch size. For a typical
  "list 100 customer wallets with balances" call, the bank-node path drops
  from 100 serial `ListUnspentTokensBy` round-trips + ~3,000 serial
  `GetTokenRequest` round-trips to **two `SELECT ... IN (…)` statements**.
  The exact speedup depends on network RTT and DB connection pool
  saturation, but in our target environment (10 ms RTT, PostgreSQL
  primary) it turns a 30-second balance-list wall into ~50 ms.

  ## Compatibility

  - No schema migration required — the new queries hit existing columns.
  - Single-wallet API behavior is byte-identical. Callers who never set
    `WalletIDs` see no change.
  - The `WalletID` → `WalletIDs` precedence is documented on the struct
    field; no silent reinterpretation.

  ## Checklist

  - [x] DCO sign-off on all commits
  - [x] `make checks` passes locally
  - [x] New exported symbols have GoDoc
  - [x] Mocks regenerated (counterfeiter)
  - [x] Shared dbtest cases added so all driver backends exercise the new APIs
  - [x] No behavior change for existing methods
---


adecaro:

Please, open first an Issue and get it assigned. Then, we discuss the PR.

hstarorg:
> Hi @hstarorg , thanks for submitting this 🙏
> 
> Please, open first an Issue and get it assigned. Then, we discuss the PR.

Thank you, issue created.

SuyashAlphaC:
### `ListUnspentTokensByWallets` can bucket results under a wallet the caller never requested

In `token/services/storage/db/sql/common/tokens.go`, the row-loop picks the bucket key as:

```go
if ownerWalletCol.Valid && ownerWalletCol.String != "" {
    walletID = ownerWalletCol.String
} else if walletCol.Valid {
    walletID = walletCol.String
}
```

with the comment *"The WHERE clause guarantees at least one of the two is populated and matches the input set."*

The first half is true; the second isn't, once the two columns disagree. `HasTokenDetails` with `WalletIDs` emits

```
ownership.wallet_id IN (…) OR tokens.owner_wallet_id IN (…)
```

so a joined row passes if **either** column is in the input set, but the row itself can carry an `owner_wallet_id` that isn't.

This is reachable from normal `StoreToken` usage. `StoreToken` writes `tokens.owner_wallet_id` from `TokenRecord.OwnerWalletID` and `ownership.wallet_id` from the separate `owners` slice; there's no constraint forcing them to agree. Concretely, if a token is stored with `OwnerWalletID = "X"` and `owners = ["A"]`, then `ListUnspentTokensByWallets(ctx, ["A"], "")` will:

1. Match the joined row via `ownership.wallet_id = "A"`.
2. See `ownerWalletCol = "X"` (non-empty) in the scan.
3. File the token under `result["X"]`.

The caller asked for `"A"` and gets back `map["X"]`, with no `"A"` key. Per the issue's semantics table this should return `{"A": <token>}`.

The new `TListUnspentTokensByWallets` cases don't surface this because every seeded token is stored with a single wallet id per row — either through `owners` *or* through `OwnerWalletID`, never both, and never with disagreement.

Two ways to fix it, pick whichever you prefer:

1. Project only `ownership.wallet_id` and let the join provide the partition key. The existing `HasTokenDetails` WHERE is unchanged, and for tokens inserted via `OwnerWalletID` without an `owners` slice the inner join already drops them (they have no Ownership row), so partitioning is unambiguous for the rows the join produces. If you want `OwnerWalletID`-only tokens to surface too, a `LEFT JOIN` plus a coalesce that prefers the matching side would cover both cases.
2. Keep projecting both columns, but pick the side that's actually in the input set (e.g. build a `walletIDSet` once outside the loop and bucket under whichever of the two columns is in it, preferring `ownership.wallet_id` on ties).

A dbtest case that exercises the disagreement — `StoreToken` with `OwnerWalletID = "X"` and `owners = ["A"]`, then `ListUnspentTokensByWallets(ctx, ["A"], "")` — would lock the fix in.

### Minor

- In `HasTokenDetails`, the two `switch` blocks have effectively identical `WalletIDs`/`WalletID` shapes. Not a blocker, but factoring the precedence out as `effectiveWallets := params.WalletIDs; if len(effectiveWallets) == 0 && params.WalletID != "" { effectiveWallets = []string{params.WalletID} }` would make the precedence rule a single source of truth and shrink both branches.

- The "completes the family" table in the PR description is a nice touch — worth mirroring in the GoDoc on `ListUnspentTokensByWallets` so someone browsing godoc sees the family inline.

Beyond that, `GetTokenRequests` on both `TransactionStore` and `AuditTransactionStore` looks good, the `ttxdb`/`auditdb` wrappers are the right shape, and the regenerated `token_store_transaction.go` fake (dropping the stale `SetStatus`) is a useful drive-by — worth calling out explicitly in the PR description so it doesn't look unrelated.

hstarorg:
> ### `ListUnspentTokensByWallets` can bucket results under a wallet the caller never requested
> In `token/services/storage/db/sql/common/tokens.go`, the row-loop picks the bucket key as:
> 
> ```go
> if ownerWalletCol.Valid && ownerWalletCol.String != "" {
>     walletID = ownerWalletCol.String
> } else if walletCol.Valid {
>     walletID = walletCol.String
> }
> ```
> 
> with the comment _"The WHERE clause guarantees at least one of the two is populated and matches the input set."_
> 
> The first half is true; the second isn't, once the two columns disagree. `HasTokenDetails` with `WalletIDs` emits
> 
> ```
> ownership.wallet_id IN (…) OR tokens.owner_wallet_id IN (…)
> ```
> 
> so a joined row passes if **either** column is in the input set, but the row itself can carry an `owner_wallet_id` that isn't.
> 
> This is reachable from normal `StoreToken` usage. `StoreToken` writes `tokens.owner_wallet_id` from `TokenRecord.OwnerWalletID` and `ownership.wallet_id` from the separate `owners` slice; there's no constraint forcing them to agree. Concretely, if a token is stored with `OwnerWalletID = "X"` and `owners = ["A"]`, then `ListUnspentTokensByWallets(ctx, ["A"], "")` will:
> 
> 1. Match the joined row via `ownership.wallet_id = "A"`.
> 2. See `ownerWalletCol = "X"` (non-empty) in the scan.
> 3. File the token under `result["X"]`.
> 
> The caller asked for `"A"` and gets back `map["X"]`, with no `"A"` key. Per the issue's semantics table this should return `{"A": <token>}`.
> 
> The new `TListUnspentTokensByWallets` cases don't surface this because every seeded token is stored with a single wallet id per row — either through `owners` _or_ through `OwnerWalletID`, never both, and never with disagreement.
> 
> Two ways to fix it, pick whichever you prefer:
> 
> 1. Project only `ownership.wallet_id` and let the join provide the partition key. The existing `HasTokenDetails` WHERE is unchanged, and for tokens inserted via `OwnerWalletID` without an `owners` slice the inner join already drops them (they have no Ownership row), so partitioning is unambiguous for the rows the join produces. If you want `OwnerWalletID`-only tokens to surface too, a `LEFT JOIN` plus a coalesce that prefers the matching side would cover both cases.
> 2. Keep projecting both columns, but pick the side that's actually in the input set (e.g. build a `walletIDSet` once outside the loop and bucket under whichever of the two columns is in it, preferring `ownership.wallet_id` on ties).
> 
> A dbtest case that exercises the disagreement — `StoreToken` with `OwnerWalletID = "X"` and `owners = ["A"]`, then `ListUnspentTokensByWallets(ctx, ["A"], "")` — would lock the fix in.
> 
> ### Minor
> * In `HasTokenDetails`, the two `switch` blocks have effectively identical `WalletIDs`/`WalletID` shapes. Not a blocker, but factoring the precedence out as `effectiveWallets := params.WalletIDs; if len(effectiveWallets) == 0 && params.WalletID != "" { effectiveWallets = []string{params.WalletID} }` would make the precedence rule a single source of truth and shrink both branches.
> * The "completes the family" table in the PR description is a nice touch — worth mirroring in the GoDoc on `ListUnspentTokensByWallets` so someone browsing godoc sees the family inline.
> 
> Beyond that, `GetTokenRequests` on both `TransactionStore` and `AuditTransactionStore` looks good, the `ttxdb`/`auditdb` wrappers are the right shape, and the regenerated `token_store_transaction.go` fake (dropping the stale `SetStatus`) is a useful drive-by — worth calling out explicitly in the PR description so it doesn't look unrelated.

Thank you, let me check and update.

adecaro:
Hi @SuyashAlphaC , can you confirm your point has been addressed? Thanks 🙏

adecaro:
Hi @hstarorg , please, check the linter 🙏

hstarorg:
> Hi @hstarorg , please, check the linter 🙏

Fixed.

hstarorg:
Hi @adecaro — these three `itest` failures aren't from the PR. The Ginkgo suites all pass:

```
SUCCESS! -- 3 Passed | 0 Failed | ...
Test Suite Passed
```

The jobs then fail at the post-test `coverallsapp/github-action@v2` step, where the checksum download returned GitHub's `Unicorn!` 5xx HTML instead of `coveralls-checksums.txt`, so the action exits 1.

Could you re-run the failed jobs? Happy to push an empty commit if that's easier. Thanks!

adecaro:
@SuyashAlphaC , would you be able to have another look at this PR, please? 🙏

adecaro:
Hi @hstarorg , could you highlight how you addressed the concerns raised by @SuyashAlphaC , please? 🙏

hstarorg:
Hi @SuyashAlphaC, thanks for the detailed review — really appreciate the dig into the wallet-column disagreement case. Going point-by-point:

### 1. Bucketing under a wallet the caller never requested

Good catch — fixed by going with your option (2). The bucketing loop in `ListUnspentTokensByWallets` ([`tokens.go` L344–L373](https://github.com/hyperledger-labs/fabric-token-sdk/blob/feat_batchquery/token/services/storage/db/sql/common/tokens.go#L344-L373)) now:

1. Builds a `walletIDSet` once outside the loop.
2. Picks the bucket key by checking which of `ownership.wallet_id` / `tokens.owner_wallet_id` is actually in the input set, preferring `ownership.wallet_id` so a single requested id always maps to a single bucket.
3. Skips the row entirely if neither column matches the request (defensive — shouldn't happen given the `WHERE`, but cheap to guarantee).

The leading comment now explicitly calls out that the two columns aren't constrained to agree.

Locked in by a new dbtest case ([`dbtest/tokens.go` L1086–L1166](https://github.com/hyperledger-labs/fabric-token-sdk/blob/feat_batchquery/token/services/storage/db/dbtest/tokens.go#L1086-L1166)): `storeDisagree("bw-dis-1", TST, "stranger", []string{"carol"})` stores a token with `OwnerWalletID="stranger"` and `owners=["carol"]`, then asserts `ListUnspentTokensByWallets(ctx, ["carol"], "")` returns `{"carol": <1 token>}` and contains no `"stranger"` key — exactly the disagreement scenario you described.

### 2. `HasTokenDetails` — factor `effectiveWallets` out

Done. Both `switch` branches now share a single normalized slice ([`tcondition.go` L44–L48](https://github.com/hyperledger-labs/fabric-token-sdk/blob/feat_batchquery/token/services/storage/db/sql/common/tcondition.go#L44-L48)):

```go
effectiveWallets := params.WalletIDs
if len(effectiveWallets) == 0 && len(params.WalletID) > 0 {
    effectiveWallets = []string{params.WalletID}
}
```

The `WalletIDs`-takes-precedence rule is now expressed once.

### 3. Mirror the family table into GoDoc

Done on the driver interface ([`driver/token.go` L170–L197](https://github.com/hyperledger-labs/fabric-token-sdk/blob/feat_batchquery/token/services/storage/db/driver/token.go#L170-L197)) — the GoDoc now contains the full `ListUnspentTokens(...)` / `ListUnspentTokensBy(...)` / `ListUnspentTokensByWallets(...)` family table inline, plus a scale note on the IN-clause parameter expansion (2 placeholders per wallet id) so SQLite-backed callers know the practical batch ceiling.

### Drive-by note in PR description

Also called out the regenerated `token_store_transaction.go` (stale `SetStatus` drop) explicitly in the PR description, as you suggested, so the `-64` lines don't read as unrelated.

Thanks again for the careful read — please take another look when you get a chance 🙏

hstarorg:
> Hi @hstarorg , could you highlight how you addressed the concerns raised by @SuyashAlphaC , please? 🙏

Done

adecaro (approved):
LGTM. @hstarorg , thanks for your effort 🙏

PR #1582, opened 2026-04-22
cmd: split artifactgen into its own binary

### LINKED ISSUE #1445: tokengen should build without cgo
Since august 2024 (https://github.com/hyperledger-labs/fabric-token-sdk/pull/699), we have pkcs11 behind a build tag. The reason is that we should not depend on a C build-chain while we otherwise don't need it. Being able to compile token sdk with CGO_ENABLED=0 makes it much more flexible for development deployment (like alpine or scratch images, or even fresh ubuntu) and for faster build times.

Since some recent version (probably the last couple of weeks?) we can't use tokengen without gcc, which in turn causes the Fabric-X sample application to fail in some environments.

Most likely some imports can be restructured or more things put behind the existing build flag to keep the token sdk gcc independent.

```
CGO_ENABLED=0 go build ./cmd/tokengen
# github.com/hyperledger/fabric-lib-go/bccsp/pkcs11
../../../../pkg/mod/github.com/hyperledger/fabric-lib-go@v1.1.3-0.20240523144151-25edd1eaf5f5/bccsp/pkcs11/pkcs11.go:41:21: undefined: pkcs11.Ctx
../../../../pkg/mod/github.com/hyperledger/fabric-lib-go@v1.1.3-0.20240523144151-25edd1eaf5f5/bccsp/pkcs11/pkcs11.go:51:23: undefined: pkcs11.SessionHandle
../../../../pkg/mod/github.com/hyperledger/fabric-lib-go@v1.1.3-0.20240523144151-25edd1eaf5f5/bccsp/pkcs11/pkcs11.go:52:22: undefined: pkcs11.SessionHandle
../../../../pkg/mod/github.com/hyperledger/fabric-lib-go@v1.1.3-0.20240523144151-25edd1eaf5f5/bccsp/pkcs11/pkcs11.go:55:32: undefined: pkcs11.ObjectHandle
../../../../pkg/mod/github.com/hyperledger/fabric-lib-go@v1.1.3-0.20240523144151-25edd1eaf5f5/bccsp/pkcs11/pkcs11.go:332:51: undefined: pkcs11.SessionHandle
../../../../pkg/mod/github.com/hyperledger/fabric-lib-go@v1.1.3-0.20240523144151-25edd1eaf5f5/bccsp/pkcs11/pkcs11.go:344:46: undefined: pkcs11.SessionHandle
../../../../pkg/mod/github.com/hyperledger/fabric-lib-go@v1.1.3-0.20240523144151-25edd1eaf5f5/bccsp/pkcs11/pkcs11.go:376:50: undefined: pkcs11.SessionHandle
../../../../pkg/mod/github.com/hyperledger/fabric-lib-go@v1.1.3-0.20240523144151-25edd1eaf5f5/bccsp/pkcs11/pkcs11.go:391:51: undefined: pkcs11.SessionHandle
../../../../pkg/mod/github.com/hyperledger/fabric-lib-go@v1.1.3-0.20240523144151-25edd1eaf5f5/bccsp/pkcs11/pkcs11.go:690:72: undefined: pkcs11.ObjectHandle
../../../../pkg/mod/github.com/hyperledger/fabric-lib-go@v1.1.3-0.20240523144151-25edd1eaf5f5/bccsp/pkcs11/pkcs11.go:699:77: undefined: pkcs11.ObjectHandle
../../../../pkg/mod/github.com/hyperledger/fabric-lib-go@v1.1.3-0.20240523144151-25edd1eaf5f5/bccsp/pkcs11/pkcs11.go:699:77: too many errors
```

SuyashAlphaC:
Closes #1445.

## Summary

The `artifacts` subcommand of `tokengen` is the only subcommand that pulls the Fabric Smart Client `integration/nwo` topology framework — and its transitive Google Cloud / benchmarking tooling — into the `tokengen` binary. The other subcommands (`pp`, `certfier`, `version`) do not need any of that.

Per @adecaro's direction on the issue thread ("_please, go ahead with the splitting of the tools_"), this PR moves the artifact generator into a standalone `cmd/artifactgen` binary and removes it from `tokengen`'s root command. No functional changes to the subcommand itself — the files are moved with `git mv` so history is preserved.

## Impact

| Metric | Before | After | Δ |
| --- | ---: | ---: | --- |
| `cmd/tokengen` transitive dep count | 1762 | 869 | **−893 (~51%)** |
| `CGO_ENABLED=0 go build ./cmd/tokengen` | passes | passes | preserved |
| `cmd/tokengen` binary linkage | statically linked | statically linked | preserved |

The removed dependency set is predominantly `integration/nwo` and its tail (`cloud.google.com/*`, `filippo.io/*`, `debug/dwarf`, …) — transitively pulled in only by `artifactgen`.

## Notes on scope

- On current `main` (v0.10.1), `CGO_ENABLED=0 go build ./cmd/tokengen` already succeeds — the fabric-smart-client work referenced in the original issue has landed. This PR is therefore the structural cleanup @adecaro asked for on the issue thread, not a build-breakage fix. I think it is still worth landing because it halves `tokengen`'s dependency surface and removes the cgo-capable `integration/nwo` tree from the binary entirely.
- `cmd/tokengen` still pulls a narrow subset of `integration/nwo` transitively via `pp` (chaincode packager) and `certfier` (topology-aware keygen). Slimming those further is a separate concern and intentionally out of scope here.
- @arner's secondary suggestion on the thread — _"maybe the `integration` package should be treated as internal?"_ — is likewise not addressed here. Happy to follow up in a separate issue if there is interest.

## Changes

- `git mv cmd/tokengen/cobra/artifactgen/** cmd/artifactgen/**` (renames only, history preserved)
- new `cmd/artifactgen/main.go` standalone entrypoint, mirroring `tokengen`'s cobra / viper wiring shape for consistency
- drop the `artifactgen/gen` import and `mainCmd.AddCommand(gen.Cmd())` from `cmd/tokengen/main.go`
- add `make artifactgen` target alongside `make tokengen`
- update both READMEs, with a cross-link from `cmd/tokengen/README.md` to `cmd/artifactgen/README.md`

## Test plan

Build:

- [x] `CGO_ENABLED=0 go build ./cmd/tokengen` passes; resulting binary is statically linked (`file` reports `statically linked`; `ldd` reports `not a dynamic executable`).
- [x] `CGO_ENABLED=0 go build ./cmd/artifactgen` passes; statically linked.
- [x] `CGO_ENABLED=0 go build ./...` passes for the whole repo (no regression anywhere outside `cmd/`).

Tests and linters:

- [x] `go test -count=1 ./cmd/tokengen/... ./cmd/artifactgen/...` — all 12 packages green.
- [x] `go vet ./...` — clean.
- [x] `make checks` — clean (license, gofmt, goimports, vet, misspell, ineffassign, staticcheck).
- [x] `make lint-auto-fix` — reports `0 issues.`

Runtime smoke-tests:

- [x] `tokengen --help` lists only `gen`, `update`, `pp`, `certifier-keygen`, `version`.
- [x] `tokengen artifacts` returns a clean cobra error: `Error: unknown command "artifacts" for "tokengen"`.
- [x] `artifactgen artifacts --help` preserves the original flag set (`-t/--topology`, `-o/--output`, `-p/--port`) and their defaults.
- [x] `artifactgen artifacts` (no args) returns the original behaviour: `Error: expecting topology file path`.


adecaro:
Hi @SuyashAlphaC , thanks for this effort.
Please, update the makefile to make sure tokengen is build with `CGO_ENABLED=0` so that we are sure it works. I would also include the building of the tool into the CI so that we are sure we are not breaking this condition in the future.
What do you think?

Thanks 🙏

SuyashAlphaC:
Thanks for the pointer — agreed on both, done in `c42d9a0c`:

- **`Makefile`**: `make tokengen` and `make artifactgen` now invoke `go install` with `CGO_ENABLED=0`, so the invariant is enforced at developer-experience level regardless of the user's environment.
- **CI**: added a new `cgo-check` job in `.github/workflows/tests.yml` that runs `CGO_ENABLED=0 go build ./cmd/tokengen` and the same for `./cmd/artifactgen` on every push and pull request. It runs independently of `checks` so a regression on the no-cgo invariant surfaces fast and in isolation.

Two small judgement calls I made, happy to revise if you'd prefer otherwise:

1. I applied the `CGO_ENABLED=0` pin to **both** `tokengen` and `artifactgen`. You only named `tokengen`, but `artifactgen` also builds cgo-free today and it seemed more consistent to guard both. If you'd rather leave `artifactgen` as a plain `go install`, one-line revert.
2. The CI job uses raw `CGO_ENABLED=0 go build ./cmd/tokengen` rather than `make tokengen`. Reason: this way CI still catches a regression even if someone later edits the Makefile to remove the pin. The Makefile change is the ergonomics layer; the CI step is the invariant enforcement. Again, happy to switch to calling `make` if you'd prefer the single source of truth.

Locally verified: `make tokengen` / `make artifactgen` produce statically-linked binaries (`ldd` reports "not a dynamic executable"), `CGO_ENABLED=0 go build ./...` passes for the whole repo, `make checks` is clean.

adecaro:
Hi @SuyashAlphaC , this is interesting, both tools don't need CGO anymore?

adecaro:
Hi @arner , please, can you review this and confirm that it satisfies your requirements? Thanks 🙏

SuyashAlphaC:
Correct — neither `tokengen` nor `artifactgen` pulls any cgo-dependent code today, on any of the four subcommands (`gen`/`update`/`pp/*`, `certifier-keygen`, `version`) or on `artifacts` itself.

Evidence from the rebased branch (commit `bb1be546`):

```
$ CGO_ENABLED=0 go build -o /tmp/tokengen    ./cmd/tokengen
$ CGO_ENABLED=0 go build -o /tmp/artifactgen ./cmd/artifactgen
$ file /tmp/tokengen /tmp/artifactgen
/tmp/tokengen:    ELF 64-bit LSB executable, x86-64, ..., statically linked, ...
/tmp/artifactgen: ELF 64-bit LSB executable, x86-64, ..., statically linked, ...
$ ldd /tmp/tokengen
    not a dynamic executable
$ ldd /tmp/artifactgen
    not a dynamic executable
```

The `pkcs11`-adjacent packages still appear in `artifactgen`'s transitive dep graph (via `integration/nwo` → `fabric-smart-client`), but they compile cleanly under `CGO_ENABLED=0` because the cgo-using bits are guarded by build tags — same situation as `tokengen`. That's why the new `cgo-check` CI job builds both without failing.

If you'd prefer to keep the scope strictly to `tokengen` and drop the `artifactgen` pin + its CI step, happy to do that — it's a one-line revert on the Makefile and two steps on the workflow. My reasoning for including `artifactgen` was consistency (same class of tool, same no-cgo guarantee) rather than a hard requirement of issue #1445.

adecaro (approved):
LGTM. @arner , please, have your pass. Thanks 🙏

adecaro:
all right, I talked with @arner  and he is fine. I'll merge as soon as the CI gives the okay 😄

Issue #889, opened 2025-02-28
`ttx` service's todos


adecaro:
This Github Issue asks the following:
- Create a Github Issue for each TODO in the service
- Address each TODO with a relative PR.

AkramBitar:
Moving to backlog since the only issue that we have https://github.com/LFDT-Panurus/panurus/issues/301 is in backlog

PR #1588, opened 2026-04-23
fix(ttx/recipients): audit initiator/responder validation and remove stale TODO

### LINKED ISSUE #918: ttx service: deprecated todos
The following todos can be removed as not relevant anymore:
- // TODO: check the other values too

SuyashAlphaC:
## Summary

Closes #918 (last child of #889).

Per the direction @adecaro gave in [this comment](https://github.com/hyperledger-labs/fabric-token-sdk/issues/918#issuecomment) — "analyze the views in `recipients.go`" and "make sure both the initiator and the responder perform all the checks required" — this PR treats the stale `// TODO: check the other values too` on line 350 as the starting point for a symmetric validation audit of the initiator/responder pair, rather than as a one-line comment deletion.

## What the audit found

`recipients.go` has four receive paths where `RecipientData` crosses a trust boundary:

| # | Where | Direction | What is (or isn't) validated today |
|---|---|---|---|
| 1 | `RequestRecipientIdentityView.callWithRecipientData` line 215+ | initiator receives from responder | `Identity↔AuditInfo` match via `wm.RegisterRecipientIdentity → MatchIdentity`. No view-level consistency check relating the requested peer to the received identity. |
| 2 | `RespondRequestRecipientIdentityView.Call`, `RecipientData != nil` branch (the TODO's home) | responder receives from initiator | `w.Contains(Identity)`. Other fields **are not** trusted locally in this branch — they are echoed back on the wire and never persisted. |
| 3 | `RespondRequestRecipientIdentityView.handleMultisig` line 400+ | responder receives multisig data | `Identity↔AuditInfo` match via `RegisterRecipientIdentity`. Sub-identities are registered individually, so each one is matched too. |
| 4 | `ExchangeRecipientIdentities*View` pair | both sides receive from each other | `Identity↔AuditInfo` match via `RegisterRecipientIdentity`. |

The underlying `wm.RegisterRecipientIdentity(...)` chain already performs the cryptographic `MatchIdentity(Identity, AuditInfo)` check in `token/services/identity/wallet/service.go`, so the "AuditInfo consistent with Identity" invariant on all receive paths was already enforced at the wallet layer. The view layer nonetheless had three concrete gaps, one of which was the TODO.

### 1. Replace the TODO at `recipients.go:350` with an explanation of why it's obsolete

The responder never persists `AuditInfo`, `TokenMetadata`, or `TokenMetadataAuditInfo` in this branch — it echoes them back on the wire and binds the endpoint resolver using `Identity` alone. There is nothing local to validate the other fields against, so the TODO is a false alarm. The new comment makes that invariant explicit so a future reader doesn't re-discover the confusion.

### 2. Nil-guard on `RespondExchangeRecipientIdentitiesView.Call`

Previously, `request.RecipientData.Identity` was dereferenced (line 606) *before* any code that could have rejected a nil `RecipientData`. A peer sending `ExchangeRecipientRequest{RecipientData: nil}` would panic the responder rather than surface an error. `RegisterRecipientIdentity`'s own `ErrNilRecipientData` guard never got a chance to fire because the nil-deref happened first. Added a view-level check so the panic becomes `errors.Errorf("missing recipient data in exchange request")`.

### 3. Forward `TokenMetadataAuditInfo` in the rebuilt `RecipientData` literal

The same function rebuilds a `RecipientData` struct by hand with three of the four fields copied from `request.RecipientData`, silently dropping `TokenMetadataAuditInfo`. Audit information that the wire carried (and that the peer spent effort producing) was being lost at the responder. The literal now forwards all four fields.

### 4. Extend `TestRecipientRequest_BytesRoundTrip`

The existing test set `Identity` and `AuditInfo` and asserted those two survived a round-trip. Extended to also set and assert `TokenMetadata` and `TokenMetadataAuditInfo`, closing a pre-existing wire-level coverage gap that parallels change 3. (The sibling `TestExchangeRecipientRequest_BytesRoundTrip` already covers all four.)

## What is intentionally *not* in this PR

The audit surfaced two further considerations that I believe need maintainer design input and so are out of scope here:

- **Peer-identity consistency on the initiator side** (boundary 1 and 4a): the initiator asks for an identity "from this peer" but never asserts the received `Identity` is related to the peer it asked. Whether such a check is possible or even desirable depends on the TMS's anonymity model (in an idemix-style setup it may be by-design that no linkability exists). This is a design question, not a bug fix.
- **Freshness / integrity of the multisig second message** (boundary 3): there is no binding between the initial `RecipientRequest` exchange and the subsequent `MultisigRecipientData` message. If that's the intended trust model, fine; if not, it needs a separate design discussion.

Both would dilute the scope of #918 and are better raised as follow-up issues if you think they're worth pursuing.

## Impact

- Behaviour change 1: `RespondExchangeRecipientIdentitiesView` now returns a descriptive error on a malformed (nil-`RecipientData`) request instead of panicking. Error path only; no change on the happy path.
- Behaviour change 2: `TokenMetadataAuditInfo` now survives an exchange-recipient-identities round trip. Any downstream consumer that expected the field to be absent post-exchange would break — but the field was silently dropped by bug, not by design, so this is restoration of intended behaviour.
- Changes 3 (the TODO replacement) and 4 (the test extension) are no-ops for runtime behaviour.

## Test plan

Run locally on `918-ttx-recipients-audit` branch off `main@c9c9ff53`:

- [x] `go vet ./token/services/ttx/...` — clean
- [x] `CGO_ENABLED=0 go build ./...` — full-module build clean (preserves the #1582 invariant)
- [x] `go test ./token/services/ttx/...` — all pass, including the extended `TestRecipientRequest_BytesRoundTrip`
- [x] `make checks` — license / gofmt / goimports / go vet / misspell / ineffassign / staticcheck all clean
- [x] `make lint-auto-fix` — `0 issues`

## Related

- Follow-up to #1582.
- Last remaining child of the `ttx` TODO cleanup project #889 — merging this lets #889 close.

### DIFF
--- a/token/services/ttx/recipients.go
+++ b/token/services/ttx/recipients.go
@@ -347,7 +347,6 @@ func (s *RespondRequestRecipientIdentityView) Call(context view.Context) (interf
 		if !w.Contains(context.Context(), recipientIdentity) {
 			return nil, errors.Errorf("cannot find identity [%s] in wallet [%s:%s]", recipientIdentity, wallet, recipientRequest.TMSID)
 		}
-		// TODO: check the other values too
 	} else {
 		logger.DebugfContext(context.Context(), "generate_identity")
 		// otherwise generate one fresh
@@ -596,13 +595,19 @@ func (s *RespondExchangeRecipientIdentitiesView) Call(context view.Context) (int
 		return nil, err
 	}
 
+	if request.RecipientData == nil {
+		return nil, errors.Errorf("missing recipient data in exchange request")
+	}
 	ts, err := token.GetManagementService(context, token.WithTMSID(request.TMSID))
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get token management service")
 	}
 	other := request.RecipientData.Identity
 	if err := ts.WalletManager().RegisterRecipientIdentity(context.Context(), &RecipientData{
-		Identity: other, AuditInfo: request.RecipientData.AuditInfo, TokenMetadata: request.RecipientData.TokenMetadata,
+		Identity:               other,
+		AuditInfo:              request.RecipientData.AuditInfo,
+		TokenMetadata:          request.RecipientData.TokenMetadata,
+		TokenMetadataAuditInfo: request.RecipientData.TokenMetadataAuditInfo,
 	}); err != nil {
 		return nil, err
 	}
--- a/token/services/ttx/recipients_test.go
+++ b/token/services/ttx/recipients_test.go
@@ -94,8 +94,10 @@ func TestRecipientRequest_BytesRoundTrip(t *testing.T) {
 		TMSID:    token.TMSID{Network: "net2", Channel: "ch2", Namespace: "ns2"},
 		WalletID: []byte("wallet-id"),
 		RecipientData: &RecipientData{
-			Identity:  view.Identity("bob"),
-			AuditInfo: []byte("bob-audit"),
+			Identity:               view.Identity("bob"),
+			AuditInfo:              []byte("bob-audit"),
+			TokenMetadata:          []byte("bob-token-meta"),
+			TokenMetadataAuditInfo: []byte("bob-meta-audit"),
 		},
 		MultiSig: true,
 	}
@@ -113,6 +115,8 @@ func TestRecipientRequest_BytesRoundTrip(t *testing.T) {
 	require.NotNil(t, decoded.RecipientData)
 	assert.Equal(t, original.RecipientData.Identity, decoded.RecipientData.Identity)
 	assert.Equal(t, original.RecipientData.AuditInfo, decoded.RecipientData.AuditInfo)
+	assert.Equal(t, original.RecipientData.TokenMetadata, decoded.RecipientData.TokenMetadata)
+	assert.Equal(t, original.RecipientData.TokenMetadataAuditInfo, decoded.RecipientData.TokenMetadataAuditInfo)
 }
 
 func TestRecipientRequest_FromBytes_InvalidInput(t *testing.T) {


token/services/ttx/recipients.go
```diff
@@ -600,9 +603,15 @@ func (s *RespondExchangeRecipientIdentitiesView) Call(context view.Context) (int
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get token management service")
 	}
+	if request.RecipientData == nil {
```
adecaro: let's move this check before `ts, err := token.GetManagementService(context, token.WithTMSID(request.TMSID))` so to avoid to allocate resources that might not be needed.

token/services/ttx/recipients.go
```diff
@@ -347,7 +347,10 @@ func (s *RespondRequestRecipientIdentityView) Call(context view.Context) (interf
 		if !w.Contains(context.Context(), recipientIdentity) {
 			return nil, errors.Errorf("cannot find identity [%s] in wallet [%s:%s]", recipientIdentity, wallet, recipientRequest.TMSID)
 		}
-		// TODO: check the other values too
+		// The other fields of recipientData (AuditInfo, TokenMetadata, TokenMetadataAuditInfo)
```
adecaro: I disagree with this analysis. At the current stage, an attacher can just send a recipient data with only the identity set and get back all the rest, it should not happen. 

I would also update the protocol to avoid to echo back recipient data at all. An ack is enough. 

We still have issues, but I would start with these fixes and then move to patch the rest. 

What do you think?

token/services/ttx/recipients.go
SuyashAlphaC: Done — reordered in the push just now.

token/services/ttx/recipients.go
SuyashAlphaC: Fair — the analysis comment overclaims. Dropped it in the push just now (no replacement comment; the TODO removal itself closes the literal #918 ask).

You're right that the echo is a problem in its own right: whatever the initiator puts on the wire (or doesn't) gets blessed on the round-trip and fed into `RegisterRecipientIdentity` on the initiator side, with no local source of truth. Replacing the echo with an ack is the clean fix — the initiator already has the `RecipientData` when it pre-supplied one, and for the generated case the responder only needs to send back `Identity` so the endpoint resolver can bind.

On scope, two options:

1. **Keep this PR focused on the small fixes** (nil-guard + ordering, `TokenMetadataAuditInfo` forwarding, drop the stale TODO). Do the echo → ack protocol change in a follow-up PR against a fresh issue, since it touches the wire format and both caller sides of `RequestRecipientIdentityView` / `RespondRequestRecipientIdentityView` (and possibly the exchange pair too).
2. **Fold the echo removal into this PR.** I can do it, but it'll grow the diff meaningfully and the wire change deserves its own commit for bisect-ability.

I'd lean toward (1) because it keeps #918 closable on a small diff and gives the protocol change its own reviewable unit — but happy to go with (2) if you'd prefer to bundle.

adecaro: HI @SuyashAlphaC , was the first line written by an AI? 😅 

Anyhow, I think this thread shows that the Issue was wrong. This TODO is quite relevant and needs to be addressed.

I would suggest the following: I'll close that issues as not relevant. Then, I'll create another issue describing the actual problems.  Would that be okay?

SuyashAlphaC: tbh, I am using grammarly for rephrasing my comments.

SuyashAlphaC: Yes, please — close #918 and reopen with the real problem framing. You're right
that the audit shows the issue was mis-scoped.

I just have a quick question: do you want this PR retargeted against the new issue once it's up
(possibly trimmed), or closed and started fresh? Happy with either — just want the
three concrete fixes (nil-guard, `TokenMetadataAuditInfo`, TODO) to land somewhere.

SuyashAlphaC: I'm myself learning how to be professional in terms of writting comments, having conversations. You know, having a good level of formal communication skills do indeed matter. So just getting trained  from grammarly, new phrases each time.

adecaro: Hi @SuyashAlphaC , yes, we can repurpose this PR. I'll create the issue worst case on Monday.

SuyashAlphaC: Hi @adecaro — gentle follow-up on your note about repurposing this PR: whenever you have a moment, could you please open/link the new issue so I can retarget this PR accordingly?  
    Happy to trim/scope the diff to match that issue.

adecaro: Hi @SuyashAlphaC , apologies. Time flies as they say. I'll work on this ASAP. Thanks for being patient 🙏

adecaro:
Hi @SuyashAlphaC , here is the Issue: https://github.com/hyperledger-labs/fabric-token-sdk/issues/1606
If you respond to it, I can assign it to you. 

So, if it is okay for you, I would close this PR, and the related Issue, and address the rest in the context of the other issue. What do you think?

SuyashAlphaC:
Thanks @adecaro — this works for me.  
I’ve responded on #1606; please assign it to me and I’ll handle the work there.  
Agree to close this PR and #918 and continue in the new issue.

adecaro:
closing this as agreed

PR #1596, opened 2026-04-24
feat(storage): advisory lock when creating tables #1595

### LINKED ISSUE #1595: feat(storage): advisory lock when creating tables
In a replicated setting, when multiple replicas are trying to instantiate the storage services, they might end up creating tables at the same time. In this case, Postgres can generate an error like: `duplicate key value violates unique constraint "pg_type_typname_nsp_index" (SQLSTATE 23505)`. To avoid this, it is sufficient to enhance the tables creation script with the acquisition of an advisory lock.

adecaro:
This PR closes #1595 by prefixing the SQL script for table creation with an advisory lock statement.

### DIFF
--- a/token/services/storage/db/sql/postgres/transactions.go
+++ b/token/services/storage/db/sql/postgres/transactions.go
@@ -24,8 +24,18 @@ import (
 	sqlcommon "github.com/hyperledger-labs/fabric-token-sdk/token/services/storage/db/sql/common"
 )
 
-// AuditTransactionStore is an alias for common.TransactionStore.
-type AuditTransactionStore = sqlcommon.TransactionStore
+// AuditTransactionStore wraps common.TransactionStore to add advisory lock to schema creation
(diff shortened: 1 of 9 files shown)


token/services/storage/db/sql/postgres/transactions.go
```diff
@@ -58,18 +87,28 @@ func NewTransactionStoreWithNotifier(dbs *scommon.RWDB, tableNames sqlcommon.Tab
...
 		postgres.NewPaginationInterpreter(),
 	)
+	if err != nil {
+		return nil, err
+	}
+
+	return &AuditTransactionStore{
+		TransactionStore: baseStore,
+		writeDB:          dbs.WriteDB,
+		lockID:           createTableLockID("audittx"),
```
AkramBitar, line 110: Maybe to add constant for these dbs "audittx". What do you think?

AkramBitar (approved):

PR #1607, opened 2026-04-28
docs(ttx): document recipient identity message flows (#1606)

### LINKED ISSUE #1606: bug(ttx/recipients.go): should not echo back information
Currently, then the responder hits the following line:

```go
	// if the initiator send a recipient data, check that the identity has been already registered locally.
	if recipientRequest.RecipientData != nil {
```

It checks if the identity is known and if it is known, then the responder echos back the entire RecipientData (including private information)
Then, an attacher can just send a recipient data with only the identity set and get back all the rest, it should not happen.
The protocol should only send back an ack.

The scope of this Issue is to create a flow diagram of the messages exchanged by the initiator and responder views defined in `ttx/recipients.go` and to be added to the documentation of the `ttx` service. 
This way, we can better analyze the flow of information and find critical security issues.

SuyashAlphaC:
## Summary
  Closes #1606.
  Documents the initiator/responder message flows implemented in `token/services/ttx/recipients.go` in `docs/services/ttx.md`:
  - `RequestRecipientIdentityView` / `RespondRequestRecipientIdentityView` — request/response, echo vs wallet-generated paths, optional multisig follow-up
  - `ExchangeRecipientIdentitiesView` / `RespondExchangeRecipientIdentitiesView` — exchange request/response and bindings
  Also describes which `RecipientData` fields may appear on the wire and clarifies the echo path (responder does not merge wallet fields into the echoed object). Notes that tightening the protocol (e.g. minimal or ack-only 
  responses) is follow-up implementation work beyond this documentation change.
  ## Test plan
  - [x] `make checks`

### DIFF
--- a/docs/services/ttx.md
+++ b/docs/services/ttx.md
@@ -70,6 +70,95 @@ The `RequestRecipientIdentityView` allows an initiator to request a fresh recipi
 *   The initiator receives the necessary audit information to verify the identity's validity against the TMS public parameters.
 *   Both parties update their local `IdentityInfo` stores to map the new identity to the correct FSC node.
 
+### Recipient Protocol Flows (`recipients.go`)
+
+The recipient identity protocols are implemented in `token/services/ttx/recipients.go`. The diagrams below document the on-wire messages exchanged by the initiator and responder views.
+`RecipientData` can carry `Identity`, `AuditInfo`, `TokenMetadata`, and `TokenMetadataAuditInfo`.
+
+Wire messages use JSON sessions (`token/services/utils/json/session`); the diagrams name the Go types being sent or received.
+
+**Response paths (today).** In `RespondRequestRecipientIdentityView`, after the wallet lookup:
+
+- If `recipientRequest.RecipientData != nil`, the responder checks `OwnerWallet.Contains` for `RecipientData.Identity`, then sends **the same** `RecipientData` value back on the session (echo path). The responder does **not** substitute wallet-held `AuditInfo` / metadata into that payload; what goes on the wire is the initiator-supplied structure (after the contains check).
+- If `recipientRequest.RecipientData == nil`, the responder calls `OwnerWallet.GetRecipientData` and sends that **wallet-produced** `RecipientData` (fresh path).
+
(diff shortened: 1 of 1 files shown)


adecaro:
Hi @SuyashAlphaC , great effort. I double checked the diagrams and they match. The echo is not actually leaking any auditInfo (or any other secret) that the initiator didn't know before (The initiator can still check if a given identity exist at the responder. We can avoid this by forcing the initiator to send a signature on a fresh nonce). 

The diagram though identifies other potential issues:
- We need to double check the implementation of `RegisterRecipientIdentity`. There is an issue there in the order of operations that might leave the DBs in a dirty state
- We need a way to cleanup in case of a failure. Even checking if the recipient identities are actually bound.

I can merge this, but I want to understand first your plan for the next PRs.

Thanks for this great effort 🙏

SuyashAlphaC:
Thanks for the review 🙏 — glad the diagrams line up with the code.

  I agree on the echo branch: it’s the initiator-supplied ```RecipientData``` after ```Contains```, not values merged in from the wallet. I still like the longer-term move toward a slimmer wire response. Your nonce/signature idea I’d keep as a separate protocol discussion as it changes what both sides prove (freshness, binding, replay behavior), touches initiator and responder and likely compatibility with older peers, and needs an agreed message shape—not something to fold into a small cleanup PR next to persistence ordering.

  On RegisterRecipientIdentity: I looked at ```token/services/identity/wallet/service.go``` — it registers the identity with the provider, then ```MatchIdentity```, then ```RegisterRecipientData```. If step 1 succeeds and step 2 fails, we can end up half-done depending on what the provider persists. I’d like to trace the provider implementation next and open a focused PR on either reordering, compensating cleanup, or documenting the invariants + tests.

  On the views: there’s also ordering risk around ```session.Send``` vs ```endpoint.Bind / multisig``` — happy to do a pass on failure paths once you say whether you want that before or after the registration work.

  Rough order I’d suggest: land this docs PR, then a small PR on registration safety, then view-layer cleanup, then the bigger ack / protocol change — unless you’d rather flip the last two. Happy to follow whatever priority you
  prefer.

adecaro (approved):
LGTM

adecaro:
Perfect, let's proceed as you planned. Thanks for the effort 🙏

PR #1614, opened 2026-04-29
 fix(identity): order recipient registration after MatchIdentity and add rollback

### LINKED ISSUE #1615: Harden RegisterRecipientIdentity: validate match before provider, rollback on partial failure
## Background
  Third-party recipient registration goes through the wallet service, which talks to the configured `IdentityProvider` and `Deserializer`. The order of steps and behaviour on failure should avoid leaving the provider in a half-applied state.
  ## Problem
  Today the flow can register the recipient identity with the provider **before** `MatchIdentity` runs. If matching fails, the provider may already have recorded partial side effects (for example in-memory marks).
  If `RegisterRecipientData` fails after `RegisterRecipientIdentity` has succeeded, there is no defined way to undo the provider-side effects of that first step.
  ## Goal
  1. Run **`MatchIdentity` before** any **`IdentityProvider.RegisterRecipientIdentity`** call, so a mismatch never touches provider state.
  2. Keep persisting recipient data via **`RegisterRecipientData`** after a successful registration step.
  3. If **`RegisterRecipientData` fails** after **`RegisterRecipientIdentity` succeeded**, implementations that store partial state in **`RegisterRecipientIdentity`** must be able to **roll back** that partial registration (for 
  example clearing a cache mark). Providers that have nothing to undo should not be forced into a heavy contract beyond documenting a no-op.
  4. Add **tests** that lock in: no provider registration on failed match; rollback (or equivalent) when data registration fails after identity registration succeeds, for providers that support rollback.
  ## Scope (code)
  - `token/services/identity/wallet/service.go` — registration sequence and error handling  
  - `token/services/identity/provider.go` (and related types) — concrete provider behaviour where rollback applies  
  - Tests under `token/services/identity/wallet` and `token/services/identity` as appropriate  
  ## Non-goals
  - Changing the on-wire TTX recipient protocol or multisig flows (separate work).  
  - Broader persistence or store-layer integrity (unless required to implement the rollback above).

SuyashAlphaC:
Closes #1615

 ## Summary
  Reorders `RegisterRecipientIdentity` so **`MatchIdentity` runs before any `IdentityProvider` registration**, and invokes **`RecipientRegistrationRollback`** when **`RegisterRecipientData` fails** after a successful 
  **`RegisterRecipientIdentity`**, so the identity provider is not left in a half-applied state when supported.
  Follow-up to the discussion on [PR #1607](https://github.com/hyperledger-labs/fabric-token-sdk/pull/1607).
  ## Problem
  Previously the flow was:
  1. `RegisterRecipientIdentity` (identity provider)
  2. `MatchIdentity` (deserializer)
  3. `RegisterRecipientData` (identity provider / storage)
  If **`MatchIdentity` failed**, the provider had **already** run **`RegisterRecipientIdentity`**. For implementations that record partial side effects there (for example in-memory marks), that is unnecessary work and can leave **inconsistent state** if a later step never succeeds.
  If **`RegisterRecipientData` failed** after **`RegisterRecipientIdentity` succeeded**, there was **no rollback** of provider-side partial registration.
  ## Solution
  - Run **`MatchIdentity` first** so a mismatch **does not touch** the identity provider.
  - Keep **`RegisterRecipientIdentity`** then **`RegisterRecipientData`**.
  - If **`RegisterRecipientData` returns an error** after **`RegisterRecipientIdentity` succeeded**, call **`RollbackPartialRecipientRegistration`** when the provider implements **`identity.RecipientRegistrationRollback`** (the concrete **`identity.Provider`** clears the recipient **`isMeCache`** entry; with the default **`NoCache`** this is a no-op but remains correct for real caches).
  ## Changes
  | Area | Change |
  |------|--------|
  | `token/services/identity/wallet/service.go` | New order + rollback on `RegisterRecipientData` failure |
  | `token/services/identity/recipient_registration_rollback.go` | New optional rollback interface |
  | `token/services/identity/provider.go` | Implement rollback + compile-time assertion |
  | `token/services/identity/wallet/service_test.go` | Tests for ordering and rollback |
  | `token/services/identity/provider_test.go` | Smoke test for rollback hook |
  ## Test plan
  - [x] `go test ./token/services/identity/wallet/... ./token/services/identity -count=1 -short`
  - [x] `go vet ./...` — clean.
  - [x] `make checks` — clean (license, gofmt, goimports, vet, misspell, ineffassign, staticcheck).
  - [x] `make lint-auto-fix` — reports `0 issues.`

  ## Runtime smoke-tests:
  - [x] `tokengen --help` lists only `gen`, `update`, `pp`, `certifier-keygen`, `version`.
  - [x] `tokengen artifacts` returns a clean cobra error: `Error: unknown command "artifacts" for "tokengen"`.
  - [x] `artifactgen artifacts --help` preserves the original flag set (`-t/--topology`, `-o/--output`, `-p/--port`) and their defaults.
  - [x] `artifactgen artifacts` (no args) returns the original behaviour: `Error: expecting topology file path`.

### DIFF
--- a/token/services/identity/provider.go
+++ b/token/services/identity/provider.go
@@ -24,6 +24,8 @@ import (
 var (
 	// This makes sure that Provider implements driver.IdentityProvider
 	_ driver.IdentityProvider = &Provider{}
+	// This makes sure that Provider can roll back partial recipient registration
+	_ RecipientRegistrationRollback = &Provider{}
 )
 
 // StorageProvider returns storage services scoped to a specific token
@@ -225,6 +227,13 @@ func (p *Provider) RegisterRecipientIdentity(ctx context.Context, id driver.Iden
 	p.isMeCache.Add(id.UniqueID(), false)
 
 	return nil
+}
+
+// RollbackPartialRecipientRegistration clears in-memory marks written by
+// RegisterRecipientIdentity when RegisterRecipientData did not complete.
+func (p *Provider) RollbackPartialRecipientRegistration(ctx context.Context, id driver.Identity) {
+	p.Logger.DebugfContext(ctx, "rolling back partial recipient registration for identity [%s]", id)
+	p.isMeCache.Delete(id.UniqueID())
 }
 
 // RegisterIdentityDescriptor stores the given identity descriptor in the configured storage.
--- a/token/services/identity/provider_test.go
+++ b/token/services/identity/provider_test.go
@@ -174,6 +174,21 @@ func TestProvider_EnrollmentIDHelpers(t *testing.T) {
 	assert.Equal(t, "rh2", erh)
 }
 
+func TestProvider_RollbackPartialRecipientRegistration(t *testing.T) {
+	storage := &idmock.Storage{}
+	des := &idmock.Deserializer{}
+	nbs := &idmock.NetworkBinderService{}
+	eidu := &idmock.EnrollmentIDUnmarshaler{}
+
+	p := identity.NewProvider(logging.MustGetLogger(), storage, des, nbs, eidu)
+
+	id := driver.Identity("recipient_id")
+	require.NoError(t, p.RegisterRecipientIdentity(t.Context(), id))
+
+	var rb identity.RecipientRegistrationRollback = p
+	rb.RollbackPartialRecipientRegistration(t.Context(), id)
+}
+
 func TestProvider_GetAuditInfo(t *testing.T) {
 	storage := &idmock.Storage{}
 	des := &idmock.Deserializer{}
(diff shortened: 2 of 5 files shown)


adecaro:
Hi @SuyashAlphaC , please, open an Issue that describes what we want to do here. Thanks 🙏

adecaro (approved):
LGTM. Thanks for the effort @SuyashAlphaC .

As a side note, currently `isMeCache` is instantiated as `no cache` meaning that we are not really using any caching to speed up lookups, for example. 
Would you like to investigate if this cache makes sense at all? We either remove it or enable it. Pay attention that ephemeral identities like the idemix identities are created and used only once.

PR #1616, opened 2026-04-29
replace mutex with context aware semaphore

### LINKED ISSUE #1609: Context-Aware Lock Acquisition with Bounded Hold Time [CRITICAL]
**Description**
The `AcquireLocks` implementation accepts a `context.Context` but passes it only to the logger; the actual `sync.RWMutex.Lock()` calls are unconditionally blocking. Replace the bare mutex acquisition with a context-deadline-aware mechanism (e.g. a channel-based trylock loop that respects `ctx.Done()`, or a semaphore from `golang.org/x/sync/semaphore`). Every call to `AcquireLocks` must propagate the caller's context so that if the context is cancelled or its deadline fires, the acquisition fails fast and returns an error instead of blocking indefinitely. This directly eliminates the indefinite-hold scenario that is the root cause of the lock-starvation threats, without requiring any change to the caller contract.

**Short Summary**
If a mutex is locked and not released (either bug or attack) following attempts to lock the mutex will hang and not return an error message to the user (or release other resources). For example this happens in the Auditor (function name: Audit).

Check the code and validate that the mutex is always released (view timeout, network timeout, crash).

HayimShaul:
address issue #1609

### DIFF
--- a/token/services/storage/auditdb/store.go
+++ b/token/services/storage/auditdb/store.go
@@ -23,6 +23,7 @@ import (
 	dbdriver "github.com/hyperledger-labs/fabric-token-sdk/token/services/storage/db/driver"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/storage/db/multiplexed"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/storage/ttxdb"
+	"golang.org/x/sync/semaphore"
 )
 
 var (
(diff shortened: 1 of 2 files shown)


token/services/storage/auditdb/store.go
```diff
@@ -266,15 +267,11 @@ func (d *StoreService) GetTokenRequest(ctx context.Context, txID string) ([]byte
...
-func (d *StoreService) GetTokenRequests(ctx context.Context, txIDs []string) (map[string][]byte, error) {
-	return d.db.GetTokenRequests(ctx, txIDs)
-}
-
 // AcquireLocks acquires locks for the passed anchor and enrollment ids.
 // This can be used to prevent concurrent read/write access to the audit records of the passed enrollment ids.
+// The function respects context cancellation and deadlines, returning an error if the context is cancelled
+// or times out before all locks can be acquired. This prevents indefinite blocking and enables fast failure
+// in case of lock contention or deadlock scenarios.
 func (d *StoreService) AcquireLocks(ctx context.Context, anchor string, eIDs ...string) error {
```
adecaro, line 282: please, add unit-tests for this function. Let's make sure the code works as expected. Thanks.

adecaro (changes requested):
Thanks @HayimShaul , please, add unit-tests.

adecaro:
Hi @HayimShaul , any update on this? thanks 🙏

adecaro (approved):
LGTM

PR #1624, opened 2026-05-02
Extend IssuerWallet with balance and redemption queries

### LINKED ISSUE #872: issuer wallet: how much this wallet issued?

### LINKED ISSUE #890: Enhance Redeem with the signature of the issuer

### LINKED ISSUE #1062: Redeem: ensure that the auditor can check the identity of the signing issuer

### LINKED ISSUE #1063: Redeem: The transactions db should correctly track the redeem operation, including the identity of the issuer in the storage
We must ensure that the transactions db tracks correctly the redeem operation. We might have to include the identity of the issuer in the storage as well

Soumya8898:
### Summary
Added changes to close #872 

This PR extends the `IssuerWallet` interface with five new methods that give issuers full visibility into their token lifecycle — from issuance through redemption — with flexible query filtering and sorting.

### Motivation

The existing `IssuerWallet` only exposed `HistoryTokens`, which lists issued tokens. There was no way for an issuer to:
- Query the **total value** of tokens they have issued
- List or sum tokens that have been **redeemed**
- Compute the **net outstanding** supply still in circulation
- Filter any of the above by **token type**, **issuer identity**, or **time range**
- Control the **sort order** of returned results

Issuers operating multi-wallet nodes or managing multiple token types had no first-class API for these common reporting needs.

### New API Surface

Five methods added to `IssuerWallet` (driver and public interfaces):

| Method | Returns | Description |
|--------|---------|-------------|
| `IssuedBalance(ctx, opts)` | `uint64` | Sum of all non-deleted tokens issued by this wallet |
| `RedeemedTokens(ctx, opts)` | `*IssuedTokens` | List of redeemed tokens originally issued by this wallet |
| `RedeemedBalance(ctx, opts)` | `uint64` | Sum of amounts of redeemed tokens |
| `OutstandingBalance(ctx, opts)` | `uint64` | Net in circulation: `IssuedBalance − RedeemedBalance` |

All methods accept `*ListTokensOptions` which now supports:

| Option | Type | Effect |
|--------|------|--------|
| `TokenType` | `token.Type` | Filter by token type (e.g. `"USD"`) |
| `SortBy` | `SortField` | `SortByDefault` or `SortByQuantity` |
| `SortDirection` | `SortDirection` | `SortAscending` or `SortDescending` |
| `From` / `To` | `*time.Time` | Restrict to tokens stored within a time range |

The issuer identity filter is applied automatically — each wallet scopes queries to its own identity.

### Data Model Changes

Added two fields to `IssuedToken`:
- `IsSpent bool` — whether the token has been deleted/spent
- `SpentBy string` — the transaction ID that spent the token

These fields enable the SQL layer to JOIN tokens with transactions and identify redemptions (`action_type = Redeem`).

### SQL Implementation

- **`IssuedBalance`**: `SELECT SUM(amount)` over non-deleted tokens with `issuer = true`, filtered by token type, issuer identity, and time range
- **`ListRedeemedTokens`**: `SELECT ... FROM tokens JOIN transactions` where `is_deleted = true AND action_type = Redeem`, with optional `ORDER BY amount ASC|DESC`
- **`RedeemedBalance`**: Same join as above but returns `SUM(amount)`
- **`OutstandingBalance`**: Computed at the wallet layer as `IssuedBalance − RedeemedBalance` (no extra SQL query)

All SQL conditions are built dynamically — empty/nil filters are skipped, so callers only supply what they need.

### Testing

- Unit tests added for `IssuedBalance`, `RedeemedTokens`, `RedeemedBalance`, and `OutstandingBalance`
- Covers: happy path, error propagation from vault, identity filtering, and edge cases (redeemed > issued guard)
- Uses existing counterfeiter mock pattern (`mock.IssuerTokenVault`)


adecaro:
Hi @Soumya8898 , thanks for this effort. I'll review ASAP.

A quick note already: Please, extend the fungible integration test `integration/token/fungible` (TestAll function), to include calls to the new APIs and test against expected values.

Thanks a lot 🙏

token/driver/vault.go
```diff
@@ -84,6 +85,20 @@ type QueryEngine interface {
...
+	// ListRedeemedTokens returns issued tokens that were spent by a Redeem action.
+	// If tokenType is non-empty, only tokens of that type are included.
+	// If issuerRaw is non-empty, only tokens issued by that identity are included.
+	// If from/to are non-nil, only tokens stored within that time range are included.
+	ListRedeemedTokens(ctx context.Context, tokenType token.Type, issuerRaw Identity, from, to *time.Time, sortBy SortField, sortDirection SortDirection) (*token.IssuedTokens, error)
+	// RedeemedBalance returns the sum of amounts of issued tokens spent by a Redeem action.
+	// If tokenType is non-empty, only tokens of that type are included.
+	// If issuerRaw is non-empty, only tokens issued by that identity are included.
+	// If from/to are non-nil, only tokens stored within that time range are included.
+	RedeemedBalance(ctx context.Context, tokenType token.Type, issuerRaw Identity, from, to *time.Time) (uint64, error)
```
adecaro, line 104: please, introduce a BalanceOpts where we collect all the possible options for all balance functions, we can pass it also to `Balance`.

Soumya8898: Splitted the options struct into two types: ListTokensOptions, used by methods that return ordered lists, and BalanceOpts, used by methods that return a single scalar total. All balance methods (OwnerWallet.Balance, IssuerWallet.IssuedBalance, RedeemedBalance, and OutstandingBalance) now accept *BalanceOpts, which includes only TokenType, From, and To, and does not contain any sorting fields.

token/wallet.go
```diff
@@ -45,6 +46,42 @@ func WithContext(ctx context.Context) ListTokensOption {
 	}
 }
 
+// WithSortBy returns a list token option that specifies the sort field and direction.
+func WithSortBy(field driver.SortField, dir driver.SortDirection) ListTokensOption {
```
adecaro, line 42: this function is not currently used. Please, execercise it either in a unit-test or an integration test. Thanks.

token/wallet.go
```diff
@@ -45,6 +46,42 @@ func WithContext(ctx context.Context) ListTokensOption {
...
+	return func(o *BalanceOpts) error {
+		o.TokenType = tokenType
+
+		return nil
+	}
+}
+
+// WithTimeRange returns a balance option that restricts the computation to
+// tokens stored within [from, to]. Nil values are treated as unbounded.
+func WithTimeRange(from, to *time.Time) BalanceOption {
```
adecaro, line 72: same thing for this

token/vault.go
```diff
@@ -121,6 +121,18 @@ func (q *QueryEngine) ListHistoryIssuedTokens(ctx context.Context) (*token.Issue
 	return q.qe.ListHistoryIssuedTokens(ctx)
 }
 
+func (q *QueryEngine) IssuedBalance(ctx context.Context, tokenType token.Type, issuerRaw driver.Identity, from, to *time.Time) (uint64, error) {
```
adecaro, line 128: please, add GoDocs to the new functions introduced

token/wallet.go
```diff
@@ -45,6 +46,42 @@ func WithContext(ctx context.Context) ListTokensOption {
...
+}
+
+// BalanceOpts contains options for balance queries.
+type BalanceOpts = driver.BalanceOpts
+
+// BalanceOption is a function that configures BalanceOpts
+type BalanceOption func(*BalanceOpts) error
+
+// WithBalanceType returns a balance option that filters by the passed token type.
+func WithBalanceType(tokenType token.Type) BalanceOption {
```
adecaro: to avoid confusion, I would call with WithBalanceTokenType

Soumya8898: changed it

token/vault.go
Soumya8898: Added

token/wallet.go
Soumya8898: Wrote tests that call the function and assert the values propagate correctly.

token/wallet.go
Soumya8898: Added

Soumya8898:
Hi @adecaro can you please give it another review. Thanks

adecaro:
Hi @Soumya8898 , there is something failing in the integration tests, please, have a look. Thanks 🙏

adecaro:
Hi @Soumya8898 , thanks for the effort so far 🙏 
There are still issues with the tests. Do you need support?

Soumya8898:
Hi @adecaro I am doing a last round of verification of all the changes, if unable to resolve it, will surely ask you for help. Thanks

Soumya8898:
> Hi @Soumya8898 , thanks for the effort so far 🙏 There are still issues with the tests. Do you need support?

Hi @adecaro i think i need support here, can you please share your views like wht is going wrong in the changes for which these panic calls are getting triggered.

adecaro:
Hi @Soumya8898 , I think I got the problem. The question is: Which are the tokens that represent a redeem? Those whose owner field is empty. Now, this means that the query performed by RedeemedBalance on the token store needs to be adjusted. Please, double check also that `extractTransferOutputs` returns two outputs as well. I haven't checked, but it is worth having a look. Why that function? Because it is used to extract the tokens from a token request. When a finality listener (`token/services/ttx/finality/listener.go`) gets invoked, `AppendValid` is invoked and there inputs and outputs are extracted from the token request. Hope it helps. This is getting interesting, actually 😄 

Thanks for the effort @Soumya8898 🙏

Soumya8898:
Hi @adecaro Thanks for these catches, let me give them a look will update here.

Soumya8898:
> Hi @Soumya8898 , I think I got the problem. The question is: Which are the tokens that represent a redeem? Those whose owner field is empty. Now, this means that the query performed by RedeemedBalance on the token store needs to be adjusted. Please, double check also that `extractTransferOutputs` returns two outputs as well. I haven't checked, but it is worth having a look. Why that function? Because it is used to extract the tokens from a token request. When a finality listener (`token/services/ttx/finality/listener.go`) gets invoked, `AppendValid` is invoked and there inputs and outputs are extracted from the token request. Hope it helps. This is getting interesting, actually 😄
> 
> Thanks for the effort @Soumya8898 🙏

Hi @adecaro 
The tokens with owner field as empty represents a redeem, so the query performed by RedeemedBalance is well-adjusted because the implementation works like tokens that were deleted (is_deleted=true) and whose spent_by transaction has NO output rows (LEFT JOIN → IS NULL) are redeemed tokens.

For extractTransferOutputs, 2 outputs will be produced for partial-redeem case, but in case of exact redeem it is 1. So It's actually getting interesting 😄, I rechecked my code with all the above insights but like those are already implmented, some other thing is there that I am missing, checking it.

Soumya8898:
Hi @adecaro I tried like all possible solution that I can think of to resolve this, but unable to. Can you please guide me here, like how to proceed?

adecaro:
Hi @Soumya8898 , I'll have a look ASAP. Thank you so much for the effort so far, I think we will learn something interesting here 🙏

adecaro:
Hi @Soumya8898 , 

I think the problem is in `token/request.go#extractTransferOutputs`. If the transferAction redeems tokens, then `transferAction..GetIssuer()` is not nil and this is the one that needs to be put in the Output field.

Now, you still need to pay attention to the query to retrieve the amount redeemed. Either you store negative values for these redeemed tokens or you need to introduce an additional flag in the table that marks them as redeemed.

Does this make sense?

You are touching quite deep code in the token-sdk, thanks for the patience 🙏

Soumya8898:
> Hi @Soumya8898 ,
> 
> I think the problem is in `token/request.go#extractTransferOutputs`. If the transferAction redeems tokens, then `transferAction..GetIssuer()` is not nil and this is the one that needs to be put in the Output field.
> 
> Now, you still need to pay attention to the query to retrieve the amount redeemed. Either you store negative values for these redeemed tokens or you need to introduce an additional flag in the table that marks them as redeemed.
> 
> Does this make sense?
> 
> You are touching quite deep code in the token-sdk, thanks for the patience 🙏

Hi @adecaro thanks for pointing these out. Let me do a deeper dive to better understand these specific cases, and then I’ll clarify things here before proceeding with the implementation. This issue has already taught me a lot, and I’m still actively working through it to resolve it.

Thanks a ton for sharing your insights, pointing out the specific implementations where the error can be triggered 🙏, and also for being patient while I work toward closing this issue 🙂

Soumya8898:
So @adecaro according to my understanding I think When a token is redeemed (burned), the individual output's metadata doesn't include who originally issued it. So need to check: if the per-output issuer is empty but the overall transfer action knows the issuer (because it's a redeem), using that action-level issuer instead. This ensures burn outputs get tagged with their issuer, which is required for RedeemedBalance queries to work correctly.

What's your thought on this?

adecaro:
HI @Soumya8898 , in a transfer operation that redeem tokens, the operation will carry the identity of the issuer that authorized the redeem. This is enough. We cannot really derive the identity of the issuer that authorized the issuance because the tokens spent during a redeem can be the results of many transfer operations. 

What do you think?

adecaro:
Hi @Soumya8898 , it is time to resurrect the work here. If it is okay for you, I will fork your branch (so keeping the history) and continue the work.

Would that be okay for you?

Thanks a lot for this initial work 🙏

Soumya8898:
> Hi @Soumya8898 , it is time to resurrect the work here. If it is okay for you, I will fork your branch (so keeping the history) and continue the work.
> 
> Would that be okay for you?
> 
> Thanks a lot for this initial work 🙏

Hey @adecaro yes you can continue the work. Thanks 🙏

adecaro:
closing this, it will be reworked later

PR #1662, opened 2026-05-06
fix(sql): replace non-atomic check-then-insert with atomic upsert in WalletStore.StoreIdentity

### LINKED ISSUE #1661: [BUG]: WalletStore.StoreIdentity uses non-atomic check-then-insert, risks duplicate key errors under concurrency
### 👾 Description

WalletStore.StoreIdentity in token/services/storage/db/sql/common/wallet.go
performs a non-atomic "check-then-insert" pattern:

  if db.IdentityExists(ctx, identity, wID, roleID) {
      return nil
  }
  // ... INSERT ...

Between the IdentityExists read and the subsequent INSERT, another goroutine
or node can insert the same row, causing a duplicate-key database error. The
code's own comment acknowledges the fix with "// TODO AF Use upsert".

Expected behaviour: concurrent calls to StoreIdentity with the same identity
should be idempotent and succeed without error.

Actual behaviour: under concurrent wallet registration the second writer hits
a duplicate-key constraint and StoreIdentity returns a database error.

### 🛠️ Steps to reproduce

1. Run two goroutines concurrently, both calling StoreIdentity with the
   same (identity, walletID, roleID) tuple against a real SQL backend.
2. Observe that one call returns a duplicate-key database error instead
   of silently succeeding.

The race is confirmed by the TODO comment left by the original author at:
token/services/storage/db/sql/common/wallet.go line 90

### 📋 Logs and Screenshots

// TODO AF Use upsert   ← existing comment at wallet.go:90
pq: duplicate key value violates unique constraint "wallets_pkey"

### 🌐 Network Environment

main (confirmed at commit 9c124d13)

### 🐹 Go Version

go1.24+

### 💻 Operating System

Linux (Ubuntu/Debian)

### ➕ Additional context

Fix: replace the IdentityExists + INSERT with a single atomic
INSERT ... ON CONFLICT (identity_hash, wallet_id, role_id) DO NOTHING
for both the PostgreSQL and SQLite backends. This makes the operation
atomic and removes the race entirely.

Affected file:
  token/services/storage/db/sql/common/wallet.go:89

Rama542:
Fixes #1661

## Problem

`WalletStore.StoreIdentity` used a non-atomic check-then-insert pattern:

```go
// TODO AF Use upsert
if db.IdentityExists(ctx, identity, wID, roleID) {
    return nil
}
// ... INSERT ...
```

Between the `IdentityExists` SELECT and the subsequent INSERT, a concurrent
goroutine or node can insert the same row, causing a duplicate-key constraint
error. The original author already acknowledged this with the `// TODO AF Use upsert`
comment at `wallet.go:90`.

## Fix

Replace the two-step read-then-write with a single atomic statement:

```go
q.InsertInto(db.table.Wallets).
    Fields(identity_hash, meta, wallet_id, role_id, created_at, enrollment_id).
    Row(identity.UniqueID(), meta, wID, roleID, time.Now().UTC(), eID).
    OnConflictDoNothing().
    Format()
```

`INSERT ... ON CONFLICT DO NOTHING` is natively supported by both PostgreSQL
and SQLite, and the FSC query builder already exposes `OnConflictDoNothing()`.
The operation is now atomic at the database level — no race window, no duplicate-key
errors under concurrent writers.

## Changes

| File | Change |
|------|--------|
| `token/services/storage/db/sql/common/wallet.go` | Remove `IdentityExists` guard and `// TODO AF Use upsert` comment; add `.OnConflictDoNothing()` to the INSERT chain |
| `token/services/storage/db/sql/common/wallet_test_utils.go` | Update `TestStoreIdentity` mock (remove stale SELECT expectation, add `ON CONFLICT DO NOTHING` to INSERT regex); add `TestStoreIdentityIdempotent` |
| `token/services/storage/db/sql/sqlite/wallet_test.go` | Register `TestStoreIdentityIdempotent` |
| `token/services/storage/db/sql/postgres/wallet_test.go` | Register `TestStoreIdentityIdempotent` |

## Test plan

- [x] `go test ./token/services/storage/db/sql/common/...` — all pass including `TestStoreIdentityIdempotent`
- [x] `go test ./token/services/storage/db/sql/sqlite/...` — all pass
- [x] `golangci-lint run ./token/services/storage/db/sql/...` — 0 issues

`TestStoreIdentityIdempotent` calls `StoreIdentity` twice with the same
identity; the second call returns 0 rows affected (conflict silently ignored)
and must still return `nil` — which it does.

### DIFF
--- a/token/services/storage/db/sql/common/wallet.go
+++ b/token/services/storage/db/sql/common/wallet.go
@@ -87,14 +87,10 @@ func (db *WalletStore) GetWalletIDs(ctx context.Context, roleID int) ([]driver.W
 }
 
 func (db *WalletStore) StoreIdentity(ctx context.Context, identity token.Identity, eID string, wID driver.WalletID, roleID int, meta []byte) error {
-	// TODO AF Use upsert
-	if db.IdentityExists(ctx, identity, wID, roleID) {
-		return nil
-	}
-
 	query, args := q.InsertInto(db.table.Wallets).
 		Fields("identity_hash", "meta", "wallet_id", "role_id", "created_at", "enrollment_id").
 		Row(identity.UniqueID(), meta, wID, roleID, time.Now().UTC(), eID).
+		OnConflictDoNothing().
 		Format()
 	logging.Debug(logger, query)
 
--- a/token/services/storage/db/sql/common/wallet_test_utils.go
+++ b/token/services/storage/db/sql/common/wallet_test_utils.go
@@ -107,14 +107,9 @@ func TestStoreIdentity(t *testing.T, store walletStoreConstructor) {
 	walletID := driver.WalletID("my wallet")
 	roleID := 5
 
-	mockDB.
-		ExpectQuery("SELECT wallet_id FROM WALLETS WHERE \\(identity_hash = \\$1\\) AND \\(wallet_id = \\$2\\) AND \\(role_id = \\$3\\)").
-		WithArgs(tokenID.UniqueID(), walletID, roleID).
-		WillReturnRows(mockDB.NewRows([]string{"wallet_id"}))
-
 	mockDB.ExpectExec("INSERT INTO WALLETS "+
 		"\\(identity_hash, meta, wallet_id, role_id, created_at, enrollment_id\\) "+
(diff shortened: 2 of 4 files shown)


adecaro (approved):
Hi @Rama542 , LGTM. Thanks for the effort 🙏

PR #1663, opened 2026-05-06
fix(cmd-tokengen-unit-tests): do not rewrite existing testdata

adecaro:
fix test that was rewriting content in the `testdata` folder at each execution

PR #1672, opened 2026-05-06
fix(multisig): apply timeout when waiting for co-signer responses in RequestSpendView

### LINKED ISSUE #1671: [BUG]: RequestSpendView in multisig blocks forever — WithTimeout() exists but is never applied to the channel wait
### 👾 Description

RequestSpendView.Call() in token/services/ttx/multisig/spend.go sends a
SpendRequest to every co-signer of a multisig token and then waits for
all replies on an unbuffered channel with no timeout:

  for range counter {
      // TODO: put a timeout
      a := <-answerChannel   // blocks forever if any party goes silent
  }

The struct already has a `timeout time.Duration` field and a public
WithTimeout(time.Duration) setter, but the timeout is never applied to
the actual channel receive. If any remote co-signer crashes, drops the
network connection, or simply does not respond, the loop blocks the
calling goroutine permanently no error is returned, no cleanup runs.

Expected behaviour: if a co-signer does not reply within the configured
timeout, RequestSpendView should return a descriptive error and unblock
the caller.

Actual behaviour: the caller hangs indefinitely when any co-signer is
unresponsive, consuming a goroutine and its associated network session
forever.

### 🛠️ Steps to reproduce

1. Create a multisig token with two or more parties.
2. Start a spend flow via RequestSpendView.
3. Make one co-signer unavailable (kill the process or drop the network).
4. Observe that the initiating node never returns from Call() no error,
   no timeout, no progress.

The TODO comment at spend.go:155 explicitly acknowledges the missing
timeout:
  // TODO: put a timeout

### 📋 Logs and Screenshots

No error is logged. The goroutine simply parks on the channel receive:
  // TODO: put a timeout
  a := <-answerChannel

A goroutine dump (SIGQUIT / runtime/pprof) will show the goroutine
blocked at spend.go:156 indefinitely.

### 🌐 Network Environment

main (confirmed at token/services/ttx/multisig/spend.go)

### 🐹 Go Version

go1.24+

### 💻 Operating System

Linux (Ubuntu/Debian)

### ➕ Additional context

Proposed fix wire the existing timeout field into the wait loop:

  timer := time.NewTimer(c.timeout)
  defer timer.Stop()
  for range counter {
      select {
      case a := <-answerChannel:
          if a.err != nil {
              return nil, errors.Wrapf(a.err, "failed from [%s]", a.party)
          }
          if a.response.Err != nil {
              return nil, errors.Wrapf(a.response.Err, "got failure from [%s]", a.party)
          }
      case <-timer.C:
          return nil, errors.Errorf("timed out waiting for co-signer response after %s", c.timeout)
      }
  }

A default timeout (e.g. 30s) should also be set in NewRequestSpendView
so callers that forget WithTimeout() are protected automatically.

Affected file:
  token/services/ttx/multisig/spend.go:153-156

Rama542:
Closes #1671

## Problem

`RequestSpendView.Call` spawns a goroutine per co-signer and then collects
their replies from a buffered channel. The collect loop had no timeout:

```go
for range counter {
    // TODO: put a timeout
    a := <-answerChannel   // blocks forever if any party goes silent
}
```

If even one remote co-signer crashes or drops the connection, the calling
goroutine parks on the channel receive indefinitely no error is returned,
no cleanup runs, and the network session is never released. The struct
already had a `timeout time.Duration` field and a `WithTimeout` setter, but
neither was wired into the wait loop.

## Fix

- Add `defaultSpendRequestTimeout = 30s` and set it in `NewRequestSpendView`
  so every caller is protected even without explicitly calling `WithTimeout`
- Replace the bare channel receive with a `select` that races the answer
  channel against a single shared timer, returning a descriptive error when
  the deadline is exceeded

```go
timer := time.NewTimer(c.timeout)
defer timer.Stop()
for range counter {
    select {
    case a := <-answerChannel:
        // handle response
    case <-timer.C:
        return nil, errors.Errorf("timed out after %s waiting for co-signer response", c.timeout)
    }
}
```

The answer channel is already buffered (`len(c.parties)`), so goroutines
spawned by `collectSpendRequestAnswers` can complete their sends and exit
cleanly even after the caller returns on timeout no goroutine leak.

## Changes

| File | Change |
|------|--------|
| `token/services/ttx/multisig/spend.go` | Add `defaultSpendRequestTimeout` constant; set it in `NewRequestSpendView`; replace bare channel receive with `select` + timer |
| `token/services/ttx/multisig/spend_test.go` | New file tests nil-token constructor, `WithTimeout` fluent setter, and the timer branch of the `select` |

## Test plan

- [x] `go test ./token/services/ttx/multisig/...` all pass
- [x] `golangci-lint run ./token/services/ttx/multisig/...` 0 issues

Closes #1663

### DIFF
--- a/token/services/ttx/multisig/spend.go
+++ b/token/services/ttx/multisig/spend.go
@@ -151,25 +151,31 @@ func (c *RequestSpendView) Call(context view.Context) (interface{}, error) {
 		counter++
 	}
 
-	timer := time.NewTimer(c.timeout)
+	return nil, waitForAnswers(answerChannel, counter, c.timeout)
+}
+
+// waitForAnswers drains count replies from ch within the given timeout.
+// It returns the first transport or application error, or a timeout error if
+// no reply arrives in time.
+func waitForAnswers(ch <-chan *answer, count int, timeout time.Duration) error {
+	timer := time.NewTimer(timeout)
 	defer timer.Stop()
-	for range counter {
-		logger.DebugfContext(context.Context(), "Wait for answer")
+
+	for range count {
 		select {
-		case a := <-answerChannel:
-			logger.DebugfContext(context.Context(), "Received answer")
+		case a := <-ch:
 			if a.err != nil {
-				return nil, errors.Wrapf(a.err, "got failure [%s] from [%s]", a.party.String(), a.err)
+				return errors.Wrapf(a.err, "got failure [%s] from [%s]", a.party.String(), a.err)
 			}
 			if a.response.Err != nil {
-				return nil, errors.Wrapf(a.response.Err, "got failure [%s] from [%s]", a.party.String(), a.response.Err)
+				return errors.Wrapf(a.response.Err, "got failure [%s] from [%s]", a.party.String(), a.response.Err)
 			}
 		case <-timer.C:
-			return nil, errors.Errorf("timed out after %s waiting for co-signer response", c.timeout)
+			return errors.Errorf("timed out after %s waiting for co-signer response", timeout)
 		}
 	}
 
-	return nil, nil
+	return nil
 }
 
 func (c *RequestSpendView) WithTimeout(timeout time.Duration) *RequestSpendView {
--- a/token/services/ttx/multisig/spend.go
+++ b/token/services/ttx/multisig/spend.go
@@ -81,6 +81,11 @@ type answer struct {
 	party    view.Identity
 }
 
+// defaultSpendRequestTimeout is the total time RequestSpendView waits for all
+// co-signers to reply before returning an error. Callers may override it via
+// WithTimeout.
+const defaultSpendRequestTimeout = 30 * time.Second
+
 // RequestSpendView sends a SpendRequest to all parties and waits for their responses
 type RequestSpendView struct {
 	unspentToken *token.UnspentToken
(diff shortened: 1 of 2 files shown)


adecaro:
Hi @Rama542 , please, open an Issue and get it assigned. Then link it to the PR. Thanks for the understanding 🙏

Rama542:
Hi @adecaro 👋

I have already opened the issue for this bug:
Closes #1671

Linking it to this PR now. Please let me know if anything else 
is needed. Thank you!

adecaro:
Hi @Rama542 , thanks for the effort. I think this deserves to become some utility struct because we use this kind of pattern in multiple place. I would introduce under `token/services/utils` something like `AnswersCollector` that also listen to when a context is done. What do you think? Use generics, if needed.

Thanks much 🙏

adecaro:
Hi @Rama542 , could you make the timeout configurable, please? 
Many thanks for your effort 🙏

adecaro:
Hi @Rama542 , any chance to finish this? 🙏

adecaro:
Hi @Rama542 , if it okay for you, I'll fork this and continue it. 
Let me know by tomorrow if this is fine for you. Many thanks 🙏

PR #1677, opened 2026-05-06
feat(auditor): add metrics instrumentation to SetStatus, GetStatus and GetTokenRequest

### LINKED ISSUE #1674: [ENHANCEMENT]: Add observability metrics to SetStatus, GetStatus, and GetTokenRequest in auditor.Service
### Problem

The auditor.Service in token/services/auditor/auditor.go instruments its
write operations with Prometheus metrics — Audit(), Append(), and Release()
all record duration histograms and error counters. But the three query
methods SetStatus, GetStatus, and GetTokenRequest have zero instrumentation.

In production these are the most frequently called methods on the audit
path every finality listener update goes through SetStatus, and every
audit query goes through GetStatus and GetTokenRequest. Without metrics,
operators have no way to detect slow database queries, alert on elevated
error rates, or observe p99 latency on these critical paths. The blind spot
makes it much harder to debug audit issues in a live network.

### Solution

Add duration histograms and error counters for the three uninstrumented
methods, following the exact same pattern already used in the same file:

In metrics.go add to the Metrics struct:
  SetStatusDuration        metrics.Histogram
  SetStatusErrors          metrics.Counter
  GetStatusDuration        metrics.Histogram
  GetStatusErrors          metrics.Counter
  GetTokenRequestDuration  metrics.Histogram
  GetTokenRequestErrors    metrics.Counter

In auditor.go wire them at each call site:
  func (a *Service) SetStatus(ctx context.Context, ...) error {
      start := time.Now()
      defer func() { a.metrics.SetStatusDuration.Observe(time.Since(start).Seconds()) }()
      if err := a.auditDB.SetStatus(ctx, txID, status, message); err != nil {
          a.metrics.SetStatusErrors.Add(1)
          return err
      }
      return nil
  }

Same pattern for GetStatus and GetTokenRequest.
This is purely additive no existing behaviour changes.

### Alternatives

I considered adding a single catch-all middleware wrapper around auditDB
calls, but that would require changes across multiple layers. The per-method
approach is simpler, more explicit, and consistent with how Audit() and
Append() are already instrumented in the same file.

### Additional Context

Affected files:
  token/services/auditor/auditor.go  SetStatus:189, GetStatus:195, GetTokenRequest:200
  token/services/auditor/metrics.go  Metrics struct and newMetrics factory

The full infrastructure is already in place:
  - metrics.Provider is injected via NewService()
  - noopProvider handles the case where no provider is configured
  - Histogram bucket configuration already defined for similar methods

This requires no new dependencies and no interface changes.
I am happy to open a PR for this if the team agrees with the approach.

Rama542:
Closes #1674

I was going through the auditor service and noticed something a bit inconsistent.
Audit(), Append() and Release() all have proper duration histograms and error
counters, but the three query methods SetStatus, GetStatus and GetTokenRequest
have nothing at all. In production these are actually called very frequently since
every finality update goes through SetStatus and every audit query hits GetStatus
or GetTokenRequest, so there is a real blind spot there when trying to debug slow
queries or unusual error spikes.

The fix is straightforward since the pattern is already established in the same
file. I added duration histograms and error counters for all three methods and
wired them in exactly the same way Audit() and Append() already do it. No
behaviour changes, purely additive.

Changes:
- metrics.go: added SetStatusDuration, SetStatusErrors, GetStatusDuration,
  GetStatusErrors, GetTokenRequestDuration, GetTokenRequestErrors to the
  Metrics struct and registered them in newMetrics
- auditor.go: wired the new metrics into SetStatus, GetStatus and
  GetTokenRequest using the existing start/defer/observe pattern
- auditor_internal_test.go: updated counter and histogram call counts in
  TestNewMetrics_WithProvider from 3 counters / 2 histograms to 6 counters /
  5 histograms, and added the new metric calls to TestMetricsProviderCall

Tests pass and golangci-lint reports 0 issues.

### DIFF
--- a/token/services/auditor/auditor.go
+++ b/token/services/auditor/auditor.go
@@ -187,18 +187,44 @@ func (a *Service) Release(ctx context.Context, tx Transaction) {
 
 // SetStatus sets the status of the audit records with the passed transaction id to the passed status
 func (a *Service) SetStatus(ctx context.Context, txID string, status storage.TxStatus, message string) error {
-	return a.auditDB.SetStatus(ctx, txID, status, message)
+	start := time.Now()
+	defer func() { a.metrics.SetStatusDuration.Observe(time.Since(start).Seconds()) }()
+	if err := a.auditDB.SetStatus(ctx, txID, status, message); err != nil {
+		a.metrics.SetStatusErrors.Add(1)
+
+		return err
+	}
+
+	return nil
 }
 
 // GetStatus return the status of the given transaction id.
 // It returns an error if no transaction with that id is found
 func (a *Service) GetStatus(ctx context.Context, txID string) (TxStatus, string, error) {
-	return a.auditDB.GetStatus(ctx, txID)
+	start := time.Now()
+	defer func() { a.metrics.GetStatusDuration.Observe(time.Since(start).Seconds()) }()
+	status, message, err := a.auditDB.GetStatus(ctx, txID)
+	if err != nil {
+		a.metrics.GetStatusErrors.Add(1)
+
+		return status, message, err
+	}
+
+	return status, message, nil
 }
 
 // GetTokenRequest returns the token request bound to the passed transaction id, if available.
 func (a *Service) GetTokenRequest(ctx context.Context, txID string) ([]byte, error) {
-	return a.auditDB.GetTokenRequest(ctx, txID)
+	start := time.Now()
+	defer func() { a.metrics.GetTokenRequestDuration.Observe(time.Since(start).Seconds()) }()
+	data, err := a.auditDB.GetTokenRequest(ctx, txID)
+	if err != nil {
+		a.metrics.GetTokenRequestErrors.Add(1)
+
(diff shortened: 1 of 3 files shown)


Rama542:
Hi @adecaro 👋

This PR closes #1674.

Fixed a race condition in the unit tests. 
All tests now pass locally. Please review when you get a chance 🙏

AkramBitar:
@Rama542

I would like to thanks you a lot for working on that PR. 
Did you have the chance to see why these tests fail?

Regards,
Akram

adecaro:
Hi @Rama542 , not sure if you still have cycles, but I would suggest to investigate the use of https://github.com/hexdigest/gowrap to generate wrappers with the instrumentation we need. What do you think?
I can formalize the scope in a Github Issue, if you want.

adecaro:
I'll close this though in the meantime to wait for a more systematic approach. Thanks for the understanding 🙏

PR #1681, opened 2026-05-07
perf(storage): split UnspentTokensIteratorBy into UNION ALL of two index-friendly branches

### LINKED ISSUE #1680: [BUG]: UnspentTokensIteratorBy bypasses idx_owner_wallet_part, scans every live owner-token
### 👾 Description

`UnspentTokensIteratorBy` in `token/services/storage/db/sql/common/tokens.go` joins the `tokens` and `ownership` tables and emits a cross-table `OR` predicate via `HasTokenDetails`:

```sql
SELECT t.tx_id, t.idx, owner_raw, token_type, quantity
FROM tokens t
LEFT JOIN tkn_own o ON t.tx_id = o.tx_id AND t.idx = o.idx
WHERE (o.wallet_id = $1 OR t.owner_wallet_id = $1)
  AND t.owner = true
  AND t.is_deleted = false;
```

PostgreSQL's planner cannot use the partial index `idx_owner_wallet_part (owner_wallet_id, token_type) WHERE is_deleted = false AND owner = true` (added in #1612) under that OR. It falls back to:

1. Bitmap index scan on `(is_deleted, owner)` → returns every live owner-token (hundreds of thousands of rows in production).
2. Hash join with the (typically empty) `ownership` table.
3. Post-filter with the `OR`, returning a single match.

This runs once per `selector.Select()`, so prepare latency degrades linearly with table size.

**Severity:** Medium

**Type:** Performance Bug

### 🛠️ Steps to reproduce

1. Set up a Fabric Token SDK deployment with the Postgres backend.
2. Populate `tokens` with a representative live-row count (we observed it on a node with ~220k active tokens; effect is visible at much lower counts as soon as the planner switches plans).
3. Issue `wallet.ListUnspentTokens()` for any wallet (one row per `selector.Select` call in the transfer hot path).
4. `EXPLAIN ANALYZE` the resulting query — plan hash-joins ~220k rows with the empty ownership table, then filters by the OR.

```
Hash Right Join  (cost=344818..344841 rows=1054) (actual time=216..247 rows=92)
  Filter: ((o.wallet_id = $1) OR (t.owner_wallet_id = $1))
  ->  Seq Scan on tkn_own o  (rows=0)
  ->  Hash
        ->  Bitmap Heap Scan on tokens t  (rows=219968)
              Recheck Cond: ((NOT is_deleted) AND owner)
              ->  Bitmap Index Scan on idx_spent_tokens
Planning Time: 0.84 ms
Execution Time: ~247 ms (warm cache)
```

The same query rewritten as two `UNION ALL` branches — one filtering `tokens.owner_wallet_id`, one filtering `ownership.wallet_id` — lets each side use its own index and runs in **0.66 ms warm** (~370× faster on this dataset). Fix proposed in the linked PR.

### 📋 Logs and Screenshots

`EXPLAIN ANALYZE` on a deployment with 219,968 active owner-tokens, wallet with 94 matching tokens:

```
-- LEFT JOIN + OR (current)
Execution Time: 247 ms warm / 7,330 ms cold

-- UNION ALL (proposed)
Execution Time: 0.66 ms warm
  Append:
    branch 1 -> Index Scan on idx_owner_wallet_part   (94 rows)
    branch 2 -> Bitmap Index Scan on tkn_own_pkey      (0 rows)
```

### 🌐 Network Environment

Local (NWO / Integration Test) — observed under cbdc stress harness against a Fabric-x committer-backed deployment, but the regression is in the SQL plan and reproduces on any Postgres-backed node once the `tokens` table grows.

EvanYan1024:
## Summary

Rewrites `UnspentTokensIteratorBy` so the SQL planner can pick the partial
index `idx_owner_wallet_part (owner_wallet_id, token_type) WHERE is_deleted=false AND owner=true`
that #1612 added. The original cross-table `OR` predicate forces the planner
to seq-scan every live owner-token; this PR replaces that with a single
statement of two `UNION ALL` branches that each filter only one side of the
original condition.

Closes #1680.

## Background

`UnspentTokensIteratorBy` is on the hot path of every transfer: the selector
calls it to populate its in-memory token index. After #1612 added the
partial index, the predicate that should have driven it (`owner_wallet_id = $1
AND is_deleted=false AND owner=true`) was buried inside a cross-table OR:

```sql
WHERE (o.wallet_id = $1 OR t.owner_wallet_id = $1)
  AND t.owner = true
  AND t.is_deleted = false;
```

Postgres cannot index-walk under that OR. On a deployment with ~220k live
owner-tokens the resulting `Hash Right Join` ran in **247 ms warm / 7,330 ms
cold** per call, dominating prepare latency.

### `token/services/storage/db/sql/common/tokens.go`

`UnspentTokensIteratorBy` now emits two `UNION ALL` branches into a shared
query builder, then filters duplicates in a small iterator wrapper:

1. **Branch 1 — directly owned**: `WHERE owner_wallet_id = $1`. The partial
   index drives this in microseconds. The `LEFT JOIN ownership` is kept so
   the iterator can surface the same `(token, ownership)` rows it did
   before — including the multi-ownership case where one token has several
   ownership rows. `StoreToken` permits `tokens` rows with `owner_wallet_id`
   set but no ownership entry (when its `owners` parameter is empty); those
   rows surface here, replacing what the original OR predicate caught.
2. **Branch 2 — delegation-reachable**: `WHERE ownership.wallet_id = $1`.
   Returns zero rows on deployments where ownership delegation is not
   configured, at which point the branch is essentially free.

Implementation notes worth flagging:

- Both branches share a single `common.Builder` so placeholder numbering
  (`$1`, `$2`, …) increments continuously; args concatenate in order. Single
  statement → single connection from the pool, which avoids the deadlock
  any eager two-query split would hit on a constrained pool under load.
- `UNION ALL`, not `UNION`. `UNION` would force PG to hash every row over
  five columns (including `bytea owner_raw`) for a per-row dedup pass that
  almost never finds collisions. Each branch instead selects
  `ownership.wallet_id` as a sixth trailing column used solely for app-side
  dedup in `dedupedTokenRowsIterator`.
- App-side dedup keys on `(tx_id, idx, ownership.wallet_id)`. This drops
  the rare cross-branch duplicate (a row matching both
  `tokens.owner_wallet_id = $1` and `ownership.wallet_id = $1`) while
  preserving distinct multi-ownership rows. NULL is namespaced separately
  from the empty string so the two cannot collide.
- No parens around the SELECT operands of `UNION ALL` — SQLite rejects
  parenthesised SELECTs around `UNION`; PostgreSQL accepts both forms.
  Neither branch has `ORDER BY` / `LIMIT`, so binding is unchanged.

### `token/services/storage/db/sql/common/query_test.go`

New `TestUnionAll_Compile` verifies the shared-builder pattern: two
`q.Select()` queries combined via `FormatTo` produce a single statement
with continuous placeholder numbering, ordered args, and no parens around
the operands.

## Measured impact

`EXPLAIN ANALYZE` on a deployment with 219,968 active owner-tokens, wallet
with 94 matching tokens (warm shared buffers):

| Plan                         | Time    | Notes                                            |
|------------------------------|---------|--------------------------------------------------|
| Current (\`LEFT JOIN + OR\`) | 247 ms  | Hash Right Join over 219,968 rows, post-filter   |
| Current (cold, first run)    | 7,330 ms| Same plan, 35,549 buffer reads                   |
| **Proposed (\`UNION ALL\`)** | **0.66 ms** | Append: branch 1 hits \`idx_owner_wallet_part\`, branch 2 hits \`tkn_own_pkey\` |

## Compatibility

- Pure SQL rewrite, no schema or migration change.
- The two-branch SQL is valid on PostgreSQL and SQLite (both supported by
  the existing CI matrix).
- `dedupedTokenRowsIterator` preserves the pre-PR row semantics: one row
  per `(token, ownership-wallet)` pair, including multi-ownership tokens
  (covered by the existing `TestTokens` case where `tx100:1` is co-owned
  by alice and bob) and `owners=nil` tokens (`tx2000`).

## Test plan

- [x] `go build ./...`
- [x] `go test ./token/services/storage/db/sql/common/...` — all existing
      cases pass, including `TestTokens` which exercises multi-ownership
      and no-ownership rows. New `TestUnionAll_Compile` covers the
      shared-builder behaviour.
- [x] `EXPLAIN ANALYZE` against a real Postgres deployment (numbers above).
- [x] Commit signed off (`-s`) per CONTRIBUTING.md.

### DIFF
--- a/token/services/storage/db/sql/common/query_test.go
+++ b/token/services/storage/db/sql/common/query_test.go
@@ -10,6 +10,7 @@ import (
 	"testing"
 
 	q "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/sql/query"
+	qcommon "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/sql/query/common"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/sql/query/cond"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/sql/sqlite"
 	"github.com/stretchr/testify/assert"
@@ -73,4 +74,33 @@ func TestDelete_Compile(t *testing.T) {
 		Format(sqlite.NewConditionInterpreter())
 	assert.Equal(t, "DELETE FROM users WHERE id = $1", query)
 	assert.Equal(t, 1, args[0])
+}
+
+// TestUnionAll_Compile verifies that two SELECT queries can be combined into
+// a single UNION ALL statement via a shared builder, with placeholder
+// numbering continuing across branches and args concatenated in order.
+// This is the pattern used by UnspentTokensIteratorBy.
+func TestUnionAll_Compile(t *testing.T) {
+	ci := sqlite.NewConditionInterpreter()
+
+	branch1 := q.Select().
+		FieldsByName("id", "name").
+		From(q.Table("users")).
+		Where(cond.Eq("id", 1))
+
+	branch2 := q.Select().
+		FieldsByName("id", "name").
+		From(q.Table("admins")).
+		Where(cond.Eq("name", "alice"))
+
+	sb := qcommon.NewBuilder()
+	branch1.FormatTo(ci, sb)
+	sb.WriteString(" UNION ALL ")
+	branch2.FormatTo(ci, sb)
+	query, args := sb.Build()
+
+	assert.Equal(t,
+		"SELECT id, name FROM users WHERE id = $1 UNION ALL SELECT id, name FROM admins WHERE name = $2",
+		query)
+	assert.Equal(t, []any{1, "alice"}, args)
 }
--- a/token/services/storage/db/sql/common/tokens.go
+++ b/token/services/storage/db/sql/common/tokens.go
@@ -143,34 +143,159 @@ func (db *TokenStore) UnspentTokensIterator(ctx context.Context) (tdriver.Unspen
 	return db.UnspentTokensIteratorBy(ctx, "", "")
 }
 
-// UnspentTokensIteratorBy returns an iterator of unspent tokens owned by the passed id and whose type is the passed on.
-// The token type can be empty. In that case, tokens of any type are returned.
+// UnspentTokensIteratorBy returns an iterator of unspent tokens owned by the
+// passed wallet id and of the passed type. Empty tokenType returns all types.
+//
+// Implemented as a single SQL with two UNION ALL branches so each side can
+// use its own index, instead of a cross-table OR predicate that PostgreSQL
+// cannot resolve with the partial index on (owner_wallet_id, token_type)
+// WHERE is_deleted=false AND owner=true:
+//
+//  1. tokens directly owned: filters tokens.owner_wallet_id, hits the partial
+//     index in microseconds. Joins ownership by primary key to preserve the
+//     pre-PR semantic that a token must have at least one ownership row to
+//     be visible (StoreToken can persist a tokens row without an ownership
+//     row when the owners slice is empty, and the original INNER JOIN
+//     intentionally excluded those).
+//  2. tokens reachable via the ownership-delegation table: joins ownership
+//     -> tokens by primary key. Returns zero rows when delegation is not
+//     configured, at which point that branch is essentially free.
+//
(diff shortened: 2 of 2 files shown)


adecaro (approved):
Super interesting catch, thanks @EvanYan1024 🙏

PR #1597, opened 2026-04-24
token: introduce policy-based identity (#1586)

sid200727:
Closes #1586

Introduces `PolicyIdentity` — a new identity type whose ownership is governed by a boolean expression over a set of component identities.

## Changes
- `token/core/common/boolexpr/` — recursive descent parser for expressions of the form `$0 OR ($1 AND $2)`. 30 tests.
- `token/driver/wallet.go` — `PolicyIdentityType = 6`, `PolicyIdentityTypeString = "policy"`
- `token/services/identity/boolpolicy/identity.go` — `PolicyIdentity` struct with ASN.1 serialization, AuditInfo, InfoMatcher
- `token/services/identity/boolpolicy/sig.go` — `PolicyVerifier` + sparse `PolicySignature` envelope (ASN.1). 27 tests.
- `token/services/identity/boolpolicy/deserializer.go` — `TypedIdentityDeserializer` + `AuditInfoDeserializer`, mirroring multisig
- `token/core/fabtoken/v1/driver/deserializer.go` — registration
- `token/core/zkatdlog/nogh/v1/driver/deserializer.go` — registration

## Design decisions (per @adecaro)
- No external dependencies — custom parser inspired by policydsl
- ASN.1 serialization throughout (ledger-safe, minimal overhead)
- Own signature envelope, kept fully separate from MultiIdentity for legacy compatibility

PR #1691, opened 2026-05-08
fix(multisig,boolpolicy): verify spend tx matches approved SpendRequest

### LINKED ISSUE #1682: [SECURITY]: Multisig EndorseSpendView signs received tx without checking it matches the approved SpendRequest
In `token/services/ttx/multisig/spend.go`, `EndorseSpendView.Call` receives a
transaction from the initiator and signs it without verifying that the
transaction actually corresponds to the `SpendRequest` it just endorsed.

The check is missing — there's a `// TODO: check tx matches request` at the
exact spot where it should happen:

```go
func (a *EndorseSpendView) Call(context view.Context) (interface{}, error) {
    if err := session.JSON(context).Send(&SpendResponse{}); err != nil {
        return nil, errors.Wrap(err, "failed to send response")
    }

    tx, err := ttx.ReceiveTransaction(context)
    if err != nil {
        return nil, errors.Wrap(err, "failed to receive transaction")
    }

    // TODO: check tx matches request

    _, err = context.RunView(ttx.NewEndorseView(tx))
    ...
}
```

`a.request` is already on the view (set by `NewEndorseSpendView`), so the
information is available — the validation just isn't done.

### Why this is a problem

Multisig assumes each signer consents to a *specific* spend. With this gap, a
malicious or buggy initiator can send `SpendRequest{Token: T_small}` for
review, get co-owners to approve, then send a different `tx` that consumes a
larger token co-owned by the same group. Each signer signs whatever arrives,
and on-chain there's no way to tell the request and the tx diverged.

It's also reachable without malice — any bug in the initiator's tx-assembly
path that produces a mismatch will go uncaught.

### Proposed fix

At the TODO site, inspect the inputs of the received tx (via the same
`AuditRecord` path the auditor uses) and confirm the token ID being spent
matches `a.request.Token.Id`. If not, return an error before calling
`EndorseView`.

### Acceptance criteria

- [ ] `EndorseSpendView.Call` rejects (with a clear error) when the received
      tx does not spend the token named in the request.
- [ ] Test: mismatched request/tx → endorsement is refused.
- [ ] Test: matching request/tx → flow succeeds as before.

### Out of scope

- Multi-token spends (would require `SpendRequest` to carry a set of token
  IDs — separate protocol change).
- The unrelated `// TODO: put a timeout` at line 155 (covered by #1671).

Related: #1638, #1645 (similar validation hardening).

SuyashAlphaC:
Fixes #1682.                                                                                                                                                                                                                              
                                                                                                                                                                                                                                            
  `EndorseSpendView` in `multisig/spend.go` (where the original TODO lives) and `boolpolicy/spend.go` both received a tx from the initiator and signed it without checking it actually consumed the token from the `SpendRequest` the       
  co-owner approved.
                                                                                                                                                                                                                                            
  Added a `verifySpendTxMatchesRequest` helper in each package that pulls the audit record off the received tx and rejects the flow before `EndorseView` if any input id differs from `a.request.Token.Id`. The comparison itself sits in a 
  pure `verifyInputIDsMatchExpected` so the rules are easy to unit-test without constructing a real `ttx.Transaction`.
                                                                                                                                                                                                                                            
  Scope note: only the multisig variant was named in the issue, but `boolpolicy/spend.go` (added recently in #1597) has the identical gap, so I fixed both. Holding back the policy fix would have left the same vulnerability open under a 
  different package name.
                                                                                                                                                                                                                                            
 Added a "Spend Coordination Wire Flow" section to `docs/services/ttx.md` with a Mermaid diagram covering the request → delivery → verification phases shared by both variants.                          
  
  ### Test plan                                                                                                                                                                                                                             
  - [x] Unit tests for both variants: matching / mismatched / nil / empty / multi-input cases
  - [x] `go vet` and `gofmt` clean                                                                                                                                                                                                          
  - [x] Existing tests in both packages still pass


SuyashAlphaC:
Hi @adecaro — when you have a moment, would appreciate a review on this. Closes #1682 (the multisig spend trust gap you green-lit). Same fix is applied to `boolpolicy/spend.go` since it has the identical pattern after #1597; explained in the PR description. Docs updated with a sequence diagram per your ask. Thanks!

SuyashAlphaC:
Hi @adecaro — pushed a fix for the CI failure (`AuditRecord` was the wrong API since the co-owner isn't the auditor; switched to `Request.Inputs` which doesn't need the audit store). Tests are green now. Whenever you have a moment, would appreciate a review on this. Thanks 🙏

SuyashAlphaC:
Hey @adecaro 
Can you please review this PR once 🙏

token/services/ttx/boolpolicy/spend.go
```diff
@@ -223,9 +224,53 @@ func (a *EndorseSpendView) Call(context view.Context) (interface{}, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to receive transaction")
 	}
+	// Reject any received tx that consumes a different token from the one this
```
adecaro: I would use a different approach here: I would return here tx and let the caller check the content of the transaction. If the transaction is fine, then the caller will call ttx.NewEndorseView(tx). 

We need to make sure the caller can perform the following checks:
- The token request contains the token named by SpendRequest
- The token request does not contain any other token that belong to this node. This is important to avoid that the initiator tries to spend also other tokens belonging to the responder. You can check extractRequiredSigners as a reference.

What do you think? The same reasoning applies to multisig.

adecaro:
Hi @SuyashAlphaC , thank you for your effort 🙏 
I left a comment, please, let me know what you think?

token/services/ttx/boolpolicy/spend.go
SuyashAlphaC: Thanks @adecaro 🙏 —  I agree with the direction.

  You’re right that my current version doesn’t fully match what you suggested: right now I validate inside EndorseSpendView and enforce all inputs == request.Token.Id. Your proposal is better: validate first, then endorse only if the tx passes the right checks.

  I’ll rework this for both multisig and boolpolicy so we do exactly that:

  • separate receive/validate from endorse, so ttx.NewEndorseView(tx) is called only after validation succeeds;
  • replace the current “all inputs equal expected” check with two checks:
    1. the tx contains the token from SpendRequest;
    2. the tx does not include any additional inputs owned by this responder (same spirit as extractRequiredSigners).

  I’ll also adjust tests to cover:
  • missing expected token -> reject;
  • expected token present with no extra responder-owned inputs -> accept;
  • expected token present but extra responder-owned input(s) -> reject.

Does this aligns with you views? Let me know, I'll push the updates

SuyashAlphaC:
Hey @adecaro ,
I have replied to your views and waiting for your response to push the suggested updates.
Thanks 🙏

token/services/ttx/boolpolicy/spend.go
adecaro: the important thing is that the developer should be able to perform any check that the business logic requries.

adecaro:
Hi @SuyashAlphaC , please, go ahead. Thanks much 🙏

token/services/ttx/boolpolicy/spend.go
SuyashAlphaC: So the design should expose the primitives and let the caller's business logic drive the checks, with the two I mentioned being one possible policy. I'll factor it that way.

SuyashAlphaC:
Hi @adecaro — pushed the refactor based on your review. A summary of what I did : 
  - `EndorseSpendView` → `ReceiveSpendTxView` in both `multisig/spend.go` and `boolpolicy/spend.go`. The view now sends the ACK, receives the tx, and **returns it without endorsing or applying any embedded checks** — exactly as you suggested.
  - Removed the hardcoded `verifySpendTxMatchesRequest` helper; there's no library-side policy anymore.
  - The integration callers (`MultiSigAcceptSpendView`, `PolicyAcceptSpendView`) now do the two checks you flagged — (1) tx consumes the SpendRequest token, (2) no other input is owned by this responder (via `tms.SigService().IsMe`, mirroring `extractRequiredSigners`) — and only then call `ttx.NewEndorseView(tx)` themselves.
  - Diagram in `docs/services/ttx.md` updated to reflect the application-driven check phase.

  Whenever you have a moment, would appreciate a re-review. Thanks 🙏

adecaro (approved):
Hi @SuyashAlphaC , it looks great. Thanks a lot and thanks for updating the documentation accordingly 🙏

PR #1692, opened 2026-05-08
bug(balance): to return big.Int to avoid overflows

### LINKED ISSUE #1693: bug(balance): to return big.Int to avoid overflows
Currently, if the sum of the tokens owned by a wallet cannot be stored in a uint64, the DB query will fail. 
This needs to be fixed.

adecaro:
This PR migrates the token balance representation from `uint64` to `*big.Int` throughout the codebase to support arbitrary precision arithmetic for token amounts, and improves context propagation in the storage layer.

### Core Driver Interfaces (`token/driver`)
- Updated all `Balance()` method signatures to return `*big.Int` instead of `uint64`
- Affected interfaces: `QueryEngine`, `Vault`, `WalletBalance`, and related types

### Public API Layer (`token/sdk`)
- Updated `Balance()` methods in public-facing APIs to return `*big.Int`
- Maintained backward compatibility in method signatures

### Service Layer (`token/services`)
- Updated TTX, Selector, and Vault services to work with `*big.Int`
- Modified balance calculation and comparison logic throughout

### Mock Regeneration
- All mocks regenerated to reflect new `*big.Int` signatures using `counterfeiter`

### Integration Tests
- Updated all integration tests to use proper `*big.Int` comparisons
- Replaced numeric literals with `big.NewInt()` calls
- Updated assertions to use `big.Int.Cmp()` for comparisons

### Schema Updates (`token/services/storage/db/sql/common/tokens.go`)
- Changed `amount` column from `BIGINT` to `NUMERIC(78, 0)` in Tokens table
- Supports arbitrary precision integers up to 78 digits
- Enables amounts exceeding uint64 maximum (2^64-1 ≈ 1.8×10^19)

### Storage Layer Implementation
- Updated `balance()` method to directly query NUMERIC values using `BigInt` scanner
- Removed uint64 conversion logic (`SetUint64`)
- Leverages existing `BigInt` type that implements `sql.Scanner` interface
- Properly handles `sql.ErrNoRows` for empty results

### Affected Tables
- `Tokens.amount`: NUMERIC(78, 0)
- `Transactions.amount`: NUMERIC(78, 0) 
- `Movements.amount`: NUMERIC(78, 0)

### QueryUnique to QueryUniqueContext Migration
Replaced all `common.QueryUnique` calls with `common.QueryUniqueContext` across the storage layer (11 total replacements):

- `token/services/storage/db/sql/common/tokens.go` (4 replacements):
  - `IsMine()`, `Balance()`, `PublicParams()`, `PublicParamsByHash()`, `ExistsCertification()`
- `token/services/storage/db/sql/common/identity.go` (2 replacements):
  - `ConfigurationExists()`, `GetIdentityInfo()`
- `token/services/storage/db/sql/common/keystore.go` (1 replacement):
  - `Get()`
- `token/services/storage/db/sql/common/wallet.go` (3 replacements):
  - `GetWalletID()`, `LoadMeta()`, `IdentityExists()`
- `token/services/storage/db/sql/common/transactions.go` (1 replacement):
  - `GetTokenRequest()`

### Benefits
- Enables proper context propagation for cancellation and timeouts
- Improves tracing and observability throughout storage operations
- Consistent context handling across all database queries

### Storage Documentation (`docs/services/storage.md`)
- Added new **"Amount Storage Strategy"** section explaining NUMERIC(78,0) design
- Updated **Transaction & Audit Store** section to clarify amount storage in Transactions and Movements tables
- Updated **Token Store** section with detailed explanation of amount storage and big.Int handling
- Documented automatic conversion between database NUMERIC and in-memory `*big.Int` via `BigInt` scanner

### Database Schema Diagram (`docs/imgs/storage_db.puml`)
- Updated PlantUML diagram to reflect NUMERIC(78,0) for all amount columns
- Changed `Tokens.amount`, `Transactions.amount`, and `Movements.amount` from BIGINT to NUMERIC(78,0)


integration/token/fungible/views/accept.go
```diff
@@ -44,7 +46,7 @@ func (a *AcceptCashView) Call(context view.Context) (interface{}, error) {
 		}
 		balance, err := ttx.MyWallet(context, token.WithTMSID(tx.TMSID())).Balance(context.Context(), ttx.WithType(output.Type))
 		assert.NoError(err, "failed retrieving balance for type [%s]", output.Type)
-		assert.True(balance <= 3000, "cannot have more than 3000 unspent quantity for type [%s]", output.Type)
+		assert.True(balance.Cmp(big.NewInt(3000)) <= 0, "cannot have more than 3000 unspent quantity for type [%s]", output.Type)
```
AkramBitar, line 49: Why the 3000 is hard coded number?
Maybe this need to be in a define or in a config?

docs/imgs/storage_db.puml
```diff
@@ -24,7 +24,7 @@ package "Transaction & Audit Store (TTXDB / AuditDB)" {
         sender_eid : TEXT
         recipient_eid : TEXT
         token_type : TEXT <<NOT NULL>>
-        amount : BIGINT <<NOT NULL>>
+        amount : NUMERIC(78,0) <<NOT NULL>>
```
AkramBitar, line 27: Why 78? 
I am wondering why the amount could not be decimal number?

integration/token/interop/views/balance.go
```diff
@@ -67,7 +66,7 @@ func (b *BalanceView) Call(context view.Context) (interface{}, error) {
 	assert.NoError(err, "failed to compute the sum of the htlc expired tokens")
 
 	return BalanceResult{
-		Quantity: strconv.FormatUint(balance, 10),
+		Quantity: balance.String(),
```
AkramBitar, line 69: Why Quantity is string and not bigint?

token/services/storage/db/dbtest/tokens.go
```diff
@@ -896,7 +896,7 @@ func TQueryTokenDetails(t *testing.T, db TestTokenDB) {
 	assertEqual(t, tx1, res[0])
 	balance, err := db.Balance(ctx, "alice", "TST1")
 	require.NoError(t, err)
-	assert.Equal(t, res[0].Amount, balance)
+	assert.Equal(t, balance.Uint64(), res[0].Amount.Uint64())
```
AkramBitar, line 899: Uint64() converts a big number (bigInt) into a 64-bit integer, but NUMERIC(78,0) can store numbers much larger than that. So if the value is too big it overflows

token/wallet.go
```diff
@@ -290,15 +291,16 @@ func (o *OwnerWallet) ListUnspentTokensIterator(ctx context.Context, opts ...Lis
...
 }
 
-// Balance returns the sun of the amounts, with 64 bits of precision, of the tokens with type and EID equal to those passed as arguments.
-func (o *OwnerWallet) Balance(ctx context.Context, opts ...ListTokensOption) (uint64, error) {
+// Balance returns the sum of the amounts of the tokens with type and EID equal to those passed as arguments.
+// The result is returned as a *big.Int to support arbitrary precision and prevent overflow.
+func (o *OwnerWallet) Balance(ctx context.Context, opts ...ListTokensOption) (*big.Int, error) {
 	compiledOpts, err := CompileListTokensOption(opts...)
 	if err != nil {
-		return 0, err
```
AkramBitar, line 297: Need to check that the caller code can accept nil instead of 0

token/core/zkatdlog/nogh/v1/transfer/mock/signing_identity.go
```diff
@@ -8,199 +8,155 @@ import (
 )
```
AkramBitar, line 9: General question not related to the mock.
I am wondering how this fix (moving to bigInt and NUMERIC(78,0) can effect the performance and especially the sign and verify of tokens with bigger amounts.

docs/imgs/storage_db.puml
```diff
@@ -34,7 +34,7 @@ package "Transaction & Audit Store (TTXDB / AuditDB)" {
         tx_id : TEXT <<FK, NOT NULL>>
         enrollment_id : TEXT
         token_type : TEXT <<NOT NULL>>
-        amount : BIGINT <<NOT NULL>>
+        amount : NUMERIC(78,0) <<NOT NULL>>
```
AkramBitar, line 37: What about backward compatibility?
What about migrating the current values in the database to the new shcema definition?

AkramBitar (changes requested):

integration/token/fungible/views/accept.go
adecaro: It is just an existing integration test. 3000 is arbitrary and functional to the test.

docs/imgs/storage_db.puml
adecaro: the ZK proofs do not support decimal numbers. If one wants something like 1.225, then one must store the number 1225 and keep in the application how many digits are . are used.

integration/token/interop/views/balance.go
adecaro: this is just an integration test where the result is returned as string.

token/services/storage/db/dbtest/tokens.go
adecaro: point taken but in this test we control the values and Uint54 will not overflow in this case.

token/wallet.go
adecaro: sure, the code has been updated accordingly.

token/core/zkatdlog/nogh/v1/transfer/mock/signing_identity.go
adecaro: it does affect performance, but postgres can still handle these number pretty well that's why we went for 78 digits corresponding to roughly numbers of 255 bits.

docs/imgs/storage_db.puml
adecaro: sure, after the release of v1, we will have to provide also migration scripts.

AkramBitar (approved):
LGTM

PR #1697, opened 2026-05-08
feat(ttx): WithApprovalMetadata

### LINKED ISSUE #1684: feat(ttx): WithApprovalMetadata
Currently, we don't have a way to pass additional metadata to the approvers. Metadata could be used by the application to deliver to the approvers extra information.

To achieve this, we would need to update the network service. RequestApproval should take in input also this metadata. Each driver will decide how to deliver this information to the approvers depending on the specific backend.

At the `ttx` service level, an option `WithApprovalMetadata` can be introduced. This option will be used with the collect endorsement view.

atharrva01:
Closes #1684.

I added a `WithApprovalMetadata` option to `CollectEndorsementsView` so applications can pass extra context to approvers. The metadata flows through the full call chain down to the network driver.

- **FSC path** , serialised as a JSON blob into a `approval_metadata` transient field; the responder extracts it into `Request.ApprovalMetadata`
- **Chaincode path**,  each key injected as a separate `WithTransientEntry`
- Backward compatible: no metadata means no extra transient field, count stays at 2

### DIFF
--- a/token/core/common/driver/mock/network_driver.go
+++ b/token/core/common/driver/mock/network_driver.go
@@ -205,7 +205,7 @@ type Network struct {
 		result1 [][]byte
 		result2 error
 	}
-	RequestApprovalStub        func(view.Context, *tokena.ManagementService, []byte, view.Identity, driver.TxID) (driver.Envelope, error)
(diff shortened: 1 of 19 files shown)


adecaro (approved):
Just right. Thanks a lot @atharrva01

PR #1698, opened 2026-05-08
token: return wallet balances as big integers

### LINKED ISSUE #1693: bug(balance): to return big.Int to avoid overflows

malsomesh9:
## Summary
- change owner-wallet and token-vault balance APIs to return `*big.Int`
- compute SQL balances by streaming `amount` rows into a Go `big.Int` accumulator instead of relying on `SUM(amount)`
- update the balance views and checks to serialize balances as decimal strings
- add a regression test that covers balances larger than `uint64`

## Testing
- go test ./token
- go test ./token/services/identity/role
- go test ./token/services/storage/db/sql/common
- go test ./token/services/storage/db/sql/sqlite
- go test ./integration/token/fungible/views ./integration/token/interop/views ./integration/token/dvp/views

Fixes #1693

### DIFF
--- a/integration/token/dvp/views/balance.go
+++ b/integration/token/dvp/views/balance.go
@@ -9,7 +9,6 @@ package views
 import (
 	"encoding/json"
 	"fmt"
-	"strconv"
 
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/assert"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
@@ -45,7 +44,7 @@ func (b *BalanceView) Call(context view.Context) (interface{}, error) {
 		return nil, err
 	}
 
-	return Balance{Quantity: strconv.FormatUint(balance, 10), Type: b.Type}, nil
+	return Balance{Quantity: balance.String(), Type: b.Type}, nil
 }
 
 type BalanceViewFactory struct{}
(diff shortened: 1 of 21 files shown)


SuyashAlphaC:
Hey @malsomesh9  and @adecaro , I noticed  a few things while reading through:                                                                                                                                                    
  - The per-row Amount column in ```TokenRecord``` is still ```uint64```, so this fixes the case where many tokens sum past ```MaxUint64```, but a single token bigger than MaxUint64 still can't be stored at all.                                                                           
  - Also wondering about perf — going from one SUM(amount) row to N rows is going to add up for wallets with a lot of unspent tokens. Did you happen to benchmark it? ```SUM(amount::numeric)``` on Postgres / ```SUM(CAST(amount AS DECIMAL))``` on SQLite would let the DB keep doing the aggregation at arbitrary precision and you'd just parse one string at the end.           

                                                                                                                                                                                                    
 **Small thing** : scanning amount into sql.NullString instead of sql.NullInt64 makes us depend on the driver's int→string conversion, which is a bit different across drivers. Was there a reason for the string form, or just convenience?

adecaro:
Hi @malsomesh9 , I have already submitted a PR for this. Sorry, you cannot take a task that is not assigned to you. Thanks for the understanding. I'm closing this.

PR #1701, opened 2026-05-13
feature(evm-network-driver): initial design

adecaro:
This PR proposes a design for an EVM-based network driver.

This feature will be developed in the branch `feature/evm-network-driver`.

### DIFF
--- a/x/token/services/network/evm/eth_network_driver_design.md
+++ b/x/token/services/network/evm/eth_network_driver_design.md
@@ -199,7 +199,7 @@ address admin;
 
 **Signature Verification Logic**:
 1. Recover signer address from each signature using `ecrecover`
-2. Verify each recovered address is in the endorser set
+2. Verify each recovered address is unique and in the endorser set
 3. Check that number of valid signatures meets threshold
 4. Return true if all checks pass
 
--- a/x/token/services/network/evm/eth_network_driver_design.md
+++ b/x/token/services/network/evm/eth_network_driver_design.md
@@ -1,0 +1,2264 @@
+# Ethereum Network Driver Design Document
+
+## 1. Executive Summary
+
+This document specifies the design and implementation of an Ethereum/EVM network driver for the Hyperledger Fabric Token SDK. The driver implements **Approach 2** (Pre-Order Execution with FSC Endorsers) from the existing network-ethereum.md documentation, enabling token operations on Ethereum networks while preserving the SDK's off-chain validation model.
+
+### Key Design Decisions
+
+- **Validation Model**: Off-chain validation in Go within FSC nodes, on-chain signature verification and state application
+- **Token Support**: Both fabtoken and zkatdlog drivers supported from v1
+- **Identity**: secp256k1 keys for EIP-712 endorsement signatures, configured under TMS network services path
+- **Signature Envelope**: EIP-712 for endorser signatures with domain separation by chainID and contract address
+- **Contract Architecture**: Two-contract split (EndorsementVerifier + TokenState) for separation of concerns
+- **Backend**: EVMClient interface abstraction (fabric-x-evm used for NWO test network bootstrapping)
+- **Token Storage**: Each TMS uses a single TokenState contract; all tokens (fabtoken and zkatdlog) stored in one contract
+- **Transaction ID**: Deterministic internal ID using `keccak256(nonce || creator || chainID || contractAddr)`
+- **Contract Deployment**: Contracts deployed externally via deployment tools; driver reads addresses from configuration
+- **StateDelta Translator**: Dedicated translator component analogous to RWSet translator, consuming validated token actions to produce deterministic state transitions
+- **Spent ID Encoding**: Unified spent tracking using `keccak256` with domain separators (`SERIAL_NUMBER` prefix) to prevent collisions between token IDs and serial numbers
+- **Metadata Keys**: Deterministic hashing using `keccak256(namespace || key)` to prevent cross-TMS collisions
+- **Public Parameters**: StateDelta includes `publicParamsHash` for endorser verification of parameter consistency
+- **Finality Tracking**: Hybrid architecture combining receipt polling, indexed log queries, reorg detection, and status caching for efficiency
+- **Error Handling**: Typed errors with consistent wrapping patterns, recovery strategies, and retry logic for transient failures
+- **Endorser Discovery**: Dynamic endorser registry via contract queries for runtime flexibility
+- **Configuration Validation**: Comprehensive validation with sensible defaults for production readiness
+
+### Out of Scope for v1
+
+- Cross-chain interoperability
+- Contract upgradeability
+- Advanced gas estimation (basic heuristic provided)
+- Full recovery mechanisms (basic recovery included, advanced features deferred to v2)
+- Batch operations (single transaction per state delta in v1)
+- State delta compression
+
+### Design Review Status
+
+**Last Review**: 2026-05-13  
+**Status**: Comprehensive gap analysis completed with critical enhancements integrated  
+**Review Document**: See `eth_network_driver_design_review.md` for detailed gap analysis and recommendations  
+**Critical Updates**: StateDelta translator interface, finality architecture clarification, error taxonomy, security considerations
+
+---
+
+## 2. Architecture Overview
+
+### 2.0 Token Driver Support
+
+The Ethereum network driver supports both token driver types from v1:
+
+**fabtoken (Cleartext Tokens)**:
+- Token amounts and types are visible on-chain
+- Simpler validation logic
+- Lower computational overhead
+- Suitable for use cases where privacy is not required
+
+**zkatdlog (Privacy-Preserving Tokens)**:
+- Token amounts hidden using zero-knowledge proofs
+- Requires cryptographic validation of ZK proofs
+- Higher computational overhead
+- Suitable for privacy-sensitive applications
+
+The driver abstracts the token-specific logic through the existing driver interface. Token validation remains off-chain in Go (within FSC/TTX flows), while the on-chain contracts handle signature verification and state application regardless of token type.
+
+### 2.1 High-Level Flow
+
+```mermaid
+sequenceDiagram
+    participant App as Application/TTX
+    participant Driver as Ethereum Driver
+    participant E1 as FSC Endorser 1
+    participant E2 as FSC Endorser 2
+    participant Node as Ethereum Node
+    participant Contract as TokenState Contract
+    participant Verifier as EndorsementVerifier
+
+    Note over App,Verifier: Token Transfer Flow
+    
+    App->>Driver: RequestApproval(tokenRequest)
+    Driver->>Driver: Compute state delta
+    
+    par Collect Endorsements
+        Driver->>E1: Request endorsement
+        E1->>E1: Validate in Go
+        E1->>E1: Sign EIP-712 digest
+        E1-->>Driver: Signature 1
+    and
+        Driver->>E2: Request endorsement
+        E2->>E2: Validate in Go
+        E2->>E2: Sign EIP-712 digest
+        E2-->>Driver: Signature 2
+    end
+    
+    Driver->>Driver: ABI-encode applyStateDelta()
+    Driver->>Driver: Build & sign Ethereum tx
+    
+    App->>Driver: Broadcast(envelope)
+    Driver->>Node: eth_sendRawTransaction
+    Node->>Contract: Execute applyStateDelta()
+    Contract->>Verifier: verify(signatures)
+    Verifier-->>Contract: Valid
+    Contract->>Contract: Check double-spend
+    Contract->>Contract: Apply state delta
+    Contract->>Contract: Emit StateCommitted event
+    
+    Node->>Node: Mine block
+    Driver->>Driver: Poll receipt
+    Driver->>Contract: getTokenRequestHash(txID)
+    Driver-->>App: OnStatus(Valid, requestHash)
+```
+
+### 2.2 Component Architecture
+
+```
+token/services/network/ethereum/
+├── driver.go              # Driver registration
+├── network.go             # Network interface implementation
+├── ledger.go              # Ledger interface implementation
+├── envelope.go            # Ethereum transaction envelope
+├── client.go              # JSON-RPC client abstraction
+├── eip712.go              # EIP-712 signing utilities
+├── config.go              # Configuration structures
+├── endorsement/
+│   └── esp.go             # Endorsement service provider
+├── finality/
+│   └── poller.go          # Receipt polling for finality
+└── contract/
+    └── abi.go             # Contract ABI encoding/decoding
+```
+
+---
+
+## 3. Smart Contracts
+
+### 3.1 Contract Architecture
+
+Two separate contracts provide separation of concerns and enable endorser policy updates without redeploying token state:
+
+1. **EndorsementVerifier**: Manages authorized endorsers and verifies signatures
+2. **TokenState**: Stores token data and delegates signature verification
+
+**Deployment Model**: Contracts are deployed externally using deployment tools (e.g., Foundry, Hardhat) before the driver starts. The driver reads contract addresses from configuration under `token.tms.<tms-id>.services.network.ethereum.*`.
+
+**Token Storage Model**: Each TMS refers to a single TokenState contract. All tokens managed by that TMS are stored in this one contract, regardless of token type (fabtoken or zkatdlog). The contract address is specified in the configuration:
+
+```yaml
+token:
+  tms:
+    mytms:
+      namespace: token
+      services:
+        network:
+          ethereum:
+            contracts:
+              tokenState: 0x1234...           # Single TokenState for this TMS
+              endorsementVerifier: 0x5678...  # EndorsementVerifier contract
+```
+
+Multiple TMS instances can be configured, each with its own TokenState contract if needed for organizational separation.
+
+### 3.2 EndorsementVerifier Contract
+
+**Purpose**: Centralized signature verification and endorser management.
+
+**Interface**:
+```solidity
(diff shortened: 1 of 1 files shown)


adecaro:
@atharrva01 @Storm1289 , please have a look at the design document. Your input is very valuable 🙏

atharrva01:
@adecaro,  I was going through the RequestApproval flow and got stuck on one thing. The diagram shows the endorser receives the token request, the pre-computed StateDelta, and the EIP-712 digest, and then "verifies stateDelta correctness" before signing. But I'm not sure what that verification actually looks like in practice.

Is the endorser supposed to independently recompute the StateDelta from the token request using the translator and check that it matches what the initiator sent? Or does it just validate the token request on its own and then sign whatever digest it's given?

Asking because if the endorser doesn't recompute the delta, a malicious initiator could send a valid token request but a tampered StateDelta, wrong output amounts, missing spentIDs, etc. The signature would end up covering a delta that doesn't actually match what was validated. Wanted to make sure I'm understanding this correctly before I start working on the translator.

Storm1289:
Hi @adecaro , just had a closer look at Section 8.2 (JSON-RPC Implementation) and Section 8.3 (fabric-x-evm Integration).
Section 8.2 looks fine as-is. The JSON-RPC Client lives inside the SDK driver itself, so there’s nothing extra the gateway needs to provide directly there. The JSON-RPC methods it depends on are already available in `fabric-x-evm` , and the recent geth-style error-code work should make failure classification straightforward on the driver side.
Section 8.3 and Section 10.2 are where the gaps are a bit more noticeable:
1. No clean NWO( Network Orchestrator) bootstrap path yet for `fabric-x-evm` (config YAML emission, contract-address wiring, endorser identity provisioning, etc.)
2. No forge deploy hook yet for `EndorsementVerifier` / `TokenState`
3. `eth_getLogs` topic filtering hasn’t really been exercised end-to-end yet, which Section 5.2’s anchor → receipt flow depends on heavily
4. Some assumptions in Section 6 (Finality Tracking) also overlap with the persistent mempool / isPending work still in progress on the gateway side.

Mostly wanted to flag these so we can figure out which need attention now and which can come later ,that way nothing surprises anyone down the line

x/token/services/network/evm/eth_network_driver_design.md
```diff
@@ -0,0 +1,2264 @@
...
+```solidity
+mapping(address => bool) endorsers;
+address[] endorserList;
+uint256 threshold;
+address admin;
+```
+
+**Signature Verification Logic**:
+1. Recover signer address from each signature using `ecrecover`
+2. Verify each recovered address is in the endorser set
```
arner: and unique... so we don't count 3 signatures of the same party 3 times.

x/token/services/network/evm/eth_network_driver_design.md
```diff
@@ -0,0 +1,2264 @@
...
+4. If verification fails, revert with "Invalid signatures"
+5. Check all spentIDs are not already spent (double-spend check)
+6. If double-spend detected, revert with "Double spend detected"
+7. Mark all spentIDs as spent
+8. Store all output tokens
+9. Store all metadata key-value pairs
+10. Store token request hash
+11. Emit StateCommitted event with success=true
+
+### 3.4 Token ID Mapping
```
arner, line 284: There is also an ethereum transaction id, generated very similarly to this, do we want to separate that from this internal id?

atharrva01:
@Storm1289 the `eth_getLogs` gap you flagged connects directly to what I was asking about earlier, Section 5.2's anchor → receipt resolution depends on log filtering working end-to-end, so if that path hasn't been tested yet, I'm not sure how much we can rely on that part of the design.

Two things I wanted to check:

Should we add a fallback for unreliable log queries, like storing `ethTxHash` somewhere the recipient can access off-chain? Or is this a hard dependency on fabric-x-evm shipping `eth_getLogs` support first?

On the `finality`/`isPending` overlap, if there's active work on the gateway side, I'd rather design Section 6 to align with it. Is there a doc or thread I can look at?

Storm1289:
@atharrva01 yeah, those line up pretty well.

On `eth_getLogs`: this is already implemented in `fabric-x-evm` (`gateway/api/eth.go:410`), and the current filter handling already supports indexed-topic alternatives, which is exactly what Section 5.2 needs for `StateCommitted(bytes32 indexed txID, ...)`. So I would not treat it as a hard “ship first” dependency what’s mainly missing right now is end-to-end coverage for that exact filter shape. That test coverage can land independently.

On the fallback path: I’d keep Section 5.2’s `anchor → log → receipt` flow as the canonical source of truth. An off-chain `ethTxHash` cache can still be useful as a recovery hint, but I would avoid making it authoritative; otherwise recipients implicitly depend on the initiator for finality information, which the current design explicitly tries to avoid.

For Section 6 alignment, the persistent-mempool / `isPending` work is already tracked in:

* `PR #127` — closed, but the discussion contains the lifecycle design
* `Issue #116` — original `TransactionByHash` pending-state bug
* `Issue #131` — `TxQueue` refactor (`pending` + `in-progress` tracking)
* `Issue #132` — wiring `isPending` into `TransactionByHash`
* `Issue #134` — gateway ↔ chain integration
* `Issue #50` — durable mempool with retry + MVCC handling

The lifecycle model there is:

`pending → in-progress → committed | failed | superseded`

TransactionByHash().isPending is already exposing that lifecycle externally as the queryable state signal.

So if Section 6 models the same lifecycle, the finality poller should fit naturally on top of the existing queue + receipt flow instead of introducing a second parallel transaction-state system. That keeps the gateway as the single source of truth for tx state and avoids duplicated lifecycle tracking logic.

atharrva01:
Thanks for the context @Storm1289 , really helpful.

On Section 6, I've been looking at how the FabricX finality queue works (spent a fair bit of time in that code) and what you're describing actually maps onto it pretty naturally. If the poller uses TransactionByHash().isPending as the signal instead of hitting eth_getTransactionReceipt directly, the lifecycle transitions basically become the events the queue processes. OnlyOnceListener handles the rest. Gateway stays the single source of truth, nothing gets duplicated.

Is that roughly what you had in mind or am I missing something?

Storm1289:
@atharrva01 ,Yeah, that's basically it.
Using isPending as the primary signal, OnlyOnceListener for exactly-once notification, gateway as single source of truth that lines up with what's being built on the gateway side.

Two small things worth factoring in:

1. isPending=false is necessary but not sufficient, once it flips, the poller still has to look at blockNumber or the receipt to tell committed from failed from dropped. So the loop is roughly: isPending=true → wait; isPending=false with blockNumber set → fetch receipt; isPending=false with no blockNumber → treat as dropped.
2. Replacement transactions (hyperledger/fabric-x-evm#62) introduce a superseded case where the old hash gets a synthetic status=0 receipt. Section 6 should resolve to that rather than time out if it's watching the old hash.

Other than that, the FabricX finality queue mapping over is the right read, the lifecycle is exactly what that queue would process if it had an EVM backend underneath.

adecaro:
Hi @arner , please, don't forget to review this. Many thanks 🙏

adecaro:
@alexandrosfilios , would you like to give it a look as well, please? 🙏

x/token/services/network/evm/eth_network_driver_design.md
adecaro: indeed, very important point

x/token/services/network/evm/eth_network_driver_design.md
adecaro: Good sanity check. Let's see if I can derive a convincing argument here. Image Alice is transferring money to Bob. Bob might only see the token request and therefore the internal ID. How would then Bob learn what happened to the transactions and the relative tokens if we store them using the Eth transaction ID? Bob cannot, right? This is the reason  StateDelta carries a txID. We could even call it TokenRequestAnchor to be less ambiguos. What do you think?

arner: Yes, good point. In hindsight, pre-generating the ethereum id only creates coupling that is not helpful but can only get in the way.

x/token/services/network/evm/eth_network_driver_design.md
```diff
@@ -0,0 +1,2264 @@
...
+- **Endorser Discovery**: Dynamic endorser registry via contract queries for runtime flexibility
+- **Configuration Validation**: Comprehensive validation with sensible defaults for production readiness
+
+### Out of Scope for v1
+
+- Cross-chain interoperability
+- Contract upgradeability
+- Advanced gas estimation (basic heuristic provided)
+- Full recovery mechanisms (basic recovery included, advanced features deferred to v2)
+- Batch operations (single transaction per state delta in v1)
```
alexandrosfilios, line 33: Batch operations being out of scope is fine, but it's worth making an explicit design decision here between two approaches for v1:
1. raw ETH transactions via `eth_sendRawTransaction`: simple, but one ETH tx per token operation will incure a high cost on mainnet, and the submitter needs ETH to sponsor the txs. In this case, it may be worth designing the contract interface now so a future `applyStateDeltaBatch(StateDelta[])` can be added without a breaking change, e.g. making the single-delta path internally a wrapper around a batch loop. Retrofitting batching later if the interface is rigid would require a contract migration.
2. ERC-4337 UserOperations via `eth_sendUserOperation`: each token operation becomes a UserOp, external bundlers (Pimlico, Alchemy, etc.) handle packing multiple UserOps into a single ETH transaction without us implementing bundling, and a paymaster contract can sponsor gas so FSC nodes don't need to hold ETH. Option 2 would also resolve the tx sponsoring gap. The choice affects the `Broadcast()` interface and the contract architecture (EntryPoint integration), so it's harder to swap later.

x/token/services/network/evm/eth_network_driver_design.md
```diff
@@ -0,0 +1,2264 @@
...
+│  - Decouples polling from listener notification             │
+└─────────────────────────────────────────────────────────────┘
+```
+
+**Configuration**:
+```yaml
+ethereum:
+  finality:
+    pollInterval: 2s        # How often to check for receipts
+    confirmationDepth: 12   # Blocks to wait (1 for devnet, 12+ for mainnet)
```
alexandrosfilios, line 1357: This is a PoW-era heuristic. Does it still make sense on PoS Ethereum? The network now provides `safe` and `finalized` block tags natively. If we want to match FabricX semantics and avoid having to handle reorgs, finalized (~2 epochs, ~12.8 min) seems like the right default for v1, even if it's slower. It might also be worth exposing two levels (soft/included vs. hard/finalized) so callers can choose, rather than baking one assumption into the driver. Otherwise anything using `safe` or `latest` would need reorg reconciliation which feels like a lot of extra complexity for v1.

x/token/services/network/evm/eth_network_driver_design.md
```diff
@@ -0,0 +1,2264 @@
...
+
+- `IssueAction.GetMetadata()`
+- `TransferAction.GetMetadata()`
+
+Unlike Fabric RWSet translation, where metadata is written under derived keys, Ethereum translation should turn each metadata entry into a deterministic `(key, value)` pair in the `StateDelta`.
+
+Conceptually:
+
+```go
+for key, value := range action.GetMetadata() {
```
alexandrosfilios, line 724: Should there be a `MAX_METADATA_VALUE_SIZE` guard here? For zkatdlog actions `GetMetadata()` can return cryptographic material that's potentially large, and right now there's nothing stopping a value from being arbitrarily big before it hits the contract. Or is the assumption that we trust the issuer/transferrer not to abuse this?

x/token/services/network/evm/eth_network_driver_design.md
```diff
@@ -0,0 +1,2264 @@
...
+## 3. Smart Contracts
+
+### 3.1 Contract Architecture
+
+Two separate contracts provide separation of concerns and enable endorser policy updates without redeploying token state:
+
+1. **EndorsementVerifier**: Manages authorized endorsers and verifies signatures
+2. **TokenState**: Stores token data and delegates signature verification
+
+**Deployment Model**: Contracts are deployed externally using deployment tools (e.g., Foundry, Hardhat) before the driver starts. The driver reads contract addresses from configuration under `token.tms.<tms-id>.services.network.ethereum.*`.
```
alexandrosfilios, line 142: Not sure if this is in scope for this doc since the contract implementation details aren't really defined here, but worth flagging: if each TMS gets its own `TokenState` and potentially its own `EndorsementVerifier` (the config format allows per-TMS addresses for both), deploying full bytecode every time gets expensive and operationally messy. A proxy pattern would help on two fronts:
1. cheap per-TMS deployment via clones/proxies instead of full bytecode, and
2. version updates: a beacon proxy in particular would let you upgrade all TMS contract instances in one operation by pointing the beacon to a new implementation, rather than migrating each one individually. Cheap per-TMS upgrades would also be possible though with transparent proxies. If we are convinced that no upgrades are needed, though, also a minimal clone will still be a cost benefit.

Contract upgradeability is listed as out of scope for v1, but picking the right proxy pattern upfront avoids a painful migration later.

x/token/services/network/evm/eth_network_driver_design.md
adecaro: Hi @alexandrosfilios , interesting point. Think about StateDelta as the result of the translator of a token request. It can contain as many operations as required by a given business process. 

Anyhow, Option 2 seems quite nice, indeed. I'll include a reference to ERC-4337.

x/token/services/network/evm/eth_network_driver_design.md
adecaro: this is interesting and depend on the specific DLT (Ethereum, Besu, FabricX+EVM). I think this can be improved a lot as the driver evolves. Good point though.

x/token/services/network/evm/eth_network_driver_design.md
adecaro: No, we cannot trust them and should be the job of the approver to impose these limits. Good point 👍

atharrva01:
@adecaro, I was reading through translator.go and rwset.go to understand how to build the EVM translator, and I think there's a semantic mismatch in how §5.1 models spent-tracking.

The two token modes use opposite polarities:

Graph-revealing (fabtoken): each output is stored under two keys, CreateOutputKey(txID, index) and CreateOutputSNKey(txID, index, outputBytes), where CreateOutputSNKey hashes the serialized output bytes into the key (rwset.go:35-36). checkInputs requires the output-SN key to exist before spending it (translator.go:443-450).
spendInputs then deletes both keys (translator.go:465-489). Spent = absence. AreTokensSpent returns true when len(v) == 0 (translator.go:192).

Graph-hiding (zkatdlog): checkInputs requires the input-SN key to not exist (translator.go:428-432). spendInputs writes the key (translator.go:492-506). Spent = presence. AreTokensSpent returns true when len(v) != 0 (translator.go:183).

The design's mapping(bytes32 => bool) spent and append-only spentIDs in §3.3 only capture the graph-hiding (presence) case. For fabtoken, the mandatory StateMustExist(outputSNKey) check at translator.go:443-450 — which verifies the input actually exists before it can be spent, would have no contract-level equivalent.

Also, the existing HashedKeyTranslator at rwset.go:130-226 already uses SHA-256 with a per-key-type 1-byte domain separator for collision separation, which is a stricter version of what the "SERIAL_NUMBER" prefix in §5.1 is trying to do. Should the EVM KeyTranslator follow the same scheme rather than introducing a new one?

Is the intention that input-existence is enforced entirely off-chain by endorsers, with the contract only handling double-spend? If so I think §5.1 should state that explicitly, because as written it reads as if the single spentIDs set handles both modes equivalently.

adecaro:
> @adecaro, I was reading through translator.go and rwset.go to understand how to build the EVM translator, and I think there's a semantic mismatch in how §5.1 models spent-tracking.
> 
> The two token modes use opposite polarities:
> 
> Graph-revealing (fabtoken): each output is stored under two keys, CreateOutputKey(txID, index) and CreateOutputSNKey(txID, index, outputBytes), where CreateOutputSNKey hashes the serialized output bytes into the key (rwset.go:35-36). checkInputs requires the output-SN key to exist before spending it (translator.go:443-450). spendInputs then deletes both keys (translator.go:465-489). Spent = absence. AreTokensSpent returns true when len(v) == 0 (translator.go:192).

The `OutputKey` is used to be able to address any token via <transaction id:index>. These must always be there for each output.

Now, if the action is `graph hiding`, we will have only `serial numbers` that should not exist already in the recorded list of serial numbers. In this case, the action does return an empty list of inputs cause the action is graph hiding.

If the action is not graph hiding, then the action will return an empty list of serial numbers and a non-empty list of inputs.

> 
> Graph-hiding (zkatdlog): checkInputs requires the input-SN key to not exist (translator.go:428-432). spendInputs writes the key (translator.go:492-506). Spent = presence. AreTokensSpent returns true when len(v) != 0 (translator.go:183).
> 
> The design's mapping(bytes32 => bool) spent and append-only spentIDs in §3.3 only capture the graph-hiding (presence) case. For fabtoken, the mandatory StateMustExist(outputSNKey) check at translator.go:443-450 — which verifies the input actually exists before it can be spent, would have no contract-level equivalent.
> 
> Also, the existing HashedKeyTranslator at rwset.go:130-226 already uses SHA-256 with a per-key-type 1-byte domain separator for collision separation, which is a stricter version of what the "SERIAL_NUMBER" prefix in §5.1 is trying to do. Should the EVM KeyTranslator follow the same scheme rather than introducing a new one?

I think we don't need to use HashedKeyTranslator here.

> 
> Is the intention that input-existence is enforced entirely off-chain by endorsers, with the contract only handling double-spend? If so I think §5.1 should state that explicitly, because as written it reads as if the single spentIDs set handles both modes equivalently.

So, it makes sense to extend StateDelta with both a list of serial numbers that should not exists and a list of inputs that should exists. This way, we would be able to support both graph hiding and graph non-hiding.

What do you think?

PR #1702, opened 2026-05-13
fsc v0.11.0

adecaro:


token/services/identity/x509/crypto/ecdsa_test.go
```diff
@@ -136,9 +136,7 @@ func TestPemDecodeKey(t *testing.T) {
 		assert.True(t, ok)
 
 		// Verify the decoded key matches the original
-		assert.Equal(t, priv.D, decodedPriv.D)
-		assert.Equal(t, priv.X, decodedPriv.X)
-		assert.Equal(t, priv.Y, decodedPriv.Y)
+		assert.True(t, priv.Equal(decodedPriv))
```
mbrandenburger, line 139: are we missing the check of the public key?

mbrandenburger (approved):

token/services/identity/x509/crypto/ecdsa_test.go
adecaro: it is all included, I did check the implementation.

PR #1706, opened 2026-05-13
security: enforce strict input validation and payload bounds

### LINKED ISSUE #1608: Enforce Strict Input Validation and Payload Bounds on Token Store Writes [MED]
**Description**
Even after a delivery event has been authenticated, the finality listener must not blindly commit its content to the Token Store. Before invoking `StoreToken` or `DeleteTokens`, enforce explicit bounds and well-formedness checks on the incoming data: maximum payload size for raw token bytes (`ledger`, `ledger_metadata`), maximum number of token outputs per transaction, maximum batch size for bulk delete operations, and rejection of structurally invalid or oversized `wallet_id` / `owner_raw` / `issuer_raw` fields. These checks should be applied as a validation stage in the `tokens.Service.Append` call chain, before any SQL statement is issued, ensuring that even a legitimate but misbehaving peer cannot cause unbounded storage or CPU consumption in `tokendb`.

**Short Summary**
Resource exhaustion / DoS attack. Malicious users can create toxic transaction that tokenSDK will submit to the ledger and receive back finality events that cause this issue. A solution can be limiting the message size, metadata size, etc. and some validation checks.
To address this issue we need:
(1) To prevent the Ledger from sending large messages in finality we need to limit the protobuf messages. This needs to be updated in the documentation accordingly.
(2) When Alice transfers to Bob, bob needs to validate that Alice's token is not too big.

SurbhiAgarwal1:
## Description
This PR implements strict input validation and payload bounds within the `tokens.Service.Append` method to protect the system against resource exhaustion and potential Denial of Service (DoS) attacks from malformed or oversized finality data.

## Changes
- **ValidationConfig**: Introduced a configurable validation layer with safe default limits:
    - `MaxTokenPayloadSize`: 1MB (covers `TokenOnLedger` and `TokenOnLedgerMetadata`)
    - `MaxTokenOutputsPerTx`: 1000 outputs
    - `MaxBulkDeleteSize`: 10,000 IDs
    - `MaxWalletIDSize`: 1024 bytes
    - `MaxOwnerRawSize`: 2048 bytes
    - `MaxIssuerRawSize`: 2048 bytes
- **AppendRequest**: Refactored the internal data flow to use a request-based structure for validated processing.
- **Strict Validation Logic**: Implemented `validateAppendRequest` which performs in-memory checks *before* any database transactions or storage operations are initiated.
- **Enhanced Testing**: Added a comprehensive test suite in `token/services/tokens/validation_test.go` covering edge cases, limit violations, and structural validation.

## Verification Results
- **Unit Tests**: All tests in `token/services/tokens` passed successfully.
- **Linting**: Code is clean according to `golangci-lint` (v2) standards, including `nlreturn` and `testifylint` compliance.

Fixes: #1608


token/services/tokens/tokens.go
```diff
@@ -125,6 +142,40 @@ func (t *Service) AppendValid(ctx context.Context, tx dbdriver.Transaction, txID
...
+			MaxActionCount:       1000,
+		}
+		if t.TMSProvider != nil {
+			tms, err := t.TMSProvider.GetManagementService(token.WithTMSID(t.tmsID))
+			if err == nil && tms != nil && tms.Configuration() != nil {
+				if vConfig, err := tms.Configuration().GetValidationConfig(); err == nil {
+					t.Config = vConfig
+				}
+			}
+		}
```
HayimShaul: it is better to call this initialization from TMSProvider in the constructor of ValidateConfig, or initialize t.Config when creating a new service.

Now there's a race condition where one thread may init t.Config and another thread may loose context after initing t.Config with the defaults but before reading from TMSProvider

plan.md
```diff
@@ -0,0 +1,17 @@
+# Implementation Plan - Optimize and Fix Validation Bounds for Idemix Identities
+
```
HayimShaul: remove this file from the PR

token/config.go
```diff
@@ -22,10 +22,35 @@ func NewConfiguration(cm driver.Configuration) *Configuration {
...
 
 // UnmarshalKey takes a single key and unmarshals it into a Struct
 func (m *Configuration) UnmarshalKey(key string, rawVal any) error {
 	return m.cm.UnmarshalKey(key, rawVal)
 }
+
+// GetValidationConfig returns the validation configuration
+func (m *Configuration) GetValidationConfig() (driver.ValidationConfig, error) {
+	config := driver.ValidationConfig{
+		MaxTokenPayloadSize:  2 * 1024 * 1024,
```
HayimShaul: Please document these new limits in docs/configuration.md

token/services/tokens/tokens.go
```diff
@@ -125,6 +142,40 @@ func (t *Service) AppendValid(ctx context.Context, tx dbdriver.Transaction, txID
...
+			MaxIssuerRawSize:     256 * 1024,
+			MaxTokenRequestSize:  2 * 1024 * 1024,
+			MaxActionCount:       1000,
+		}
+		if t.TMSProvider != nil {
+			tms, err := t.TMSProvider.GetManagementService(token.WithTMSID(t.tmsID))
+			if err == nil && tms != nil && tms.Configuration() != nil {
+				if vConfig, err := tms.Configuration().GetValidationConfig(); err == nil {
+					t.Config = vConfig
+				}
```
HayimShaul: If there reading configuration failed and we fall back to default values, the user must be notified about it

token/services/tokens/tokens.go
SurbhiAgarwal1: Done. I've moved the initialization of t.Config directly into the NewService constructor. This ensures the configuration is fully loaded from the TMSProvider before the service is returned, eliminating any potential race conditions.

plan.md
SurbhiAgarwal1: Done. I have removed the plan.md file from this PR.

token/config.go
SurbhiAgarwal1: Done. I have documented the validation limits and their safe default values in docs/configuration.md under a new Optional: token.validation section.

token/services/tokens/tokens.go
SurbhiAgarwal1: Done. I've added explicit logger.Warnf statements. If reading the configuration from the TMSProvider fails or if the TMS is missing, it now logs a warning to notify the user before falling back to the default validation limits.

token/services/tokens/tokens.go
```diff
@@ -160,6 +211,49 @@ func (t *Service) AppendValid(ctx context.Context, tx dbdriver.Transaction, txID
...
+			return errors.Errorf("token metadata too large at index %d: %d > %d", i, len(tok.TokenOnLedgerMetadata), t.Config.MaxTokenPayloadSize)
+		}
+
+		// OwnerWalletID can be empty in audit/sync flows where the wallet is not local.
+
+		if len(tok.OwnerWalletID) > t.Config.MaxWalletIDSize {
+			return errors.Errorf("wallet_id too large at index %d: %d > %d", i, len(tok.OwnerWalletID), t.Config.MaxWalletIDSize)
+		}
+
+		if len(tok.Tok.Owner) > t.Config.MaxOwnerRawSize {
```
HayimShaul: is tok.Tok.Owner allowed to be nil?
if so, please document that. if not, please add a check for that

SurbhiAgarwal1: Thanks for the review and for catching that, @HayimShaul!

`tok.Tok.Owner` is not allowed to be nil (or empty) for unspent tokens being appended to the local store. (Redeemed tokens, which *do* have an empty owner, are explicitly skipped during the `Parse()` phase and never reach this validation logic).

I've pushed a commit that adds a strict check for `len(tok.Tok.Owner) == 0` to immediately return an error, and I've updated the test suite in `validation_test.go` to explicitly cover this missing owner case. Let me know if everything looks good now!

HayimShaul:
Looks good to me.
@adecaro do you want to have another quick look?

HayimShaul (approved):

token/core/common/validator.go
```diff
@@ -102,21 +115,34 @@ func (v *Validator[P, T, TA, IA, DS]) SetMinProtocolVersion(version uint32) {
...
 		return nil, nil, errors.Wrap(err, "failed to unmarshal token request")
 	}
 
+	if v.ValidationConfig.MaxActionCount > 0 && len(tr.Issues)+len(tr.Transfers) > v.ValidationConfig.MaxActionCount {
+		return nil, nil, errors.Errorf("too many actions: %d > %d", len(tr.Issues)+len(tr.Transfers), v.ValidationConfig.MaxActionCount)
+	}
+
 	// Validate protocol version
 	if tr.Version == 0 {
-		return nil, nil, driver.ErrInvalidVersion
```
AkramBitar, line 119: Version 0 is now accepted, but it used to be rejected. Why this is removed? This makes validation weaker, which is the opposite of what this PR is for, and it's not related to payload size limits.

token/services/tokens/tokens.go
```diff
@@ -85,10 +95,54 @@ type Service struct {
...
 func NewService(tmsID token.TMSID, TMSProvider TMSProvider, networkProvider NetworkProvider, storage *DBStorage, requestsCache Cache) *Service {
-	return &Service{tmsID: tmsID, TMSProvider: TMSProvider, NetworkProvider: networkProvider, Storage: storage, RequestsCache: requestsCache}
+	s := &Service{
+		tmsID:           tmsID,
+		TMSProvider:     TMSProvider,
+		NetworkProvider: networkProvider,
+		Storage:         storage,
+		RequestsCache:   requestsCache,
+		Config: ValidationConfig{
+			MaxTokenPayloadSize:  2 * 1024 * 1024,
```
AkramBitar: Defaults duplicated in 4 places

1) token/driver/validator.go:
const (
    MaxTokenPayloadSize  = 2 * 1024 * 1024
    MaxTokenOutputsPerTx = 1000
    ...
)

2) token/config.go — GetValidationConfig:
config := driver.ValidationConfig{
    MaxTokenPayloadSize:  2 * 1024 * 1024,
    MaxTokenOutputsPerTx: 1000,
    ...
}

3) token/core/common/validator.go — NewValidator:
ValidationConfig: driver.ValidationConfig{
    MaxTokenPayloadSize:  2 * 1024 * 1024,
    MaxTokenOutputsPerTx: 1000,
    ...
},

4) token/services/tokens/tokens.go — NewService:
Config: ValidationConfig{
    MaxTokenPayloadSize:  2 * 1024 * 1024,
    MaxTokenOutputsPerTx: 1000,
    ...
},

One option to fix this is to define once in token/driver/validator.go, reuse everywhere:

// DefaultValidationConfig holds the default resource limits.
var DefaultValidationConfig = ValidationConfig{
    MaxTokenPayloadSize:  2 * 1024 * 1024,
    MaxTokenOutputsPerTx: 1000,
    MaxBulkDeleteSize:    10000,
    MaxWalletIDSize:      1024,
    MaxOwnerRawSize:      256 * 1024,
    MaxIssuerRawSize:     256 * 1024,
    MaxTokenRequestSize:  2 * 1024 * 1024,
    MaxActionCount:       1000,
}

token/core/common/validator.go
```diff
@@ -251,6 +289,27 @@ func (v *Validator[P, T, TA, IA, DS]) VerifyIssue(
...
+
+	// Check outputs
+	if v.ValidationConfig.MaxTokenPayloadSize > 0 || v.ValidationConfig.MaxOwnerRawSize > 0 {
+		outputs := action.GetOutputs()
+		for i, output := range outputs {
+			raw, err := output.Serialize()
+			if err != nil {
+				return errors.Wrapf(err, "failed to serialize output at index %d", i)
+			}
+			if v.ValidationConfig.MaxTokenPayloadSize > 0 && len(raw) > v.ValidationConfig.MaxTokenPayloadSize {
```
AkramBitar, line 306: A limit of 0 means two opposite things:
- common/validator.go: 0 = no limit (checks are guarded with if MaxX > 0)
- tokens.go validateAppendRequest: 0 = reject everything (no guard, so len(req.Tokens) > 0 always fails)

So an operator who sets maxTokenOutputsPerTx: 0 expecting "unlimited" would instead block every append.

I think we need to fix this and add the same > 0 guard in tokens.go so 0 means "unlimited" in both places.

AkramBitar (changes requested):

AkramBitar:
Hello @SurbhiAgarwal1

I would like to thank you a lot for your work on that PR. I appreciate it a lot. 
I added few new comments. 
Could your please let me know if you have the time to look at these comments and rebase the PR (have some conflicts)?

Thanks a lot,
Akram

adecaro:
Hi @SurbhiAgarwal1 , I'll close this cause it has been super-seeded by #1948. Thanks for your effort and thanks for the understanding.

PR #1707, opened 2026-05-14
limit the number of tokens a transaction retrieves and locks

### LINKED ISSUE #1641: Bound All Work Triggered by a Single Token Selection Request [HIGH]
**Description**
Conduct an architectural review of every code path reachable from a `Selector.Select` call  --  token iteration depth, number of lock attempts, retry and back-off cycles, and lock-store growth  --  and impose hard, configurable upper bounds on each dimension. Any request that would exceed these limits must be aborted early and all acquired locks released immediately. This is the primary defence against algorithmic attacks: no matter how a request is crafted (adversarial `OwnerFilter`, extreme quantity, or high concurrency), the system must shed load and return an error rather than consuming unbounded memory, CPU, storage, or wall-clock time.

**Short Summary**
An attacker can create a transaction that selects a large amount of tokens from database. This can exhaust the CPU and memory of the view service.

HayimShaul:
Fixes #1641

Imposes hard, configurable upper bounds on every dimension of work reachable from a `Selector.Select` call, so that a request cannot consume unbounded memory, CPU, storage or wall-clock time no matter how it is crafted. A request that would exceed a bound is aborted and its locks released.

## Bounds

All under the `token.selector.limits` config key:

| Limit | Default | Bounds |
|---|---|---|
| `maxTokensPerSelection` | 10000 | tokens iterated per selection cycle; also pushed down into SQL as a row limit |
| `maxLockAttempts` | 50000 | lock attempts per selection |
| `maxRetries` | 10 | outer retry/back-off cycles |
| `maxLocksPerTransaction` | 5000 | locks held concurrently by one transaction |
| `selectionTimeout` | derived, see below | wall-clock time for the whole selection |

The row limit is pushed into the query rather than applied in Go, so a wallet with millions of tokens no longer streams all of them into the view service. Tokens come back largest-first (`ORDER BY amount DESC`) so a selection reaches its target in the fewest rows.

Adds the `token.SelectorTimedOut` sentinel, so callers can tell a timeout apart from a permanent failure. `token.SelectorRateLimited` (pre-existing) is now what both built-in lockers return when a transaction hits `maxLocksPerTransaction`, making that a fail-fast condition rather than something callers retry.

Configuration is validated at startup; an invalid or unparseable `token.selector` block logs and falls back to defaults.

## Behaviour changes reviewers should know about

- **`selectionTimeout` default is no longer a flat 30s.** It is `30s + maxRetries * retryInterval`, i.e. **80s** with the defaults. A flat 30s expires while the retries meant to resolve contention are still backing off, turning ordinary contention into a timeout the caller cannot resolve. An explicitly configured value is used verbatim. A non-positive value means no timeout.
- **A full page of tokens with insufficient funds now reports the iteration limit** instead of retrying. The query ordering is deterministic, so a retry re-reads the same page; the wallet simply holds more tokens than the selection is permitted to examine.
- **Contention no longer surfaces as an untyped error.** The iteration budget applies per retry cycle, so the typed reason (`SelectorSufficientButLockedFunds`) reaches the caller instead of being replaced by a generic limit error on a later cycle.

## Docs

`docs/security/selector_resource_limits.md` documents each limit, its default, when it triggers, and tuning guidance.

## Testing

Unit tests per limit, plus a `dbtest` case that exercises the limited query against **both SQLite and PostgreSQL** — the two dialects emit placeholders differently, so the shared case is what keeps the pushed-down limit honest on both.

---

<sub>Follow-up commit `d1cb0422` addresses the review in [this comment](https://github.com/LFDT-Panurus/panurus/pull/1707#issuecomment-5205446095): the pushed-down limit was emitted with a `?` placeholder (invalid on PostgreSQL), was counted before de-duplication, and the release-before-back-off in `StubbornSelector` had been dropped. It also restores the two test relaxations those regressions required.</sub>


adecaro:
Hi @HayimShaul , should these changes apply also to `sherdlock`?

token/services/selector/sherdlock/manager.go
```diff
@@ -36,6 +36,10 @@ func NewManager(
 	maxRetriesAfterBackOff int,
 	leaseExpiry time.Duration,
 	leaseCleanupTickPeriod time.Duration,
+	maxTokensPerSelection int,
```
adecaro: please, have a Config struct to pass here so we put in it all the configurations

token/services/selector/simple/manager.go
```diff
@@ -22,6 +22,12 @@ type Manager struct {
 	timeout              time.Duration
 	requestCertification bool
 	precision            uint64
+
+	// Resource limits
+	maxTokensPerSelection int
```
adecaro, line 37: what's the relation of these values with the existing numRetry and timeout?

token/services/selector/sherdlock/selector.go
```diff
@@ -55,13 +64,52 @@ type StubbornSelector struct {
...
 			m.metrics.SelectionDuration.Observe(time.Since(start).Seconds())
+			m.metrics.SelectionOutcome.With(outcomeLabel, "error").Add(1)
+
+			return nil, nil, errors.Errorf(
+				"token selection aborted: exceeded max retry cycles (%d) after examining %d tokens and %d lock attempts",
+				m.maxRetryCycles, m.tokensIteratedCount, m.lockAttemptsCount,
+			)
+		}
+
+		if tokens, quantity, err := m.selectWithoutMetrics(timeoutCtx, ownerFilter, q, tokenType); err == nil || !errors.Is(err, token.SelectorSufficientButLockedFunds) {
```
adecaro: selectWithoutMetrics calls UnlockAll in case of a failure, but if the error is due context cancellation, UnlockAll will also fail. Maybe we don't need to call UnlockAll from inside selectWithoutMetrics

token/services/selector/sherdlock/selector_shutdown_test.go
```diff
@@ -46,13 +46,7 @@ func TestStubbornSelector_ContextCancellation(t *testing.T) {
...
 		// fires in the backoff select before time.After can.
 		sel := NewStubbornSelector(
-			logger,
-			mockFetcher,
-			mockLck,
-			64,
-			time.Hour,
-			10,
-			m,
+			logger, mockFetcher, mockLck, 64, time.Hour, 10, 10000, 50000, 10, 30*time.Second, m,
```
adecaro: why hardcoded values here?

adecaro:
@HayimShaul , don't we need to enforce limit also at the fetcher level `token/services/selector/sherdlock/fetcher.go` when the token db is queried?

A more general question, if we say that we can only load 100 tokens at the time from the storage, what if for these 100 tokens the selection request cannot be satisfied, but if had picked the tokens differently we would have. We might reject requests that can be satisfied.

HayimShaul:
> Hi @HayimShaul , should these changes apply also to `sherdlock`?

done

token/services/selector/sherdlock/selector.go
HayimShaul: fixed

token/services/selector/simple/manager.go
HayimShaul: consolidated numRetry and maxRetry

token/services/selector/sherdlock/selector_shutdown_test.go
HayimShaul: not sure I understand. This is a test. Do you want to make them configurable?

adecaro: my bad, I didn't notice it was a test.

docs/security/selector_resource_limits.md
```diff
@@ -0,0 +1,337 @@
+# Selector Resource Limits - Security Guide
+
+> **Performance Note**: Database-level LIMIT optimization has been implemented to reduce I/O overhead.
+
+
+## Overview
+
+The Fabric Token SDK selector service implements hard resource limits to prevent algorithmic attacks that could exhaust system resources (CPU, memory, storage) through maliciously crafted token selection requests.
```
AkramBitar: Change all the occurrence of Fabric Token SDK to panurus

AkramBitar (changes requested):

HayimShaul:
> @HayimShaul , don't we need to enforce limit also at the fetcher level `token/services/selector/sherdlock/fetcher.go` when the token db is queried?
> 
> A more general question, if we say that we can only load 100 tokens at the time from the storage, what if for these 100 tokens the selection request cannot be satisfied, but if had picked the tokens differently we would have. We might reject requests that can be satisfied.

done. we take the biggest tokens

token/services/selector/sherdlock/fetcher.go
```diff
@@ -103,9 +103,10 @@ func NewMixedFetcher(tokenDB TokenDB, m *Metrics, cacheSize int64, freshnessInte
 }
 
 // UnspentTokensIteratorBy returns an iterator for unspent tokens, trying cached results first, falling back to database query.
-func (f *mixedFetcher) UnspentTokensIteratorBy(ctx context.Context, walletID string, currency token2.Type) (Iterator[*token2.UnspentTokenInWallet], error) {
+// The limit parameter is ignored by sherdlock as it uses its own token limiting mechanism.
```
AkramBitar, line 106: **The limit is dropped here — and sherdlock is the default driver.**

All three fetchers document that they ignore it, and `selectInternal` passes `0` at both call sites (lines 273, 310). `simple/selector.go` passes `s.maxTokensPerSelection`.

So the vector #1641 names — an unbounded DB query — is bounded only in the non-default driver, and every production caller (`role/wallets.go`, `htlc/wallet.go`, `nfttx/filter.go`, `multisig/wallet.go`, `boolpolicy/wallet.go`) passes `0`.

Either push the limit into sherdlock's query, or drop the parameter and skip the API break. A test asserting the limit reaches SQL on the sherdlock path would have caught this.

token/services/selector/simple/inmemory/locker.go
```diff
@@ -175,6 +181,25 @@ func (d *locker) lockInShard(ctx context.Context, s *shard, owner string, id *to
 	if s.pruned {
 		return "", errShardPruned
 	}
+
+	// Check lock count limit for this transaction (if configured). A single
+	// selection locks tokens for one owner, so all of a transaction's locks
+	// live in this owner's shard and counting within it is per-transaction.
+	if d.maxLocksPerTx > 0 {
+		txLockCount := 0
+		for _, entry := range s.locked {
```
AkramBitar: **O(n²) on the hot path.** This scans the whole owner shard on every `TryLock`. With defaults (10 000 iterations, up to 5 000 locks) that is tens of millions of comparisons per selection, inside the shard lock — an algorithmic cost added by the defence against algorithmic attacks.

Fix: keep a `map[txID]int`, incremented on lock and decremented on unlock. O(1).

token/sdk/network/selector.go
```diff
@@ -48,5 +60,5 @@ func (s *LockerProvider) New(network, channel, namespace string) (selector.Locke
 		return nil, err
 	}
 
-	return inmemory.NewLocker(db, s.sleepTimeout, s.validTxEvictionTimeout), nil
+	return inmemory.NewLockerWithLimits(db, s.sleepTimeout, s.validTxEvictionTimeout, s.maxLocksPerTx), nil
```
AkramBitar, line 63: `maxLocksPerTransaction` only reaches the `simple` driver's in-memory locker. Sherdlock uses the SQL-backed lock store, so the default driver has no lock-count bound — same gap as the token limit.

token/services/selector/sherdlock/selector.go
```diff
@@ -159,8 +266,22 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
...
+	if s.cache == nil {
+		s.logger.DebugfContext(ctx, "Initializing token cache for first selection")
+		if s.cache, err = s.fetcher.UnspentTokensIteratorBy(ctx, owner.ID(), tokenType, 0); err != nil {
+			return nil, nil, 0, errors.Wrapf(err, "failed to initialize token cache for [%s:%s]", owner.ID(), tokenType)
+		}
+	}
+
 	sum, selected, tokensLockedByOthersExist, immediateRetries := token2.NewZeroQuantity(s.precision), collections.NewSet[*token2.ID](), true, 0
 	for {
+		// Double-check cache is not nil (defensive programming against race conditions)
```
AkramBitar: **This is a race with a nil check in front of it, not a fixed race.** `mu` exists specifically to protect `cache` against concurrent `Close()`, but `selectInternal` reads and assigns `s.cache` without taking it (also lines 273, 310), and `tokensIteratedCount`/`lockAttemptsCount` are unsynchronized too.

Take `mu`, or make the counters atomic. Worth a `-race` test calling `Close()` concurrently with `Select` — nothing exercises that today.

token/services/selector/sherdlock/selector.go
```diff
@@ -55,13 +64,71 @@ type StubbornSelector struct {
...
+
+					return nil, nil, errors.Wrapf(
+						token.SelectorSufficientButLockedFunds,
+						"token selection aborted: exceeded timeout (%v) after examining %d tokens and %d lock attempts",
+						m.selectionTimeout, m.tokensIteratedCount, m.lockAttemptsCount,
+					)
+				}
+				m.metrics.SelectionOutcome.With(outcomeLabel, "error").Add(1)
+
+				return nil, nil, fmt.Errorf(
```
AkramBitar: AGENTS.md: *"Never use `fmt.Errorf` (or `fmt` at all) to build or wrap errors."* Use the FSC `errors` package, as the rest of this file does. Same at line 226.

Separately, wrapping a timeout as `SelectorSufficientButLockedFunds` (line ~100) overstates what happened — a timeout does not establish that funds were sufficient, and that error tells callers to retry, which inverts load shedding. A distinct sentinel would be clearer.

token/services/selector/config/driver.go
```diff
@@ -105,3 +129,93 @@ func (c *Config) GetFetcherCacheMaxQueries() int {
...
+	return c.GetLimits().MaxLocksPerTransaction
+}
+
+// GetSelectionTimeout returns the wall-clock timeout for selection operations
+func (c *Config) GetSelectionTimeout() time.Duration {
+	return c.GetLimits().SelectionTimeout
+}
+
+// Validate checks that the configuration is valid
+func (c *Config) Validate() error {
```
AkramBitar, line 264: `Validate()` calls `GetLimits()`, which already coerces every non-positive value to a default — so the five `<= 0` checks can never fire and only the two relational checks can. Validate the raw `c.Limits` instead.

Also in this file: `GetNumRetries` now falls back to `MaxRetries`, changing the default from 3 to 10 (worth a note in the PR body); `Limits.MaxRetryCycles` is introduced and marked `Deprecated:` in the same PR — just drop it; and `maxRetriesAfterBackoff` plus `maxRetryCycles` both bound the same loop, with the inner guard unreachable at equal defaults.

go.mod
```diff
@@ -32,7 +32,7 @@ require (
 	golang.org/x/exp v0.0.0-20260709172345-9ea1abe57597
 	golang.org/x/sync v0.22.0
 	google.golang.org/protobuf v1.36.11
-	modernc.org/sqlite v1.53.0
+	modernc.org/sqlite v1.51.0
```
AkramBitar: This downgrades `modernc.org/sqlite` from v1.53.0 to v1.51.0 — unrelated to the change and probably a stale-branch artifact. Worth dropping from the diff.

AkramBitar:
@adecaro 

I would like to ask your help to review this PR. 

Regards,
Akram

token/services/storage/db/sql/common/tokens_limit_test.go
```diff
@@ -0,0 +1,132 @@
+/*
```
adecaro, line 1: **Nit: attribution mismatch with the commit message.**

The final commit message credits this file with "a real-database exercise of the limited query that runs on both SQLite and PostgreSQL," but this test is `sqlmock`-based, not a live-DB test. The actual dual-backend coverage that would have caught the Postgres `?`-placeholder syntax error lives in `dbtest/tokens.go`'s `TUnspentTokensIteratorByLimit`, run through the shared `dbtest` harness. Worth correcting the attribution so a future reader knows where the real guarantee lives.

token/services/selector/sherdlock/bounded_locker_test.go
```diff
@@ -0,0 +1,226 @@
+/*
```
adecaro, line 1: **Nit: coverage gap — the real SQL-backed lock store is never exercised here.**

Every test in this file drives `NewBoundedLocker` against `mocks.FakeLocker`, so `maxLocksPerTx` enforcement is only verified against the wrapper's own logic. The production wiring in `service.go:157` (`NewBoundedLocker(tokenLockStoreService, s.maxLocksPerTx)`) looks correct today, but nothing here would catch a future regression in how the real lock store is wired to the bound. An integration-style test against the actual SQL lock store would close that gap — this was the exact class of bug (limit configured but not reaching the default driver) the earlier review round caught.

adecaro (commented):
## Review: does this fully address #1641?

@HayimShaul

**Yes — every Blocking finding from the earlier review round is fixed at `694ad64f7`, on both the sherdlock (default) and simple drivers; what's left is test-coverage depth, not correctness.**

Checked out `1641_bound_work_by_request` at `694ad64f7` and verified:

```
CI checks / utest (race + regression) / itest (~50 legs incl. fabricx-dlog-t3) / bench-check / cgo-check / CodeQL — all SUCCESS
DCO — both commits signed off
make lint (full repo) — 17 pre-existing revive findings, none in this PR's changed files
Read in full: sherdlock/{selector,manager,bounded_locker,fetcher,service}.go, simple/{selector,manager}.go,
  simple/inmemory/locker.go, selector/config/driver.go, sdk/network/selector.go, token/selector.go
git merge-tree HEAD origin/main — 1 real conflict (dlogx/dlog_test.go), 2 clean auto-merges
```

I re-checked every item @AkramBitar's prior review flagged as Blocking against the current source rather than trusting the commit message's claims at face value: the sherdlock SQL `LIMIT` placeholder bug, `StubbornSelector` dropping locks before backoff, pre-dedup row limiting, the O(n²) in-memory locker scan, the `cache`/`Close()` race, `fmt.Errorf` usage, `Validate()`'s unreachable checks, the dead `MaxRetryCycles` field, the `sqlite` downgrade, sherdlock's missing lock-count bound, and the timeout/sufficient-funds mislabeling. All ten check out fixed by reading the code, not just the diff — details and file:line for each are in the local write-up. I'm not repeating any of those threads below.

CI is fully green, including the `fabricx-dlog-t3` leg the final two commits specifically target.

### What it fixes

| # | Finding | Verified |
|---|---|---|
| 1641.1 | Bound token-iteration depth per `Select` | ✅ `sherdlock/selector.go` (`tokensIterated > s.maxTokensPerSelection`) and `simple/selector.go` (`tokensIteratedCount`), both abort with a distinct error before further work |
| 1641.2 | Bound lock attempts per `Select` | ✅ `maxLockAttempts` counters in both selectors |
| 1641.3 | Bound retry/back-off cycles | ✅ `MaxRetriesAfterBackOff` (sherdlock) / `Limits.MaxRetries` (simple), plus `SelectionTimeout` as a wall-clock backstop |
| 1641.4 | Bound lock-store growth, on the default driver too | ✅ `sherdlock/bounded_locker.go`, wired to the real SQL lock store at `service.go:157` — this was the specific gap Akram's review caught at an earlier commit, now closed |
| 1641.5 | Abort early and release acquired locks on limit breach | ✅ confirmed via the resolved `UnlockAll`-on-cancellation thread and reading the defer path in `selectWithoutMetrics` |

All five rows check out against the current source.

### Non-blocking

- **Attribution nit:** the final commit message credits "a real-database exercise ... on both SQLite and PostgreSQL" to the new limit tests, but `tokens_limit_test.go` is `sqlmock`-based — the actual dual-backend coverage lives in `dbtest/tokens.go`'s `TUnspentTokensIteratorByLimit`. Worth fixing the attribution so the next reader knows where the guarantee actually lives.
- **Coverage gap:** `sherdlock/bounded_locker_test.go` only exercises the wrapper against `mocks.FakeLocker`; nothing drives `maxLocksPerTx` through the real SQL-backed lock store end-to-end. The wiring in `service.go` is correct today, but an integration-style test here would catch a future wiring regression that the wrapper unit tests can't see.
- **Rebase needed:** `git merge-tree` against current `main` (5 commits ahead) shows one real conflict in `integration/token/fungible/dlogx/dlog_test.go`; `wallets.go` and `auditor_test.go` also moved on `main` but auto-merge cleanly.

### Recommendation

No correctness blockers. Rebase past the `dlogx/dlog_test.go` conflict, and consider the two test-depth nits above if there's time before merge — neither should hold this up.

Effi-S (approved):

AkramBitar (approved):
@HayimShaul 

See my comments below.

Thanks.
Akram

AkramBitar:
Thanks a lot for the effort you've put into this — bounding the work a single request can trigger is the right thing to be doing, and there's a lot of careful plumbing and test coverage here.

Below is what I think needs addressing before merge. I've split it into things I'd consider blocking and smaller ones that are quick fixes.

## Blocking

**1. `Validate()` compares the raw config fields instead of the resolved limits it just computed.**
`token/services/selector/config/driver.go:269`. Line 267 does `limits := c.GetLimits()` with a comment saying the relational checks should use the numbers the runtime will actually apply, but both conditions test `c.Limits.*` and only the error *messages* use `limits.*`. So a partially-specified config skips the check entirely:

```yaml
token.selector.limits.maxLockAttempts: 100
# maxTokensPerSelection left unset
```

`c.Limits.MaxTokensPerSelection == 0`, the guard is skipped, `Validate` returns nil — and the effective values are `maxLockAttempts=100` vs `maxTokensPerSelection=10000`, which is exactly the combination `simple/service.go:57` describes as aborting every selection partway through. Any wallet with more than 100 tokens then fails with "exceeded max lock attempts". Same hole for `maxLocksPerTransaction: 20000` with `maxTokensPerSelection` unset. `driver_limits_test.go` only exercises fully-specified configs, so this isn't covered.

**2. `ForgetTx` on `Close` makes the lock ceiling per-action rather than per-transaction.**
`token/services/selector/sherdlock/manager.go:102`. `token/request.go:1621-1625` acquires a selector for `r.Anchor` and defers `sm.Close(anchor)` inside `prepareTransfer`, which is invoked once per `Request.Transfer` / `Request.Redeem`. For a multi-action transfer (the `TransferView` pattern in `integration/token/fungible/views/transfer.go`), action 1's locks remain in the store while its counter is dropped, so action 2 starts from a zero count — a transaction with N actions can hold up to `N * maxLocksPerTransaction`. That's the same "fresh budget" failure mode the `EvictStaleTxState` comment at lines 105-110 is written to prevent. This one probably needs a design decision rather than a patch: either the counter has to outlive the per-action selector, or the ceiling needs to be documented as per-action.

**3. The new DB-side row limit never takes effect on the default driver.**
`token/services/selector/sherdlock/selector.go:302` passes `limit = 0`, and all three sherdlock fetchers explicitly ignore the parameter anyway (`fetcher.go:106`, `:140`, `:338`); `lazyFetcher` delegates to `SpendableTokensIteratorBy`, which takes no limit at all. Since `sherdlock` is `defaultDriver` (`config/driver.go:19`), the query on the default path is still unbounded. The iteration cap at `selector.go:316` does bound how far the selector walks, so this isn't "no protection" — but the `LIMIT` that the PR adds only actually applies on the non-default `simple` driver. Is a follow-up intended for fetcher support, or should sherdlock bound the query itself here?

**4. Concurrent `Select` on one `Selector` looks racy after the `next()` helper was removed.**
`token/services/selector/sherdlock/selector.go:263`. The old `next()` held `s.mu` across `s.cache.Next()`, which serialized concurrent readers. Now both callers take the *read* side of `selectMu`, snapshot the same iterator into a local, and call `cache.Next()` concurrently — a data race on the underlying `sql.Rows`. On top of that, `swapCache` (line 379) closes `s.cache` while a peer goroutine may still hold that pointer in its local `cache` variable, which would be a use-after-close.

I derived reachability from `Manager`'s `lazy2.Provider` handing back the same `Selector` instance for a given `transaction.ID` rather than from an actual concurrent caller, so please sanity-check whether two goroutines can really select under one anchor in practice. `selector_concurrency_test.go` covers Close-vs-Select but not Select-vs-Select.

## Quick fixes

**5. The reclaim path leaks the previous holder's `txLocks` count.**
`token/services/selector/simple/inmemory/locker.go:279`. The stated invariant is that the counter is decremented on every delete, but the reclaim branch does a raw `delete(s.locked, k)` and then falls through to `s.locked[k] = &lockEntry{TxID: txID}` / `s.txLocks[txID]++` at 306-307, so `s.txLocks[observedTxID]` is never decremented. Tx `A` locks N tokens, `A` becomes `Deleted`, `B` reclaims them one by one, and `txLocks[A]` stays at N indefinitely. On a hot wallet whose shard never empties (so `pruneEmptyShard` never runs) that's a dead entry per reclaimed token. Calling `s.deleteLocked(k)` instead should be all that's needed.

**6. A freshly created counter is born stale, so a live reservation can be evicted.**
`token/services/selector/sherdlock/bounded_locker.go:133`. `counter()` returns `new(txLockCounter)` with `lastLock == 0`, and `Lock` only calls `c.touch(time.Now())` at line 74 — *after* the CAS reservation loop. `staleAt(cutoff)` is `lastLock.Load() < cutoff.UnixNano()`, which is true for zero. If `Manager.runCleanupTick` runs `EvictStaleTxState` in the window between `LoadOrStore` and `touch`, `CompareAndDelete` removes the counter the reservation was just made on, and the next `Lock` for that tx creates a fresh one — the transaction gets a full `maxLocksPerTx` budget on top of the locks it already holds. The comment at 118-120 says this race is prevented, but staleness is judged only on `lastLock`. Setting `lastLock` at construction (or before the CAS loop) closes it. Narrow window, but a one-line fix.

## Non-blocking, worth a look

- `sherdlock/selector.go:316` — zero-valued `MaxTokensPerSelection` / `MaxLockAttempts` abort every selection on the first token, so `NewManager(&Config{Fetcher: f, Locker: l, Precision: p})` is unusable. `withSelectionTimeout` (362-373) has a comment explaining why an omitted zero value has to stay harmless; suggest giving the two new int limits the same treatment and reading `<= 0` as unlimited.
- `config/driver.go:116` — dropping `defaultNumRetries = 3` makes `GetNumRetries()` fall back to `GetLimits().MaxRetries` (10), which derives `SelectionTimeout = 10*5s + 30s = 80s`. Every deployment that never set `numRetries` moves from ~15s to ~80s worst-case selection latency. If that's intended, it's worth a line in the docs diff.
- `sherdlock/service.go:50` and `simple/service.go:62` — on `Validate()` failure the code does `cfg = &config.Config{}`, which discards `retryInterval`, `leaseExpiry`, the fetcher-cache settings and every `rateLimit*` field (since `loader.limiter` is built from `cfg` on the next line). One mistyped limit silently turns off per-wallet rate limiting. Resetting just `cfg.Limits` would match the stated intent.

AkramBitar (dismissed):
See my comments

Effi-S:
> Thanks a lot for the effort you've put into this — bounding the work a single request can trigger is the right thing to be doing, and there's a lot of careful plumbing and test coverage here.
> 
> Below is what I think needs addressing before merge. I've split it into things I'd consider blocking and smaller ones that are quick fixes.
> 
> ## Blocking
> **1. `Validate()` compares the raw config fields instead of the resolved limits it just computed.** `token/services/selector/config/driver.go:269`. Line 267 does `limits := c.GetLimits()` with a comment saying the relational checks should use the numbers the runtime will actually apply, but both conditions test `c.Limits.*` and only the error _messages_ use `limits.*`. So a partially-specified config skips the check entirely:
> 
> ```yaml
> token.selector.limits.maxLockAttempts: 100
> # maxTokensPerSelection left unset
> ```
> 
> `c.Limits.MaxTokensPerSelection == 0`, the guard is skipped, `Validate` returns nil — and the effective values are `maxLockAttempts=100` vs `maxTokensPerSelection=10000`, which is exactly the combination `simple/service.go:57` describes as aborting every selection partway through. Any wallet with more than 100 tokens then fails with "exceeded max lock attempts". Same hole for `maxLocksPerTransaction: 20000` with `maxTokensPerSelection` unset. `driver_limits_test.go` only exercises fully-specified configs, so this isn't covered.
> 
> **2. `ForgetTx` on `Close` makes the lock ceiling per-action rather than per-transaction.** `token/services/selector/sherdlock/manager.go:102`. `token/request.go:1621-1625` acquires a selector for `r.Anchor` and defers `sm.Close(anchor)` inside `prepareTransfer`, which is invoked once per `Request.Transfer` / `Request.Redeem`. For a multi-action transfer (the `TransferView` pattern in `integration/token/fungible/views/transfer.go`), action 1's locks remain in the store while its counter is dropped, so action 2 starts from a zero count — a transaction with N actions can hold up to `N * maxLocksPerTransaction`. That's the same "fresh budget" failure mode the `EvictStaleTxState` comment at lines 105-110 is written to prevent. This one probably needs a design decision rather than a patch: either the counter has to outlive the per-action selector, or the ceiling needs to be documented as per-action.
> 
> **3. The new DB-side row limit never takes effect on the default driver.** `token/services/selector/sherdlock/selector.go:302` passes `limit = 0`, and all three sherdlock fetchers explicitly ignore the parameter anyway (`fetcher.go:106`, `:140`, `:338`); `lazyFetcher` delegates to `SpendableTokensIteratorBy`, which takes no limit at all. Since `sherdlock` is `defaultDriver` (`config/driver.go:19`), the query on the default path is still unbounded. The iteration cap at `selector.go:316` does bound how far the selector walks, so this isn't "no protection" — but the `LIMIT` that the PR adds only actually applies on the non-default `simple` driver. Is a follow-up intended for fetcher support, or should sherdlock bound the query itself here?
> 
> **4. Concurrent `Select` on one `Selector` looks racy after the `next()` helper was removed.** `token/services/selector/sherdlock/selector.go:263`. The old `next()` held `s.mu` across `s.cache.Next()`, which serialized concurrent readers. Now both callers take the _read_ side of `selectMu`, snapshot the same iterator into a local, and call `cache.Next()` concurrently — a data race on the underlying `sql.Rows`. On top of that, `swapCache` (line 379) closes `s.cache` while a peer goroutine may still hold that pointer in its local `cache` variable, which would be a use-after-close.
> 
> I derived reachability from `Manager`'s `lazy2.Provider` handing back the same `Selector` instance for a given `transaction.ID` rather than from an actual concurrent caller, so please sanity-check whether two goroutines can really select under one anchor in practice. `selector_concurrency_test.go` covers Close-vs-Select but not Select-vs-Select.
> 
> ## Quick fixes
> **5. The reclaim path leaks the previous holder's `txLocks` count.** `token/services/selector/simple/inmemory/locker.go:279`. The stated invariant is that the counter is decremented on every delete, but the reclaim branch does a raw `delete(s.locked, k)` and then falls through to `s.locked[k] = &lockEntry{TxID: txID}` / `s.txLocks[txID]++` at 306-307, so `s.txLocks[observedTxID]` is never decremented. Tx `A` locks N tokens, `A` becomes `Deleted`, `B` reclaims them one by one, and `txLocks[A]` stays at N indefinitely. On a hot wallet whose shard never empties (so `pruneEmptyShard` never runs) that's a dead entry per reclaimed token. Calling `s.deleteLocked(k)` instead should be all that's needed.
> 
> **6. A freshly created counter is born stale, so a live reservation can be evicted.** `token/services/selector/sherdlock/bounded_locker.go:133`. `counter()` returns `new(txLockCounter)` with `lastLock == 0`, and `Lock` only calls `c.touch(time.Now())` at line 74 — _after_ the CAS reservation loop. `staleAt(cutoff)` is `lastLock.Load() < cutoff.UnixNano()`, which is true for zero. If `Manager.runCleanupTick` runs `EvictStaleTxState` in the window between `LoadOrStore` and `touch`, `CompareAndDelete` removes the counter the reservation was just made on, and the next `Lock` for that tx creates a fresh one — the transaction gets a full `maxLocksPerTx` budget on top of the locks it already holds. The comment at 118-120 says this race is prevented, but staleness is judged only on `lastLock`. Setting `lastLock` at construction (or before the CAS loop) closes it. Narrow window, but a one-line fix.
> 
> ## Non-blocking, worth a look
> * `sherdlock/selector.go:316` — zero-valued `MaxTokensPerSelection` / `MaxLockAttempts` abort every selection on the first token, so `NewManager(&Config{Fetcher: f, Locker: l, Precision: p})` is unusable. `withSelectionTimeout` (362-373) has a comment explaining why an omitted zero value has to stay harmless; suggest giving the two new int limits the same treatment and reading `<= 0` as unlimited.
> * `config/driver.go:116` — dropping `defaultNumRetries = 3` makes `GetNumRetries()` fall back to `GetLimits().MaxRetries` (10), which derives `SelectionTimeout = 10*5s + 30s = 80s`. Every deployment that never set `numRetries` moves from ~15s to ~80s worst-case selection latency. If that's intended, it's worth a line in the docs diff.
> * `sherdlock/service.go:50` and `simple/service.go:62` — on `Validate()` failure the code does `cfg = &config.Config{}`, which discards `retryInterval`, `leaseExpiry`, the fetcher-cache settings and every `rateLimit*` field (since `loader.limiter` is built from `cfg` on the next line). One mistyped limit silently turns off per-wallet rate limiting. Resetting just `cfg.Limits` would match the stated intent.

1. config.Validate() now checks resolved GetLimits() values instead of raw c.Limits.*, so partially-specified configs (e.g. maxLockAttempts=100 with maxTokensPerSelection unset) are rejected. Added two test cases for the holes the reviewer identified.
5. Reclaim path in simple/inmemory/locker.go uses deleteLocked() so the previous holder's txLocks count is decremented.
6. boundedLocker.counter() stamps lastLock at construction, closing the EvictStaleTxState race.

- NB1 sherdlock selector treats non-positive maxTokensPerSelection/maxLockAttempts as unlimited.
- NB3 Both selector services reset only cfg.Limits on Validate() failure, preserving rateLimit*/retryInterval/etc.
- NB2 Docs: corrected stale maxRetries default (3→10) + migration note on the latency change.

docs/security/selector_resource_limits.md
```diff
@@ -1,13 +1,335 @@
...
+**Error message**: `"token selection aborted: exceeded max token iteration limit (10000 tokens)"`
+
+**Database optimization**: This limit is enforced at **two levels**:
+1. **Database query level**: SQL queries include `LIMIT ?` clause to prevent fetching excess rows
+2. **Application level**: Iterator counter provides defense-in-depth safety check
+
+**Performance impact**: Database-level LIMIT significantly reduces I/O overhead:
+- Without LIMIT: Database may scan millions of rows, returning them one-by-one
+- With LIMIT: Database stops after returning configured number of rows
+- Result: ~10x faster query execution for large token sets
```
adecaro: **Should fix: this "Database optimization" claim is false for the default (sherdlock) driver.**

All three sherdlock fetchers explicitly document that they ignore the `limit` parameter (`fetcher.go:106-107`, `140-141`, `339-340`), and `cachedFetcher.update()` runs an unfiltered `SpendableTokensIteratorBy(ctx, "", "")` full-table scan on every cache refresh (`fetcher.go:260`). Only `simple.Selector` actually pushes `maxTokensPerSelection` into the query (`simple/selector.go:144`). As written this section describes the system as a whole, but it's only true for the non-default driver.

Suggest scoping this section explicitly to `simple`, or noting that sherdlock's `maxTokensPerSelection` is an in-process CPU/lock-attempt bound only, not a DB-query or memory bound.

docs/security/selector_resource_limits.md
```diff
@@ -1,13 +1,335 @@
...
+
+### Default Secure Configuration
+
+```yaml
+token:
+  selector:
+    limits:
+      maxTokensPerSelection: 10000      # 10k tokens ≈ 1MB memory
+      maxLockAttempts: 50000            # 5x iteration limit
+      maxRetryCycles: 10                # Reasonable for transient issues
```
adecaro: **Nit: `maxRetryCycles` isn't a real config key.**

The field is `maxRetries` (`Limits.MaxRetries`, yaml tag `maxRetries`). Since config unmarshalling ignores unknown keys, copying this example verbatim silently keeps the default retry count instead of setting the one shown here. Same typo recurs at lines 158, 186, 213, 226.

token/services/selector/sherdlock/selector.go
```diff
@@ -185,59 +295,85 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
...
-				return nil, nil, immediateRetries, errors.Wrapf(err, "failed to add quantity")
+
+			// Check lock attempt limit (non-positive means "unlimited", as above)
+			lockAttempts++
+			if s.maxLockAttempts > 0 && lockAttempts > s.maxLockAttempts {
+				return nil, nil, immediateRetries, tokensIterated, lockAttempts, errors.Errorf(
+					"token selection aborted: exceeded max lock attempts (%d) after examining %d tokens",
+					s.maxLockAttempts, tokensIterated,
+				)
 			}
```
adecaro, line 334: **Should fix: neither abort branch has a test.**

`grep -rn "exceeded max token iteration\|exceeded max lock attempts" token/services/selector/sherdlock/` matches only this file. `manager_unit_test.go:33` wires `MaxTokensPerSelection` into a `Config` but nothing drives a selection past it to confirm the abort fires. The `simple` driver has exactly this coverage (`simple/selector_limits_test.go`'s `TestSelector_TokenIterationLimit`/`TestSelector_LockAttemptLimit`), but sherdlock's cache-swap-on-retry path (lines 301-309 above) is different enough that those tests give no confidence this code path is wired correctly.

token/services/selector/config/driver.go
```diff
@@ -30,17 +29,32 @@ const (
...
+// Limits defines resource limits for token selection to prevent algorithmic attacks
+type Limits struct {
+	MaxTokensPerSelection  int           `yaml:"maxTokensPerSelection,omitempty"`
+	MaxLockAttempts        int           `yaml:"maxLockAttempts,omitempty"`
+	MaxRetries             int           `yaml:"maxRetries,omitempty"`
+	MaxLocksPerTransaction int           `yaml:"maxLocksPerTransaction,omitempty"`
+	SelectionTimeout       time.Duration `yaml:"selectionTimeout,omitempty"`
+}
+
 type Config struct {
```
adecaro, line 59: **Nit: missing Godoc.**

`Limits` right above has a doc comment; `Config` doesn't. AGENTS.md requires Godoc on exported symbols — human-review-only, easy to miss since it's not lint-enforced.

adecaro (commented):
## Review: does this fully address #1641?

@HayimShaul

**This is a much stronger patch than the version I (and others) reviewed earlier — approve, with two should-fix comments.**

Checked out this branch at `14d0930` and verified:

```
go build ./...                                              # clean
go vet ./token/...                                           # clean
go build ./integration/...                                   # clean
go test -count=1 ./token/services/selector/...                # pass
go test -count=1 ./token/services/storage/db/sql/common/...   # pass
go test -count=1 ./token/...                                  # pass (full tree)
grep -rn "fmt.Errorf" token/services/selector/                # no matches
```

I re-checked every issue raised in the prior review rounds against this commit rather than trusting
the PR description. Thirteen of them are genuinely fixed, not just superficially patched — the
O(n²) lock counting is now O(1) via `shard.txLocks`, the `Close()`/iterator race is closed with a
drain-lock (`selectMu`), the timeout is a distinct sentinel instead of being conflated with
"sufficient but locked funds," `Validate()` now checks resolved rather than raw config values so a
partially-specified limits block can't sneak through, and the bounded locker finally has a test
against the real SQL-backed store instead of only a fake.

CI is fully green, including `utest (unit-tests-race)` which actually ran for 28 minutes and
passed (not just compiled), and every `itest` leg across dlog/fabtoken/fabricx/dloghsm/interop/
update/dvp/nft.

### What it fixes

| # | Finding | Verified |
|---|---|---|
| 1 | Limit silently dropped on default (sherdlock) driver | Now enforced in-process via `tokensIterated`/`lockAttempts` counters in `selectInternal` — but see Blocking-adjacent note below, the DB query itself is still unbounded for this driver |
| 2 | O(n²) lock counting in `inmemory` locker | `shard.txLocks map[string]int`, O(1) |
| 3 | Race between `Close()` and in-flight selection | `selectMu sync.RWMutex` drain lock |
| 4 | Timeout conflated with locked-funds error | Distinct `token.SelectorTimedOut` sentinel |
| 5 | `fmt.Errorf` usage | Removed, confirmed via grep |
| 6 | Scattered constructor config | Consolidated into `sherdlock.Config` |
| 7 | Rate-limit denial mishandled by selector | Hard-stops via `errors.Is(lockErr, token.SelectorRateLimited)` |
| 8 | `Validate()` checked raw config | Now validates `GetLimits()` (resolved) values |
| 9 | Bounded locker only tested against a fake | New `bounded_locker_sql_test.go` against real SQL store |

All nine rows above check out against the current code, not just the diff.

### Should fix 1 — security doc overstates DB-level enforcement for the default driver

`docs/security/selector_resource_limits.md:50-57` says the token-iteration limit is enforced "at
two levels," including a `LIMIT ?` SQL clause, and claims ~10x faster queries as a result. That's
accurate for the `simple` driver (`simple/selector.go:144` does pass the limit into the query), but
it's false for `sherdlock` — the *default* driver. All three sherdlock fetchers document that they
ignore the `limit` parameter (`fetcher.go:106-107`, `140-141`, `339-340`), and `cachedFetcher.update()`
does an unfiltered `SpendableTokensIteratorBy(ctx, "", "")` full-table scan on every cache refresh
(`fetcher.go:260`). `maxTokensPerSelection` only bounds the in-process counting loop over whatever
that unbounded fetch already returned — it doesn't stop a wallet with millions of tokens from being
pulled into memory in the first place, which is exactly the attack this doc's own threat model
section describes.

### Should fix 2 — sherdlock's new abort logic has zero direct test coverage

`selectInternal`'s two new limit checks (`selector.go:320-334`) aren't exercised by any test —
`grep -rn "exceeded max token iteration\|exceeded max lock attempts" token/services/selector/sherdlock/`
matches only the production file. `manager_unit_test.go:33` wires `MaxTokensPerSelection` into a
`Config` but never drives a selection past it. The `simple` driver has exactly this coverage
(`simple/selector_limits_test.go`'s `TestSelector_TokenIterationLimit`/`TestSelector_LockAttemptLimit`),
but sherdlock's cache-swap-on-retry path (`selector.go:301-309`) is different enough that passing
simple's tests says nothing about sherdlock's behavior here.

### Non-blocking

- **Doc nit:** the same file's YAML examples and prose (lines 158, 186, 200, 213, 226) use a key
  `maxRetryCycles` that doesn't exist — the actual field is `maxRetries`. Config unmarshalling
  silently ignores unknown keys, so anyone copying these examples gets the default retry count
  regardless of what they set.
- **Doc nit:** `config.Config` (`driver.go:55`) is missing a Godoc comment; the `Limits` struct
  right above it has one.

### Recommendation

Good to merge once the security doc is corrected to scope the DB-level LIMIT claim to the `simple`
driver (or the claim is made accurate for sherdlock too) — that's the one thing here with a real
chance of misleading an operator's security posture. The test-coverage gap for sherdlock's abort
paths is worth a fast follow if not blocking. Everything else checks out.

token/services/selector/config/driver.go
Effi-S: Added

token/services/selector/sherdlock/selector.go
Effi-S: Added

docs/security/selector_resource_limits.md
Effi-S: replaced all six `maxRetryCycles` with `maxRetries`

docs/security/selector_resource_limits.md
Effi-S: Both false claims are now corrected

AkramBitar (commented):
## Review scope

Reviewed `git diff origin/main...` for this PR — single commit `394d432ac`, 77 files, +3944/-638. `go build ./...`, `go vet ./token/...` and the selector + sql unit tests all pass on the PR head, so everything below is a logic, reachability or configuration argument rather than a compile/test failure.

14 findings are posted as inline comments. Two of them are blocking:

**Blocking 1 — tightening the limit stops the node from starting.** `config/driver.go` validates the *defaults-applied* limits and enforces `MaxLocksPerTransaction <= MaxTokensPerSelection`, but `defaultMaxLocksPerTransaction = 5000`. Setting only `token.selector.limits.maxTokensPerSelection: 1000` — exactly what `docs/security/selector_resource_limits.md` tells operators to do to tighten the bound — resolves locks to 5000, fails validation, `NewService` returns an error, and the dig container fails to build.

**Blocking 2 — the `simple` selector aborts on the first token when limits are unset.** `simple/selector.go` is missing the `> 0` guards that the sherdlock equivalents have. `simple.NewManager` is exported and takes the limits as bare ints, so any caller that does not pass them fails every selection with 'exceeded max token iteration limit (0 tokens)'.

## Please split out the unrelated changes

`Aries: true` is added to the dlog/dvp/nft/interop integration suites, which switches the Idemix backend those CI jobs exercise, and `integration/token/fungible/topology/topology.go` restructures bootstrap-node/TMS membership. Neither relates to selector limits, and a CI regression in those jobs will be misattributed to this PR.

## Correct as-is, no action needed

- `integration/token/fungible/views/transfer.go` — the additional-recipients loop was using the outer `recipient`/`t.RecipientEID` instead of `actionRecipient`/`action.RecipientEID`. Real bug, correct fix.
- `token/request.go` — moving `defer sm.Close(...)` to after the error check is correct.

## Checked and dismissed, no need to re-litigate

- Scanning `NUMERIC(78,0)` into `*string` is safe via `database/sql`'s `asString`.
- `boundedLocker` does not double-count a same-tx re-lock: the `token_locks` PK `(tx_id, idx)` gives `UniqueKeyViolation` -> `ErrTokenAlreadyLocked` -> the reservation is rolled back.
- `ORDER BY amount DESC` is unambiguous across the `UNION`; only one output column is named `amount`.
- No migration gap: `amount NUMERIC(78,0) NOT NULL` and `validateTokenAmount` both pre-exist on `origin/main`.
- `UnlockAll` before backoff matches base behaviour.
- `reclaim := s.maxRetries == 1 || actualRetries > 1` is equivalent to the old `i > 0`.
- `selectMu`->`mu` lock ordering is consistent; no deadlock.
- The variadic `expectedErrorMsgs` in `TransferCash` are all-of, so splitting the message did not weaken the assertion.
- `Convert` has no other in-tree callers, so the signature change breaks nothing here.
- `tokens_limit_test.go` and the new `TUnspentTokensIteratorByLimit` case are solid on placeholder dialect, `UNION` vs `UNION ALL`, stmt-cache bypass and DESC ordering.

token/services/selector/config/driver.go
```diff
@@ -169,3 +199,89 @@ func (c *Config) GetRateLimitBurst() int {
...
+	// e.g. maxLockAttempts=100 with maxTokensPerSelection unset resolves to
+	// maxTokensPerSelection=10000, an impossible combination that would abort
+	// every selection partway through, so it must be rejected here.
+	limits := c.GetLimits()
+
+	if limits.MaxLockAttempts < limits.MaxTokensPerSelection {
+		return errors.Errorf("maxLockAttempts (%d) should be >= maxTokensPerSelection (%d)",
+			limits.MaxLockAttempts, limits.MaxTokensPerSelection)
+	}
+	if limits.MaxLocksPerTransaction > limits.MaxTokensPerSelection {
```
AkramBitar, line 288: **HIGH — blocking: the documented way to tighten the limit makes the node refuse to start.**

`Validate()` runs on the *defaults-applied* limits and enforces `MaxLocksPerTransaction <= MaxTokensPerSelection`, but `defaultMaxLocksPerTransaction = 5000`.

Failure path: an operator follows `docs/security/selector_resource_limits.md` ('decrease for tighter security') and sets only `token.selector.limits.maxTokensPerSelection: 1000`. `MaxLocksPerTransaction` is then resolved to its default 5000, `5000 <= 1000` fails, `NewService` returns an error (`simple/service.go:63`, `sherdlock/service.go:52`), and the dig container fails to build — the node does not start.

Suggested direction: either clamp the locks default to the resolved `MaxTokensPerSelection` when it was not explicitly set, or validate only explicitly-set values and apply defaults afterwards.

token/services/selector/simple/selector.go
```diff
@@ -75,42 +117,69 @@ func (s *selector) selectByID(ctx context.Context, ownerFilter token.OwnerFilter
...
 				return nil, nil, errors.Wrap(err, "token selection failed")
 			}
 			if t == nil {
 				break
 			}
 
+			// Check token iteration limit (only count actual tokens, not nil)
+			s.tokensIteratedCount++
+			s.tokensIteratedThisCycle++
+			if s.tokensIteratedThisCycle > s.maxTokensPerSelection {
```
AkramBitar: **HIGH — blocking: missing `> 0` guard makes an unset limit abort on the very first token.**

This check is `if s.tokensIteratedThisCycle > s.maxTokensPerSelection`, and line 193 does the same for `maxLockAttempts`. The sherdlock equivalents at `sherdlock/selector.go:320,329` correctly guard with `s.maxTokensPerSelection > 0 && ...`.

Failure path: `simple.NewManager` (`simple/manager.go:48`) is exported and takes these limits as bare ints. Any caller constructing it without the new limits gets `1 > 0` on the first token, so every selection aborts with 'exceeded max token iteration limit (0 tokens)'. `maxRetries == 0` at line 130 has the same problem — it aborts and calls `UnlockByTxID` before the first cycle runs.

Suggested direction: mirror sherdlock's `limit > 0 &&` guards here (and for `maxRetries`), so that zero means unbounded.

token/services/storage/db/sql/common/tokens.go
```diff
@@ -267,24 +294,30 @@ func buildUnspentTokensIteratorByQuery(db *TokenStore, walletID string, tokenTyp
...
+	if limit > 0 {
+		sb.WriteString(" UNION ")
+	} else {
+		sb.WriteString(" UNION ALL ")
+	}
 	branch2.FormatTo(db.ci, sb)
+	if limit > 0 {
+		// Order by amount descending: largest tokens first so the selector
+		// reaches the target amount with the fewest rows.
+		sb.WriteString(" ORDER BY amount DESC LIMIT ")
```
AkramBitar, line 318: **MEDIUM — `UNION` + `LIMIT` bounds rows, not distinct tokens, so a funded wallet can report insufficient funds.**

Branch 1 (lines 279-282) constrains `tokens.owner_wallet_id` but not `ownership.wallet_id`, and `dedupedTokenRowsIterator` keys on `(tx_id, idx, wallet_id)` (line 424), so a token with K ownership rows still produces K rows through the `UNION`.

Failure path: with shared ownership — `StoreToken(..., []string{alice, bob})` — a token owned by alice yields 2 rows with an identical `amount`. They sort adjacently under `ORDER BY amount DESC`, so `LIMIT 2` surfaces only **one** distinct token. The simple selector sees a short page and reports `SelectorInsufficientFunds` (line 291) even though the wallet holds plenty.

The new `TUnspentTokensIteratorByLimit` fixture in `dbtest/tokens.go` only ever stores `[]string{wallet}`, which is why the suite cannot catch this.

Suggested direction: dedupe before applying `LIMIT` (constrain `ownership.wallet_id` in branch 1, or wrap the `UNION` in a subquery that collapses on `(tx_id, idx)` and apply `LIMIT` outside), and extend the fixture with a multi-owner token.

token/services/selector/sherdlock/selector.go
```diff
@@ -143,32 +231,54 @@ func (s *Selector) Select(ctx context.Context, owner token.OwnerFilter, q string
...
+	// not blocked for the whole retry budget.
+	s.selectMu.RLock()
+	defer s.selectMu.RUnlock()
+
+	// Take mu to snapshot s.cache into a local variable, then release the lock
+	// and work exclusively through the local copy. This eliminates the data race
+	// between selectInternal reads of s.cache and a concurrent Close() write.
+	// On the cache-reload path we take mu again to publish the new iterator so
+	// that a concurrent Close() can still call Close() on it.
+	s.mu.Lock()
```
AkramBitar, line 262: **MEDIUM — the iterator lock was dropped; two concurrent selections can now drive one `sql.Rows`.**

The deleted `Selector.next()` (base `sherdlock/selector.go:232`) held `s.mu` around **every** `s.cache.Next()`. The replacement snapshots `cache` under `mu`, releases it, then calls `cache.Next()` unguarded for the whole iteration. `selectMu` is only `RLock`ed here, so it serialises against `Close()` but not against another `selectInternal`.

Failure path: `Manager.NewSelector` (`manager.go:89`) returns the *same* `*Selector` from `selectorCache` for a given txID. Two concurrent `Select` calls on one transaction therefore iterate the same `sql.Rows`, and one of them can `swapCache` (line 306) and close the iterator the other still holds a local pointer to — 'sql: Rows are closed', or torn reads.

Suggested direction: keep the lock held across `Next()` as base did, or take `selectMu.Lock()` (not `RLock`) so selections on one selector serialise.

token/services/selector/simple/selector.go
```diff
@@ -43,9 +43,25 @@ type selector struct {
...
 	timeout              time.Duration
 	requestCertification bool
+
+	// Resource limits to prevent algorithmic attacks
+	maxTokensPerSelection int
+	maxLockAttempts       int
+	selectionTimeout      time.Duration
+
+	// Resource tracking counters (reset per selection)
+	tokensIteratedCount int
```
AkramBitar, line 56: **MEDIUM — new data race: per-selection counters moved from locals onto the shared `*selector`.**

`tokensIteratedCount`, `lockAttemptsCount` and `tokensIteratedThisCycle` were locals in base; they are now `*selector` fields reset at `Select` entry (lines 74-76).

Failure path: a `*selector` handed in via `transferOpts.Selector` and used from two goroutines now has unsynchronised int read/write on every token. The race detector will flag it, and the pre-PR code was race-free by construction.

Suggested direction: keep them as locals threaded through the loop, or carry them in a per-call state struct.

token/services/selector/simple/selector.go
```diff
@@ -119,6 +188,18 @@ func (s *selector) selectByID(ctx context.Context, ownerFilter token.OwnerFilter
 				return nil, nil, errors.Wrap(err, "failed to convert quantity")
 			}
 
+			// Check lock attempt limit
+			s.lockAttemptsCount++
```
AkramBitar, line 192: **MEDIUM — `lockAttemptsCount` is cumulative across retry cycles while the iteration counter is per-cycle.**

`tokensIteratedThisCycle` is deliberately reset each cycle (lines 150-152); this counter is not.

Failure path: with the shipped defaults (`maxRetries=10`, `maxTokensPerSelection=10000`, `maxLockAttempts=50000`) a contended selection may attempt up to 100 000 locks across 10 cycles, so it trips the lock-attempt ceiling around cycle 6 and returns the unsentinelled error from line 177 instead of ever reaching the typed `SelectorSufficientButLockedFunds` at line 319. `Validate()`'s `MaxLockAttempts >= MaxTokensPerSelection` invariant does not account for the retry multiplier.

Suggested direction: either reset this per cycle to match, or make the invariant `MaxLockAttempts >= MaxTokensPerSelection * MaxRetries`.

token/services/selector/simple/selector.go
```diff
@@ -183,17 +264,51 @@ func (s *selector) selectByID(ctx context.Context, ownerFilter token.OwnerFilter
 		if target.Cmp(potentialSumWithLocked) <= 0 && potentialSumWithLocked.Cmp(sum) != 0 {
 			// funds are potentially enough but they are locked
 			logger.DebugfContext(ctx, "token selection: sufficient funds but partially locked")
+		} else if target.Cmp(potentialSumWithLocked) > 0 {
+			// Insufficient funds with no locked tokens. Whether that is the
+			// final answer depends on how much of the wallet this cycle got to
+			// see, which is decided per cycle: the query is capped at
+			// maxTokensPerSelection rows, and the cumulative counter grows past
+			// that cap across cycles even when every cycle saw the full set.
+			if s.tokensIteratedThisCycle >= s.maxTokensPerSelection {
```
AkramBitar: **MEDIUM — off-by-one at the boundary: `>=` here vs `>` at line 173.**

Failure path: a wallet holding **exactly** `maxTokensPerSelection` tokens with genuinely insufficient funds returns a full page that *is* the entire wallet. This early exit then reports 'exceeded max token iteration limit' instead of `token.SelectorInsufficientFunds`, and the error carries no sentinel (see the comment on line 177), so the caller cannot classify it.

Suggested direction: make the two comparisons consistent, and treat 'page was exactly the limit but is also the whole wallet' as insufficient funds.

token/services/selector/simple/selector.go
```diff
@@ -75,42 +117,69 @@ func (s *selector) selectByID(ctx context.Context, ownerFilter token.OwnerFilter
...
 			}
 
+			// Check token iteration limit (only count actual tokens, not nil)
+			s.tokensIteratedCount++
+			s.tokensIteratedThisCycle++
+			if s.tokensIteratedThisCycle > s.maxTokensPerSelection {
+				s.locker.UnlockIDs(ctx, id, toBeSpent...)
+				s.locker.UnlockIDs(ctx, id, toBeCertified...)
+
+				return nil, nil, errors.Errorf(
```
AkramBitar, line 177: **MEDIUM — limit-exceeded errors carry no `token.Selector*` sentinel.**

This site and line 280, plus `sherdlock/selector.go:321,330`, all return a plain `errors.Errorf`.

Failure path: `integration/token/fungible/views/transfer.go:317` falls through to `default:` and reports 'system failure', and the retry filter at lines 290-296 cannot recognise these either. An operator-configured limit being hit is indistinguishable from a genuine bug, both to callers and in logs.

Suggested direction: add a sentinel (e.g. `token.SelectorResourceLimitExceeded`) alongside the existing `SelectorInsufficientFunds` / `SelectorSufficientButLockedFunds`, wrap all four sites with it, and handle it explicitly in the view.

token/services/selector/sherdlock/fetcher.go
```diff
@@ -103,9 +103,10 @@ func NewMixedFetcher(tokenDB TokenDB, m *Metrics, cacheSize int64, freshnessInte
...
+func (f *mixedFetcher) UnspentTokensIteratorBy(ctx context.Context, walletID string, currency token2.Type, limit int) (Iterator[*token2.UnspentTokenInWallet], error) {
```
AkramBitar: **MEDIUM — the headline DB-level bound is inert on the default driver.**

All three fetchers accept `limit int` and ignore it (this one, plus lines 182 and 381), and `sherdlock/selector.go:302` passes `0`.

Consequence: sherdlock is `defaultDriver` (`config/driver.go:19`), so on a default deployment nothing pushes a limit into the query — only the `simple` driver (`simple/selector.go:144`) actually does. The 4-arg interface change and all the regenerated mocks buy nothing for the default path, which is the one the PR title describes.

Suggested direction: either thread the limit through to the sherdlock queries, or drop the parameter from the interface for now and say so in the PR description so the bound is not assumed to be in force.

token/services/selector/config/driver.go
```diff
@@ -30,17 +29,36 @@ const (
 	// defaultRateLimitBurstFactor multiplies the rate to obtain the burst capacity applied
 	// when rate limiting is enabled without an explicit rateLimitBurst.
 	defaultRateLimitBurstFactor = ratelimit.DefaultBurstFactor
+
+	// Security limits to prevent algorithmic attacks
+	defaultMaxTokensPerSelection  = 10000            // Max tokens to iterate per selection
+	defaultMaxLockAttempts        = 50000            // Max lock attempts per selection (5x iteration limit)
+	defaultMaxRetries             = 10               // Max outer retry loops
```
AkramBitar, line 36: **MEDIUM — silent worst-case latency change for deployments that configure nothing.**

`defaultMaxRetries` goes 3 -> 10, and an unset `selectionTimeout` is now extended to `maxRetries * retryInterval + 30s` (lines 225-238), which is 80s with the defaults.

Consequence: every existing deployment that sets neither `numRetries` nor `limits.*` moves from a ~15s worst case to ~80s of retrying before a transfer fails. That is a real latency and backpressure change for callers that treat selection failure as fast-fail. It is described in `docs/security/selector_resource_limits.md` but is not gated behind any opt-in.

Suggested direction: keep `defaultMaxRetries` at 3, or call the new worst case out explicitly in the PR description and release notes.

token/services/selector/sherdlock/bounded_locker.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return b.Locker.UnlockByTxID(ctx, consumerTxID)
+}
+
+// ForgetTx drops the lock counter for consumerTxID without touching the store.
+// Manager calls it when a selector is closed: after a successful selection the
+// locks must survive (the transaction still needs them) but the counter is dead
+// weight, and nothing else would ever remove it on a replica that does not win
+// cleanup leadership.
+func (b *boundedLocker) ForgetTx(consumerTxID transaction.ID) {
```
AkramBitar, line 96: **LOW — `maxLocksPerTransaction` is enforced per action, not per transaction.**

`prepareTransfer` (`token/request.go:1625`) defers `sm.Close(anchor)` for every transfer action it creates a selector for, and `Manager.Close` (`sherdlock/manager.go:101-103`) calls `ForgetTx`, which drops this counter while the locks themselves survive by design.

Consequence: a request with N transfer actions gets a fresh `maxLocksPerTx` budget N times, so the effective ceiling is N x the configured value. The field name, the config key and the docs all say per transaction.

Suggested direction: either key the counter so it survives per-action `Close`, or rename the field, config key and docs to say per action.

token/services/selector/simple/inmemory/locker.go
```diff
@@ -208,6 +238,25 @@ func (d *locker) lockInShard(ctx context.Context, s *shard, owner string, id *to
 	if s.pruned {
 		return "", errShardPruned
 	}
+
+	// Check lock count limit for this transaction (if configured). A single
+	// selection locks tokens for one owner, so all of a transaction's locks
+	// live in this owner's shard and counting within it is per-transaction.
+	if d.maxLocksPerTx > 0 {
+		if txLockCount := s.txLocks[txID]; txLockCount >= d.maxLocksPerTx {
```
AkramBitar, line 246: **LOW — same ceiling escaped a different way: the counter is per owner shard.**

`txLocks` lives in the per-**owner** shard (lines 59 and 67), so a transaction spending from two wallets gets `maxLocksPerTx` in each shard.

The comment at line 244 ('all tokens a single selection locks live in this owner's shard') is true per `Select` call but not per transaction, which is what the limit claims to bound. Same underlying issue as the sherdlock `boundedLocker` comment — worth fixing both consistently, or renaming the limit to match what it actually enforces.

token/services/selector/sherdlock/manager.go
```diff
@@ -36,29 +64,20 @@ type Manager struct {
...
+		metrics:                cfg.Metrics,
 		cancel:                 cancel,
 		cleanerDone:            make(chan struct{}),
 		selectorCache: lazy2.NewProvider(func(txID transaction.ID) (TokenSelectorUnlocker, error) {
-			return NewSherdSelector(txID, fetcher, locker, precision, backoff, maxRetriesAfterBackOff, m), nil
+			return NewSherdSelector(txID, cfg.Fetcher, cfg.Locker, cfg.Precision, cfg.Backoff, cfg.MaxRetriesAfterBackOff, cfg.MaxTokensPerSelection, cfg.MaxLockAttempts, cfg.SelectionTimeout, cfg.Metrics), nil
 		}),
 	}
-	if leaseCleanupTickPeriod > 0 && leaseExpiry > 0 {
+	if cfg.LeaseCleanupTickPeriod > 0 && cfg.LeaseExpiry > 0 {
```
AkramBitar: **LOW — the new `counts sync.Map` has no reaper when lease cleanup is disabled.**

`runCleanupTick` is the only driver of `boundedLocker.EvictStaleTxState`, and this `cleaner` goroutine only starts when `LeaseCleanupTickPeriod > 0 && LeaseExpiry > 0`.

Failure path: an operator who disables lease cleanup leaves `counts` with nothing to evict it. Any txID whose selector is neither `Close`d nor `UnlockByTxID`ed — for example a caller supplying its own selector via `transferOpts.Selector` — leaks a `txLockCounter` for the process lifetime.

Suggested direction: start a reaper for `counts` independently of the lease-cleanup config, or bound the map size.

integration/nwo/token/topology/options.go
```diff
@@ -148,18 +148,51 @@ func ToOptions(o *node.Options) *Options {
...
+		} else {
+			panic(fmt.Sprintf("invalid nested mapping type: %T", mappingVal))
+		}
+	} else {
+		// Use the entire map if no "mapping" key exists
+		source = m
+	}
+
+	for k, v := range source {
+		opts.Mapping[k.(string)] = v
```
AkramBitar: **LOW — unchecked type assertion, newly reachable after widening `Convert` to `map[any]any`.**

`opts.Mapping[k.(string)] = v` panics on a non-string key.

Failure path: a YAML topology with a non-string key under `token.mapping` (e.g. `1:` or `true:`) now panics with an opaque 'interface conversion: interface {} is int, not string' instead of the informative `panic(fmt.Sprintf("invalid nested mapping type: %T", ...))` used eight lines earlier. The base signature (`map[string]any`) made this impossible.

Suggested direction: comma-ok the assertion and panic with the same informative message used above.

token/services/selector/config/driver.go
Effi-S: GetLimits() clamps an unset MaxLocksPerTransaction to min(default, resolved
MaxTokensPerSelection), so lowering only maxTokensPerSelection no longer fails Validate() at startup.

token/services/selector/simple/selector.go
Effi-S: Added > 0 && guards to the iteration-limit, lock-attempt and maxRetries checks so an unset
limit (0) means unbounded instead of aborting on the first token.

token/services/selector/sherdlock/selector.go
Effi-S: On the limited path branch 1 now projects the constant owner_wallet_id so a co-owned token's
multiple ownership rows collapse to one UNION row and LIMIT counts distinct tokens; fixture gains a
multi-owner token that fails pre-fix (token [co:0] returned more than once).

token/services/selector/simple/selector.go
Effi-S: selectInternal takes selectMu.Lock() (not RLock) so two concurrent selections on one selector
serialise instead of racing on the same sql.Rows.

token/services/selector/simple/selector.go
Effi-S: Per-selection counters moved off *selector into a per-call stack struct, removing the data race
when a selector is shared across goroutines.

token/services/selector/simple/selector.go
Effi-S: Made the two boundary comparisons consistent and treat "full page that is the whole wallet" as
insufficient funds (sentinelled), not limit-exceeded.

token/services/selector/sherdlock/fetcher.go
Effi-S: Dropped the inert limit parameter from the sherdlock TokenFetcher interface, impls, caller and
regenerated mock (sherdlock enforces its bound in-code, not via the DB query).

token/services/selector/config/driver.go
Effi-S: Reverted defaultMaxRetries 10 → 3, restoring the prior ~45s (was documented 80s) default
worst-case latency; docs updated.

token/services/selector/sherdlock/bounded_locker.go
Effi-S: Documented that the counter is a per-selection ceiling (reset on per-action Close) rather than
a hard per-transaction cap — renaming the config key would be breaking.

token/services/selector/simple/inmemory/locker.go
Effi-S: Corrected the misleading comment/docs to state the counter is per-owner-shard, so a
multi-wallet selection gets the budget per shard.

token/services/selector/sherdlock/manager.go
Effi-S: Added a standalone counterReaper that evicts orphaned lock counters (idle > 1h) even when lease
cleanup is disabled, bounding replica-local growth.

integration/nwo/token/topology/options.go
Effi-S: Comma-ok'd the k.(string) key assertion and panic with the same informative invalid nested
mapping type: %T message used above.

PR #1708, opened 2026-05-14
fix(recovery): unblock queue head by marking NotFound orphans Deleted after grace period

### LINKED ISSUE #1709: recovery: queue head can be permanently blocked by orphan NotFound transactions
## Problem

The recovery loop's queue can be permanently blocked when the oldest batch of `status=1` rows in the audit DB contains orphan transactions — entries whose audit log was persisted but whose transaction never reached the ledger (NotFound on `GetTransactionStatus`).

## Mechanism

1. `ClaimPendingTransactions` selects `ORDER BY stored_at ASC LIMIT BatchSize`, always pulling the oldest 100 rows.
2. `TTXRecoveryHandler.Recover` calls `GetTransactionStatus`; for an orphan this returns NotFound and the error is propagated.
3. `ReleaseRecoveryClaim` clears `claimed_by` / `expires_at` but does not touch `status`, so the row remains `status=1` and is immediately re-eligible.
4. The next sweep selects the same 100 rows. Newer rows in the queue never get scanned.

## Impact

Verified on a deployment with ~250k `status=1` audit rows: 99/99 txid overlap between two sweeps 10 seconds apart, with the recovery log spinning at ~1224 WARN/min on the same set indefinitely. The recovery worker remains busy but makes no forward progress for any newer pending row.

## Proposed fix

Introduce a `NotFoundGracePeriod` (default 30 min) in `recovery.Config`: when `GetTransactionStatus` returns NotFound and `stored_at` is older than the grace period, mark the row `Deleted` via `SetStatus` so it leaves the scan range. Transient NotFound that resolves within the grace period is unaffected. Setting `NotFoundGracePeriod=0` preserves the original "never mark Deleted, retry forever" behaviour.

Implementation in #1708.

## Environment

- `recovery.Storage.ClaimPendingTransactions` (token/services/storage/recovery/manager.go)
- `TTXRecoveryHandler.Recover` (network/fabric)
- Deployment uses fabric-x committer; orphans arise when the audit log commit succeeds but the corresponding ledger commit is rejected by the orderer.

EvanYan1024:
Fixes #1709

## Summary

The recovery loop's queue is permanently blocked when the oldest batch of
`status=1` rows contains orphan transactions whose audit log was persisted
but never reached the ledger.

This PR adds `NotFoundGracePeriod` (default 30 min) to `recovery.Config`:
when `GetTransactionStatus` returns NotFound and `stored_at` is older than
the grace period, the row is marked `Deleted` via `SetStatus`, freeing it
from the `ORDER BY stored_at ASC LIMIT BatchSize` scan range.

Configurable via `services.network.fabric.recovery.notFoundGracePeriod`;
set to 0 to disable (preserves prior behaviour).

## Mechanism (why the queue blocks)

1. `ClaimPendingTransactions` selects `ORDER BY stored_at ASC LIMIT BatchSize`,
   always pulling the oldest 100 rows.
2. `TTXRecoveryHandler.Recover` calls `GetTransactionStatus` which returns
   NotFound for orphans, propagating the error.
3. `ReleaseRecoveryClaim` clears `claimed_by` / `expires_at` but does not
   touch `status`, so the row remains `status=1` and immediately re-eligible.
4. Next sweep selects the same 100 rows. Newer rows in the queue never get scanned.

Verified on a deployment with ~250k `status=1` audit rows: 99/99 txid
overlap between two sweeps 10 seconds apart, with the recovery log spinning
at 1224 WARN/min on the same set indefinitely.

## Behaviour after fix

- Orphans older than the grace period get `SetStatus(Deleted)` and leave the
  scan range; queue head advances naturally.
- Transient NotFound (committer restart, brief namespace propagation lag)
  does not trigger the promotion as long as it resolves within the grace period.
- Backwards compatible: `NotFoundGracePeriod=0` restores the original
  "never mark Deleted, retry forever" behaviour.

## API changes

- `recovery.Storage` interface adds `SetStatus(ctx, txID, status, message) error`.
  The existing `transactionDB` adapter in `network/fabric` already satisfies
  this method, so no integration code changes are required.
- `recovery.Config` adds `NotFoundGracePeriod time.Duration`.

## Test plan

- [x] `go build ./token/services/storage/recovery/...`
- [x] `go vet ./token/services/storage/recovery/...`
- [x] Existing mock and unit tests compile against the updated `Storage` interface
- [ ] CI green

### DIFF
--- a/token/services/storage/recovery/config.go
+++ b/token/services/storage/recovery/config.go
@@ -35,18 +35,25 @@ type Config struct {
 	AdvisoryLockID int64
 	// InstanceID identifies the current replica as the lease owner.
 	InstanceID string
+	// NotFoundGracePeriod: when GetTransactionStatus returns a NotFound error and
+	// the tx was stored more than this duration ago, the recovery loop marks the
+	// row as Deleted instead of leaving it for another retry. Prevents the queue
+	// from being permanently blocked by orphan transactions (broadcast failures
+	// that never reached the ledger). Zero disables this behaviour.
+	NotFoundGracePeriod time.Duration
 }
 
 // DefaultConfig returns the default recovery configuration
 func DefaultConfig() Config {
 	return Config{
-		Enabled:        true,
-		TTL:            30 * time.Second,
-		ScanInterval:   5 * time.Second,
-		BatchSize:      defaultBatchSize,
-		WorkerCount:    defaultWorkers,
-		LeaseDuration:  defaultLeaseDuration,
-		AdvisoryLockID: defaultLockID,
+		Enabled:             true,
+		TTL:                 30 * time.Second,
+		ScanInterval:        5 * time.Second,
+		BatchSize:           defaultBatchSize,
+		WorkerCount:         defaultWorkers,
+		LeaseDuration:       defaultLeaseDuration,
+		AdvisoryLockID:      defaultLockID,
+		NotFoundGracePeriod: 30 * time.Minute,
 	}
 }
 
@@ -88,6 +95,13 @@ func LoadConfig(cfg *config.Configuration) (Config, error) {
 	}
 	if config.InstanceID != "" {
 		result.InstanceID = config.InstanceID
+	}
+	// NotFoundGracePeriod accepts an explicit zero to disable the orphan
+	// promotion, so check IsSet rather than the Go zero value. Without this
+	// gate, setting notFoundGracePeriod: 0 in config would silently fall back
+	// to the 30 min default and the documented opt-out would be unreachable.
+	if cfg.IsSet(ConfigKeyRecovery + ".notFoundGracePeriod") {
+		result.NotFoundGracePeriod = config.NotFoundGracePeriod
 	}
 
 	return result, nil
--- a/token/services/storage/recovery/manager.go
+++ b/token/services/storage/recovery/manager.go
@@ -10,11 +10,13 @@ import (
 	"context"
 	"fmt"
 	"math/rand"
+	"strings"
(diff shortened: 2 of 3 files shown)


adecaro:
Hi @EvanYan1024 , thanks for submitting this. Great catch 🙏 

I would take the change and modify `ClaimPendingTransactions`. There is no need to return a TransactionRecrod there, actually. We just need the transaction ID and when it was stored last time.
What do you think?

EvanYan1024:
> Hi @EvanYan1024 , thanks for submitting this. Great catch 🙏您好，感谢您的提交。发现得真好🙏
> 
> I would take the change and modify `ClaimPendingTransactions`. There is no need to return a TransactionRecrod there, actually. We just need the transaction ID and when it was stored last time.我会接受修改并修改 `ClaimPendingTransactions` 。实际上，这里不需要返回 TransactionRecord 对象。我们只需要交易 ID 和上次存储时间即可。 What do you think?你怎么认为？

Thanks @adecaro! Confirmed — recovery only reads `record.TxID` and
  `record.Timestamp` (manager.go:258-259); the other 12 fields go unused.

  I'd like to do that simplification, but as a follow-up PR rather than
  folding it into this one — the type change ripples through 5 layers
  (driver / SQL / postgres / store / adapter) plus mock regen, which is
  a different review surface from this queue-head fix. Happy to open
  the follow-up right after this merges. WDYT?

adecaro:
Perfect, let's do that. Thanks @EvanYan1024 🙏

adecaro (approved):
LGTM

PR #1712, opened 2026-05-17
fix: add signature request verification in EndorseView (#1611)

### LINKED ISSUE #1611: Signer-Side Independent Verification of the Transaction Content [CRITICAL]
**Description**
When a remote signer (owner or issuer) receives a `SignatureRequest`, it must independently verify that the `Request` bytes it is being asked to sign are consistent with the transaction it already holds. Concretely, the `EndorseView` should recompute `MarshalToSign()` from its own local copy of the transaction (`s.tx`) and compare the result against `signatureRequest.Request` before producing a signature. If they do not match, the request must be rejected. This closes the gap currently marked with `// TODO: check what is signed...` and ensures that a malicious initiator cannot substitute a different payload (e.g., a different amount, recipient, or token type) into the `Request` field while keeping the outer transaction envelope intact.

**Summary**
EndorseView signs transactions without verifying the initiator's signature request matches the locally-computed transaction data, enabling transaction tampering attacks.
Proposed solution: Add byte comparison check before signing:
`if !bytes.Equal(signatureRequest.Request, tokenRequestToSign) { return error }` (in endorse.go line 139).

Need to check the entire code for cases where data is signed. In all these cases the signer should verify the data it is signing and the identity it is using to sign (both are read from the network) match what the signer expects.

AkramBitar:
Addresses a security vulnerability where remote signers would accept and sign transactions without verifying that the SignatureRequest.TX bytes match their locally stored transaction.

See more information in: #1611

### DIFF
--- a/token/services/ttx/endorse.go
+++ b/token/services/ttx/endorse.go
@@ -120,6 +120,14 @@ func (s *EndorseView) handleSignatureRequests(context view.Context) error {
 			}
 		}
 
+		// Verify the transaction in the signature request matches our local transaction
+		if !bytes.Equal(signatureRequest.TX, s.tx.FromRaw) {
+			return errors.Errorf(
+				"signature request transaction mismatch for signer [%s]",
+				signerIdentity,
+			)
+		}
+
 		// check for the expected identity
 		if !signatureRequest.Signer.Equal(signerIdentity) {
 			return errors.Wrapf(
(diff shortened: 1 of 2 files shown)


AkramBitar:
This is not security issue. See https://github.com/hyperledger-labs/fabric-token-sdk/issues/1611
See https://github.com/hyperledger-labs/fabric-token-sdk/issues/1611#issuecomment-4476371474

PR #1715, opened 2026-05-18
refactor(recovery): return RecoveryClaim from ClaimPendingTransactions

### LINKED ISSUE #1714: recovery: ClaimPendingTransactions over-projects rows the caller never reads
## Problem

`ClaimPendingTransactions` returns `[]*TransactionRecord`, but the only consumer — `recovery.Manager` in `token/services/storage/recovery/manager.go` — reads exactly two fields: `TxID` and the stored-at timestamp. The other ten fields of `TransactionRecord` (`ActionType`, `SenderEID`, `RecipientEID`, `TokenType`, `Amount`, `Status`, `ApplicationMetadata`, `PublicMetadata`, `RecoveryClaimedBy`, `RecoveryClaimExpiresAt`) are always discarded by the caller.

## Affected paths

The over-projection materialises as wasted work in every recovery sweep:

- **SQL `SELECT` projection** in both `db/sql/common/transactions.go` and `db/sql/postgres/transactions.go` lists 10 columns; only 2 are consumed.
- **Postgres CTE-based claim query** does a second `INNER JOIN` against the `requests` table purely to fetch `status` + metadata columns the caller throws away.
- **Two JSON unmarshals per row** (`application_metadata`, `public_metadata`) executed by the SQL adapter, results never read.
- **Internal `pendingTx` struct** in `recovery.Manager` exists only to extract `{TxID, StoredAt}` from the bloated `TransactionRecord` before fanning out to workers — it would be redundant if the storage layer already returned the minimal projection.

## Proposed refactor

Introduce a dedicated lightweight type at the driver layer:

```go
type RecoveryClaim struct {
    TxID     string
    StoredAt time.Time
}
```

…and trim the SQL projection in both implementations to `SELECT tx_id, stored_at`. The internal `pendingTx` struct can then be deleted and `*RecoveryClaim` used directly as the work-channel element type.

## Context from review thread

This follow-up was explicitly requested by reviewer @adecaro on #1708 (the queue-head fix that introduced grace-period handling):

> Hi @EvanYan1024, thanks for submitting this. Great catch 🙏
>
> I would take the change and modify `ClaimPendingTransactions`. There is no need to return a TransactionRecord there, actually. We just need the transaction ID and when it was stored last time. What do you think?
>
> — @adecaro, [comment](https://github.com/hyperledger-labs/fabric-token-sdk/pull/1708#issuecomment-4474440103)

I agreed but proposed splitting it out so the type-change ripple wouldn't be folded into the queue-head fix:

> Thanks @adecaro! Confirmed — recovery only reads `record.TxID` and `record.Timestamp` (manager.go:258-259); the other 12 fields go unused.
>
> I'd like to do that simplification, but as a follow-up PR rather than folding it into this one — the type change ripples through 5 layers (driver / SQL / postgres / store / adapter) plus mock regen, which is a different review surface from this queue-head fix. Happy to open the follow-up right after this merges. WDYT?
>
> — @EvanYan1024, [comment](https://github.com/hyperledger-labs/fabric-token-sdk/pull/1708#issuecomment-4474884180)

> Perfect, let's do that. Thanks @EvanYan1024 🙏
>
> — @adecaro, [comment](https://github.com/hyperledger-labs/fabric-token-sdk/pull/1708#issuecomment-4474890925)

This issue tracks that follow-up.

## Out of scope

- The dedupe-by-TxID step in `Manager.recoverTransactions` stays in the manager (not pushed into SQL) — the `transactions` table still produces one row per movement/output for a given txID, and pushing that into SQL is a separate query-shape change.
- The `NotFoundGracePeriod` orphan-promotion logic from #1708 is unchanged.

EvanYan1024:
Closes #1714

Follow-up to #1708.

## Summary

`ClaimPendingTransactions` previously returned `[]*TransactionRecord`, but the only consumer (`recovery.Manager`) reads just `TxID` and `StoredAt` — the other ten fields (action type, amounts, application/public metadata, status, ...) were always discarded.

Introduce a dedicated lightweight type in the driver layer, mirroring the existing `RecoveryClaimParams` input type:

```go
type RecoveryClaim struct {
    TxID     string
    StoredAt time.Time
}
```

…and project only the two columns the recovery loop actually needs from SQL.

## Concrete savings

- **No more `transactions` JOIN.** Both the postgres atomic claim and the common permissive query now read `tx_id, stored_at` directly from `requests`. The postgres CTE collapses to a single `UPDATE ... RETURNING (tx_id, stored_at)`; the previous outer `SELECT` existed only to recover `stored_at` via a second join.
- **Two JSON unmarshals per row removed** (`application_metadata`, `public_metadata`).
- **Eight unused columns dropped** from the `SELECT` projection (`action_type`, `sender_eid`, `recipient_eid`, `token_type`, `amount`, `status`, `application_metadata`, `public_metadata`).
- **Internal `pendingTx` struct in `recovery.Manager` removed** — `*RecoveryClaim` is now used directly as the work-channel element, eliminating a redundant intermediate type.
- **Dedupe-by-`TxID` block in `Manager.recoverTransactions` removed** (see "Semantic change" below).

## Semantic change — please confirm

`olderThan` is now compared against **`requests.stored_at` (system insert time)** instead of **`transactions.stored_at` (the caller-supplied `TransactionRecord.Timestamp`)**.

Rationale: the recovery loop's notion of "how long has this row been stuck in pending" — which `NotFoundGracePeriod` in #1708 already relies on — is a system-time question, not a business-time one. The previous JOIN against `transactions` was using a caller-controlled field whose meaning depends on whoever calls `AddTransaction`, making the recovery clock implicitly dependent on the producer's choice of `Timestamp`. Anchoring on `requests.stored_at` makes the filter robust against that.

In typical usage `AddTokenRequest` and `AddTransaction` are committed in the same SQL transaction, so the two columns are within milliseconds of each other and observable behaviour is identical. The change matters only when a caller deliberately writes a backdated `TransactionRecord.Timestamp` — which the recovery loop should arguably not depend on.

Flagging this explicitly because it goes a bit beyond the literal description in #1714. Happy to revert this specific point if @adecaro / maintainers prefer to keep the historical behaviour.

## Dedupe block removed

`Manager.recoverTransactions` used to dedupe by `TxID` because the previous SQL joined against `transactions` (which fans out to one row per movement/output for a single tx_id). The new query reads from `requests` where `tx_id` is the primary key, so each returned claim is unique by construction and the seen-map became dead weight.

## Scope (5 layers)

1. `db/driver/{ttx,audit}.go` — new `RecoveryClaim` type + interface signature.
2. `db/sql/common/transactions.go` — query reads `tx_id, stored_at` from `requests` only; no join.
3. `db/sql/postgres/transactions.go` — CTE collapses to `UPDATE ... RETURNING`; dead `unmarshalMetadata` + `encoding/json` import removed.
4. `{ttxdb,auditdb}/store.go` — `StoreService` signature + re-exported type alias.
5. `network/fabric/network.go` + `recovery/manager.go` — adapter and `recovery.Storage` interface signatures; `pendingTx` and the dedupe-by-`TxID` block deleted, `RecoveryClaim` consumed directly by the worker goroutines.

Mocks regenerated via `counterfeiter v6.12.2` (`recovery/mock/storage.go`, `auditor/mock/audit_transaction_store.go`).

## Behaviour preserved

- The `NotFoundGracePeriod` orphan-promotion logic from #1708 is unchanged (and now lines up with `olderThan` semantically — both are system-time anchored).

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./token/services/storage/recovery/...`
- [x] `go test ./token/services/storage/...` — sqlite, ttxdb, auditdb, tokendb, walletdb all green
- [x] `go test ./token/services/auditor/... ./token/services/network/fabric/...`
- [x] `make testing-docker-images && go test -race ./token/services/storage/db/sql/postgres -count=1` — full postgres suite green (this is the path that was red on the first CI run; tests now seed `requests.stored_at` via a new `ageRequests` helper so the filter actually matches)
- [ ] CI green

### DIFF
--- a/token/services/storage/db/sql/common/transactions.go
+++ b/token/services/storage/db/sql/common/transactions.go
@@ -327,22 +327,19 @@ func (db *TransactionStore) AcquireRecoveryLeadership(ctx context.Context, lockI
 
 // ClaimPendingTransactions returns a claimed batch of Pending transactions.
 // The default SQL implementation is permissive and does not persist recovery claims.
-// Only tx_id and stored_at are projected — the recovery loop does not need the
-// rest of the row, and skipping them avoids two metadata JSON unmarshals per row.
+// tx_id and stored_at are projected directly from the requests table — the
+// transactions table is no longer joined since it carries no information
+// the recovery loop needs and adding it would re-introduce a fan-out by
+// movement/output that the caller would have to dedupe by tx_id.
 func (db *TransactionStore) ClaimPendingTransactions(ctx context.Context, params dbdriver.RecoveryClaimParams) ([]*dbdriver.RecoveryClaim, error) {
-	transactionsTable, requestsTable := q.Table(db.table.Transactions), q.Table(db.table.Requests)
 	query, args := q.Select().
-		Fields(
-			transactionsTable.Field("tx_id"), transactionsTable.Field("stored_at"),
-		).
-		From(transactionsTable.Join(requestsTable,
-			cond.Cmp(transactionsTable.Field("tx_id"), "=", requestsTable.Field("tx_id"))),
-		).
+		FieldsByName("tx_id", "stored_at").
+		From(q.Table(db.table.Requests)).
 		Where(cond.And(
 			cond.Eq("status", dbdriver.Pending),
-			cond.Lt(common3.FieldName(db.table.Transactions+".stored_at"), params.OlderThan),
+			cond.Lt("stored_at", params.OlderThan),
 		)).
-		OrderBy(q.Asc(transactionsTable.Field("stored_at"))).
+		OrderBy(q.Asc(common3.FieldName("stored_at"))).
 		Limit(params.Limit).
 		Format(db.ci)
 
--- a/token/services/storage/db/sql/common/transactions.go
+++ b/token/services/storage/db/sql/common/transactions.go
@@ -327,14 +327,13 @@ func (db *TransactionStore) AcquireRecoveryLeadership(ctx context.Context, lockI
 
 // ClaimPendingTransactions returns a claimed batch of Pending transactions.
 // The default SQL implementation is permissive and does not persist recovery claims.
-func (db *TransactionStore) ClaimPendingTransactions(ctx context.Context, params dbdriver.RecoveryClaimParams) ([]*dbdriver.TransactionRecord, error) {
+// Only tx_id and stored_at are projected — the recovery loop does not need the
+// rest of the row, and skipping them avoids two metadata JSON unmarshals per row.
+func (db *TransactionStore) ClaimPendingTransactions(ctx context.Context, params dbdriver.RecoveryClaimParams) ([]*dbdriver.RecoveryClaim, error) {
 	transactionsTable, requestsTable := q.Table(db.table.Transactions), q.Table(db.table.Requests)
 	query, args := q.Select().
 		Fields(
-			transactionsTable.Field("tx_id"), common3.FieldName("action_type"), common3.FieldName("sender_eid"),
-			common3.FieldName("recipient_eid"), common3.FieldName("token_type"), common3.FieldName("amount"),
-			requestsTable.Field("status"), requestsTable.Field("application_metadata"),
-			requestsTable.Field("public_metadata"), transactionsTable.Field("stored_at"),
+			transactionsTable.Field("tx_id"), transactionsTable.Field("stored_at"),
 		).
 		From(transactionsTable.Join(requestsTable,
 			cond.Cmp(transactionsTable.Field("tx_id"), "=", requestsTable.Field("tx_id"))),
@@ -353,19 +352,8 @@ func (db *TransactionStore) ClaimPendingTransactions(ctx context.Context, params
 		return nil, err
 	}
 
-	results := common.NewIterator(rows, func(r *dbdriver.TransactionRecord) error {
-		var amount BigInt
-		var appMeta []byte
-		var pubMeta []byte
-		if err := rows.Scan(&r.TxID, &r.ActionType, &r.SenderEID, &r.RecipientEID, &r.TokenType, &amount, &r.Status, &appMeta, &pubMeta, &r.Timestamp); err != nil {
-			return err
-		}
-		r.Amount = amount.Int
-
-		return errors2.Join(
-			unmarshal(appMeta, &r.ApplicationMetadata),
-			unmarshal(pubMeta, &r.PublicMetadata),
-		)
+	results := common.NewIterator(rows, func(r *dbdriver.RecoveryClaim) error {
+		return rows.Scan(&r.TxID, &r.StoredAt)
 	})
 
 	return iterators.ReadAllPointers(results)
--- a/token/services/storage/recovery/manager.go
+++ b/token/services/storage/recovery/manager.go
@@ -244,22 +244,14 @@ func (m *Manager) recoverTransactions(ctx context.Context) error {
 		go m.worker(ctx, &workerWG, work, errCh)
 	}
 
-	// ClaimPendingTransactions returns one RecoveryClaim per ledger transaction
-	// row, and a single txID can produce multiple rows (one per movement/output).
-	// Dedupe by TxID before fanning out so two workers do not concurrently call
-	// Recover/SetStatus/ReleaseRecoveryClaim against the same transaction. Keep
-	// the claim with the earliest StoredAt so the grace period decision uses
-	// the row's true age.
-	seen := make(map[string]*ttxdb.RecoveryClaim, len(records))
-	for _, record := range records {
-		if record == nil {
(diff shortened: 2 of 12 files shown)


token/services/storage/db/sql/common/transactions.go
```diff
@@ -327,14 +327,13 @@ func (db *TransactionStore) AcquireRecoveryLeadership(ctx context.Context, lockI
 
 // ClaimPendingTransactions returns a claimed batch of Pending transactions.
 // The default SQL implementation is permissive and does not persist recovery claims.
-func (db *TransactionStore) ClaimPendingTransactions(ctx context.Context, params dbdriver.RecoveryClaimParams) ([]*dbdriver.TransactionRecord, error) {
+// Only tx_id and stored_at are projected — the recovery loop does not need the
+// rest of the row, and skipping them avoids two metadata JSON unmarshals per row.
+func (db *TransactionStore) ClaimPendingTransactions(ctx context.Context, params dbdriver.RecoveryClaimParams) ([]*dbdriver.RecoveryClaim, error) {
 	transactionsTable, requestsTable := q.Table(db.table.Transactions), q.Table(db.table.Requests)
```
adecaro: I think we don't need anymore the transactions table. We have everything we need in the requests table. So, we can simplify further this query and the custom Postgres implementation. What do you think?

token/services/storage/recovery/manager.go
```diff
@@ -244,23 +244,23 @@ func (m *Manager) recoverTransactions(ctx context.Context) error {
...
 	}
 
-	// ClaimPendingTransactions returns one TransactionRecord per ledger
-	// transaction row, and a single txID can produce multiple rows (one per
-	// movement/output). Dedupe by TxID before fanning out so two workers do
-	// not concurrently call Recover/SetStatus/ReleaseRecoveryClaim against
-	// the same transaction. Keep the earliest StoredAt to make the grace
-	// period decision use the row's true age.
-	seen := make(map[string]time.Time, len(records))
+	// ClaimPendingTransactions returns one RecoveryClaim per ledger transaction
```
adecaro: If we remove the Join with the transactions table, we don't need this anymore. Would you agree?

adecaro:
Thanks a lot, @EvanYan1024 . I have left a few comments to the code to simplify it even further.

I would also probe your opinion on introducing a new state for this transactions that are kind of orphaned. This would allows us to find them more easily and give them a second chance.
What do you think?

token/services/storage/recovery/manager.go
EvanYan1024: yes

token/services/storage/db/sql/common/transactions.go
EvanYan1024: sure

EvanYan1024:
> I would also probe your opinion on introducing a new state for this transactions that are kind of orphaned. This would allows us to find them more easily and give them a second chance.

@adecaro to make sure I'm reading you right — is the proposal to change the SetStatus(ctx, txID, storage.Deleted, ...) call I added in #1708 (recovery/manager.go:331, on
   the NotFoundGracePeriod path) so it sets a new dedicated Orphan state instead? That way Deleted stays reserved for explicit operator/user deletion, and "tried to
  commit, never landed on chain within the grace window" gets its own unambiguous signal.

adecaro:
> > I would also probe your opinion on introducing a new state for this transactions that are kind of orphaned. This would allows us to find them more easily and give them a second chance.
> 
> @adecaro to make sure I'm reading you right — is the proposal to change the SetStatus(ctx, txID, storage.Deleted, ...) call I added in #1708 (recovery/manager.go:331, on the NotFoundGracePeriod path) so it sets a new dedicated Orphan state instead? That way Deleted stays reserved for explicit operator/user deletion, and "tried to commit, never landed on chain within the grace window" gets its own unambiguous signal.

Yes, this is what I meant. We can introduce this new state in a new PR. That's also fine.

EvanYan1024:
> > > I would also probe your opinion on introducing a new state for this transactions that are kind of orphaned. This would allows us to find them more easily and give them a second chance.
> > 
> > 
> > @adecaro to make sure I'm reading you right — is the proposal to change the SetStatus(ctx, txID, storage.Deleted, ...) call I added in #1708 (recovery/manager.go:331, on the NotFoundGracePeriod path) so it sets a new dedicated Orphan state instead? That way Deleted stays reserved for explicit operator/user deletion, and "tried to commit, never landed on chain within the grace window" gets its own unambiguous signal.
> 
> Yes, this is what I meant. We can introduce this new state in a new PR. That's also fine.

Okay. I think that's fine.

adecaro:
Hi @EvanYan1024 , please, double check the DCO and the unit-test to see if it is a flaky test or something else. Thanks 🙏

adecaro (approved):
LGTM. Thanks @EvanYan1024 🙏

PR #1700, opened 2026-05-13
feat(ttx): add versioned envelope for interactive protocol messages

SuyashAlphaC:
## Summary
  Implements **Phase 1 (Foundation)** of the TTX versioning proposal discussed in #1654.
  This introduces a session-level `Envelope` wrapper for all JSON-typed interactive protocol messages, enabling strict version negotiation and message-type discrimination — addressing the silent failure mode when nodes running different
   SDK versions exchange messages.
  ### What's included
  - **`Envelope` struct** with compact wire fields (`v`=version, `t`=type, `b`=body) in `token/services/utils/json/session/envelope.go`
  - **Strict versioning**: receivers reject unversioned messages (`ErrMissingVersion`) and version mismatches (`VersionError` with `errors.Is(ErrVersionMismatch)` support)
  - **Mandatory type discriminator** with constants for all JSON-typed flows (`TypeRecipientRequest`, `TypeWithdrawalRequest`, `TypeSpendRequest`, etc.)
  - **Session helpers**: `SendTyped`, `ReceiveTyped`, `ReceiveTypedWithTimeout` for simplified versioned message exchange
  - **Metrics instrumentation**: `EnvelopeMetrics` (sent/received counters, error counter, body size histogram) using `fabric-smart-client/platform/view/services/metrics`
  - **`VersionCompatibility` map** and `IsCompatible` helper for future multi-version support
  - **27 unit tests + 3 benchmarks** covering all error paths, concurrent operations, large payloads, and round-trip performance
  ### Design decisions (per #1654 discussion with @adecaro)
  | Question | Decision |
  |---|---|
  | Mixed-version policy | Strict — reject unversioned |
  | Field naming | Terse (`v/t/b`) |
  | Type discriminator | Mandatory (no `omitempty`) |
  | Coordination with #1622 | Proceed independently |
  | Raw-byte flows | Out of scope for v1 |

  ## Test plan
  - [x] `go vet ./...` passes
  - [x] `gofmt` clean
  - [x] All 27 unit tests pass (`go test -v -count=1 ./token/services/utils/json/session/...`)
  - [x] 3 benchmarks run successfully (WrapEnvelope, UnwrapEnvelope, RoundTrip)
  - [x] No regressions in existing `json/session` tests

Issue #1654, opened 2026-05-05
exploration(ttx): Versioning for Interactive Recipient Protocols


SuyashAlphaC:
# Versioning for TTX Interactive Protocols — First Analysis

## The Problem

Interactive views in `token/services/ttx/` exchange messages over FSC sessions via `session.Send` / `session.Receive`. There's no version or capability marker anywhere on the wire. If two nodes run different SDK versions, they connect a session and "talk" just fine — JSON silently ignores unknown fields and tolerates missing ones — but they can misinterpret each other with no error and no diagnostic.

This is going to bite more as the recipient flows pick up new fields (slimmer responses landing in #1652, future nonce/signature attestations, ack-only modes). The same concern applies to other interactive views in the package, not just recipients.

## Surface Map

Two categories of interactive flows in `ttx`:

**JSON-typed exchanges (the ones that need versioning):**

| File | Initiator view | Responder view | Wire structs |
|------|----------------|----------------|--------------|
| `recipients.go` | `RequestRecipientIdentityView` | `RespondRequestRecipientIdentityView` | `RecipientRequest` → `RecipientData` (+ optional `MultisigRecipientData`) |
| `recipients.go` | `ExchangeRecipientIdentitiesView` | `RespondExchangeRecipientIdentitiesView` | `ExchangeRecipientRequest` → `RecipientData` |
| `withdrawal.go` | `RequestWithdrawalView` | `ReceiveWithdrawalRequestView` | `WithdrawalRequest` |
| `upgrade.go` | `UpgradeTokensInitiatorView` | `UpgradeTokensResponderView` | `UpgradeTokensAgreement` ↔ `UpgradeTokensRequest` |

**Raw byte exchanges (out of scope here):** `auditor.go`, `endorse.go`, `accept.go`, `collectactions.go`, `collectendorsements.go` — these already wrap their own protocol-level structure inside `[]byte` payloads (`token.Request`, signatures), so versioning lives inside the inner payload, not on the session.

A couple of patterns from the recent history that motivate this:

- `RecipientData` picked up `TokenMetadataAuditInfo` after the original 3-field shape. Older peers serialising the original shape just get tolerated silently.
- `WithdrawalRequest` and `UpgradeTokensRequest` both embed `RecipientData` inline, so any shape change to recipients transitively touches them.
[...]

PR #1716, opened 2026-05-18
  feat(ttx): Phase 2 — migrate all JSON-typed interactive views to versioned envelope

### LINKED ISSUE #1710: ttx: Phase 2 — migrate JSON interactive protocols to versioned envelope (#1654)
This tracks Phase 2 (migration) from #1654 : move the JSON-typed TTX flows onto the Phase 1 envelope so everything on the wire is versioned + type-tagged .

  ### Scope

  • Migrate RecipientRequest / RecipientData to the envelope
  • Migrate ExchangeRecipientRequest and the exchange response path to the envelope
  • Migrate WithdrawalRequest to the envelope
  • Migrate UpgradeTokensAgreement / UpgradeTokensRequest to the envelope
  • Migrate MultisigRecipientData / PolicyRecipientData to the envelope

  ### Testing

  • Integration tests for each migrated view
  • Error-path tests (bad/missing version, wrong type, malformed envelope, etc.)

SuyashAlphaC:
Closes #1710 

Phase 2 of the versioning work tracked in #1654. Builds on the envelope layer that landed in #1700 (Phase 1).

  Every JSON-typed interactive flow in `ttx` now goes through `SendTyped` / `ReceiveTyped`, so every wire message carries the `{v, t, b}` envelope with strict version + mandatory type checking. The raw-byte flows (`auditor.go`,
  `endorse.go`, `accept.go`, `collectactions.go`, `collectendorsements.go`) are intentionally left alone — they have their own internal structure and wrapping them would conflict with signatures over the raw payload.

  ### What's migrated

  | File | Messages now enveloped |
  |------|------------------------|
  | `recipients.go` | `RecipientRequest`/`Response`, `ExchangeRecipientRequest`/`Resp`, `MultisigRecipientData`, `PolicyRecipientData` |
  | `withdrawal.go` | `WithdrawalRequest` (initiator + responder) |
  | `upgrade.go` | `UpgradeTokensAgreement` and `UpgradeTokensRequest` (both directions) |
  | `multisig/spend.go` | `SpendRequest`, `SpendResponse` |
  | `boolpolicy/spend.go` | `SpendRequest`, `SpendResponse` |

  ### Tests

  New protocol-level round-trip + error-path tests live next to each migrated package:

  - `token/services/utils/json/session/envelope_protocol_test.go` — exercises every type constant end-to-end through the session helpers and the four documented error paths (wrong version, wrong type, missing version, malformed
  envelope)
  - `token/services/ttx/envelope_protocol_test.go` — round-trip for all 8 recipients/withdrawal/upgrade message types
  - `token/services/ttx/multisig/envelope_protocol_test.go` — round-trip for both spend messages
  - `token/services/ttx/boolpolicy/envelope_protocol_test.go` — same for the policy variant

  ### Documentation

`docs/services/ttx.md` gets three new sections:

  - **Interactive Protocol Versioning** — wire shape, the full type-constant registry, strict-mode error semantics with the sentinel errors, version-compatibility rules, and the explicit out-of-scope rationale for the raw-byte flows
  - **Withdrawal Flow** — sequence diagram + prose for `RequestWithdrawalView` / `ReceiveWithdrawalRequestView`
  - **Token Upgrade Flow** — sequence diagram + prose for the agreement → request round-trip, including the challenge-ID check the responder enforces

  ### Test plan

  - [x] `go build ./...` clean
  - [x] `go vet ./token/services/...` clean
  - [x] `gofmt -l` clean
  - [x] `go test -count=1 ./token/services/utils/json/session/... ./token/services/ttx/ ./token/services/ttx/multisig/... ./token/services/ttx/boolpolicy/...` — all green

  Refs: #1654, #1700, #1710


SuyashAlphaC:
Hi @adecaro — Phase 2 is up for review.
  
  All four JSON-typed flows in `ttx` are now on the envelope (recipients, withdrawal, upgrade, multisig + boolpolicy spend), with the strict mandatory-Type semantics from your evaluation. Protocol round-trip and error-path tests added next to each migrated package.

  Per your "document all interactive protocols" ask, `docs/services/ttx.md` gains three new sections — the envelope layer itself, the withdrawal flow, and the token-upgrade flow.

  Whenever you have a moment, would appreciate a re-review. Thanks 🙏

adecaro:
Hi @SuyashAlphaC , great effort, thanks much 🙏 

I think we need to target all calls to `SendWithContext` or `Send`. In the cases, we are just sending raw bytes, we should introduce a struct and a type to cover that case. What do you think?

An example of these calls is in `func (s *EndorseView) handleSignatureRequests(context view.Context) error {`.

SuyashAlphaC:
Thanks for the review @adecaro . I totally agree — we should envelope every session message, including the ones we treated as raw bytes. For each of those I’ll add a small wire struct + type constant (e.g. signature response, signature request, tx payload, ack) and switch the send/receive sites to SendTyped/ReceiveTyped, starting with ```EndorseView.handleSignatureRequests``` and the matching paths in collectendorsements, auditor, accept, and collectactions. I will push the changes soon. 🙏

token/services/utils/json/session/envelope.go
```diff
@@ -59,6 +60,19 @@ const (
 	// multisig/spend.go and boolpolicy/spend.go
 	TypeSpendRequest  = "spend_req"
 	TypeSpendResponse = "spend_resp"
+
+	// collectendorsements.go, endorse.go, accept.go, auditor.go, receivetx.go
+	TypeSignatureRequest    = "sig_req"
```
adecaro: please, move all these consts closer to where they are used. This package should not carry service specific constants.

adecaro:
Hi @SuyashAlphaC , this PR is getting to a very good place. Thanks a lot for this effort 🙏 

I would say, let's take the change to:
- update the documentation under `docs/services/ttx.md` with the typed messages used by the service and if you can for each view to the expected flow of messages
- let's plug also EnvelopeMetrics, quite good metrics there. You can get an instance of the metrics provider using FSC `platform/view/services/metrics#GetProvider`.

Thanks again 🙏

SuyashAlphaC:
hey @adecaro , I have pushed the changes. Please review the changes whenever you have time.
Thanks 🙏

adecaro:
Hi @SuyashAlphaC , great. Thanks 🙏 

Let's refine the part relative to the metrics. You can do the following:
- Register in DIG (token/sdk/dig/sdk.go) the sessions metrics like `p.Container().Provide(session.NewEnvelopeMetrics)`. This in the `Install` function.
- Then, we can make it available to the view context by additing: `digutils.Register[*session.EnvelopeMetrics](p.Container())`.
- Now, in the views, metrics.go, we add the following function: 
```go
var (
	envelopeMetricsType = reflect.TypeOf((*EnvelopeMetrics)(nil))
)

func GetEnvelopeMetrics(sp token.ServiceProvider) (*EnvelopeMetrics, error) {
	s, err := sp.GetService(envelopeMetricsType)
	if err != nil {
		return nil, err
	}
	nip, ok := s.(*EnvelopeMetrics)
	if !ok {
		panic("implementation error, type must be *EnvelopeMetrics")
	}

	return nip, nil
}
```
At this point, GetEnvelopeMetrics can be invoked from any view to obtain the envelope metrics as from `SendTyped` for example. 


Ah, another thing I have just noticed: Instead of having:

```go
	jsonSession := session2.NewFromSession(context, session)
	if err := session2.SendTyped(jsonSession, context.Context(), signatureRequest, TypeSignatureRequest); err != nil {
		return nil, errors.Wrap(err, "failed sending transaction content")
	}
```

we can introduce a `NewTypedSession(context, session)` so we don't have to call static methods.

What do you think?

adecaro (approved):
LGTM. Great effort @SuyashAlphaC . Thanks a lot 🙏

PR #1717, opened 2026-05-18
Aggregated Protobuf Best Practices Analysis & Improvement Plan #1622

### LINKED ISSUE #993: dlog v1 finalization: an extra signer in the metadata should be an AuditableIdentity

### LINKED ISSUE #1622: Aggregated Protobuf Best Practices Analysis & Improvement Plan
This  issue provides a consolidated analysis of the Protobuf definitions within the `fabric-token-sdk` project. It outlines specific areas for improvement to ensure consistency, maintainability, and performance, while respecting external project constraints.
The goal is to get ready for the v1 release of the Token-SDK.

### Observations
Current package names are inconsistent and often lack namespacing or versioning:
- Generic names like `package protos;` are used in `token/driver/protos/request.proto`.
- Specific names like `package fabtoken;` (in `ftpp.proto`) or `package identity_idemix;` (in `idemix_config.proto`) are used without a common root.
- No versioning (e.g., `v1`, `v2`) is present in the package declarations.

### Recommendations
- **Hierarchical Namespacing**: Use a consistent root: `package fabric_token_sdk.<subsystem>.<version>;`.
- **Versioning**: Always include a version (e.g., `v1`) to allow for breaking changes without disrupting existing integrations.

### Implementation Details: Directory Structure
To support versioned packages, the directory structure should reflect the package name:
```text
token/driver/protos/v1/request.proto  -> package fabric_token_sdk.token.driver.v1;
token/common/v1/common.proto          -> package fabric_token_sdk.common.v1;
```

---

### Observations
Core primitives like `Identity` and `TokenID` are redefined in multiple files with inconsistent field names (`id` vs `tx_id`). The `Identity` message is also overly simplistic, using a generic `bytes raw` field without type information.

### Recommendations
- **Consolidation**: Move common messages to a `common` package to reduce duplication and ensure a single source of truth for SDK-wide types.
- **Structured Identity**: Enhance the `Identity` message to include an optional `type` or clear documentation indicating that `raw` contains a self-describing serialized identity (e.g., Fabric's `SerializedIdentity`).

### Implementation Details: `token/common/v1/common.proto`
```proto
syntax = "proto3";
package fabric_token_sdk.common.v1;

option go_package = "github.com/hyperledger-labs/fabric-token-sdk/token/common/v1;commonv1";

message Identity {
  // raw contains the serialized identity.
  // It is expected to be a self-describing format like Fabric's SerializedIdentity.
  bytes raw = 1;
}

message TokenID {
  string tx_id = 1;
  uint32 index = 2;
}
```

---

### Observations
- **Default Behavior**: `ActionType` in `request.proto` starts with `ISSUE = 0`. In proto3, the first enum value is the default, making it impossible to distinguish between an unset field and an explicit "Issue" action.
- **Naming Collisions**: Enum values like `ISSUE` are very generic and can collide in languages with flat enum namespaces.

### Recommendations
- **Zero-Value Unspecified**: The first value (index 0) must always be `ACTION_TYPE_UNSPECIFIED = 0;`.
- **Prefixing**: Prefix enum values with the enum name in `SCREAMING_SNAKE_CASE`.

### Implementation Details
```proto
enum ActionType {
  ACTION_TYPE_UNSPECIFIED = 0;
  ACTION_TYPE_ISSUE = 1;
  ACTION_TYPE_TRANSFER = 2;
}
```

---

### Observations
- **Cryptic Abbreviations**: `ipk` (Issuer Public Key), `sk` (Secret Key), and `ski` (Subject Key Identifier) are used in some configurations.
- **Typos**: `curver_id` instead of `curve_id` in `noghpp.proto`.
- **Consistency**: Inconsistent naming for identical concepts, e.g., `tx_id` vs `id` for transaction identifiers, and `application` vs `extra_data` vs `metadata` for map fields.

### Recommendations
- **Use Descriptive Names**: `issuer_public_key` instead of `ipk`, `secret_key` instead of `sk`.
- **Exception for Idemix**: `idemix_config.proto` must **NOT** be modified for these rules. Its field names (e.g., `ipk`, `sk`, `ski`) and numbering are inherited directly from the [IBM Idemix](https://github.com/IBM/idemix) library.
- **Fix Typos**: Rename `curver_id` to `curve_id`.
- **Unify IDs**: Standardize on `tx_id` for transaction identifiers.
- **Unified Metadata**: Use the name `metadata` for all map fields intended for application-specific or extension data.

---

### Observations
- **Version Types**: Mix of `uint32` and `uint64` for version fields.

### Recommendations
- **Standardize Versions**: Use `uint32` for all version fields unless `uint64` is strictly required.

---

### A. Buf Integration
Integrate `buf` for linting, formatting, breaking change detection, and documentation enforcement.

**Implementation Details (`buf.yaml`)**:
```yaml
version: v2
modules:
  - path: token/core/fabtoken/protos
  - path: token/core/zkatdlog/nogh/protos
  - path: token/driver/protos
  - path: token/services/identity/x509/crypto/protos
  - path: token/services/identity/idemix/crypto/protos
lint:
  use:
    - DEFAULT
    - PACKAGE_DIRECTORY_MATCH
    - PACKAGE_VERSION_SUFFIX
    - FIELD_LOWER_SNAKE_CASE
    - ENUM_ZERO_VALUE_SUFFIX
    - ENUM_VALUE_PREFIX
    - COMMENTS # Enforce documentation standards
  ignore:
    - token/services/identity/idemix/crypto/protos/idemix_config.proto
```

### B. Validation Constraints
Integrate `buf-validate` to enforce business rules directly in the schema (e.g., non-empty strings, byte length limits).

### C. Field Deprecation and Reservation
Establish clear guidelines for safely removing fields to prevent tag reuse collisions.

**Implementation Details**:
When removing a field, reserve both its number and name:
```proto
message TransferMetadata {
  reserved 4, 5, 6, 7;
  reserved "old_field_name";
}
```

---

## Implementation Roadmap

0.  **Phase 0: Cleanup (Immediate)**
    - Apply `buf format -w` to all files to fix indentation and formatting inconsistencies.
    - Fix the `curver_id` typo in `noghpp.proto`.
    - Standardize `go_package` options to a consistent path format.

1.  **Phase 1: Foundation (Short Term)**
    - Create `token/common/v1/common.proto` with core primitives.
    - Initial `buf.yaml` setup with Idemix exclusions.

2.  **Phase 2: Standardization (Short Term)**
    - Implement hierarchical versioned packages (e.g., `fabric_token_sdk.token.driver.v1`).
    - Restructure directories to match package names.
    - Standardize field naming to `lower_snake_case` (excluding Idemix).
    - Expand cryptic abbreviations (excluding Idemix).

3.  **Phase 3: Optimization & Safety (Medium Term)**
    - Transition `Token.quantity` from hex `string` to `bytes` (big-endian).
    - Implement `ENUM_ZERO_VALUE_UNSPECIFIED` across all enums.
    - Since these are breaking changes, they should be performed as part of the `v1` to `v2` package migration or within the new `v1` structure if the SDK is still in early development.

4.  **Phase 4: Continuous Quality (Long Term)**
    - Implement breaking change detection in CI (`buf breaking --against-input '.git#branch=main'`).
    - Generate API documentation from proto comments.

adecaro:


Storm1289:
Hi @adecaro, went through the new auditor.

Input commitments aren't re-checked -`validateTransferInputs` (auditor.go:403) only verifies owner/sender bytes; `actionInput.Token.Data` is never recomputed and compared the way `InspectOutput` does for outputs (auditor.go:585). A tampered input commitment would pass audit, which is what  #998 was about.

A few smaller things:
- Single sender hardcoded at auditor.go:418 (`len(Senders) != 1`) - breaks multi-sig inputs?
- TokenID check is conditional at auditor.go:439 (`if TokenID != nil`) - nil TokenID silently skips. Hard error instead?
- Empty receiver identity silently accepted in the transfer path at auditor.go:561 - `InspectIdentity` then fails late with "identity is nil". Fail early?

Thanks 🙏

adecaro:
> Hi @adecaro, went through the new auditor.
> 
> Input commitments aren't re-checked -`validateTransferInputs` (auditor.go:403) only verifies owner/sender bytes; `actionInput.Token.Data` is never recomputed and compared the way `InspectOutput` does for outputs (auditor.go:585). A tampered input commitment would pass audit, which is what #998 was about.
> 
> A few smaller things:
> 
> * Single sender hardcoded at auditor.go:418 (`len(Senders) != 1`) - breaks multi-sig inputs?
> * TokenID check is conditional at auditor.go:439 (`if TokenID != nil`) - nil TokenID silently skips. Hard error instead?
> * Empty receiver identity silently accepted in the transfer path at auditor.go:561 - `InspectIdentity` then fails late with "identity is nil". Fail early?
> 
> Thanks 🙏

Great catches, @Storm1289. Indeed. Those points need to be addressed 🙏

docs/driverapi.md
```diff
@@ -99,128 +99,161 @@ A `TokenRequest` is the primary structure submitted to the ledger. It consists o
...
-- **Boundary Ambiguity**: Lack of delimiter creates potential for hash collision attacks
-- **No Input Validation**: Anchor parameter not validated for size or content
-- **Binary Data in Logs**: Error messages may expose sensitive data
+The computation follows these steps:
 
-**Status**: ⚠️ **DEPRECATED** - Use Protocol V2 for new deployments
+1. **Action Serialization**: All actions in the token request are serialized in their original order, preserving the sequence of issues and transfers as they appear in the request.
 
-### Protocol V2 (Recommended)
+2. **ASN.1 Encoding**: The actions are encoded using a structured ASN.1 format that includes both the action type and data:
```
AkramBitar, line 203: I am wondering why we need to support two protocols?

docs/driverapi.md
```diff
@@ -99,128 +99,161 @@ A `TokenRequest` is the primary structure submitted to the ledger. It consists o
...
-- No breaking changes to existing deployments
+3. **Type Mapping**: Action types from the protobuf enum are mapped to ASN.1 integers:
+   - `ACTION_TYPE_ISSUE` (protobuf value 1) → ASN.1 INTEGER 0
+   - `ACTION_TYPE_TRANSFER` (protobuf value 2) → ASN.1 INTEGER 1
+
+4. **Anchor Binding**: The anchor (transaction ID) is included in the outer ASN.1 structure, creating a cryptographic binding between the signature and the specific transaction.
+
+### Security Properties
+
+Protocol V1 provides several critical security guarantees:
```
AkramBitar, line 229: What about Protocol V2?

token/core/common/auditor.go
```diff
@@ -0,0 +1,364 @@
...
+					return nil, errors.Errorf("duplicate token ID [%s] found in metadata at action index [%d] for tx [%s]", idKey, i, anchor)
+				}
+				tokenIDMap[idKey] = id
+				tokenIDs = append(tokenIDs, id)
+			}
+		}
+
+		// Extract TokenIDs from issue action inputs (for token upgrades/conversions)
+		if action.IssueMetadata != nil {
+			for _, input := range action.IssueMetadata.Inputs {
```
AkramBitar, line 245: Do we need to check that action.IssueMetadata.Inputs is not nil?

token/core/common/auditor.go
```diff
@@ -0,0 +1,364 @@
...
+	}
+
+	logger.DebugfContext(ctx, "[%s] retrieving [%d] audit tokens...", anchor, len(tokenIDs))
+	tokens, err := queryEngine.ListAuditTokens(ctx, tokenIDs...)
+	if err != nil {
+		return nil, errors.WithMessagef(err, "failed to retrieve audit tokens for tx [%s]", anchor)
+	}
+
+	// Build the token map
+	auditTokens := make(map[*token.ID]*token.Token, len(tokens))
```
AkramBitar: The map keys are pointer addresses (*token.ID), so equal values at different addresses are different keys. Is that ok?

token/core/common/auditor.go
```diff
@@ -0,0 +1,364 @@
...
+	for i, tok := range tokens {
+		if tok != nil && i < len(tokenIDs) {
```
AkramBitar: I assume this is in order to prevents a panic, but it can ignores extra entries in tokens

token/core/common/auditor.go
```diff
@@ -0,0 +1,364 @@
...
+			auditTokens[tokenIDs[i]] = tok
```
AkramBitar: Does tokenIDs[i] can be nil for some reason?

token/core/common/auditor.go
```diff
@@ -0,0 +1,364 @@
...
+
+	// Validate and extract token type from inputs (if any exist)
+	for i, inputMetadata := range metadata.Inputs {
+		if inputMetadata == nil {
+			continue
+		}
+
+		// Verify input token exists in auditTokens map
+		if inputMetadata.TokenID != nil {
+			inputToken, exists := auditTokens[inputMetadata.TokenID]
```
AkramBitar: Why are we comparing pointer memory addresses instead of the actual data values?

token/core/common/auditor.go
```diff
@@ -0,0 +1,364 @@
...
+					return errors.Errorf(
+						"token type mismatch in issue action: input [%d] has type [%s] but expected [%s]",
+						i, inputToken.Type, actionTokenType,
+					)
+				}
+			}
+		}
+	}
+
+	// Note: Output token type validation is driver-specific and handled by the driver's
```
AkramBitar, line 456: What about inputToken.Type = ""? Is that valid?

See ValidateTransferActionTokenTypes method it checks:

if actionTokenType == "" {
			actionTokenType = inputToken.Type
		} else if actionTokenType != inputToken.Type {
			return errors.Errorf(
				"token type mismatch in transfer action: input [%d] has type [%s] but expected [%s]",
				i, inputToken.Type, actionTokenType,
			)
		}

token/core/common/validator.go
```diff
@@ -134,11 +134,25 @@ func (v *Validator[P, T, TA, IA, DS]) VerifyTokenRequestFromRaw(ctx context.Cont
...
-	for _, sig := range tr.AuditorSignatures {
-		signatures = append(signatures, sig.Signature)
+	auditorSignatures := make([][]byte, 0, len(tr.Signatures))
+	actionSignatures := make([][]byte, 0, len(tr.Signatures))
+	for _, sig := range tr.Signatures {
+		if sig == nil {
+			continue
+		}
+		if sig.Auditor != nil {
+			auditorSignatures = append(auditorSignatures, sig.Auditor.Signature)
```
AkramBitar, line 144: Do we need to check sig.Auditor.Signature if nil before appending?

token/core/common/validator.go
```diff
@@ -134,11 +134,25 @@ func (v *Validator[P, T, TA, IA, DS]) VerifyTokenRequestFromRaw(ctx context.Cont
...
+
+			continue
+		}
+		if sig.Action != nil {
+			actionSignatures = append(actionSignatures, sig.Action.Signature)
```
AkramBitar, line 149: Do we need to check sig.Auditor.Signature if nil before appending?

token/core/zkatdlog/nogh/v1/testutils/support.go
```diff
@@ -0,0 +1,244 @@
...
+// testTokenLoader is a minimal mock TokenLoader for testing
+type testTokenLoader struct {
+	tokens map[string]v1.LoadedToken
+}
+
+func (t *testTokenLoader) LoadTokens(ctx context.Context, ids []*token2.ID) ([]v1.LoadedToken, error) {
+	result := make([]v1.LoadedToken, len(ids))
+	for i, id := range ids {
+		key := id.TxId
+		if id.Index != 0 {
```
AkramBitar: What about = 0?

AkramBitar (changes requested):

docs/driverapi.md
adecaro: Good question. In this paragraph we are talking about how to generate the message to sign. It is important that the message has no ambiguity and cannot be manipulated. ASN1 is the standard in the cryptographic context to get these kind of encodings.

docs/driverapi.md
adecaro: I'm making Protocol V2 the new Protocol V1. We are restarting from zero in preparation of Panurus v1

token/core/common/auditor.go
adecaro: if Inputs is empty, nothing will fail here

token/core/common/auditor.go
adecaro: good catch, I'll simplify the loop

token/core/common/auditor.go
adecaro: good observation, at the current stage it is not forbidden to have an empty type. What counts is that all tokens have the same type

token/core/common/validator.go
adecaro: actually, let me add a validation function to the token request so that we can reject quickly a request that is not well formed.

token/core/common/validator.go
adecaro: covered by my answer to the comment above.

token/core/common/validator.go
adecaro: okay, now, when we unmarshal we do the checks and we are good here.

token/core/zkatdlog/nogh/v1/testutils/support.go
adecaro: good catch. Even though this is mock, better to use the same sematic of the production code.

token/core/common/auditor.go
adecaro: switched to string, indeed!

token/core/common/auditor.go
adecaro: indeed, switched to strings

AkramBitar (approved):
LGTM
Great work @adecaro thanks a million.

PR #1718, opened 2026-05-18
Fix #1635: Add Comprehensive Tests for Auditor Lock Management

### LINKED ISSUE #1635: Mandatory Scoped Lock Ownership via Structured Session Object [HIGH]
**Description**
Refactor the `Audit` → `Append`/`Release` protocol so that lock ownership is encapsulated in a short-lived, scoped object (e.g. an `AuditSession` struct returned by `Audit`) whose finalisation path unconditionally calls `ReleaseLocks`. The session object should carry a context with a hard deadline; when the context fires or the object is garbage-collected (via a `runtime.SetFinalizer` safety net), the release path runs automatically. This removes the reliance on callers to remember to invoke `Release` and makes the release path robust against panics, early returns, and deliberate omission  --  the primary attack vector described in the lock-starvation threats.

**Short Summary**
Prevent lock-leaks. create a mechanism to automatically release locks even if panicked or buggy code. One solution is to have:

```
// Audit returns a session that auto-releases locks
session := auditor.Audit(tx)  // Acquires locks
defer session.Close()         // Guaranteed cleanup

session.Append(data)          // Uses locks
// Locks automatically released when session.Close() runs


type AuditSession struct {
    locks   []Lock
    ctx     context.Context
    cancel  context.CancelFunc
}

func (s *AuditSession) Close() {
    s.cancel()           // Cancel context
    s.ReleaseLocks()     // Free locks
}

// Safety net: if caller forgets Close(), GC triggers cleanup
func init() {
    runtime.SetFinalizer(session, func(s *AuditSession) {
        s.Close()
    })
}

```

AkramBitar:
**Problem**
The Audit() method acquires locks on enrollment IDs but relies on callers to manually call Release(). If Release() is forgotten or code panics before defer Release() is set up, locks are held forever, causing DoS.

**Approach**
Our approach is to ensure Release() is called via defer immediately after successful Audit(). This is the developer and reviewer's responsibility to enforce during code reviews.

To support this:

1) Enhanced Documentation: Added comprehensive documentation to Audit() and Release() methods explaining the required defer Release() pattern with code examples
2) Added 7 Unit Tests: Created tests verifying lock behavior in all scenarios (success, failure, panic, context cancellation) to serve as reference implementations and catch regressions

After reviewing the code, PR #1616 solved context cancellation during lock acquisition using semaphore.Weighted with automatic rollback. This PR ensures locks are released after successful acquisition via the defer pattern.

Tests added in token/services/auditor/auditor_test.go:

Locks acquired only after successful AuditRecord() (no leaks on early errors)
Locks properly released on success
Context cancellation doesn't leak locks (semaphore auto-rollback)
Release() safe to call multiple times or without prior Audit()
Panic recovery: defer Release() executes even when code panics, preventing lock leaks
All 68 tests pass (61 existing + 7 new). Concurrent lock tests already exist at storage layer.

Correct Usage Pattern
inputs, outputs, err := auditor.Audit(ctx, tx)
if err != nil {
    return err  // No locks acquired, safe to return
}
defer auditor.Release(ctx, tx)  // MUST be here, right after error check
// ... subsequent processing ...

### DIFF
--- a/token/services/auditor/auditor.go
+++ b/token/services/auditor/auditor.go
@@ -113,8 +113,22 @@ func (a *Service) Validate(ctx context.Context, request *token.Request) error {
 }
 
 // Audit extracts the list of inputs and outputs from the passed transaction.
-// In addition, the Audit locks the enrollment named ids.
-// Release must be invoked in case
+// In addition, Audit acquires locks on the enrollment IDs involved in the transaction.
+// The caller MUST call Release() to unlock these enrollment IDs after processing.
+//
+// IMPORTANT: The defer Release() statement MUST be placed immediately after checking
+// the error returned by Audit(). This ensures locks are released even if subsequent
+// operations fail. Example:
+//
+//	inputs, outputs, err := auditor.Audit(ctx, tx)
+//	if err != nil {
+//	    return errors.Wrap(err, "audit failed")
+//	}
+//	defer auditor.Release(ctx, tx)
+//
+// Note: The semaphore-based locking mechanism handles context cancellation during
+// lock acquisition (see PR #1616), ensuring proper cleanup in case of timeouts or
+// cancellations.
 func (a *Service) Audit(ctx context.Context, tx Transaction) (*token.InputStream, *token.OutputStream, error) {
 	start := time.Now()
 	logger.DebugfContext(ctx, "audit transaction [%s]....", tx.ID())
@@ -203,6 +217,7 @@ func (a *Service) GetTokenRequest(ctx context.Context, txID string) ([]byte, err
 	return a.auditDB.GetTokenRequest(ctx, txID)
 }
 
+// Check performs a health check on the auditor service and returns any issues found.
 func (a *Service) Check(ctx context.Context) ([]string, error) {
 	return a.checkService.Check(ctx)
 }
@@ -212,22 +227,30 @@ type requestWrapper struct {
 	tms dep.TokenManagementService
 }
 
+// newRequestWrapper creates a new requestWrapper that wraps a token request with its associated
+// token management service for enhanced audit record processing.
 func newRequestWrapper(r *token.Request, tms dep.TokenManagementService) *requestWrapper {
 	return &requestWrapper{r: r, tms: tms}
 }
 
+// ID returns the unique identifier (anchor) of the wrapped token request.
 func (r *requestWrapper) ID() token.RequestAnchor {
 	return r.r.ID()
 }
 
+// Bytes returns the serialized byte representation of the wrapped token request.
 func (r *requestWrapper) Bytes() ([]byte, error) { return r.r.Bytes() }
 
+// AllApplicationMetadata returns all application-specific metadata associated with the token request.
 func (r *requestWrapper) AllApplicationMetadata() map[string][]byte {
 	return r.r.AllApplicationMetadata()
 }
 
+// PublicParamsHash returns the hash of the public parameters used in the token request.
 func (r *requestWrapper) PublicParamsHash() token.PPHash { return r.r.PublicParamsHash() }
 
+// AuditRecord retrieves the audit record for the wrapped token request and completes any
+// inputs with missing enrollment IDs by querying the token vault.
 func (r *requestWrapper) AuditRecord(ctx context.Context) (*token.AuditRecord, error) {
 	record, err := r.r.AuditRecord(ctx)
 	if err != nil {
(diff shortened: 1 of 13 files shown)


integration/token/fungible/views/auditor.go
```diff
@@ -57,9 +57,11 @@ func (a *AuditView) Call(context view.Context) (interface{}, error) {
 	// extract inputs and outputs
 	logger.Debugf("AuditView: audit [%s]", tx.ID())
 	inputs, outputs, err := auditor.Audit(context.Context(), tx)
-	assert.NoError(err, "failed retrieving inputs and outputs")
-	logger.Debugf("AuditView: audit done [%s]", tx.ID())
+	if err != nil {
```
adecaro: what was wrong with the assert?

integration/token/interop/views/auditor.go
```diff
@@ -40,7 +41,9 @@ func (a *AuditView) Call(context view.Context) (interface{}, error) {
 
 	// extract inputs and outputs
 	inputs, outputs, err := auditor.Audit(context.Context(), tx)
-	assert.NoError(err, "failed retrieving inputs and outputs")
+	if err != nil {
```
adecaro: what was wrong with the assert?

adecaro:
Hi @AkramBitar , thanks for pushing this 🙏 
I don't see `AuditSession` in the code. Did you choose a different direction? In that case, I would ask you to update title and Issue. Thanks.

AkramBitar:
> Hi @AkramBitar , thanks for pushing this 🙏 I don't see `AuditSession` in the code. Did you choose a different direction? In that case, I would ask you to update title and Issue. Thanks.

@adecaro thanks a lot for looking at this PR.

The approach is to defer Release() immediately after a successful Audit(), with enforcement handled during development and code reviews.

To support this:

Added clear Audit()/Release() documentation with required defer Release() examples
Added unit tests covering success, failure, panic, and context cancellation scenarios to verify lock behavior and prevent regressions
I updated the issue and the PR accordingly.

adecaro:
Hi @AkramBitar , would you mind updating also the documentation under `docs/services/auditor.md`? Many thanks 🙏

token/services/auditor/auditor.go
```diff
@@ -111,8 +111,22 @@ func (a *Service) Validate(ctx context.Context, request *token.Request) error {
...
+//	inputs, outputs, err := auditor.Audit(ctx, tx)
+//	if err != nil {
+//	    return errors.Wrap(err, "audit failed")
+//	}
+//	defer auditor.Release(ctx, tx)
+//
+// Note: The semaphore-based locking mechanism handles context cancellation during
+// lock acquisition (see PR #1616), ensuring proper cleanup in case of timeouts or
+// cancellations.
 func (a *Service) Audit(ctx context.Context, tx Transaction) (*token.InputStream, *token.OutputStream, error) {
```
adecaro, line 132: could you add also GoDocs to the rest of the functions of the package?

token/services/ttx/auditor.go
```diff
@@ -59,6 +59,22 @@ func (a *Auditor) Validate(tx *Transaction) error {
...
+// operations fail. Example:
+//
+//	inputs, outputs, err := auditor.Audit(ctx, tx)
+//	if err != nil {
+//	    return errors.Wrap(err, "audit failed")
+//	}
+//	defer auditor.Release(ctx, tx)
+//
+// Note: The semaphore-based locking mechanism handles context cancellation during
+// lock acquisition, ensuring proper cleanup in case of timeouts or cancellations.
```
adecaro, line 79: could you add/expand the comments on all functions of this package as well?

adecaro (changes requested):
Hi @AkramBitar , thanks for effort 🙏 
Please, could you address the comments mostly related to expanded documentation? 
Thanks.

token/services/ttx/auditor.go
AkramBitar: Regarding the file [token/services/ttx/auditor.go](https://github.com/hyperledger-labs/fabric-token-sdk/pull/1718/files/f7d3ae6cf279b21d1c944c86da9cd1c8d7a74f65#diff-2ae2cb37ac2c5e33fa7731eda42db9aac46985f0d08046b315d55414c7c99739) added /expand the comments on all functions 

Regarding the ttx package: 
Added comprehensive GoDocs to the most critical and frequently used public APIs in both packages.

AkramBitar:
> Hi @AkramBitar , thanks for effort 🙏 Please, could you address the comments mostly related to expanded documentation? Thanks.

Thanks a lot @adecaro 
Added comprehensive GoDocs to the most critical and frequently used public APIs in both packages.

adecaro:
Hi @AkramBitar , please, can you fix the DCO? Thanks 🙏

adecaro:
Hi @AkramBitar , should the Issue be updated as well? The PR and Issue look disconnected.

AkramBitar:
> Hi @AkramBitar , should the Issue be updated as well? The PR and Issue look disconnected.

Hi @adecaro I updated the issue to reflect changes that we did under this PR.

AkramBitar:
> Hi @AkramBitar , please, can you fix the DCO? Thanks 🙏

Done. Thanks a lot.

adecaro (approved):
LGTM

PR #1722, opened 2026-05-21
feat(recovery): promote NotFound orphans to dedicated Orphan TxStatus

### LINKED ISSUE #1723: recovery: TxStatus.Deleted overloads ledger-rejected and never-landed transactions
## Problem

`TxStatus.Deleted` currently represents two semantically distinct outcomes that callers cannot tell apart without inspecting the free-text message column:

1. **Ledger-rejected / invalid** — set by the finality listener when the token request hash check fails (`token/services/ttx/finality/listener.go:138`, `token/services/ttx/finality/recovery.go:122`) or when the network returns `Invalid` (`listener.go:148`, `recovery.go:133`). The transaction was actively rejected; there is nothing to retry.
2. **Never reached the ledger (orphan)** — set by the recovery loop when `GetTransactionStatus` keeps returning `NotFound` past `NotFoundGracePeriod` (`token/services/storage/recovery/manager.go:331`, introduced in #1708). The transaction's audit log was persisted but the broadcast / commit never landed; in principle the request could be replayed.

These two cases require very different operator responses, but in storage they are indistinguishable beyond the message string.

## Why this matters

Without a dedicated state for the orphan case:

- Operational dashboards / queries cannot count or surface "broadcast failures" separately from "rejected transactions".
- Any future second-chance workflow (manual replay, automatic re-broadcast) has no clean column to filter on — it would have to parse the message column to decide which `Deleted` rows are candidates.
- The downstream consumers that act on `Deleted` (token-lock cleanup, in-memory locker, finality status checks) currently apply the same termination semantics to both cases, even though an orphan with retry potential should keep its input locks held until the retry window expires.

## Context from review thread

This was raised by @adecaro during review of #1715:

> I would also probe your opinion on introducing a new state for this transactions that are kind of orphaned. This would allows us to find them more easily and give them a second chance.
>
> — @adecaro, [comment](https://github.com/hyperledger-labs/fabric-token-sdk/pull/1715#issuecomment-4498027921)

I confirmed the intended scope:

> @adecaro to make sure I'm reading you right — is the proposal to change the `SetStatus(ctx, txID, storage.Deleted, ...)` call I added in #1708 (recovery/manager.go:331, on the NotFoundGracePeriod path) so it sets a new dedicated Orphan state instead? That way Deleted stays reserved for explicit operator/user deletion, and "tried to commit, never landed on chain within the grace window" gets its own unambiguous signal.
>
> — @EvanYan1024

And @adecaro agreed to land it as a follow-up PR:

> Yes, this is what I meant. We can introduce this new state in a new PR. That's also fine.

This issue tracks that follow-up.

## Proposed change

Add `Orphan` as a dedicated `TxStatus` value alongside `Deleted`, and route the recovery loop's NotFound-past-grace path to it. Keep `Deleted` reserved for ledger-rejected / hash-mismatch cases.

Implementation in #1722.

## Out of scope

- The retry / rebroadcast workflow itself (promoting `Orphan` back to `Pending` or to a dedicated `Retried` state). That requires its own design discussion and is meaningful only once the `Orphan` value exists.
- Behaviour changes to the downstream `Deleted`-checking call sites (token-lock cleanup, in-memory locker, `db/common/checks.go`, `ttx/finality.go`). #1722 keeps those identical so the state introduction is strictly additive; consumer-side semantics for `Orphan` will be tuned in follow-up PRs as the second-chance workflow takes shape.
- The residual race in the recovery `SetStatus` write described in `recovery/manager.go:318-324` — replacing the unconditional update with an atomic `status=Pending → Orphan` CAS at the SQL layer is a separate hardening step.

EvanYan1024:
Closes #1723.

Follow-up to #1715 (review thread).

Per @adecaro's suggestion there:

> I would also probe your opinion on introducing a new state for this
> transactions that are kind of orphaned. This would allows us to find them
> more easily and give them a second chance.

And the agreed plan:

> Yes, this is what I meant. We can introduce this new state in a new PR.

This PR is that new PR — just the state split, no retry mechanism yet.

## Summary

`TxStatus.Deleted` currently overloads two semantically distinct outcomes:

1. **Invalid** — token request hash mismatch (`finality/listener.go:138`,
   `finality/recovery.go:122`) or `network.Invalid` (`listener.go:148`,
   `recovery.go:133`): the ledger or local validation actively rejected the tx.
2. **Orphan** — `recovery/manager.go:331` (added in #1708): the tx never reached
   the ledger; the recovery loop observed `NotFound` past `NotFoundGracePeriod`
   and promoted the row to keep the sweep queue moving.

Operators today can only tell the two apart by parsing the free-text message
column, and any future "second-chance" workflow (re-broadcast, manual replay)
would have no clean handle to filter on.

This PR adds a dedicated `Orphan` value to `TxStatus` (=4, after `Deleted`)
and routes the recovery loop's NotFound-past-grace path to it. `Deleted`
remains reserved for explicit ledger rejection / validation failure.

## What changes

- **`token/driver/vault.go`**: new `Orphan` constant after `Deleted`.
- **`token/services/storage/db/driver/common.go`**: `TxStatusMessage` map gets
  `Orphan: "Orphan"`.
- **Re-exports**: `token/vault.go`, `token/services/storage/storage.go`,
  `token/services/storage/ttxdb/store.go`,
  `token/services/storage/auditdb/store.go`,
  `token/services/ttx/status.go`,
  `token/services/auditor/auditor.go` — each adds the one-line `Orphan = …`
  alias next to the existing `Deleted` alias.
- **`token/services/storage/recovery/manager.go`**: the grace-period path now
  calls `SetStatus(ctx, txID, storage.Orphan, …)` instead of `storage.Deleted`;
  the local `markedDeleted` flag is renamed to `markedOrphan`, log/release
  messages updated, and the `Storage.SetStatus` doc comment refreshed.
- **`token/services/storage/recovery/config.go`**: `NotFoundGracePeriod` doc
  comment updated to reflect the new target status.

## "Second chance" — why this PR keeps lock-cleanup behaviour identical

A retryable orphan still needs its input tokens locked: if the locks were
released the moment we marked the row Orphan, a subsequent retry would have
no inputs to spend. So every other `Deleted`-checking call site is **left
unchanged** in this PR:

| Site | Why it stays unchanged |
|------|------------------------|
| `db/sql/common/tokenlock.go:106`, `db/sql/postgres/tokenlock.go:80` | Status-based lock cleanup filters on `Deleted`. Orphan locks fall through to the existing `OlderThan(created_at, leaseExpiry)` timeout path — they expire on the lease clock, not immediately, preserving the retry window. |
| `selector/simple/inmemory/locker.go:178,226` | Same reasoning at the in-memory layer. |
| `db/common/checks.go:151-157` | Health-check cross-references local `Deleted` against ledger `network.Invalid`. An Orphan-vs-ledger case belongs with the operator workflow that this PR defers. |
| `ttx/finality.go:178,220` | `Finality()` returns `ErrFinalityInvalidTransaction` for `Deleted`. For Orphan it should not — the tx is not invalid, it just never landed. Letting `Finality()` time out naturally is the correct behaviour pending a retry mechanism. |

Net effect: existing consumers that filter on `Deleted` keep behaving
identically; only the recovery loop's output column changes.

## Out of scope (deferred follow-ups)

- The retry / rebroadcast workflow itself (promoting `Orphan` back to `Pending`,
  or to a separate `Retried` state).
- An admin API surface for orphan counts / inspection.
- The residual race fix mentioned in `manager.go:318-324` — replacing the
  unconditional `SetStatus` with an atomic `status=Pending → Orphan` CAS at
  the SQL layer.

Each of these wants its own design discussion and is meaningful only once the
`Orphan` value itself exists in the type system.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt -l token/` clean
- [x] `go test ./token/services/storage/recovery/...` — adds
  `TestManager_PromoteOrphanOnNotFoundPastGracePeriod` (asserts the recovery
  loop calls `SetStatus(_, _, storage.Orphan, _)` on the grace-period path) and
  `TestManager_NoPromotionWhenGracePeriodDisabled` (asserts
  `NotFoundGracePeriod=0` short-circuits the promotion).
- [x] `go test ./token/services/storage/...` — sqlite, postgres (~53s), ttxdb,
  auditdb, tokendb, walletdb all green.
- [x] `go test ./token/services/ttx/finality/... ./token/services/auditor/...
  ./token/services/selector/...` — green; verifies the intentionally-unchanged
  call sites still behave the same.
- [ ] CI green

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -197,6 +197,18 @@ token:
               # This helps with debugging and tracking which instance processed which transactions.
               instanceID:
 
+              # notFoundGracePeriod is the time after which the recovery loop promotes a
+              # transaction whose status query keeps returning NotFound to the terminal
+              # Orphan status. Without this, a transaction whose audit log was persisted
+              # but whose broadcast never reached the ledger would sit at the head of the
+              # `ORDER BY stored_at ASC LIMIT batchSize` claim query forever and prevent
+              # newer rows from being scanned. Default: 30m.
+              # The promoted row is marked Orphan (not Deleted) so operators can tell
+              # broadcast failures apart from ledger-rejected transactions.
+              # Set to 0 to disable the promotion; the row stays Pending and is re-claimed
+              # on every sweep until it either resolves or an operator intervenes.
+              notFoundGracePeriod: 30m
+
       # sections dedicated to the definition of the wallets
       wallets:
         # Default cache size reference that can be used by any wallet that supports caching.
@@ -354,6 +366,7 @@ token:
               leaseDuration: 30s
               advisoryLockID: 8389190333894887286
               instanceID:
+              notFoundGracePeriod: 30m
 ```
 
 Default values:
@@ -366,6 +379,7 @@ Default values:
 - leaseDuration: 30s
 - advisoryLockID: 8389190333894887286 (`0x74746b7265636f76`)
 - instanceID: empty, auto-generated when the recovery manager starts
+- notFoundGracePeriod: 30m (set to 0 to disable promotion to Orphan)
 
 **Parameter Relationships and Tuning:**
 
@@ -374,6 +388,7 @@ Default values:
 - **The manager validates** that `ttl`, `scanInterval`, `batchSize`, `workerCount`, and `leaseDuration` are all greater than zero.
 - **`advisoryLockID`** is used to acquire PostgreSQL advisory-lock leadership so that only one replica performs a recovery sweep at a time. The default value (8389190333894887286 or 0x74746b7265636f76) represents the ASCII string "ttkrecov" (Token Transaction Recovery) encoded as a 64-bit integer.
 - **`instanceID`** is used as the lease owner identifier for claimed transactions; if omitted, the manager generates a unique UUID automatically at startup.
+- **`notFoundGracePeriod`** controls how long a transaction whose status query keeps returning `NotFound` is left in `Pending` before being promoted to the terminal `Orphan` status. This protects the recovery sweep from being permanently blocked by transactions whose audit log was persisted but whose broadcast never reached the ledger. The promoted row is marked `Orphan` rather than `Deleted` so operators can distinguish broadcast failures from ledger-rejected transactions. Raise the default if your network has long catch-up windows after committer/orderer restarts; set it to `0` to disable the promotion entirely.
 
 **Tuning Recommendations:**
 
--- a/docs/services/recovery.md
+++ b/docs/services/recovery.md
@@ -34,8 +34,9 @@ The Handler interface defines how individual transactions are recovered. The TTX
 
 The Storage interface abstracts database operations needed for recovery:
 - `AcquireRecoveryLeadership`: Obtains distributed lock for leader election
-- `ClaimPendingTransactions`: Atomically claims a batch of transactions
+- `ClaimPendingTransactions`: Atomically claims a batch of pending transactions, returning a lightweight `RecoveryClaim` (`TxID` + `StoredAt`) for each row — the recovery loop only needs these two fields, so the SQL projection is kept narrow
 - `ReleaseRecoveryClaim`: Releases claim after processing
+- `SetStatus`: Promotes a transaction to a terminal status. Used by the recovery loop to mark `NotFound`-past-grace-period rows as `Orphan` so they exit the eligible scan range without being conflated with ledger-rejected transactions (`Deleted`)
 
 ## Database Support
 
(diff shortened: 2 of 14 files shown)


adecaro:
Hi @EvanYan1024 , thanks a lot for following up 🙏 

Please, could you also update the documentation:
- `docs/configuration.md`: update the configuration with a reference to `notFoundGracePeriod`.
- `docs/services/storage.md`: update "Recovery Process" section accordingly to changes from this and the previous PRs.

Thanks a lot 🙏

EvanYan1024:
> Hi @EvanYan1024 , thanks a lot for following up 🙏
> 
> Please, could you also update the documentation:
> 
> * `docs/configuration.md`: update the configuration with a reference to `notFoundGracePeriod`.
> * `docs/services/storage.md`: update "Recovery Process" section accordingly to changes from this and the previous PRs.
> 
> Thanks a lot 🙏

Thanks for the review, pushed docs updates: configuration.md (notFoundGracePeriod), storage.md (Recovery
  Process), recovery.md (Storage interface, config sample, lifecycle, error handling). Also extended recovery.md beyond your literal ask — happy to trim if too much.

adecaro (approved):
LGTM

PR #1726, opened 2026-05-21
check token is still unspent after locking

### LINKED ISSUE #1632: Payload Integrity Validation Before Lock Persistence [HIGH]
**Description**
Before the SDK persists any lock record, it must validate that the referenced `(tx_id, token_index)` pair corresponds to a real, unspent token that exists in the token store and is owned by the requesting wallet. Lock requests referencing non-existent, already-spent, or out-of-range token IDs must be rejected immediately. This eliminates the ability to inflate the lock table with phantom entries and ensures that every lock record has a legitimate, bounded backing asset.

**Short Summary**
When adding a lock to TokenLocks table, we don't check that the token exists and is unspent. This can lead to phantom locks and resource exhaustion.

HayimShaul:
When dealing with a transaction a token is checked for being unspent and then is locked for the transaction.
This PR checks the token is still unspent after the lock to catch a case where the token was spent in the time window between checking it was unspent and locking it.
Addresses issue #1632

### DIFF
--- a/integration/token/fungible/views/transfer.go
+++ b/integration/token/fungible/views/transfer.go
@@ -284,6 +284,7 @@ func (t *TransferWithSelectorView) Call(context view.Context) (any, error) {
 				if t.Retry {
 					time.Sleep(10 * time.Second)
 				}
+
 				continue
 			}
 
@@ -293,6 +294,7 @@ func (t *TransferWithSelectorView) Call(context view.Context) (any, error) {
 			// If an error occurs and retry has been asked, then wait first a bit
 			if err != nil && t.Retry {
 				time.Sleep(10 * time.Second)
+				
 				continue
 			}
 
--- a/integration/token/fungible/views/transfer.go
+++ b/integration/token/fungible/views/transfer.go
@@ -290,11 +290,11 @@ func (t *TransferWithSelectorView) Call(context view.Context) (any, error) {
 
 			// Select the request amount of tokens of the given type
 			ids, sum, err = selector.Select(context.Context(), ttx.GetWallet(context, t.Wallet), amount.Decimal(), t.Type)
-			
+
 			// If an error occurs and retry has been asked, then wait first a bit
 			if err != nil && t.Retry {
 				time.Sleep(10 * time.Second)
-				
+
 				continue
 			}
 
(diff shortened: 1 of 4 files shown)


adecaro:
isn't it easier to add a 

```sql
FOREIGN KEY (tx_id, idx) REFERENCES <Tokens Table Name> (tx_id, idx)
```

to the `TokenLocks` table?

HayimShaul:
> isn't it easier to add a
> 
> ```sql
> FOREIGN KEY (tx_id, idx) REFERENCES <Tokens Table Name> (tx_id, idx)
> ```
> 
> to the `TokenLocks` table?

I don't understand how this solves the race condition. Let's discuss

adecaro:
Hi @HayimShaul ,

To the best of my understanding, the issue says:
```
Before the SDK persists any lock record, it must validate that the referenced (tx_id, token_index) pair corresponds to a real, unspent token that exists in the token store and is owned by the requesting wallet. Lock requests referencing non-existent, already-spent, or out-of-range token IDs must be rejected immediately. 
```

With a `FOREIGN KEY` we make sure that only tokens that exist in the other table can be locked. 
Checking in the code that the token exist after the locking does not guarantee that it will not be deleted right after.
No?

HayimShaul:
> Hi @HayimShaul ,
> 
> To the best of my understanding, the issue says:
> 
> ```
> Before the SDK persists any lock record, it must validate that the referenced (tx_id, token_index) pair corresponds to a real, unspent token that exists in the token store and is owned by the requesting wallet. Lock requests referencing non-existent, already-spent, or out-of-range token IDs must be rejected immediately. 
> ```
> 
> With a `FOREIGN KEY` we make sure that only tokens that exist in the other table can be locked. Checking in the code that the token exist after the locking does not guarantee that it will not be deleted right after. No?

The token is already checked for existing and being unspent before locking.
There may be a race between locking the token and spending it so we need to check the token is unspent again after locking

adecaro:
> > Hi @HayimShaul ,
> > To the best of my understanding, the issue says:
> > ```
> > Before the SDK persists any lock record, it must validate that the referenced (tx_id, token_index) pair corresponds to a real, unspent token that exists in the token store and is owned by the requesting wallet. Lock requests referencing non-existent, already-spent, or out-of-range token IDs must be rejected immediately. 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > With a `FOREIGN KEY` we make sure that only tokens that exist in the other table can be locked. Checking in the code that the token exist after the locking does not guarantee that it will not be deleted right after. No?
> 
> The token is already checked for existing and being unspent before locking. There may be a race between locking the token and spending it so we need to check the token is unspent again after locking

but the spending event can happen at any time, even after the check after the locking, no?
There is something missing here, I feel.

PR #1729, opened 2026-05-22
feat(auditor): extract Locker interface with pluggable distributed locking

### LINKED ISSUE #1651: feat(auditor): distributed locking
The auditor must acquire locks on the Enrollment IDs involved in a given transaction to make sure its decision are taken with the respect of total order of transactions. The auditor must impose this total order because at time of auditing the transaction hasn't reached yet the ordering service.

The current code, performs locking in memory limiting the possibility to distribute the load of the auditor to multiple replicas.

This Issue calls for the implementation of a fast and reliable distributed locking mechanism that allows multiple replicas of the same auditor entity to process transactions in parallel. The setting I'm considering here is one in which each replica of share the same auditing secret key.

SuyashAlphaC:
Closes #1651

  Extracts the enrollment-ID locking out of StoreService into a pluggable Locker interface so the auditor can scale to multiple replicas.

  What's in here:

  • Locker interface with AcquireLocks / ReleaseLocks / AssertLocksHeld
  • memoryLocker — drop-in replacement for the existing sync.Map behaviour, zero config change needed
  • postgresLocker — lease-table backed distributed locker for multi-replica setups. Single INSERT...ON CONFLICT upsert for atomic acquisition, background heartbeat renewal, TTL-based expiry
  • AssertLocksHeld check before DB writes in Append to catch expired leases
  • Config-driven backend selection under auditor.locker (defaults to memory, fully backward-compatible)
  • Unit tests for both locker implementations

  Based on the design plan discussed and approved in #1651.


token/services/storage/auditdb/locker_postgres.go
```diff
@@ -0,0 +1,311 @@
...
+				logger.Warnf("eid lease heartbeat failed for [%s]: %v", anchor, err)
+
+				return
+			}
+		}
+	}
+}
+
+func (p *postgresLocker) renewLeases(ctx context.Context, anchor string, expected int) error {
+	query := fmt.Sprintf( //nolint:gosec // table name from config, not user input
```
adecaro: please, use the query builder as used in the storage service

token/services/storage/auditdb/locker_postgres.go
```diff
@@ -0,0 +1,311 @@
...
+	}
+	if c.Owner == "" {
+		c.Owner = generateOwnerID()
+	}
+
+	return c
+}
+
+// postgresLocker implements Locker using a Postgres lease table.
+type postgresLocker struct {
```
adecaro: is there any postgres specific features used here? Would it work with any SQL Db?

token/services/storage/auditdb/store.go
```diff
@@ -43,8 +44,36 @@ type tokenRequest interface {
...
+			cfg, err := cp.ConfigurationFor(tmsID.Network, tmsID.Channel, tmsID.Namespace)
+			if err != nil {
+				return nil, err
+			}
+			store, err := drivers.NewAuditTransaction(fsccommon.GetPersistenceName(cfg, "auditdb.persistence"), tmsID.Network, tmsID.Channel, tmsID.Namespace)
+			if err != nil {
+				return nil, err
+			}
+			lockerCfg := DefaultLockerConfig()
+			_ = cfg.UnmarshalKey("auditor.locker", &lockerCfg)
```
adecaro, line 67: we need to update the documentation about this additional configuration

adecaro:
Hi @SuyashAlphaC , great effort. I think we are in the right direction. A few iterations and I find we should be good. Thanks much 🙏

token/services/storage/auditdb/locker_postgres.go
SuyashAlphaC: Yes, there are a few Postgres-specific features in the acquire and renew paths:

  • **TIMESTAMPTZ** column type in the schema DDL
  • **ON CONFLICT (eid) DO UPDATE … WHERE … RETURNING eid** for atomic single-query
    acquisition
  • **NOW() + $1::interval** casts for lease expiry computation

  These don't have direct equivalents in SQLite or standard SQL. That said, the simpler
  queries (releaseAnchor, AssertLocksHeld) now go through the project's query builder +
  ConditionInterpreter, so those are portable.

The Postgres-specific bits are naturally built around a shared network-accessible database for distributed locking — SQLite won't work across multiple replicas, so Postgres is used here. If we ever need another SQL backend for distributed locking we could abstract the dialect-specific bits behind an interface.

token/services/storage/auditdb/store.go
SuyashAlphaC: Yeah @adecaro 
I have added them

adecaro:
Hi @SuyashAlphaC , I few more comments to improve the PR:
- move the memory locker implementation under `locker/memory`
- move the postgres locker implementation under `locker/postgres`
- regarding the replica id, you can use https://github.com/hyperledger-labs/fabric-smart-client/blob/1f3de7f2277517800c1af6c4d0067879186cd1ea/platform/view/services/config/provider.go `ID() string`. Define a local interface with that function and resolve the dependency via DIG. Notice that you can add more arguments to `NewStoreServiceManager` and resolve the dependencies via DIG as necessary. For example, you can pass the factory there.
- Instead of getting the table name from the storage service, I would rather enhance `AuditTransactionStore` with a method that get in input a table name and returns the prefixed table name following the rules of target store. It looks to me more generic then.
- Please, augment the documentation of the auditor service too `docs/services/auditor.md`.
- The query builder has been now moved under the token-sdk and it will be removed soon from the smart client. So, please, update the imports to point to the local query builder. In doing, so, please, enhance it to make sure we can write all the queries with the query builder.

Many comments, but this PR is getting super interesting. Thanks a lot @SuyashAlphaC 🙏

SuyashAlphaC:
Thanks @adecaro — all six points are in e9b92bff (latest b9301f35):

  • lockers under locker/memory and locker/postgres
  • replica owner from FSC config ID() via ReplicaIDProvider + DIG
  • PrefixedTableName("eid_leases") on the audit store
  • docs updated in auditor.md (+ config section)
  • locker DML on the local query builder (DDL still raw SQL — no builder support there)

  CI’s green on my side. 🙏

token/services/storage/auditdb/locker/dedup/dedup.go
```diff
@@ -0,0 +1,20 @@
...
+
+package dedup
+
+import (
+	"slices"
+
+	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/collections"
+)
+
+func AndSort(source []string) []string {
```
adecaro, line 23: Please, add a comment and a unit-test.

token/services/storage/auditdb/locker/memory/memory.go
```diff
@@ -0,0 +1,69 @@
...
+// Suitable for single-replica deployments.
+type Locker struct {
+	locks sync.Map
+}
+
+func New() *Locker {
+	return &Locker{}
+}
+
+func (m *Locker) AcquireLocks(ctx context.Context, anchor string, eIDs ...string) error {
```
adecaro, line 41: Please, add comments to the functions to describe the actual implementation of the expected behavior.

token/services/storage/auditdb/locker/postgres/postgres.go
```diff
@@ -0,0 +1,288 @@
...
+		);
+		CREATE INDEX IF NOT EXISTS %s_anchor_idx     ON %s (anchor);
+		CREATE INDEX IF NOT EXISTS %s_expires_at_idx ON %s (expires_at);
+	`, p.table, p.table, p.table, p.table, p.table)
+	_, err := p.db.Exec(schema)
+
+	return errors.Wrap(err, "failed to create auditor eid lease table")
+}
+
+func (p *Locker) AcquireLocks(ctx context.Context, anchor string, eIDs ...string) error {
```
adecaro, line 107: same thing here. Please, describe the actual implementation of the expected behavior.

adecaro:
Hi @SuyashAlphaC , fantastic job. Just some additional comments on the functions and we are good to go. Thanks a lot 🙏

SuyashAlphaC:
Hey @adecaro , this PR is really getting interesting even for me. Would love your reviews on the recent changes as per your comments 🙏

adecaro (approved):
Great Job

adecaro:
Great job, @SuyashAlphaC . Now, we need an integration test to make sure everything work as expected. I'll create an Issue describing what is needed 😄

Issue #1620, opened 2026-04-30
TTX recipient responder: bind before send; simplify identity.Provider isMeCache


SuyashAlphaC:
## Background
  Recipient flows use **JSON sessions** and the **endpoint resolver** (`endpoint.Bind`) so FSC identities map to token **owner / pseudonym** identities. The **wallet layer** already orders **`MatchIdentity`** before **`RegisterRecipientIdentity`** and supports optional **rollback** when **`RegisterRecipientData`** fails after **`RegisterRecipientIdentity`** (`RecipientRegistrationRollback`). The concrete **`identity.Provider`** historically kept an **`isMeCache`** separate from the signer cache, but the default construction used a **no-op cache**, so that layer did not actually cache anything.
  ## Problems
  ### 1. Responder ordering: send before bind
  In **respond** views, the code may **send** `RecipientData` to the initiator **before** updating **local** endpoint bindings. If **`Bind`** then **fails**, the **peer** may already have the data while **this node** never finished resolver wiring — an awkward partial state.
  ### 2. Dead “is me” cache machinery
  **`isMeCache`** with a **no-op** implementation does not speed up lookups and adds mental overhead. **Enabling** a real cache without careful rules risks **wrong** results for **single-use / ephemeral** identities (for example **Idemix** nyms). The codebase should either **drop** the unused layer or replace it with a **deliberately designed** cache — not silently carry dead structure.
  ### 3. Exchange responder registration incomplete
  When the **responder** registers the **initiator’s** `RecipientData` from an **exchange** request, **all** relevant fields should be passed through to **`RegisterRecipientIdentity`** (including **`TokenMetadataAuditInfo`** when present) so behaviour matches other paths.
  ### 4. Misleading multisig bind errors
  Multisig follow-up code binds **node** identities to **recipients**; error text should describe that, not “bind me” when it is a loop over nodes.
  ### 5. Documentation drift
  **`docs/services/ttx.md`** sequence diagrams should reflect **bind-before-send** on responder paths where the implementation enforces that order.
  ## Goals / acceptance criteria
  1. **`RespondRequestRecipientIdentityView`**: call **`endpoint.Bind(context.Me(), recipientIdentity)`** (or equivalent) **before** **`session.Send(recipientData)`**; then continue **multisig** handling as today.
  2. **`RespondExchangeRecipientIdentitiesView`**: perform **both** required **`Bind`** calls **before** **`session.Send(recipientData)`**.
[...]

PR #1731, opened 2026-05-22
feat(ttx): recipient wire hardening — slim echo + nonce/signature binding

### LINKED ISSUE #1652: feat(ttx): Milestone 3 — Recipient Wire Hardening (slimmer responses + nonce/signature)
This issue tracks Milestone 3 of the recipient identity hardening roadmap agreed upon during
the review of #1607. Phased plan:

- **M0** — Baseline documentation → merged #1607
- **M1** — Registration safety → merged #1614
- **M2** — Responder hygiene + provider cleanup → merged #1628
- **M3** — Recipient wire hardening → this issue

The goal is to make recipient identity flows safer and more consistent on the wire, without expanding scope into protocol versioning (tracked separately).

---

## Maintainer Direction (per #1620)

Per @adecaro's guidance:
- This issue is structured as an umbrella with sub-tasks.
- Versioning is treated as a **separate exploration** (its own issue) and is intentionally
  out of scope here.
- Backward compatibility is intentionally deferred to that versioning exploration and is
  also out of scope here.
- Wire-level streamlining proceeds independently and is the focus of this umbrella.

---

### 1. Slimmer Wire Responses

Reduce unnecessary full `RecipientData` echo where safe.

`RecipientData` is defined in `token/driver/wallet.go`:
```go
type RecipientData struct {
    Identity               Identity
    AuditInfo              []byte
    TokenMetadata          []byte
    TokenMetadataAuditInfo []byte
}
```

- **Echo path** in `RespondRequestRecipientIdentityView.Call`
  (`token/services/ttx/recipients.go`): when the initiator already supplied a
  `RecipientData`, the responder echoes the full 4-field struct back. The initiator already
  has it — an ack or bare `Identity` is sufficient.
- **Exchange path** in `RespondExchangeRecipientIdentitiesView.Call`: audit which fields are
  actually consumed by the initiator after receive and strip the rest.

### 2. Optional Nonce/Signature Binding

Add freshness/replay resistance and stronger binding between request/response context and
identities.

Currently in `token/services/ttx/recipients.go`, the initiator (`callWithRecipientData`)
accepts `RecipientData` from the responder with no proof of key ownership:
```go
err = session.ReceiveWithTimeout(recipientData, 10*time.Second)
// ...
if err := wm.RegisterRecipientIdentity(context.Context(), recipientData); err != nil { ... }
```
And the responder (`RespondRequestRecipientIdentityView.Call`) sends back with no
attestation:
```go
if err := session.Send(recipientData); err != nil { ... }
```

The same gap exists in `RespondExchangeRecipientIdentitiesView.Call`.

**Proposed approach:** add a `Nonce []byte` to `RecipientRequest`, have the responder sign
`nonce || recipientIdentity` via `tms.SigService()`, and have the initiator verify before
registering.

> Scope (single-recipient / exchange / multisig paths) and struct-shape choice
> (wrapper vs extending `RecipientData`) — pending maintainer input on #1620.

### 3. Tests + Docs

- Malformed/stale message rejection tests.
- Updated Mermaid sequence diagrams in `docs/services/ttx.md`.

---

## Out of Scope (Tracked Separately)

- **Versioning exploration** for ttx interactive protocols — separate issue.
- **Compatibility strategy** for mixed-version peers — output of the versioning exploration.

---

cc @adecaro

SuyashAlphaC:
Closes #1652 (Milestone 3)

  Milestones 1 & 2 are already shipped. This wraps up the third and final one — slimmer responses + nonce/signature binding.

  What's in here:

  • Slim echo path: when the initiator already supplied RecipientData, the responder now sends back just a signature instead of echoing the full object. Saves unnecessary bytes on the wire.
  • Nonce/signature binding: every recipient request carries a 32-byte nonce; the responder signs nonce || identity and the initiator verifies before registering. Prevents identity spoofing through session compromise.
  • Both the single-recipient and exchange flows are covered.
  • Wire round-trip tests, malformed message tests, nonce helper tests.
  • Updated Mermaid diagrams and prose in docs/services/ttx.md.

  Based on the approach discussed in #1652 and #1620.


token/services/ttx/recipients.go
```diff
@@ -814,23 +871,74 @@ func (s *RespondExchangeRecipientIdentitiesView) Call(context view.Context) (int
...
+		}
+
+		return nil, nil
+	}
+
+	signer, err := w.GetSigner(ctx, identity)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to get signer for recipient identity")
+	}
+	sig, err := signer.Sign(buildAttestationMessage(nonce, identity))
```
adecaro: I think we should sign everything that is in the received request + session ID + context ID. Please, build the bytes to be signed by marshalling with asn1 a struct containing all the required fields. We need to avoid extension attacks.

token/services/ttx/nonce.go
```diff
@@ -0,0 +1,17 @@
+/*
```
adecaro: the content of the file and its name do not really match. I would move this function to where it is used. Please, don't forget to unit test the support functions

adecaro:
Super nice work, @SuyashAlphaC . I left a few comments to improve it. Thanks much 🙏

SuyashAlphaC:
@adecaro done 🙏

  - attestation's now an asn1 struct instead of nonce || identity — signs all the request fields + session id +
  context id, so it's bound to the exact request/session/context and no more ambiguous boundaries. both sides
  rebuild the same bytes from the header.
  - moved the builder into recipients.go and killed nonce.go. added tests for it (round-trip, determinism, the
  boundary case, per-field binding). nonce gen was already covered in random_test.go.
  - docs updated to match.

  one q: for RecipientData i only bind Identity, not the whole audit/metadata blob — on the fresh path the
  request has no RecipientData (responder makes it), so binding all of it would break signer/verifier symmetry.
  lmk if you want more in there.

adecaro:
Hi @SuyashAlphaC , I think we are good here. This part of the protocol is now more robust. Thanks a lot for your effort 🙏

adecaro (approved):
LGTM

adecaro:
Hi @SuyashAlphaC , could you identify other parts of the `ttx` package that would require this application level signature?

SuyashAlphaC:
Hey @adecaro, I went digging through the package. the gap is basically anywhere a flow takes a RecipientData off the session and just acts on it without an app-level sig. the two that stood out are the ones that actually move value:

  withdrawal is the obvious one. in RespondRequestWithdrawalView the issuer just registers request.RecipientData and mints to that identity, nothing signed. tamper with the identity on the session and the issuer mints straight to the attacker. and there's no nonce there today either, so it'd need freshness on top of the signature. i'd start here.

  token upgrade is the same idea (issuer issues to RecipientData.Identity), but it's in better shape already, it's got the challenge for freshness and a proof over the tokens. only thing missing is tying the recipient identity to the
  requester's key, so just the attestation, reusing the challenge that's already there.

  lastly the multisig/policy distribution. like you said the components get attested in the single-recipient phase, but the aggregate identity + node bindings that get passed around aren't signed. pretty minor.

  didn't flag the sig_req/signature/endorsement/audit messages since those are signatures already, and tx distribution + spend looked covered by the endorsed tx + on-chain validation.

  all of them can reuse the attestation from #1731. lmk if withdrawal first sounds right and i'll open an issue + PR for it 🙏

PR #1738, opened 2026-05-26
perf(ttx): probe ExternalWalletSigner before GetSigner in collectEndorsements

### LINKED ISSUE #1737: perf(ttx): CollectEndorsementsView checks GetSigner before ExternalWalletSigner, wasting BCCSP work when both are configured
## Problem

In `CollectEndorsementsView.requestSignatures` (`token/services/ttx/collectendorsements.go`),
the per-party loop currently checks for a local signer **before** the
external-wallet-signer case:

```go
// Case: there is a signer locally bound to the party, use it to generate the signature
if signer, err := c.tx.TokenService().SigService().GetSigner(context.Context(), signerIdentity); err == nil {
    // sign locally
    continue
}

// Case: there is a wallet bound to the party but the signer is not local, the signature is generated externally
if w, err := c.tx.TokenService().WalletManager().OwnerWallet(context.Context(), signerIdentity); err == nil {
    ews := c.Opts.ExternalWalletSigner(w.ID())
    if ews == nil {
        return nil, errors.Errorf("no external wallet signer found for [%s][%s]", w.ID(), signerIdentity)
    }
    // sign externally
    continue
}
```

This ordering is suboptimal when callers register an `ExternalWalletSigner`
via `Opts` — i.e. signing is delegated to a component outside the local
BCCSP:

- `GetSigner` runs an x509 parse + BCCSP key load (via
  `DeserializeSigningIdentity`) that is guaranteed to fail because the
  private key isn't in the local BCCSP, but only after paying the
  deserialization cost.
- The wallet+`ExternalWalletSigner` branch is only reached as a fallback
  **after** that failure.
- Repeated per-party for every endorsement signature request.

There is also a second, smaller behavioural wart: if a wallet exists for
the party but no `ExternalWalletSigner` is registered for it, the code
returns `no external wallet signer found` — even though a wallet without
an external signer registered is a perfectly reasonable configuration
(just use the local signer if there is one).

## Observed impact

In our deployment we register an `ExternalWalletSigner` via `Opts` for
x509-based user wallets — the private key is held outside the local
BCCSP and signing is delegated to that external signer. Because the
local BCCSP can't satisfy the lookup, `GetSigner` is guaranteed to
fail; but it only fails after the ASN.1 parse + BCCSP probe cost has
been paid, on every per-party iteration of the endorsement loop.

Pprof captures of the signing node under load showed the doomed
`GetSigner` path **dominating CPU**: cumulative time in
`DeserializeSigner` / `x509.ParseCertificate` reached **~78% at
concurrency ≈ 400** in our deployment. The cost is paid before the
loop falls through to the wallet branch that actually signs.

The cost itself is a function of `DeserializeSigningIdentity` (covered
by #1226), but the *frequency* with which it is hit in the hot path is
purely an ordering problem in `CollectEndorsementsView` and is
resolvable independently.

## Suggested fix

Reorder so the wallet+`ExternalWalletSigner` branch is probed first. When
`WalletManager.OwnerWallet` returns a wallet **and**
`Opts.ExternalWalletSigner(w.ID())` returns a non-nil signer, sign
externally and `continue` without calling `GetSigner`.

When the wallet exists but no `ExternalWalletSigner` is registered for
it, fall through to the local-signer path instead of returning an error.
This treats `ExternalWalletSigner` registration as the explicit opt-in
it already is.

## Scope

- Single file: `token/services/ttx/collectendorsements.go`
- Local-signer and remote-signer paths are unchanged.
- No new dependencies, no API changes.

## Related

#1226 — broader handling-of-signers issue covering the cost of
`DeserializeSigningIdentity` for idemix in any caller; this issue is
narrower and addresses only the case ordering in
`CollectEndorsementsView`, which is independent and resolvable without
touching the identity-layer code.

EvanYan1024:
Closes #1737

## Motivation

In our deployment we register an `ExternalWalletSigner` via `Opts` for
x509-based user wallets — the private key is held outside the local
BCCSP and signing is delegated to that external signer.

When `GetSigner` runs first, `DeserializeSigningIdentity` parses the
x509 certificate and then looks up the matching private key in the
local BCCSP. The lookup is guaranteed to fail because the key isn't
there — but only after the ASN.1 parse + BCCSP probe cost has been
paid. Every per-party iteration in the endorsement loop pays that cost
before finally falling through to the wallet branch that calls the
external signer.

Pprof captures on the signing node under load showed that wasted work
dominating CPU: cumulative time in `DeserializeSigner` /
`x509.ParseCertificate` reached **~78% at concurrency ≈ 400** in our
deployment. Reordering the cases removes the wasted work — the hot
path becomes a wallet lookup + map probe.

The reorder is a pure ordering change; it does not touch the
identity-layer code that drives the underlying cost (which is the
subject of #1226 / #1564). It is independent of those discussions and
benefits any caller that has registered an `ExternalWalletSigner`.

## Summary

In `requestSignatures` (`token/services/ttx/collectendorsements.go`),
the per-party loop currently calls `SigService.GetSigner` first, then
falls back to the `ExternalWalletSigner` branch only if `GetSigner`
returns an error.

This PR reorders the case checks so the `wallet + ExternalWalletSigner`
branch is probed first. When both are present, signing is dispatched
externally in O(1) without touching `GetSigner`. The local-signer and
remote-signer paths are unchanged.

## Behaviour change

When a wallet exists for the party but no `ExternalWalletSigner` is
registered for it, the previous code returned
`no external wallet signer found`; this PR falls through to the
local-signer path instead.

Rationale: `ExternalWalletSigner` registration via `Opts` is an explicit
opt-in, not a strict requirement implied by the presence of a wallet.
A wallet without an external signer registered simply means "use the
local signer if available, otherwise sign remotely".

Callers that register both a wallet and an `ExternalWalletSigner`
consistently are unaffected. Callers that register only a wallet and
rely on the local signer are also unaffected (and previously hit the
error path).

## Scope

Single file, +22/-17 lines. No new dependencies, no API changes, no
changes to the local-signer or remote-signer paths.

## Test plan

- [x] `gofmt` / `go vet` / `go build ./token/services/ttx/...` clean
- [ ] `make checks`
- [ ] `make unit-tests` for `token/services/ttx/`
- [ ] Integration tests (fabtoken + zkatdlog) — case order changed but
      no path was removed


adecaro (approved):
LGTM

PR #1743, opened 2026-05-26
check wallet id when receiving a transaction from the network

### LINKED ISSUE #1631: Semantic Invariant Enforcement at the Store-Service Boundary [HIGH]
**Description**
Before committing any write, the store service should validate that the payload satisfies the structural and relational invariants of the target table: a `TokenRequest` payload must deserialize without error; a `Transaction` row must reference an existing `Requests.tx_id`; a `TokenLock` may only be created for a `(tx_id, idx)` that exists in `Tokens` and is not already held by a different consumer; a `Wallet` row's `identity_hash` must resolve to a registered identity. These checks prevent malformed, duplicate, or logically impossible records from corrupting vault state, audit trails, and balance calculations. Architects should be aware that exhaustive re-validation against large datasets (e.g., checking every existing token entry on each write) can impose significant latency penalties at scale. The recommended approach is to enforce cheap structural checks unconditionally, and to scope relational/cross-table checks to the minimum set of invariants whose violation would cause downstream correctness failures -- deferring expensive consistency sweeps to background reconciliation jobs rather than the hot write path.

**Short Summary*
Table 1: Where Missing Checks Should Be Implemented

        | Validation Type | DB Schema | Application Layer | Recommendation |
        |----------------|-----------|-------------------|----------------|
        | **Unique IDs** | ✅ Implemented | N/A | **Keep in DB** - Primary keys are perfect for this |
        | **Referential Integrity** | ✅ Partial | N/A | **Add missing FKs in DB** - Cheap structural checks |
        | **Required Fields** | ✅ Implemented | N/A | **Keep in DB** - NOT NULL is efficient |
        | **Data Types** | ✅ Implemented | N/A | **Keep in DB** - Type system handles this |
        | **TokenRequest Deserialization** | ❌ Can't do | ✅ **Add here** | **Application Layer** - Requires protobuf parser |
        | **Signature Verification** | ❌ Can't do | ✅ **Add here** | **Application Layer** - Requires crypto library |
        | **Identity Validation** | ❌ Can't do | ✅ **Add here** | **Application Layer** - Requires X.509/key parsing |
        | **TokenLock Existence Check** | ✅ **Add FK** | ✅ **Add check** | **Both** - FK for structure, app for race conditions |
        | **Cross-Table Consistency** | ❌ Too complex | ✅ **Add here** | **Application Layer** - Requires aggregations |
        | **Business Logic** | ❌ Can't express | ✅ **Add here** | **Application Layer** - Requires custom rules |
        | **PP Hash Validation** | ✅ **Add FK** | ✅ **Add check** | **Both** - FK for reference, app for hash matching |

### LINKED ISSUE #1753: Check wallet_id upon receiving a transaction from the network [HIGH]
When a pending transaction is received from the network the wallet_id needs to be check that it is valid.
This addresses one of the check in issue #1631 . Specifically, all pending transactions in the database have a valid wallet id.

HayimShaul:
check wallet id when receiving a transaction from the network

Addresses one check mentioned in #1631

### DIFF
--- a/integration/token/fungible/views/auditor.go
+++ b/integration/token/fungible/views/auditor.go
@@ -28,13 +28,15 @@ type AuditView struct {
 func (a *AuditView) Call(context view.Context) (any, error) {
 	logger.Debugf("AuditView: [%s]", context.ID())
 	tx, err := ttx.ReceiveTransaction(context, TxOpts(a.TMSID, ttx.WithNoTransactionVerification())...)
-
-	assert.NoError(err, "failed receiving transaction")
-	assert.NotNil(tx, "received transaction is nil")
+	if err != nil {
+		return nil, errors.Wrap(err, "failed receiving transaction")
+	}
+	if tx == nil {
+		return nil, errors.New("received transaction is nil")
+	}
 	logger.Debugf("AuditView: [%s]", tx.ID())
 
 	w := ttx.MyAuditorWallet(context, ServiceOpts(a.TMSID)...)
-	assert.NotNil(w, "failed getting default auditor wallet")
 	if w == nil {
 		return nil, errors.Errorf("auditor wallet is nil")
 	}
@@ -42,8 +44,12 @@ func (a *AuditView) Call(context view.Context) (any, error) {
 	// Validate
 	logger.Debugf("AuditView: get auditor [%s]", tx.ID())
 	auditor, err := ttx.NewAuditor(context, w)
-	assert.NoError(err, "failed to get auditor instance")
-	assert.NoError(auditor.Validate(tx), "failed auditing verification")
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to get auditor instance")
+	}
+	if err := auditor.Validate(tx); err != nil {
+		return nil, errors.Wrap(err, "failed auditing verification")
+	}
 	logger.Debugf("AuditView: get auditor done [%s]", tx.ID())
 
 	// Check ValidationRecords
--- a/integration/token/fungible/views/auditor.go
+++ b/integration/token/fungible/views/auditor.go
@@ -30,6 +30,7 @@ func (a *AuditView) Call(context view.Context) (any, error) {
 	tx, err := ttx.ReceiveTransaction(context, TxOpts(a.TMSID, ttx.WithNoTransactionVerification())...)
 
 	assert.NoError(err, "failed receiving transaction")
+	assert.NotNil(tx, "received transaction is nil")
 	logger.Debugf("AuditView: [%s]", tx.ID())
 
 	w := ttx.MyAuditorWallet(context, ServiceOpts(a.TMSID)...)
--- a/integration/token/fungible/views/auditor.go
+++ b/integration/token/fungible/views/auditor.go
@@ -163,7 +163,6 @@ func (a *AuditView) Call(context view.Context) (any, error) {
 	}
 
 	kvsInstance := GetKVS(context)
-	assert.NotNil(kvsInstance, "failed getting KVS instance")
 	if kvsInstance == nil {
 		return nil, errors.Errorf("KVS instance is nil")
 	}
--- a/integration/token/fungible/views/auditor.go
+++ b/integration/token/fungible/views/auditor.go
@@ -35,6 +35,9 @@ func (a *AuditView) Call(context view.Context) (any, error) {
 
 	w := ttx.MyAuditorWallet(context, ServiceOpts(a.TMSID)...)
 	assert.NotNil(w, "failed getting default auditor wallet")
+	if w == nil {
+		return nil, errors.Errorf("auditor wallet is nil")
+	}
 
 	// Validate
 	logger.Debugf("AuditView: get auditor [%s]", tx.ID())
@@ -154,6 +157,10 @@ func (a *AuditView) Call(context view.Context) (any, error) {
 	}
 
 	kvsInstance := GetKVS(context)
+	assert.NotNil(kvsInstance, "failed getting KVS instance")
+	if kvsInstance == nil {
+		return nil, errors.Errorf("KVS instance is nil")
+	}
 
 	for _, rID := range inputs.RevocationHandles() {
 		rh := utils.Hashable(rID).String()
--- a/integration/token/fungible/views/issue.go
+++ b/integration/token/fungible/views/issue.go
@@ -18,9 +18,6 @@ import (
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/assert"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/id"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
-	"github.com/hyperledger-labs/fabric-token-sdk/token"
-	"github.com/hyperledger-labs/fabric-token-sdk/token/services/ttx"
-	token2 "github.com/hyperledger-labs/fabric-token-sdk/token/token"
 )
 
 // IssueCash contains the input information to issue a token
--- a/integration/token/fungible/views/issue.go
+++ b/integration/token/fungible/views/issue.go
@@ -8,6 +8,7 @@ package views
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -20,7 +21,6 @@ import (
 	"github.com/hyperledger-labs/fabric-token-sdk/token"
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/ttx"
 	token2 "github.com/hyperledger-labs/fabric-token-sdk/token/token"
-	"github.com/pkg/errors"
 )
 
 // IssueCash contains the input information to issue a token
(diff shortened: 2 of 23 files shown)


token/services/ttx/transaction.go
```diff
@@ -453,3 +461,73 @@ func (t *Transaction) appendPayload(payload *Payload) error {
 	// }
 	// return nil
 }
+
+// validateTransactionWalletIDs validates that all enrollment IDs (wallet IDs) in the transaction
```
adecaro: let's consider this case: Alice transfers to Bob 20 COINS. Alice has only one token with quantity 100. Alice then creates two outputs: One goes to Bob and contains 20, the other one goes to Alice and contains 80. Bob will not be able to match Alice's EID to one of his wallets. No?

token/services/ttx/transaction.go
```diff
@@ -453,3 +461,73 @@ func (t *Transaction) appendPayload(payload *Payload) error {
...
+	// Get wallet service to retrieve registered wallet IDs
+	walletService := tms.WalletManager()
+	if walletService == nil {
+		logger.DebugfContext(ctx, "wallet service not available, skipping wallet ID validation")
+
+		return nil
+	}
+
+	// Get registered owner wallet IDs
+	registeredWallets, err := walletService.OwnerWalletIDs(ctx)
```
adecaro: here you are assuming that the wallet ID corresponding to the EID. Is this always true?

integration/token/fungible/views/auditor.go
```diff
@@ -153,6 +163,10 @@ func (a *AuditView) Call(context view.Context) (any, error) {
 	}
 
 	kvsInstance := GetKVS(context)
+	assert.NotNil(kvsInstance, "failed getting KVS instance")
```
AkramBitar: Redundant check. assert.NotNil() and if kvsInstance == nil validate the same condition. I am wondering why we need both?

integration/token/fungible/views/issue.go
```diff
@@ -52,14 +53,22 @@ func (p *IssueCashView) Call(context view.Context) (any, error) {
 	// Notice that, this step would not be required if the issuer knew already which
 	// identity the recipient wants to use.
 	recipient, err := ttx.RequestRecipientIdentity(context, p.Recipient, ServiceOpts(p.TMSID, ttx.WithRecipientWalletID(p.RecipientWalletID))...)
-	assert.NoError(err, "failed getting recipient identity")
```
AkramBitar, line 55: Why you replaced the assert with return?

integration/token/fungible/views/utils.go
```diff
@@ -68,7 +68,10 @@ func TxOpts(tmsId *token.TMSID, opts ...ttx.TxOption) []ttx.TxOption {
 func GetKVS(sp services.Provider) *kvs.KVS {
 	kvss, err := sp.GetService(&kvs.KVS{})
 	if err != nil {
-		panic(err)
```
AkramBitar, line 71: Why you removed the panic?

token/services/interop/htlc/wallet.go
```diff
@@ -249,18 +273,77 @@ func (w *OwnerWallet) filter(ctx context.Context, tokenType token2.Type, sender
...
 func (w *OwnerWallet) filterIterator(ctx context.Context, tokenType token2.Type, sender bool, selector SelectFunction) (iterators.Iterator[*token2.UnspentToken], error) {
-	var walletID string
+	provider := w.walletIDs
+	if provider == nil {
+		provider = &tokenOwnerWalletIDProvider{wallet: w.wallet}
+	}
+
+	walletIDs := []string{provider.BaseID()}
 	if sender {
-		walletID = senderWallet(ctx, w.wallet)
```
AkramBitar, line 254: why we replaced senderWallet with append?

token/services/interop/htlc/wallet.go
```diff
@@ -249,18 +273,77 @@ func (w *OwnerWallet) filter(ctx context.Context, tokenType token2.Type, sender
...
+		walletIDs = append(walletIDs, provider.SenderID(ctx))
 	} else {
-		walletID = recipientWallet(ctx, w.wallet)
+		walletIDs = append(walletIDs, provider.RecipientID(ctx))
```
AkramBitar, line 285: why we replaced recipientWallet with append?

token/services/interop/htlc/wallet.go
```diff
@@ -249,18 +273,77 @@ func (w *OwnerWallet) filter(ctx context.Context, tokenType token2.Type, sender
...
 	}
-	it, err := w.queryEngine.UnspentTokensIteratorBy(ctx, walletID, tokenType)
-	if err != nil {
-		return nil, errors.WithMessagef(err, "failed to get iterator over unspent tokens")
+
+	// Collect all valid iterators from all wallet IDs
+	var validIterators []driver.UnspentTokensIterator
+	var errs []error
+	for _, walletID := range walletIDs {
+		it, err := w.queryEngine.UnspentTokensIteratorBy(ctx, walletID, tokenType)
```
AkramBitar, line 292: Why we need to get the UnspentTokensIteratorBy?

token/services/ttx/recipients.go
```diff
@@ -219,7 +219,12 @@ func (f *RequestRecipientIdentityView) Call(context view.Context) (any, error) {
 	multiSig := len(f.Recipients) > 1
 	for i, recipient := range f.Recipients {
 		local[i] = true
-		w, err := tms.WalletManager().OwnerWallet(context.Context(), recipient.Identity)
+		// Use wallet ID if provided, otherwise fall back to recipient identity
+		walletID := recipient.WalletID
+		if walletID == "" {
```
AkramBitar, line 224: How this worked before change when walletID == ""?

token/wallet.go
```diff
@@ -102,6 +102,9 @@ func (wm *WalletManager) OwnerWallet(ctx context.Context, id WalletLookupID) (*O
 	if err != nil {
 		return nil, err
 	}
+	if w == nil {
+		return nil, errors.Errorf("wallet not found for id [%s]", id)
+	}
```
AkramBitar, line 107: I am wondering if the caller now need to handle this returned

token/services/ttx/transaction.go
```diff
@@ -459,3 +480,117 @@ func (t *Transaction) appendPayload(payload *Payload) error {
...
+		if eid != "" {
+			txEIDs[eid] = true
+		}
+	}
+
+	// If no enrollment IDs to validate, we're done
+	if len(txEIDs) == 0 {
+		logger.DebugfContext(ctx, "no enrollment IDs to validate in transaction")
+
+		return nil
```
AkramBitar: why returned nil here?

token/services/ttx/transaction.go
```diff
@@ -459,3 +480,117 @@ func (t *Transaction) appendPayload(payload *Payload) error {
...
+
+	for eid := range txEIDs {
+		if _, isLocal := localEIDs[eid]; isLocal {
+			// EID is local and valid (registered)
+			localValidCount++
+			logger.DebugfContext(ctx, "[WalletValidation] EID [%s] is LOCAL and valid", eid)
+		} else {
+			// EID is not in our local set - could be:
+			// 1. A remote party's EID (valid - we don't validate remote parties)
+			// 2. A malformed local EID (invalid - but we can't distinguish without more context)
```
AkramBitar: Ok, so if it is invalid what is the action item that we need to do? only to have a message in the log file?

AkramBitar (changes requested):

integration/token/fungible/views/utils.go
HayimShaul: returning nil allows a graceful error handling instead of crashing the system.
calling code (e.g., auditor.go) checks for nil and returns an appropriate error

token/services/interop/htlc/wallet.go
HayimShaul: The original code was returning immediately after finding the first valid iterator (typically for the base wallet ID like 'bob'), without querying the second wallet ID (like 'htlc.recipientbob') where HTLC locked tokens are actually stored.

The Problem
In HTLC (Hash Time Locked Contract) scenarios, tokens can be stored under multiple wallet IDs:

Base wallet ID (e.g., "bob") - the user's main wallet
HTLC-specific wallet ID (e.g., "htlc.recipientbob" or "htlc.senderbob") - where locked tokens are stored
The old code would query only the first wallet ID and return, missing all tokens in the HTLC-specific wallet.

The Solution
The PR (commit d6c7fd84) changed the logic to:

Build an array of wallet IDs to query:

Always includes the base wallet ID
Adds sender-specific or recipient-specific HTLC wallet ID based on the sender flag
Query ALL wallet IDs and collect valid iterators from each

Chain iterators together using a new chainedIterator type that sequentially iterates through all of them

Graceful fallback: If one wallet ID fails (e.g., doesn't exist yet), it continues with others instead of failing completely

This ensures that HTLC queries return tokens from both the base wallet AND the HTLC-specific wallet, fixing the broken HTLC functionality in both fabtoken and dlog implementations.

The test TestOwnerWalletFilterIteratorFallsBackToBaseWalletID demonstrates this fallback behavior where the HTLC wallet ID fails but tokens are still found in the base wallet.

token/services/interop/htlc/wallet.go
HayimShaul: same answer as above

token/services/interop/htlc/wallet.go
HayimShaul: Tokens can be kept in multiple wallets.
we need to collect unspent tokens from all wallets.

token/services/ttx/recipients.go
HayimShaul: I don't know but this PR failed the tests until I made these fixes

token/wallet.go
HayimShaul: what do you mean?
the function always returned nil,err  when there was an error. This behavior did not change

token/services/ttx/transaction.go
HayimShaul: this function validates all enrollment ids. nil means there's no error. If there are no enrollment ids to validate no need to check anything

token/services/ttx/transaction.go
HayimShaul: We can't distinguish between a remote wallet and an invalid local wallet.
The issue this PR addresses requested to validate that all wallets in database are valid.
Since you can't distinguish between a malformed local and a remote we can only log what we found

AkramBitar: Ok, please add log

HayimShaul:
The issue was closed

PR #1747, opened 2026-05-27
fix: add Chrome installation for linkspector action

AkramBitar:
Linkspector's Puppeteer now requires Chrome, which isn't pre-installed on GitHub runners. Added explicit Chrome installation step to resolve the "Could not find Chrome" error.

Fixes CI/CD pipeline failure.


adecaro (approved):
LGTM

PR #1764, opened 2026-06-02
perf(ttx): index transactions.stored_at for time-ordered pagination

### LINKED ISSUE #1763: Transaction store: transactions table missing stored_at index for time-ordered pagination
### Description
The SQL transaction store (`token/services/storage/db/sql/common/transactions.go`)
creates the `transactions` table with only `idx_tx_id`. Listing transactions
paginated by time uses `ORDER BY stored_at` (descending) + offset/limit; without a
`stored_at` index the planner falls back to a full scan + sort.

### Impact
On a ~9.3M-row table in our deployment, a time-ordered pagination query ran ~3.3s
(full-table Sort), dropping to ~5.6ms (Index Scan Backward) after adding a
`( stored_at DESC )` index.

### Suggested fix
The sibling `movements` table already indexes `( enrollment_id, stored_at )`. Add the
matching `( stored_at DESC )` index on `transactions` in `GetSchema()`.

### Environment
fabric-token-sdk `main`; PostgreSQL.

EvanYan1024:
## What
Add a `( stored_at DESC )` index on the `transactions` table in the SQL transaction store schema.

## Why
The `transactions` table is queried with `ORDER BY stored_at` (descending) + offset/limit for
time-ordered pagination, but it only carried `idx_tx_id`. Without an index on `stored_at` the
planner does a full scan + sort. On a ~9.3M-row table in our deployment that pagination query
ran ~3.3s, dropping to ~5.6ms (Index Scan Backward) once the index was added.

The sibling `movements` table already indexes `( enrollment_id, stored_at )`; this adds the
matching `( stored_at DESC )` index on `transactions`.

## Testing
`go build` + `go vet` on the package. The DDL uses `CREATE INDEX IF NOT EXISTS`, so it is
idempotent and safe on existing deployments (the schema is re-run on startup).

Closes #1763

### DIFF
--- a/integration/token/fungible/support.go
+++ b/integration/token/fungible/support.go
@@ -189,16 +189,27 @@ func CheckAuditedTransactions(network *integration.Infrastructure, auditor *toke
 	for i, tx := range txs {
 		fmt.Printf("tx %d: %+v\n", i, tx)
 		fmt.Printf("expected %d: %+v\n", i, expected[i])
-		txExpected := expected[i]
-		gomega.Expect(tx.TokenType).To(gomega.Equal(txExpected.TokenType), "tx [%d][%s] expected token type [%v], got [%v]", i, tx.TxID, txExpected.TokenType, tx.TokenType)
-		gomega.Expect(strings.HasPrefix(tx.SenderEID, txExpected.SenderEID)).To(gomega.BeTrue(), "tx [%d][%s] expected sender [%v], got [%v]", i, tx.TxID, txExpected.SenderEID, tx.SenderEID)
-		gomega.Expect(strings.HasPrefix(tx.RecipientEID, txExpected.RecipientEID)).To(gomega.BeTrue(), "tx [%d][%s] tx.RecipientEID: %s, txExpected.RecipientEID: %s", i, tx.TxID, tx.RecipientEID, txExpected.RecipientEID)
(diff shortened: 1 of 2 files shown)


EvanYan1024:
The failing `itest`s here seem to trace back to the new `stored_at` index, but I think the root cause is an assumption in the test helper rather than the index itself. I pushed a tentative fix in `bceb1ec7` — I'd appreciate a sanity check on whether this is the right way to handle it.

**What seems to be going on**

`CheckAuditedTransactions` compares the auditor's records against the expected slice by strict position. But as far as I can tell, all records from a single token request share one `stored_at` value — `Transactions()` / `Movements()` in `token/services/storage/ttxdb/store.go` stamp every record of a request with the same timestamp. When `stored_at` ties, the row order seems undefined and depends on the scan path the planner picks.

Without a `stored_at` index, the scan happened to return tied rows in insertion order, matching the expected slice. The new `CREATE INDEX ... (stored_at DESC)` lets the planner use that index and return tied rows in a different order, e.g.:

```
expected[i]:  Transfer alice -> bob    USD 111
actual[i]:    Transfer alice -> alice  USD 9     (change output)
```

both stamped `…39.398809996`, just swapped.

**What I tried**

I changed `CheckAuditedTransactions` to group records by `stored_at` and match each group as an unordered set, while still asserting ordering across different timestamps. I leaned this way rather than the `CheckNext` / `CheckPrevious` pattern `CheckAcceptedTransactions` uses, since it handles groups of any size without manual marking — but I'm honestly not sure which you'd prefer.

Does this seem like a reasonable approach, or would you rather keep the index change and handle the ordering some other way? Happy to adjust.

adecaro (approved):
LGTM, thanks @EvanYan1024 🙏

PR #1770, opened 2026-06-03
Fix blocking channel receives without timeout or context cancellation

### LINKED ISSUE #1767: Blocking Channel Receives Without Timeout or Context Cancellation
This document identifies all places in the codebase where a channel receive operation (`<-ch`) is performed without a timeout or context cancellation mechanism, which could lead to indefinite blocking.

We need to fix this.

### Production Code

1. **File:** `token/services/selector/simple/inmemory/locker.go`
   **Line:** 73
   **Code:** `<-d.scanDone`
   **Context:** In the `Stop()` method, waits for the scan goroutine to signal completion via `scanDone` channel.
   **Risk:** If the scan goroutine fails to exit (e.g., due to a bug), `Stop()` will block indefinitely.
   **Fix:** Add a timeout or context to the `Stop()` method. Example:
   ```go
   func (d *locker) Stop() {
       d.stopOnce.Do(func() {
           d.cancel()
           select {
           case <-d.scanDone:
           case <-time.After(5 * time.Second):
               logger.Warnf("timeout waiting for scan goroutine to stop")
           }
       })
   }
   ```

2. **File:** `token/services/selector/sherdlock/manager.go`
   **Line:** 102
   **Code:** `<-m.cleanerDone`
   **Context:** In the `Stop()` method, waits for the cleaner goroutine to signal completion via `cleanerDone` channel.
   **Risk:** If the cleaner goroutine fails to exit, `Stop()` will block indefinitely.
   **Fix:** Similar to locker.go, add a timeout or context to the `Stop()` method.

3. **File:** `token/services/ttx/boolpolicy/spend.go`
   **Line:** 153
   **Code:** `a := <-answerChannel`
   **Context:** In a loop waiting for answers from other parties in a boolean policy spend view.
   **Risk:** If a party fails to send an answer (e.g., network issue, bug), the loop will block indefinitely.
   **Fix:** Use a timeout on the receive. Since the function already has a `WithTimeout` method in the multisig version, we can add similar functionality here or pass a context.
   **Example fix:**
   ```go
   for range counter {
       select {
       case a := <-answerChannel:
           // process answer
       case <-time.After(10 * time.Second):
           return nil, errors.New("timeout waiting for answers")
       }
   }
   ```

4. **File:** `token/services/ttx/multisig/spend.go`
   **Line:** 151
   **Code:** `a := <-answerChannel`
   **Context:** Similar to boolpolicy/spend.go, but in multisig spend view.
   **Note:** This file already has a `WithTimeout` method and a TODO comment: `// TODO: put a timeout`.
   **Fix:** Implement the timeout as indicated by the TODO. Use the existing `c.timeout` field.
   **Example fix:**
   ```go
   for range counter {
       if c.timeout > 0 {
           select {
           case a := <-answerChannel:
               // process answer
           case <-time.After(c.timeout):
               return nil, errors.New("timeout waiting for answer")
           }
       } else {
           a := <-answerChannel
           // process answer
       }
   }
   ```

5. **File:** `token/services/network/fabricx/finality/queue/queue.go`
   **Line:** 237
   **Code:** `<-done`
   **Context:** In the `Shutdown` method, waits for workers to finish when `timeout == 0`.
   **Risk:** When timeout is zero, the method blocks indefinitely waiting for workers.
   **Fix:** Even when timeout is zero, consider using a non-zero default or context. However, note that the function signature may imply that zero means "wait forever". If so, document this behavior and ensure callers are aware.
   **Alternative:** Change the behavior to use a reasonable default timeout when zero is passed, or require a positive timeout.

6. **File:** `cmd/token_validation_service/server/main.go`
   **Line:** 45
   **Code:** `<-ch`
   **Context:** Waits for an OS signal (SIGINT) to shut down the server.
   **Risk:** This is intended behavior for a server to wait until termination signal. It is acceptable in this context because the process should live until signaled.
   **Fix:** No fix needed. This is a correct use of a blocking receive.

### Test Code

7. **File:** `integration/token/fungible/tests.go`
   **Line:** 712
   **Code:** `err := <-transfer`
   **Context:** In a test, waiting for an error from a transfer operation.
   **Risk:** If the transfer operation never completes (e.g., due to a bug), the test will hang indefinitely.
   **Fix:** Add a timeout to the test using `time.After` or use the testing framework's timeout mechanisms.

8. **File:** `integration/token/fungible/tests.go`
   **Line:** 757
   **Code:** `errs = append(errs, <-transfer)`
   **Context:** Similar to above, in a loop waiting for multiple transfer errors.
   **Fix:** Same as above, add a timeout per receive.

9. **File:** `integration/token/common/views/finality.go`
   **Line:** 56
   **Code:** `if err := <-errs; err != nil {`
   **Context:** In a view that waits for the first of two error channels.
   **Risk:** If neither error channel is ever written to, the view will block indefinitely.
   **Fix:** Add a timeout to the receive, especially since this is in a test integration context.

10. **File:** `integration/token/common/views/finality.go`
    **Line:** 60
    **Code:** `return nil, <-errs`
    **Context:** Waits for the second error channel.
    **Fix:** Same as above, add a timeout.

11. **File:** `integration/nwo/txgen/service/runner/base_runner.go`
    **Line:** 66
    **Code:** `<-r.done`
    **Context:** In the `Stop` method, waits for the runner to signal completion via `done` channel.
    **Risk:** If the runner fails to shut down, `Stop()` will block indefinitely.
    **Fix:** Add a timeout to the `Stop` method, similar to the fixes for locker.go and manager.go.

## Summary of Fixes

For each blocking receive without timeout or context:

1. **Add a timeout:** Use `select` with `time.After` to limit the blocking duration.
2. **Use context:** If the function already accepts a context, use `ctx.Done()` in the select.
3. **Document assumptions:** If blocking indefinitely is intentional (like waiting for a signal), document it clearly.
4. **In tests:** Use test timeouts or contextual timeouts to prevent test suite hangs.

## Recommendation

- For production code, prefer adding timeouts or context cancellation to avoid indefinite blocking.
- For test code, ensure that all blocking operations have a timeout to prevent test suite hangs.
- Review each case to determine the appropriate timeout value based on the expected operation duration.

adecaro:
## Description

This PR fixes all 10 instances of blocking channel receive operations that could cause deadlocks by adding appropriate timeout or context cancellation mechanisms.

Resolves #1767

### Approach
- **Context-based cancellation** (preferred): Used in 3 files where `view.Context` is available
- **Named timeout constants**: Added in 5 files with descriptive comments
- All timeout values are now configurable constants instead of magic numbers

### Production Code (5 fixes)
1. `token/services/selector/simple/inmemory/locker.go` - Added `stopTimeout = 5s` constant
2. `token/services/selector/sherdlock/manager.go` - Added `stopTimeout = 5s` constant  
3. `token/services/ttx/boolpolicy/spend.go` - Uses `context.Done()` for cancellation
4. `token/services/ttx/multisig/spend.go` - Uses `context.Done()` for cancellation
5. `token/services/network/fabricx/finality/queue/queue.go` - Added `defaultShutdownTimeout = 10s` constant

### Test Code (5 fixes)
6. `integration/token/fungible/tests.go` - Added `transferTimeout = 60s` constant (2 locations)
7. `integration/token/common/views/finality.go` - Uses `context.Done()` for cancellation (2 locations)
8. `integration/nwo/txgen/service/runner/base_runner.go` - Added `shutdownTimeout = 10s` constant

## Testing

- ✅ All code passes `make lint-auto-fix` (0 issues)
- ✅ All code passes `make checks` (license, fmt, vet, staticcheck, etc.)
- ✅ Unit tests run successfully

## Impact

- **Low risk**: Changes are additive (adding cancellation/timeouts to existing code)
- **High impact**: Prevents production deadlocks and improves reliability
- **Backward compatible**: Context cancellation respects existing behavior

### DIFF
--- a/token/services/network/fabricx/finality/queue/queue.go
+++ b/token/services/network/fabricx/finality/queue/queue.go
@@ -20,6 +20,12 @@ import (
 
 var logger = logging.MustGetLogger()
 
+const (
+	// defaultShutdownTimeout is the default maximum time to wait for workers to finish during shutdown
+	// when no explicit timeout is provided. This prevents indefinite blocking if workers fail to stop.
+	defaultShutdownTimeout = 10 * time.Second
+)
+
 var (
 	// ErrQueueClosed is returned when an event is added to a closed queue
 	ErrQueueClosed = errors.New("queue is closed")
@@ -225,17 +231,15 @@ func (eq *EventQueue) Shutdown(timeout time.Duration) error {
 			close(done)
 		}()
 
-		if timeout > 0 {
-			select {
-			case <-done:
-				logger.Info("All workers shut down gracefully")
-			case <-time.After(timeout):
-				shutdownErr = ErrShutdownTimeout
-				logger.Warnf("Shutdown timeout exceeded")
-			}
-		} else {
-			<-done
+		if timeout == 0 {
+			timeout = defaultShutdownTimeout
+		}
+		select {
+		case <-done:
 			logger.Info("All workers shut down gracefully")
+		case <-time.After(timeout):
+			shutdownErr = ErrShutdownTimeout
+			logger.Warnf("Shutdown timeout exceeded")
 		}
 	})
 
--- a/token/services/selector/sherdlock/manager.go
+++ b/token/services/selector/sherdlock/manager.go
@@ -20,8 +20,10 @@ import (
 const (
 	// stopTimeout is the maximum time to wait for the cleaner goroutine to stop during shutdown.
 	// This prevents indefinite blocking if the goroutine fails to exit cleanly.
-	stopTimeout = 5 * time.Second
+	stopTimeout = 10 * time.Second
 )
+
+var ErrTimeout = errors.New("timeout occurred")
 
 type Manager struct {
 	selectorCache          lazy2.Provider[transaction.ID, TokenSelectorUnlocker]
@@ -102,14 +104,18 @@ func (m *Manager) cleaner(ctx context.Context) {
 }
 
 // Stop cancels the cleaner goroutine and waits for it to exit.
-func (m *Manager) Stop() {
+func (m *Manager) Stop() error {
+	var err error
 	m.stopOnce.Do(func() {
 		m.cancel()
 		select {
 		case <-m.cleanerDone:
 			logger.Debugf("cleaner goroutine stopped successfully")
 		case <-time.After(stopTimeout):
+			err = ErrTimeout
 			logger.Warnf("cleaner goroutine did not stop within timeout")
 		}
 	})
+
+	return err
 }
--- a/token/services/selector/sherdlock/manager.go
+++ b/token/services/selector/sherdlock/manager.go
@@ -17,6 +17,12 @@ import (
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/utils/types/transaction"
 )
 
+const (
+	// stopTimeout is the maximum time to wait for the cleaner goroutine to stop during shutdown.
+	// This prevents indefinite blocking if the goroutine fails to exit cleanly.
+	stopTimeout = 5 * time.Second
+)
+
 type Manager struct {
 	selectorCache          lazy2.Provider[transaction.ID, TokenSelectorUnlocker]
 	locker                 Locker
@@ -99,6 +105,11 @@ func (m *Manager) cleaner(ctx context.Context) {
 func (m *Manager) Stop() {
 	m.stopOnce.Do(func() {
 		m.cancel()
-		<-m.cleanerDone
+		select {
+		case <-m.cleanerDone:
+			logger.Debugf("cleaner goroutine stopped successfully")
+		case <-time.After(stopTimeout):
+			logger.Warnf("cleaner goroutine did not stop within timeout")
+		}
 	})
 }
--- a/cmd/token_validation_service/go.mod
+++ b/cmd/token_validation_service/go.mod
@@ -206,7 +206,7 @@ require (
 	github.com/prometheus/common v0.67.5 // indirect
 	github.com/prometheus/procfs v0.20.1 // indirect
 	github.com/quic-go/qpack v0.6.0 // indirect
-	github.com/quic-go/quic-go v0.59.0 // indirect
+	github.com/quic-go/quic-go v0.59.1 // indirect
 	github.com/quic-go/webtransport-go v0.10.0 // indirect
 	github.com/remyoudompheng/bigfft v0.0.0-20230129092748-24d4a6f8daec // indirect
 	github.com/rogpeppe/go-internal v1.14.1 // indirect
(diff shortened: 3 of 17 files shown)


token/services/network/fabricx/finality/queue/queue.go
```diff
@@ -20,6 +20,12 @@ import (
 
 var logger = logging.MustGetLogger()
 
+const (
+	// defaultShutdownTimeout is the default maximum time to wait for workers to finish during shutdown
+	// when no explicit timeout is provided. This prevents indefinite blocking if workers fail to stop.
+	defaultShutdownTimeout = 10 * time.Second
```
AkramBitar, line 26: I am wondering if we need to have this in config that can be change based on different deployments and workloads.

token/services/selector/sherdlock/manager.go
```diff
@@ -17,6 +17,12 @@ import (
 	"github.com/hyperledger-labs/fabric-token-sdk/token/services/utils/types/transaction"
 )
 
+const (
+	// stopTimeout is the maximum time to wait for the cleaner goroutine to stop during shutdown.
+	// This prevents indefinite blocking if the goroutine fails to exit cleanly.
+	stopTimeout = 5 * time.Second
```
AkramBitar: Why we have 5 here and the above 10?

token/services/network/fabricx/finality/queue/queue.go
```diff
@@ -225,17 +231,15 @@ func (eq *EventQueue) Shutdown(timeout time.Duration) error {
...
-			<-done
+		if timeout == 0 {
+			timeout = defaultShutdownTimeout
+		}
+		select {
+		case <-done:
 			logger.Info("All workers shut down gracefully")
+		case <-time.After(timeout):
+			shutdownErr = ErrShutdownTimeout
+			logger.Warnf("Shutdown timeout exceeded")
```
AkramBitar, line 242: Do we have test for this case?

token/services/selector/sherdlock/manager.go
```diff
@@ -99,6 +105,11 @@ func (m *Manager) cleaner(ctx context.Context) {
 func (m *Manager) Stop() {
 	m.stopOnce.Do(func() {
 		m.cancel()
-		<-m.cleanerDone
+		select {
+		case <-m.cleanerDone:
+			logger.Debugf("cleaner goroutine stopped successfully")
+		case <-time.After(stopTimeout):
```
AkramBitar, line 114: Do we have test for this as well?
Same comment for the new added cases of timeout in the code.

AkramBitar (changes requested):

token/services/network/fabricx/finality/queue/queue.go
adecaro: fair point. Notice though that Shutdown takes in input the timeout to shutdown. This default value is used only in case 0 is passed to Shutdown.

token/services/network/fabricx/finality/queue/queue.go
adecaro: yes, this part of the code is covered by the unit-tests in the package.

token/services/selector/sherdlock/manager.go
adecaro: I have updated code and tests. It is covered now. Thanks for catching this.

adecaro:
Hi @AkramBitar , I have addressed all your comments. Please, have another look. Thanks 🙏

AkramBitar (approved):

PR #1772, opened 2026-06-03
force token exists in schema

### LINKED ISSUE #1632: Payload Integrity Validation Before Lock Persistence [HIGH]

HayimShaul:
Force a correct token id in the schema

### DIFF
--- a/token/services/selector/sherdlock/manager_test.go
+++ b/token/services/selector/sherdlock/manager_test.go
@@ -96,13 +96,17 @@ func createManager(pgConnStr string, backoff time.Duration, maxRetries int) (tes
 		DataSource:   pgConnStr,
 		MaxOpenConns: 10,
 	}), &dbProvider{})
-	lockDB, err := d.NewTokenLock("")
+
+	// Create Token DB first
+	tokenDB, err := d.NewToken("")
 	if err != nil {
 		return nil, err
 	}
-	tokenDB, err := d.NewToken("")
+
+	// Create Lock DB after
+	lockDB, err := d.NewTokenLock("")
 	if err != nil {
-		return nil, errors.Join(err, lockDB.Close())
+		return nil, errors.Join(err, tokenDB.Close())
 	}
 
 	m := NewMetrics(&disabled.Provider{})
--- a/token/services/storage/db/sql/common/tokenlock.go
+++ b/token/services/storage/db/sql/common/tokenlock.go
@@ -25,6 +25,7 @@ import (
 
 type tokenLockTables struct {
 	TokenLocks string
+	Tokens     string
 	Requests   string
 }
 
@@ -52,6 +53,7 @@ func NewTokenLockStore(readDB, writeDB *sql.DB, tables TableNames, ci common3.Co
 		writeDB,
 		tokenLockTables{
 			TokenLocks: tables.TokenLocks,
+			Tokens:     tables.Tokens,
 			Requests:   tables.Requests,
 		},
 		ci), nil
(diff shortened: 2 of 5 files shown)


token/services/storage/db/sql/common/tokenlock.go
```diff
@@ -84,6 +86,10 @@ func (db *TokenLockStore) UnlockByTxID(ctx context.Context, consumerTxID transac
 }
 
 func (db *TokenLockStore) GetSchema() string {
+	// Note: We intentionally do not add a FOREIGN KEY constraint to the tokens table here.
```
AkramBitar: The application logic ensures referential integrity by only locking.
How this is enforced in the application level?
I thought that we need to enforce this by adding FK to PK relationships in the schema level?

AkramBitar (changes requested):

token/services/selector/sherdlock/manager_test.go
```diff
@@ -96,13 +96,17 @@ func createManager(pgConnStr string, backoff time.Duration, maxRetries int) (tes
 		DataSource:   pgConnStr,
 		MaxOpenConns: 10,
 	}), &dbProvider{})
-	lockDB, err := d.NewTokenLock("")
+
+	// Create Token DB first (this creates test_tokens table)
```
AkramBitar: Why the name of the table is test_tokens table? it should be Tokens table.
See https://github.com/hyperledger-labs/fabric-token-sdk/blob/main/docs/imgs/storage_db.png

By the way in the digram you can see that TokenLocks table points to tokens table, however in the code this was not implemented and you fixed this now.

token/services/selector/sherdlock/manager_test.go
```diff
@@ -96,13 +96,17 @@ func createManager(pgConnStr string, backoff time.Duration, maxRetries int) (tes
...
+	tokenDB, err := d.NewToken("")
 	if err != nil {
 		return nil, err
 	}
-	tokenDB, err := d.NewToken("")
+
+	// Create Lock DB after (this creates test_token_locks table with FK to test_tokens)
```
AkramBitar: Same comment as above regarding the table's names.

token/services/selector/sherdlock/manager_test.go
AkramBitar: Or maybe the names is ok since it is test

token/services/storage/db/sql/common/tokenlock.go
```diff
@@ -52,6 +53,7 @@ func NewTokenLockStore(readDB, writeDB *sql.DB, tables TableNames, ci common3.Co
 		writeDB,
```
AkramBitar, line 53: The list of the tables is passed to NewTokenLockStore. I did not see in your changes where this method is called with adding the Tokens table to it.

AkramBitar (changes requested):
Added some comments

token/services/selector/sherdlock/manager_test.go
AkramBitar: Change to // Create Lock DB after Token DB

token/services/storage/db/sql/common/tokenlock.go
HayimShaul: added

AkramBitar (approved):

Issue #1740, opened 2026-05-26
Database Resource Exhaustion (Denial of Service) [HIGH]


AkramBitar:
**Database Resource Exhaustion (Denial of Service)**

**Severity:** HIGH

**Threat:** Compromised node, malicious insider, buggy code, or supply chain attack floods database with requests until system crashes.

**Example:** Hacked participant node sends 10,000 token requests per second. Database accepts all until it runs out of resources and crashes the entire network.

AkramBitar:
# Database Resource Exhaustion - Primarily Analysis and Solution Proposal

## Issue
**Database Resource Exhaustion (Denial of Service) - Severity: HIGH**

Compromised node floods database with 10,000 requests/second. Without limits, database accepts all until crash.

## 1. Parameters to Address Issue

| Parameter | Purpose | Recommended | How It Helps |
|-----------|---------|-------------|--------------|
| `maxOpenConns` | Cap concurrent connections | 50 | Limits total connections, prevents connection exhaustion |
| `maxIdleConns` | Idle connections in pool | 10 | Efficient connection reuse without waste |
| `maxIdleTime` | Idle connection lifetime | 1 hour | Recycles stale connections |
| `statementTimeout` | Kill long queries | 30 seconds | Prevents runaway queries from blocking database |

**How this solves the issue**: Even if attacker sends 10,000 requests/second, only 50 can connect simultaneously, and each query is killed after 30 seconds. Database stays responsive.

## 2. Current vs Needed

| Parameter | Current | Needed |
|-----------|---------|--------|
| `maxOpenConns` | 0 (unlimited) | 50 |
| `maxIdleConns` | 2 | 10 |
| `maxIdleTime` | 1 minute | 1 hour |
| `statementTimeout` | Not set | 30 seconds |

## 3. Current Defaults (Applied When Not in Config)

**FSC**: `fabric-smart-client/.../common/config.go:18-20`
```go
const (
    DefaultMaxIdleConns = 2           // Applied if not configured
    DefaultMaxIdleTime  = time.Minute // Applied if not configured
)
```

**FSC**: `fabric-smart-client/.../postgres/config.go:50-55`
```go
if o.MaxIdleConns == nil {
    o.MaxIdleConns = CopyPtr(DefaultMaxIdleConns) // = 2
}
if o.MaxIdleTime == nil {
    o.MaxIdleTime = CopyPtr(DefaultMaxIdleTime)   // = 1 minute
}
[...]

PR #1781, opened 2026-06-08
enforce database timeouts

### LINKED ISSUE #1633: Enforce Session and Operation Timeouts [HIGH]
**Description**
Every blocking call in the endorsement flow  --  waiting for a signer reply, an auditor signature, or a distribution acknowledgement  --  must be bounded by a hard deadline. Implement per-session and per-phase timeouts (e.g., via `context.WithTimeout` or equivalent FSC session deadline APIs) so that a slow, silent, or malicious remote party cannot hold a goroutine and its associated resources indefinitely. When a deadline expires the session must be torn down and its resources released immediately, preventing accumulation of zombie sessions.

**Short Summary**
Today there is a timeout in the view level and there is no timeout in the lower level (e.g., accessing the DB). Also, check whether all resources are freed on a view timeout.

HayimShaul:
Set a timeout to database queries to protect against database queries that are maliciously designed to take a long time.
Addresses issue #1633

### DIFF
--- a/token/services/storage/db/sql/common/tokens.go
+++ b/token/services/storage/db/sql/common/tokens.go
@@ -137,6 +137,7 @@ func (db *TokenStore) DeleteTokens(ctx context.Context, deletedBy string, ids ..
 		Where(HasTokens("tx_id", "idx", ids...)).
 		Format(db.ci)
 	logging.Debug(logger, query, args)
+
 	if _, err := db.writeDB.ExecContext(ctx, query, args...); err != nil {
 		return errors.Wrapf(err, "error setting tokens to deleted [%v]", ids)
 	}
@@ -1530,6 +1531,7 @@ func (t *TokenTransaction) Delete(ctx context.Context, tokenID token.ID, deleted
 		Format(t.ci)
 
 	logging.Debug(logger, query, args)
+
 	if _, err := t.tx.ExecContext(ctx, query, args...); err != nil {
 		return errors.Wrapf(err, "error setting token to deleted [%s]", tokenID.TxId)
 	}
@@ -1549,6 +1551,7 @@ func (t *TokenTransaction) StoreToken(ctx context.Context, tr driver.TokenRecord
 		OnConflictDoNothing().
 		Format()
 	logging.Debug(logger, query, args)
+
 	if _, err := t.tx.ExecContext(ctx, query, args...); err != nil {
 		logger.Errorf("error storing token [%s] in table [%s] [%s]: [%s][%s]", tr.TxID, t.table.Tokens, query, err, string(debug.Stack()))
 
--- a/token/services/storage/db/sql/common/tokenlock_test_utils.go
+++ b/token/services/storage/db/sql/common/tokenlock_test_utils.go
@@ -7,11 +7,14 @@ SPDX-License-Identifier: Apache-2.0
 package common
 
 import (
+	"context"
 	"database/sql"
 	"testing"
+	"time"
 
 	"github.com/DATA-DOG/go-sqlmock"
 	"github.com/LFDT-Panurus/panurus/token/token"
+	fscerrors "github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"github.com/onsi/gomega"
 )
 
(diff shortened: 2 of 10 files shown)


docs/security/database_timeouts.md
```diff
@@ -0,0 +1,254 @@
...
+| **Medium** | 15 seconds | Standard queries and updates |
+| **Long** | 30 seconds | Batch operations and complex queries |
+
+### Configuration
+
+Timeouts are configured via `DBTimeoutConfig`:
+
+```go
+config := &common.DBTimeoutConfig{
+    ShortOpTimeout:  5 * time.Second,
```
AkramBitar: I am wondering if these values should be defined in the yaml file where the database connection is defined (if not defined then we can take defaults). If we need to add this in the yaml file then we need to check if this should be part of FSC.

docs/security/database_timeouts.md
```diff
@@ -0,0 +1,254 @@
...
+- **Ensures resource cleanup**: Context cancellation releases connections and locks
+- **Enables graceful degradation**: System can detect and respond to DB performance issues
+- **Provides DoS protection**: Limits impact of expensive or malicious queries
+
+## Timeout Configuration
+
+### Default Timeouts
+
+Three timeout tiers are provided based on operation complexity:
+
```
AkramBitar: Why we need to have different timeouts? (maybe only one which is the Long).

docs/security/database_timeouts.md
```diff
@@ -0,0 +1,254 @@
...
+1. **Configurable timeouts**: Load from configuration files
+2. **Adaptive timeouts**: Adjust based on observed DB performance
+3. **Circuit breakers**: Fail fast when DB is consistently slow
+4. **Metrics integration**: Track timeout rates and patterns
+5. **Retry logic**: Automatic retry with backoff for transient failures
+
+## References
+
+- [Context Package Documentation](https://pkg.go.dev/context)
+- [Hyperledger Security Policy](../../SECURITY.md)
```
AkramBitar: where I can find this file and why we need it at all.

token/services/storage/db/sql/common/tokenlock.go
```diff
@@ -62,23 +62,31 @@ func (db *TokenLockStore) CreateSchema() error {
 }
 
 func (db *TokenLockStore) Lock(ctx context.Context, tokenID *token.ID, consumerTxID transaction.ID) error {
+	// Apply short timeout for lock operation to prevent indefinite blocking
+	timeoutCtx, cancel := WithShortTimeout(ctx, nil)
```
AkramBitar: Is the context (ctx) is the same context of the view which has view timeout?
If so, then I am wondering if it is ok to overwrite it here.

token/services/storage/db/sql/common/timeout.go
```diff
@@ -0,0 +1,67 @@
...
+func DefaultDBTimeoutConfig() *DBTimeoutConfig {
+	return &DBTimeoutConfig{
+		ShortOpTimeout:  5 * time.Second,
+		MediumOpTimeout: 15 * time.Second,
+		LongOpTimeout:   30 * time.Second,
+	}
+}
+
+// WithShortTimeout wraps the context with a timeout for short operations
+// Returns the new context and a cancel function that MUST be called to release resources
```
AkramBitar: Does the cancel method that returned close the connections to the database and clean the locks?

token/services/storage/db/sql/common/tokenlock.go
```diff
@@ -62,23 +62,31 @@ func (db *TokenLockStore) CreateSchema() error {
 }
 
 func (db *TokenLockStore) Lock(ctx context.Context, tokenID *token.ID, consumerTxID transaction.ID) error {
```
AkramBitar, line 66: Please see the diagram https://github.com/hyperledger-labs/fabric-token-sdk/blob/main/docs/imgs/storage_db.png I noticed there is identity and wallet database. Did you address queries to this DB as well?

AkramBitar (changes requested):

AkramBitar:
The query timeout issue is tracked also under issue: #1740
This issue suggest to add the query timeout in the database level as a configuration parameter, while other solution can be in the application level like this PR suggests.

The timeout of the query can be defined for PostgreSQL in the database level but for SQLite we do not have this option. 

Need to discuss with @adecaro and @mbrandenburger which approach is better.

```
PostgreSQL:

ALTER ROLE app_user
SET statement_timeout = '30s';

Go:

ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
defer cancel() 

```

AkramBitar:
> The query timeout issue is tracked also under issue: #1740 This issue suggest to add the query timeout in the database level as a configuration parameter, while other solution can be in the application level like this PR suggests.
> 
> The timeout of the query can be defined for PostgreSQL in the database level but for SQLite we do not have this option.
> 
> Need to discuss with @adecaro and @mbrandenburger which approach is better.
> 
> ```
> PostgreSQL:
> 
> ALTER ROLE app_user
> SET statement_timeout = '30s';
> 
> Go:
> 
> ctx, cancel := context.WithTimeout(ctx, 30*time.Second)
> defer cancel() 
> ```

After consulting with @adecaro, 
We need to make sure that token sdk respects the context timeout. 
In case we have a context timeout (or when the context canceled) then the query should be interrupted ASAP.

token/services/storage/db/sql/common/tokens.go
```diff
@@ -116,6 +116,7 @@ func (db *TokenStore) DeleteTokens(ctx context.Context, deletedBy string, ids ..
 		Where(HasTokens("tx_id", "idx", ids...)).
 		Format(db.ci)
 	logging.Debug(logger, query, args)
+
```
AkramBitar, line 140: Could you please revert the original code since we ado not have any changes here and we have only new lines?

AkramBitar:
@HayimShaul 
The context is indeed passed through to ExecContext(ctx, ...)/QueryContext(ctx, ...), so cancellation can propagate to the driver.

Should we add a test that runs a slow blocking query (e.g. pg_sleep/SLEEP or a fake blocking driver) and asserts that QueryContext/ExecContext is actually interrupted and returns context.Canceled or context.DeadlineExceeded instead of hanging?

AkramBitar (approved):
LGTM

PR #1785, opened 2026-06-11
chore!: upgrade fabric-smart-client to v0.12.1 across all modules

mbrandenburger:
This PR updates the FSC dependency to v0.12.1.

Note that FSC v0.12.0 introduces an issue related to NWO configuration generation for fabric-x. This bug got fixed in v0.12.1

The major advantage that comes with this dependency upgrade is the removal of many indirect dependencies.

Moreover, this PR introduces a breaking change for TransferOptions.Attributes / IssueOptions.Attributes and the narrowing of WithTransferAttribute/WithIssueAttribute key types as we switch from `map[any]any` to `map[string]any`. They `any` to `any` mapping was previously dictated by the underlying yaml parser. `go.yaml.in/yaml/v3` comes with `map[string]any` and therefore some additional cleanup here.


integration/nwo/token/fabric/fabric.go
```diff
@@ -298,6 +299,78 @@ func (p *NetworkHandler) GenerateCryptoMaterial(cmGenerator generators.CryptoMat
...
+// Safe to call during extension generation because platform.go uses two sequential
+// phases: all entries are fully populated before any extension is generated.
+func (p *NetworkHandler) getCombinedWallets(tms *topology2.TMS, nodeName string) *topology2.Wallets {
+	var matching []*Entry
+	for _, entry := range p.Entries {
+		if entry.TMS.TmsID() == tms.TmsID() {
+			matching = append(matching, entry)
+		}
+	}
+	if len(matching) == 1 {
```
AkramBitar, line 326: Do we need to check the case where len(matching) = 0?

integration/nwo/token/fabric/fabric.go
```diff
@@ -298,6 +299,78 @@ func (p *NetworkHandler) GenerateCryptoMaterial(cmGenerator generators.CryptoMat
...
+	// Non-transient entries first, then stable sort by alias.
+	sort.Slice(matching, func(i, j int) bool {
+		ti, tj := matching[i].TMS, matching[j].TMS
+		if ti.Transient != tj.Transient {
+			return !ti.Transient
+		}
+
+		return ti.Alias < tj.Alias
+	})
+	combined := &topology2.Wallets{}
```
AkramBitar, line 338: In case len(matching) = 0 or all w := entry.Wallets[nodeName] in the for are nil then in this as we will return  empty &topology2.Wallets{}. I am wondering if that is a valid case?

AkramBitar (approved):
LGTM
I left some comments.

integration/nwo/token/fabric/fabric.go
mbrandenburger: if `len(matching) == 0` the code will just return `&topology2.Wallets{}`, which is reasonable to me.

integration/nwo/token/fabric/fabric.go
mbrandenburger: good catch! if there are no matching wallets that can be merged; we return an empty wallet here. I think this is a reasonable behavior. What do you suggest do return instead?

AkramBitar: I am ok with this a well

PR #1789, opened 2026-06-17
Validate and Verify Audit Information Before Registration

### LINKED ISSUE #1645: Validate and Verify Audit Information Before Registration [HIGH]
**Description**
Implement comprehensive validation of all audit information and recipient data before allowing registration or updates in the wallet. This includes verifying the cryptographic consistency of identity credentials, checking that enrollment IDs and revocation handles conform to expected formats and are properly bound to the identity being registered, and validating that audit metadata is well-formed and contains all required fields. For recipient registration operations, verify that the provided `RecipientData` includes valid cryptographic proofs (such as signatures or zero-knowledge proofs) that demonstrate the caller legitimately controls the identity being registered. Reject any registration attempts with malformed, incomplete, or unverifiable audit information to prevent injection of forged data into the audit trail.

AkramBitar:
## Problem

`RegisterRecipientIdentity` accepted recipient identity data from a counterparty without any structural validation, so nil, empty, or multi-MB blobs reached the debug log (which dereferences `data.Identity`), the deserializer, and the audit trail — a denial-of-service vector.

Fixes #1645

## Solution

A technology-agnostic structural guard in `token/services/identity/wallet`, run before any registration side effect:

1. **Basic structure** (`validateBasicStructure`) — nil/empty checks and length bounds: identity `10 B … 1 MiB`, audit info `≤ 50 KB`. The identity cap is deliberately generous so legitimate identities are never falsely rejected (composite MultiSig/Policy identities aggregate several inner identities, and X509 chains / Idemix proofs run to several KB); it exists only to reject pathological multi-MB blobs.
2. **Cryptographic binding** — the existing `Deserializer.MatchIdentity` (identity ↔ audit info), now guaranteed to run before the `IdentityProvider` is touched.

Earlier revisions of this PR also added driver-specific checks (audit-info JSON well-formedness, typed-identity decode, enrollment ID / revocation handle bounds). Those were **removed per review**: they are technology-specific, belong to the driver deserializers rather than this layer, and duplicated deserialization work already done downstream.

### Scope note: strength of `MatchIdentity` varies by driver

For idemix wallets, `MatchIdentity` is a genuine proof of ownership. For x509 wallets it is weaker: `x509.AuditInfoMatcher.Match` compares the self-reported EID in the audit info against the certificate's CN, and `crypto.GetEnrollmentID` parses that certificate with no chain or signature validation — a self-consistency check, not proof of control. `token/services/ttx/recipients.go` layers a signature-based key-ownership attestation on top of most call sites, but the per-component registration loops skip it for sub-identities. This is a pre-existing limitation outside this PR's diff, which this change does not worsen; it is worth a follow-up issue.

## Testing

- `TestValidateBasicStructure` — table-driven coverage of every bound, including the 1 MiB identity cap.
- `FuzzRegisterRecipientIdentityStructuralGuard` — drives counterparty identity and audit-info bytes through `RegisterRecipientIdentity` with an always-succeeding `IdentityProvider` and `Deserializer`, so the guard is the only source of errors: the call must fail exactly when the lengths fall outside the declared bounds, and must never panic. Wired into `.github/workflows/nightly-fuzz.yml` and the fuzz-target table in `docs/development/testing.md`.

### DIFF
--- a/token/services/identity/wallet/service.go
+++ b/token/services/identity/wallet/service.go
@@ -114,6 +114,15 @@ func (s *Service) RegisterRecipientIdentity(ctx context.Context, data *tdriver.R
 
 	s.Logger.DebugfContext(ctx, "register recipient identity [%s] with audit info [%s]", data.Identity, utils.Hashable(data.AuditInfo))
 
+	// Validate basic structure: nil/empty checks and length bounds as a DoS guard.
+	// Driver-specific validation (audit-info format, typed-identity structure, enrollment
+	// ID / revocation handle) is intentionally not performed here — it is technology-specific
+	// and is already covered downstream by MatchIdentity and the driver's own deserializers.
+	if err := validateBasicStructure(data); err != nil {
+		return errors.Wrap(err, "basic structure validation failed")
+	}
+
+	// Match identity against audit info (cryptographic binding) before any registration.
 	if err := s.Deserializer.MatchIdentity(ctx, data.Identity, data.AuditInfo); err != nil {
 		return errors.Wrapf(err, "failed to match identity to audit information for [%s]:[%s]", data.Identity, utils.Hashable(data.AuditInfo))
 	}
--- a/token/services/identity/wallet/service.go
+++ b/token/services/identity/wallet/service.go
@@ -108,12 +108,22 @@ func (s *Service) GetEIDAndRH(ctx context.Context, identity tdriver.Identity, au
 // If RegisterRecipientData fails after RegisterRecipientIdentity succeeds, the IdentityProvider
 // may implement identity.RecipientRegistrationRollback so partial registration can be undone.
 func (s *Service) RegisterRecipientIdentity(ctx context.Context, data *tdriver.RecipientData) error {
-	if data == nil {
-		return errors.Wrapf(ErrNilRecipientData, "invalid recipient data")
+	// Validate basic structure: nil/empty checks and length bounds as a DoS guard.
+	// This is the single nil guard for this entry point, so it must run before anything
+	// dereferences data. Driver-specific validation (audit-info format, typed-identity
+	// structure, enrollment ID / revocation handle) is intentionally not performed here:
+	// it is technology-specific and belongs to the driver deserializers, which MatchIdentity
+	// dispatches to below. How strong that downstream check is varies by driver — for idemix
+	// it is a cryptographic proof, for x509 only an EID/CommonName consistency check over an
+	// unvalidated certificate — so it is not a proof of ownership for every driver. Call
+	// sites that need one layer it on separately (see ttx.verifyRecipientAttestation).
+	if err := validateBasicStructure(data); err != nil {
+		return errors.Wrap(err, "basic structure validation failed")
 	}
 
 	s.Logger.DebugfContext(ctx, "register recipient identity [%s] with audit info [%s]", data.Identity, utils.Hashable(data.AuditInfo))
 
+	// Match identity against audit info (cryptographic binding) before any registration.
 	if err := s.Deserializer.MatchIdentity(ctx, data.Identity, data.AuditInfo); err != nil {
 		return errors.Wrapf(err, "failed to match identity to audit information for [%s]:[%s]", data.Identity, utils.Hashable(data.AuditInfo))
 	}
--- a/token/services/identity/wallet/service_test.go
+++ b/token/services/identity/wallet/service_test.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/LFDT-Panurus/panurus/token/driver"
 	dmock "github.com/LFDT-Panurus/panurus/token/driver/mock"
+	"github.com/LFDT-Panurus/panurus/token/services/identity"
 	idriver "github.com/LFDT-Panurus/panurus/token/services/identity/driver"
 	"github.com/LFDT-Panurus/panurus/token/services/identity/wallet"
 	wmock "github.com/LFDT-Panurus/panurus/token/services/identity/wallet/mock"
@@ -84,35 +85,45 @@ func TestRegisterRecipientIdentityFailuresAndSuccess(t *testing.T) {
 	d := &dmock.Deserializer{}
 	regSvc := wallet.NewService(&logging.MockLogger{}, ip, d, nil)
 
+	// Create a properly typed identity
+	typedID, err := identity.WrapWithType(driver.X509IdentityType, []byte("raw-identity"))
+	require.NoError(t, err)
+
 	// nil data
-	err := regSvc.RegisterRecipientIdentity(ctx, nil)
+	err = regSvc.RegisterRecipientIdentity(ctx, nil)
 	require.Error(t, err)
 
 	// RegisterRecipientIdentity fails
+	mockVerifier := &dmock.Verifier{}
+	mockVerifier.VerifyReturns(nil)
+	d.GetOwnerVerifierReturns(mockVerifier, nil)
+	d.MatchIdentityReturns(nil)
+	ip.GetEnrollmentIDReturns("alice", nil)
+	ip.GetRevocationHandlerReturns("rh", nil)
 	ip.RegisterRecipientIdentityReturns(errors.New("rri"))
-	err = regSvc.RegisterRecipientIdentity(ctx, &driver.RecipientData{Identity: driver.Identity("id")})
+	err = regSvc.RegisterRecipientIdentity(ctx, &driver.RecipientData{Identity: typedID, AuditInfo: []byte(`{"EID":"alice","RH":"rh"}`)})
 	require.Error(t, err)
 	ip.RegisterRecipientIdentityReturns(nil)
 
 	// MatchIdentity fails
 	d.MatchIdentityReturns(errors.New("mismatch"))
-	err = regSvc.RegisterRecipientIdentity(ctx, &driver.RecipientData{Identity: driver.Identity("id"), AuditInfo: []byte("ai")})
+	err = regSvc.RegisterRecipientIdentity(ctx, &driver.RecipientData{Identity: typedID, AuditInfo: []byte(`{"EID":"alice","RH":"rh"}`)})
 	require.Error(t, err)
 	d.MatchIdentityReturns(nil)
 
 	// RegisterRecipientData fails
 	ip.RegisterRecipientDataReturns(errors.New("rrd"))
-	err = regSvc.RegisterRecipientIdentity(ctx, &driver.RecipientData{Identity: driver.Identity("id"), AuditInfo: []byte("ai")})
+	err = regSvc.RegisterRecipientIdentity(ctx, &driver.RecipientData{Identity: typedID, AuditInfo: []byte(`{"EID":"alice","RH":"rh"}`)})
 	require.Error(t, err)
 	ip.RegisterRecipientDataReturns(nil)
 
 	// success
 	ip.RegisterRecipientIdentityCalls(func(context.Context, driver.Identity) error { return nil })
 	d.MatchIdentityCalls(func(context.Context, driver.Identity, []byte) error { return nil })
-	d.GetOwnerVerifierCalls(func(context.Context, driver.Identity) (driver.Verifier, error) { return &dmock.Verifier{}, nil })
+	d.GetOwnerVerifierCalls(func(context.Context, driver.Identity) (driver.Verifier, error) { return mockVerifier, nil })
 	ip.RegisterRecipientDataCalls(func(context.Context, *driver.RecipientData) error { return nil })
 
-	err = regSvc.RegisterRecipientIdentity(ctx, &driver.RecipientData{Identity: driver.Identity("id"), AuditInfo: []byte("ai")})
+	err = regSvc.RegisterRecipientIdentity(ctx, &driver.RecipientData{Identity: typedID, AuditInfo: []byte(`{"EID":"alice","RH":"rh"}`)})
 	require.NoError(t, err)
 }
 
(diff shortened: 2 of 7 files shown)


adecaro:
I need a bit more time here to reflect because it looks to me that these checks should be done at a different stage and possibly being driver dependent. When we introduce magic constants we are always at risk of setting them wrongly.

token/services/identity/wallet/service.go
```diff
@@ -114,6 +114,37 @@ func (s *Service) RegisterRecipientIdentity(ctx context.Context, data *tdriver.R
 
 	s.Logger.DebugfContext(ctx, "register recipient identity [%s] with audit info [%s]", data.Identity, utils.Hashable(data.AuditInfo))
 
+	// Step 1: Validate basic structure (nil checks, empty checks, length bounds)
+	if err := validateBasicStructure(data); err != nil {
```
adecaro, line 120: this check and those below are token driver dependant. AuditInfo might not be a json structure, or typed identity might not be used. 
The length of the enrollmenet id or revocation handle is certification authority specific. So, I don't think we can apply these checks here.
Please, investigate where these checks fit better.

adecaro (dismissed):
Hi Akram, sorry for this late reply. I left a comment on the overall approach.

token/services/identity/wallet/service.go
```diff
@@ -114,6 +114,37 @@ func (s *Service) RegisterRecipientIdentity(ctx context.Context, data *tdriver.R
...
+		if err := s.validateEnrollmentID(ctx, data); err != nil {
+			return errors.Wrap(err, "enrollment ID validation failed")
+		}
+		if err := s.validateRevocationHandle(ctx, data); err != nil {
+			return errors.Wrap(err, "revocation handle validation failed")
+		}
+	}
+
+	// Step 5: Match identity against audit info (cryptographic binding)
 	if err := s.Deserializer.MatchIdentity(ctx, data.Identity, data.AuditInfo); err != nil {
```
HayimShaul, line 127: s.validateEnrollmentID
s.validateRevocationHandle
and
s.validateRevocationHandle

all deserialize the same thing 3 times. This can be saved

token/services/identity/wallet/validation.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+// validateJSONStructure validates that audit info is valid JSON
+func validateJSONStructure(auditInfo []byte) error {
+	if !json.Valid(auditInfo) {
+		return ErrInvalidJSON
+	}
+
+	return nil
+}
```
HayimShaul: This checks the json is valid but not that it has required fields.
Is that ok?

token/services/identity/wallet/validation_test.go
```diff
@@ -0,0 +1,499 @@
...
+
+	tdriver "github.com/LFDT-Panurus/panurus/token/driver"
+	dmock "github.com/LFDT-Panurus/panurus/token/driver/mock"
+	"github.com/LFDT-Panurus/panurus/token/services/identity"
+	"github.com/LFDT-Panurus/panurus/token/services/identity/wallet"
+	"github.com/LFDT-Panurus/panurus/token/services/logging"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
```
HayimShaul, line 22: is there a test that proves a forged identity/audit-info pair is actually rejected?

token/services/identity/wallet/service.go
AkramBitar: Good point. Agreed, these checks are driver-specific and don't belong in the technology-agnostic wallet layer. I've dropped them; I think also that MatchIdentity and the driver deserializers already reject malformed/forged pairs. Kept only a generic size guard.

token/services/identity/wallet/service.go
AkramBitar: Good catch. Resolved. These checks were removed, so there's no repeated deserialization anymore.

token/services/identity/wallet/validation.go
AkramBitar: Good point — removed. Field-level validation is driver-specific and now left to MatchIdentity.

token/services/identity/wallet/validation_test.go
AkramBitar: Yes — TestRegisterRecipientIdentityForgedPair: forged pair → MatchIdentity rejects it, registration fails, nothing is persisted.

AkramBitar:
@adecaro, @HayimShaul 
Thanks a lot for the review. I answered the comments and update the code accordingly please let me know if you have any other issue.
Thanks a lot,
Akram

Effi-S (approved):

AkramBitar:
@Effi-S 

Could you please have additional review on that PR?

Regards,
Akram

token/services/identity/wallet/service.go
```diff
@@ -108,12 +108,19 @@ func (s *Service) GetEIDAndRH(ctx context.Context, identity tdriver.Identity, au
...
 // may implement identity.RecipientRegistrationRollback so partial registration can be undone.
 func (s *Service) RegisterRecipientIdentity(ctx context.Context, data *tdriver.RecipientData) error {
-	if data == nil {
-		return errors.Wrapf(ErrNilRecipientData, "invalid recipient data")
+	// Validate basic structure: nil/empty checks and length bounds as a DoS guard.
+	// This is the single nil guard for this entry point, so it must run before anything
+	// dereferences data. Driver-specific validation (audit-info format, typed-identity
+	// structure, enrollment ID / revocation handle) is intentionally not performed here
+	// — it is technology-specific and is already covered downstream by MatchIdentity
+	// and the driver's own deserializers.
```
adecaro: **Design note: "MatchIdentity and the driver's own deserializers already reject malformed/forged pairs" doesn't hold for the x509 driver.**

For idemix this is true — `idemix.Deserializer.MatchIdentity` runs a ZK-proof matcher that cryptographically verifies the binding. For x509, `AuditInfoMatcher.Match` (`token/services/identity/x509/deserializer.go:41-51`) just parses the audit info's `EID` field and string-compares it against the `CommonName` of an *unvalidated* certificate (`crypto.GetEnrollmentID` → `PemDecodeCert`, no chain/signature check). A self-signed cert with `CN=alice` plus audit info `{"EID":"alice"}` passes this check even though nothing ties either blob to a certificate the network actually trusts — it's a consistency check between two attacker-suppliable fields, not a proof of ownership.

Not a regression from this PR (the checks it removed here were equally non-cryptographic), and `token/services/ttx/recipients.go`'s `verifyRecipientAttestation` does add a real signature-based ownership proof on top for most call sites — just not all, and not in this file. Worth calling out explicitly in the PR description rather than letting the current comment stand as an unqualified claim across all drivers.

token/services/identity/wallet/service_test.go
```diff
@@ -84,35 +85,45 @@ func TestRegisterRecipientIdentityFailuresAndSuccess(t *testing.T) {
...
+
 	// nil data
-	err := regSvc.RegisterRecipientIdentity(ctx, nil)
+	err = regSvc.RegisterRecipientIdentity(ctx, nil)
 	require.Error(t, err)
 
 	// RegisterRecipientIdentity fails
+	mockVerifier := &dmock.Verifier{}
+	mockVerifier.VerifyReturns(nil)
+	d.GetOwnerVerifierReturns(mockVerifier, nil)
```
adecaro: **Nit: these mock stubs aren't exercised by `RegisterRecipientIdentity`.**

`d.GetOwnerVerifierReturns`, `ip.GetEnrollmentIDReturns`, and `ip.GetRevocationHandlerReturns` (repeated a few more times further down in this file) set up methods that the current `RegisterRecipientIdentity` implementation never calls — it only calls `validateBasicStructure`, `Deserializer.MatchIdentity`, `IdentityProvider.RegisterRecipientIdentity`, and `IdentityProvider.RegisterRecipientData`. Looks like a leftover from an earlier version of the code that did call these before the driver-specific checks were removed. Harmless, but worth cleaning up so the test setup doesn't imply verification steps that don't exist.

adecaro (commented):
## Review: does this fully address #1645?

@AkramBitar

**Approve, with two non-blocking comments — the refactor correctly resolves all three prior review threads, but the PR description still describes checks that were removed, and the "MatchIdentity already covers it" rationale doesn't hold uniformly across drivers.**

Checkout out `feature/comprehensive-recipient-validation` at `da39fc331` and verified:

```
go build ./...                                                    -> clean
go vet ./token/services/identity/wallet/...                       -> clean
go test -count=1 ./token/services/identity/wallet/...             -> ok, 12/12 tests pass
go test -run=^$ -fuzz=FuzzRegisterRecipientIdentityStructuralGuard -fuzztime=20s -> 264k execs, no panics
golangci-lint run ./token/services/identity/wallet/...            -> 0 issues
gofmt -l -s (5 changed .go files)                                 -> clean
```

I also grepped every caller of `wallet.Service.RegisterRecipientIdentity` (all via `token.WalletManager`, in `token/services/ttx/{recipients,withdrawal,upgrade}.go`) and read the concrete `Deserializer.MatchIdentity` implementations for idemix and x509, since the PR's justification for removing the driver-specific checks rests entirely on "MatchIdentity and the driver deserializers already reject malformed/forged pairs" — see the design note below.

`checks`, `utest` (race + regression), the full `itest` matrix, `bench-check`, and `cgo-check` are all green; DCO is signed.

@adecaro's CHANGES_REQUESTED thread and both of @HayimShaul's threads (repeated deserialization; JSON-valid-but-no-required-fields) are addressed in the current diff — not repeating those here.

### What it fixes

| # | Finding | Verified |
|---|---|---|
| 1645.1 | Enrollment ID / revocation handle format checks, bound to identity | Deliberately **not** implemented in the wallet layer per review feedback — correctly identified as driver-specific; delegated to `MatchIdentity` and driver deserializers |
| 1645.2 | Audit metadata well-formed, required fields present | Not implemented — dropped per @HayimShaul's thread; only a DoS-guard length check remains |
| 1645.3 | RecipientData includes valid cryptographic proof the caller controls the identity | Delegated to existing `Deserializer.MatchIdentity` — strong (ZK proof) for idemix, weak (string compare of two self-reported fields, no chain/signature validation) for x509 — see design note below |
| 1645.4 | Reject malformed/incomplete/unverifiable registrations | ✅ nil/empty/oversized inputs rejected before anything else runs (`validateBasicStructure`), confirmed by fuzz + table tests |

Row 4 is solid. Rows 1–3 reflect a scope narrowing that the reviewers themselves drove (checks were provably driver-specific and didn't belong in this layer) — reasonable, but issue #1645's actual ask (cryptographic proof of ownership) is fully met only for idemix by this change alone.

### Non-blocking

- **Stale description:** The PR's top "Solution" section (steps 2–5: JSON well-formedness, typed-identity decode, EID/RH length checks) describes an earlier version of the code — those checks were removed per review and no longer exist at `da39fc331`. The "In short" paragraph further down is accurate. Worth trimming steps 2–5 so the merged history doesn't mislead future readers.
- **Driver-dependent strength:** `x509.AuditInfoMatcher.Match` (`token/services/identity/x509/deserializer.go:41-51`) only compares a self-reported EID from audit info against a CN parsed from an unverified certificate (`crypto.GetEnrollmentID`, no chain/signature validation) — it's a self-consistency check, not proof of ownership. Not a regression (the removed checks were equally non-cryptographic), but it means "MatchIdentity already reject[s] forged pairs" overstates what happens for x509 wallets. `token/services/ttx/recipients.go` does layer a real signature-based key-ownership attestation on top of most call sites, but not all (the per-component registration loops at `recipients.go:636`/`727` skip it for sub-identities), and that's outside this PR's diff.
- **Test nit:** `service_test.go`'s added mock stubs for `GetOwnerVerifier`/`GetEnrollmentID`/`GetRevocationHandler` aren't exercised by `RegisterRecipientIdentity` at all — likely leftovers from an earlier version of the code.

### Recommendation

Good to merge once the PR description is trimmed to match the shipped code (Should-fix). The x509/MatchIdentity gap is a design note worth acknowledging in the description or a follow-up issue, not a blocker — it's a pre-existing limitation this PR doesn't worsen.

token/services/identity/wallet/service.go
AkramBitar: Agreed, and thanks for reading the concrete matchers rather than taking the claim at face value — you're right that it doesn't hold uniformly.

Addressed in both places you suggested:

1. **The comment no longer makes the unqualified claim.** It now says the driver deserializers own this validation, and that how strong the downstream check is varies by driver — a cryptographic proof for idemix, only an EID/CommonName consistency check over an unvalidated certificate for x509 — so it is not a proof of ownership for every driver, with a pointer to `ttx.verifyRecipientAttestation` for call sites that need one.
2. **The PR description** now carries a *Scope note: strength of `MatchIdentity` varies by driver* section spelling out the same thing, including that the per-component registration loops skip the attestation for sub-identities.

I'm leaving the x509 matcher itself alone in this PR, as you suggested — it's pre-existing and this change doesn't worsen it.

token/services/identity/wallet/service_test.go
AkramBitar: Good catch — they were exactly that, leftovers from the revision that decoded the typed identity and checked EID/RH before those checks were dropped.

Removed across all three tests: the `GetOwnerVerifier` stubs (both `Returns` and `Calls` forms), the `GetEnrollmentID` / `GetRevocationHandler` stubs on both the plain and rollback providers, and the orphaned `mockVerifier` setup. I also reworded the `// Create a properly typed identity` comments, which implied this layer validates the identity type — `WrapWithType` is now just a realistic way to clear the 10-byte minimum. The setup left standing is only what the entry point actually calls. 12/12 tests still pass.

AkramBitar:
@adecaro 

Thanks a lot for the review. 
Please let me know if you need additional changes in this PR. 

Regards,
Akram

Effi-S (approved):
LGTM

PR #1790, opened 2026-06-18
verify we have the private key for an identity before signing

### LINKED ISSUE #1638: Implement Ownership Verification and Field Reconstruction [HIGH]
**Description**
When a responder receives a RecipientRequest with non-nil RecipientData, implement strict ownership verification by checking not just that the identity exists in the wallet, but that the wallet possesses the private key material by attempting to retrieve the signer. Never return the requester-supplied RecipientData object directly. Instead, fetch authentic audit info and metadata from the wallet's identity provider and reconstruct a fresh RecipientData object with verified fields. Validate that the TMSID matches expected configuration and reject requests for unauthorized TMS instances. If supplied fields don't match locally stored values or ownership cannot be proven, reject the request immediately. This prevents attackers from forcing responders to claim ownership of identities they don't control.

HayimShaul:
Verify we have the secret key for an id given in Echo Path matches a key in our wallet.
Addresses issue #1638

### DIFF
--- a/token/services/ttx/recipients.go
+++ b/token/services/ttx/recipients.go
@@ -514,12 +514,51 @@ func (s *RespondRequestRecipientIdentityView) Call(context view.Context) (any, e
 	var recipientIdentity view.Identity
 	isEcho := false
 	if recipientRequest.RecipientData != nil {
-		recipientData = recipientRequest.RecipientData
-		recipientIdentity = recipientData.Identity
-		if !w.Contains(context.Context(), recipientIdentity) {
-			return nil, errors.Errorf("cannot find identity [%s] in wallet [%s:%s]", recipientIdentity, wallet, recipientRequest.TMSID)
+		// Echo path: requester supplied an identity, verify strict ownership
+		suppliedIdentity := recipientRequest.RecipientData.Identity
+
+		// Step 1: Verify the identity exists in the wallet
+		if !w.Contains(context.Context(), suppliedIdentity) {
+			return nil, errors.Errorf("cannot find identity [%s] in wallet [%s:%s]", suppliedIdentity, wallet, recipientRequest.TMSID)
 		}
+
+		// Step 2: Verify we possess the private key by attempting to retrieve the signer
+		// This proves ownership, not just visibility
+		if !w.Remote() {
+			_, err = w.GetSigner(context.Context(), suppliedIdentity)
+			if err != nil {
+				return nil, errors.Wrapf(err, "cannot prove ownership of identity [%s]: failed to retrieve signer", suppliedIdentity)
+			}
+		}
+
+		// Step 3: Never return requester-supplied data directly
+		// Fetch authentic audit info and metadata from our own wallet's identity provider
+		auditInfo, err := w.GetAuditInfo(context.Context(), suppliedIdentity)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to retrieve audit info for identity [%s]", suppliedIdentity)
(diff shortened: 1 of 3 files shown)


token/services/ttx/recipients.go
```diff
@@ -496,12 +496,51 @@ func (s *RespondRequestRecipientIdentityView) Call(context view.Context) (any, e
...
+			return nil, errors.Wrapf(err, "failed to retrieve token metadata for identity [%s]", suppliedIdentity)
+		}
+
+		tokenMetadataAuditInfo, err := w.GetTokenMetadataAuditInfo(suppliedIdentity)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to retrieve token metadata audit info for identity [%s]", suppliedIdentity)
+		}
+
+		// Step 4: Reconstruct RecipientData with verified fields from trusted sources
+		recipientData = &RecipientData{
```
AkramBitar: Why you overwrite the original value and recontract the RecipientData?

AkramBitar (changes requested):

AkramBitar:
@adecaro could you please have a look at that PR? I need your help here. Thanks a lot.

token/services/ttx/recipients.go
```diff
@@ -496,12 +496,51 @@ func (s *RespondRequestRecipientIdentityView) Call(context view.Context) (any, e
...
+
+		// Step 1: Verify the identity exists in the wallet
+		if !w.Contains(context.Context(), suppliedIdentity) {
+			return nil, errors.Errorf("cannot find identity [%s] in wallet [%s:%s]", suppliedIdentity, wallet, recipientRequest.TMSID)
+		}
+
+		// Step 2: Verify we possess the private key by attempting to retrieve the signer
+		// This proves ownership, not just visibility
+		if !w.Remote() {
+			_, err = w.GetSigner(context.Context(), suppliedIdentity)
```
adecaro, line 528: please, can you explain why w.Contains does not guarantee that we are also able to sign with this identity? Under which condition does this happen?

token/services/ttx/recipients.go
adecaro: we should not return anything to caller in this case. Indeed, the caller can be an attacker that wants to extract the rest of the information knowing only an identity.

token/services/ttx/recipients.go
HayimShaul: w.Contains queries a wallet-to-identity mapping table — it answers "was this identity ever registered/bound to this wallet?" by checking a DB row written by BindIdentity or RegisterRecipientData. GetSigner is a completely separate question: "can I produce a cryptographic signature right now?" which requires the private key to be locally available and is answered by attempting to reconstruct a live signer via DeserializeSigner against a different store (StoreSignerInfo/the local key store). The two stores are populated by entirely independent code paths, so an identity can be fully enrolled in the wallet — its bytes bound to the wallet ID, its audit info persisted — while the private key never existed on this node. The canonical example is a remote/KMS wallet (IsRemote() == true, meaning userKeySKI == nil for Idemix or sID == nil for X.509) and an Idemix pseudonym registered as a counterparty recipient (where only BindIdentity + RegisterRecipientData were called, never RegisterSigner).

token/services/ttx/recipients.go
HayimShaul: good point

AkramBitar:
@HayimShaul 

Need to check how much performance degradation we have if we call  w.GetSigner?

PR #1794, opened 2026-06-21
implement per-wallet lock quota and rate limit

### LINKED ISSUE #1640: Per-Caller and Per-Wallet Lock Quotas with Rate Limiting [HIGH]
**Description**
Impose a hard upper bound on the number of active lock entries any single caller identity (or wallet) may hold simultaneously. Requests that would exceed this quota must be rejected with an appropriate error rather than written to the store. Complement the quota with a rate limit (e.g., maximum lock-creation requests per second per identity) to prevent burst flooding. These controls should be enforced in the service layer, above the database, so they apply regardless of which storage backend is in use.

**Short Summary**
Currently we assume a rate limiter in the upper layers but we don't have a rate limiter for the view.

HayimShaul:
implement per-wallet lock quota and rate limit.
Addresses issue #1640

### DIFF
--- a/docs/security/selector_resource_limits.md
+++ b/docs/security/selector_resource_limits.md
@@ -1,0 +1,144 @@
+# Selector Resource Limits
+
+## Overview
+
+Token selection acquires a short-lived *lock* on each candidate token so that two
+concurrent transactions do not try to spend the same token. Under load, a single
+wallet can drive a large number of selection/lock requests. Applications that need to
+throttle this — to protect the lock store, to enforce fairness between wallets, or to
+integrate with an existing quota system — can do so by supplying their own `Locker`
+implementation.
+
+Panurus deliberately ships **no built-in rate limiter or quota**. Instead it
+gives you two things:
+
+1. A **wallet-id-aware lock function**. Both selector drivers (simple and sherdlock)
+   pass the wallet id the tokens are being selected for into the `Locker`'s lock
+   function, so a custom `Locker` can apply per-wallet policies.
+2. A **fail-fast contract**, `token.SelectorRateLimited`. When a `Locker` denies a
+   lock by returning an error that wraps this sentinel, the selector aborts the
+   selection immediately and returns the error to the caller instead of retrying.
+
+This keeps the Panurus minimal and lets applications reuse whatever rate-limiting
+infrastructure they already run (for example a Redis-backed limiter shared across
+processes).
+
+## The lock function
+
+Both selector drivers route through a `Locker` whose lock function receives the
+wallet id.
+
+**Simple selector** — `token/services/selector/simple/selector.go`:
+
+```go
+type Locker interface {
+    // Lock locks the token id for the consumer transaction txID on behalf of walletID
+    // (ownerFilter.ID()). Return an error wrapping token.SelectorRateLimited to deny
+    // the lock and make the selection fail fast.
+    Lock(ctx context.Context, id *token.ID, txID string, walletID string, reclaim bool) (string, error)
+    UnlockIDs(ctx context.Context, ids ...*token.ID) []*token.ID
+    UnlockByTxID(ctx context.Context, txID string)
+    IsLocked(id *token.ID) bool
+}
+```
+
+**Sherdlock selector** — the lock store it drives is
+`token/services/storage/db/driver/token.go` `TokenLockStore` (also exposed as
+`sherdlock.Locker`):
+
+```go
+type TokenLockStore interface {
+    common.DBObject
+    // Lock locks tokenID for consumerTxID on behalf of walletID. A custom store may use
+    // walletID to throttle per wallet, returning an error wrapping token.SelectorRateLimited.
+    Lock(ctx context.Context, tokenID *token.ID, consumerTxID transaction.ID, walletID string) error
+    UnlockByTxID(ctx context.Context, consumerTxID transaction.ID) error
+    Cleanup(ctx context.Context, leaseExpiry time.Duration) error
+}
+```
+
+The built-in in-memory locker and the SQL-backed `TokenLockStore` accept `walletID`
+but do not act on it — they apply no rate limiting or quota.
+
+## The fail-fast contract
+
+`token/selector.go` defines:
+
+```go
+// SelectorRateLimited is the contract error a Locker implementation returns (directly
+// or wrapped) to deny a lock for policy reasons such as rate limiting or quota.
+var SelectorRateLimited = errors.New("selection rate limit exceeded")
+```
+
+When your `Locker` returns an error `e` with `errors.Is(e, token.SelectorRateLimited)`,
+the selector:
+
+- stops iterating candidate tokens,
+- releases any tokens it already locked for this request, and
+- returns `e` to the caller.
+
+Any *other* error from the lock function keeps the existing semantics: the token is
+treated as unavailable (e.g. already locked by another transaction) and selection
+continues / retries as before.
+
+## Integrating your own rate limiting
+
+Provide a `Locker` that wraps the SDK's default locker and enforces your policy before
+delegating. Below, a Redis-backed limiter throttles per wallet; the same shape works
+for an in-process limiter, a quota table, etc.
+
+### Simple selector
+
+```go
+import (
+    "context"
+
+    "github.com/LFDT-Panurus/panurus/token"
+    "github.com/LFDT-Panurus/panurus/token/services/selector/simple"
+    tokenapi "github.com/LFDT-Panurus/panurus/token/token"
+    "github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
+)
+
+// rateLimitedLocker decorates the default simple.Locker with per-wallet throttling.
+type rateLimitedLocker struct {
+    simple.Locker            // embeds the SDK's default locker
+    limiter       RedisLimiter // your existing infrastructure
+}
+
+func (l *rateLimitedLocker) Lock(ctx context.Context, id *tokenapi.ID, txID string, walletID string, reclaim bool) (string, error) {
+    if !l.limiter.Allow(ctx, walletID) {
+        return "", errors.Wrapf(token.SelectorRateLimited, "wallet %s throttled", walletID)
+    }
+
+    return l.Locker.Lock(ctx, id, txID, walletID, reclaim)
+}
+```
+
+Wire it in by providing a `simple.LockerProvider` whose `New` returns your decorator
+instead of the default `inmemory.NewLocker`.
+
+### Sherdlock selector
+
+Provide a `TokenLockStore` (via the `tokenlockdb.StoreServiceManager` used by
+`sherdlock.NewService`) whose `Lock` enforces the limit before delegating to the
+SQL-backed store, returning an error wrapping `token.SelectorRateLimited` when a wallet
+is throttled.
+
+## Handling the error
+
+Callers should treat `token.SelectorRateLimited` as a transient, retryable-later
+condition rather than an insufficient-funds error:
+
+```go
+ids, sum, err := selector.Select(ctx, ownerFilter, amount, tokenType)
+if errors.Is(err, token.SelectorRateLimited) {
+    // back off and retry later, shed the request, or surface a 429-style response
+}
+```
+
+## Notes
+
+- Passing an empty `walletID` is valid; a `Locker` that keys its policy on wallet id
+  should treat empty as "no throttling" (the default lockers ignore it entirely).
+- Because the policy lives in your `Locker`, its scope (per process vs shared across a
+  cluster), persistence, and lifecycle are entirely under your control.
--- a/token/selector.go
+++ b/token/selector.go
@@ -25,6 +25,14 @@ var (
 	// SelectorSufficientFundsButConcurrencyIssue is returned when funds are sufficient to cover the request, but
 	// concurrency issues does not make some of the selected tokens available.
 	SelectorSufficientFundsButConcurrencyIssue = errors.New("sufficient funds but concurrency issue")
+	// SelectorRateLimited is the contract error a Locker implementation returns (directly
+	// or wrapped) to deny a lock for policy reasons such as rate limiting or quota.
+	// Both the simple and sherdlock selectors detect it via errors.Is and abort the
+	// selection immediately, returning the error to the caller instead of retrying.
+	// Panurus ships no built-in limiter: applications integrate their own
+	// (e.g. a Redis-backed limiter) by providing a Locker implementation that returns
+	// this error when a request must be throttled.
+	SelectorRateLimited = errors.New("selection rate limit exceeded")
 )
 
 // OwnerFilter tells if a passed identity is recognized
--- a/token/services/selector/sherdlock/selector.go
+++ b/token/services/selector/sherdlock/selector.go
@@ -192,7 +192,11 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter, 
 
 			immediateRetries++
 			tokensLockedByOthersExist = false
-		} else if locked := s.locker.TryLock(ctx, &t.Id); !locked {
+		} else if locked, lockErr := s.locker.TryLock(ctx, &t.Id, owner.ID()); !locked {
+			// A rate-limit denial from the locker is a hard stop: abort instead of retrying.
+			if errors.Is(lockErr, token.SelectorRateLimited) {
+				return nil, nil, immediateRetries, lockErr
+			}
 			s.logger.DebugfContext(ctx, "Tried to lock token [%v], but it was already locked by another process", t)
 			tokensLockedByOthersExist = true
 		} else {
@@ -248,13 +252,13 @@ type locker struct {
 	txID transaction.ID
 }
 
-func (l *locker) TryLock(ctx context.Context, tokenID *token2.ID) bool {
-	err := l.Lock(ctx, tokenID, l.txID)
+func (l *locker) TryLock(ctx context.Context, tokenID *token2.ID, walletID string) (bool, error) {
+	err := l.Lock(ctx, tokenID, l.txID, walletID)
 	if err != nil {
 		logger.DebugfContext(ctx, "failed to lock [%v] for [%s]: [%s]", tokenID, l.txID, err)
 	}
 
-	return err == nil
+	return err == nil, err
 }
 
 func (l *locker) UnlockAll(ctx context.Context) error {
--- a/token/services/selector/sherdlock/selector_test.go
+++ b/token/services/selector/sherdlock/selector_test.go
@@ -11,6 +11,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/LFDT-Panurus/panurus/token"
 	"github.com/LFDT-Panurus/panurus/token/services/selector/sherdlock"
 	"github.com/LFDT-Panurus/panurus/token/services/selector/sherdlock/mocks"
 	token2 "github.com/LFDT-Panurus/panurus/token/token"
@@ -36,7 +37,7 @@ func TestSelectorUnit(t *testing.T) {
 		mockIt.NextReturnsOnCall(1, nil, nil)
 
 		mockFetcher.UnspentTokensIteratorByReturns(mockIt, nil)
-		mockLocker.TryLockReturns(true)
+		mockLocker.TryLockReturns(true, nil)
 
 		tokens, sum, err := s.Select(t.Context(), &unitTestMockOwnerFilter{id: "alice"}, "50", "ABC")
 		require.NoError(t, err)
@@ -103,8 +104,8 @@ func TestStubbornSelectorUnit(t *testing.T) {
 		}
 
 		// Fails first lock attempt, succeeds on second
-		mockLocker.TryLockReturnsOnCall(0, false)
-		mockLocker.TryLockReturnsOnCall(1, true)
+		mockLocker.TryLockReturnsOnCall(0, false, nil)
+		mockLocker.TryLockReturnsOnCall(1, true, nil)
 
 		tokens, sum, err := s.Select(t.Context(), &unitTestMockOwnerFilter{id: "alice"}, "50", "ABC")
 		require.NoError(t, err)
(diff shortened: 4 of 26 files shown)


docs/security/selector_resource_limits.md
```diff
@@ -0,0 +1,233 @@
...
+- Provides smooth rate limiting with predictable behavior
+
+**Behavior**:
+- Each identity has an independent rate limit bucket
+- Requests exceeding the rate limit are immediately rejected with `ErrRateLimitExceeded`
+- Rate limit state is maintained in memory and resets on service restart
+- Empty identity strings bypass rate limiting (for backward compatibility)
+
+**Error Handling**:
+- Selector does NOT retry when rate limit is exceeded
```
AkramBitar: So, the current approach is to return error to the caller? 
In case the caller receives this error what he should do?

token/services/selector/simple/inmemory/locker.go
```diff
@@ -133,9 +179,25 @@ func (d *locker) Lock(ctx context.Context, id *token2.ID, txID string, reclaim b
...
+	}
+
+	logger.DebugfContext(ctx, "locking [%s] for [%s] by identity [%s]", id, txID, identity)
 	now := time.Now()
-	d.locked[k] = &lockEntry{TxID: txID, Created: now, LastAccess: now}
+	d.locked[k] = &lockEntry{TxID: txID, Identity: identity, Created: now, LastAccess: now}
+
+	// Update identity lock count
+	if identity != "" {
+		d.identityLockCount[identity]++
```
AkramBitar: Do we need to increment identityLockCount  under lock?

token/services/selector/simple/inmemory/locker.go
```diff
@@ -133,9 +179,25 @@ func (d *locker) Lock(ctx context.Context, id *token2.ID, txID string, reclaim b
 			return e.TxID, AlreadyLockedError
 		}
 	}
-	logger.DebugfContext(ctx, "locking [%s] for [%s]", id, txID)
+
+	// Check quota if configured and identity provided
+	if d.maxLocksPerIdentity > 0 && identity != "" {
+		currentCount := d.identityLockCount[identity]
```
AkramBitar: Do we need to get from identityLockCount under lock?

token/services/selector/simple/inmemory/locker.go
```diff
@@ -153,12 +215,20 @@ func (d *locker) UnlockIDs(ctx context.Context, ids ...*token2.ID) []*token2.ID
...
+			logger.Warnf("unlocking [%s] hold by no one, skipping", id)
 
 			continue
 		}
 		logger.DebugfContext(ctx, "unlocking [%s] hold by [%s]", id, entry)
 		delete(d.locked, k)
+
+		// Decrement identity lock count
+		if entry.Identity != "" {
+			d.identityLockCount[entry.Identity]--
```
AkramBitar: Do we need to decrement identityLockCount under lock?

token/services/selector/simple/inmemory/locker.go
```diff
@@ -173,6 +243,14 @@ func (d *locker) UnlockByTxID(ctx context.Context, txID string) {
 		if entry.TxID == txID {
 			logger.DebugfContext(ctx, "unlocking [%s] hold by [%s]", id, entry)
 			delete(d.locked, id)
+
+			// Decrement identity lock count
+			if entry.Identity != "" {
+				d.identityLockCount[entry.Identity]--
```
AkramBitar: Do we need to decrement identityLockCount under lock?

token/services/selector/simple/inmemory/locker.go
```diff
@@ -173,6 +243,14 @@ func (d *locker) UnlockByTxID(ctx context.Context, txID string) {
...
+				if d.identityLockCount[entry.Identity] <= 0 {
+					delete(d.identityLockCount, entry.Identity)
```
AkramBitar: Do we need to delete identityLockCount under lock?

token/services/selector/simple/inmemory/locker.go
```diff
@@ -39,35 +39,68 @@ type TXStatusProvider interface {
...
+	if config.RateLimit > 0 {
+		rateLimiter = NewRateLimiter(config.RateLimit, config.RateLimitBurst)
+	}
+
 	r := &locker{
 		ttxdb:                  ttxdb,
 		sleepTimeout:           timeout,
 		lock:                   &sync.RWMutex{},
 		locked:                 map[token2.ID]*lockEntry{},
+		identityLockCount:      map[string]int{},
```
AkramBitar: How big this map can grow? 
What is the criteria to clean up it?

token/services/selector/simple/inmemory/locker.go
```diff
@@ -195,6 +273,14 @@ func (d *locker) reclaim(ctx context.Context, id *token2.ID, txID string) (bool,
 	case ttxdb.Deleted:
 		delete(d.locked, *id)
 
+		// Decrement identity lock count
+		if identity != "" {
+			d.identityLockCount[identity]--
+			if d.identityLockCount[identity] <= 0 {
```
AkramBitar: Do we need to delete identityLockCount under lock?

token/services/selector/simple/inmemory/locker.go
```diff
@@ -258,6 +344,14 @@ func (d *locker) scan(ctx context.Context) {
 			// reclaimed this token and re-locked it for a new transaction.
 			if entry, ok := d.locked[s.id]; ok && entry.TxID == s.txID {
 				delete(d.locked, s.id)
+
+				// Decrement identity lock count
+				if entry.Identity != "" {
+					d.identityLockCount[entry.Identity]--
```
AkramBitar: Do we need to decrement identityLockCount under lock?

token/services/selector/simple/inmemory/locker_quota_test.go
```diff
@@ -0,0 +1,327 @@
...
+*/
+
+package inmemory
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/selector/simple"
+	token2 "github.com/hyperledger-labs/fabric-token-sdk/token/token"
```
AkramBitar: Please make sure that we have the following tests as well:
1) Test cleaning up the map identityLockCount 
2) Councarency tests

token/services/selector/simple/inmemory/ratelimiter.go
```diff
@@ -0,0 +1,153 @@
...
+// allow checks if a request can proceed and consumes a token if so
+func (tb *tokenBucket) allow() bool {
+	tb.mu.Lock()
+	defer tb.mu.Unlock()
+
+	now := time.Now()
+	elapsed := now.Sub(tb.lastRefillTime).Seconds()
+
+	// Refill tokens based on elapsed time
+	tb.tokens += elapsed * tb.refillRate
```
AkramBitar: Not clear to me, could you explain.

token/services/selector/simple/inmemory/ratelimiter.go
```diff
@@ -0,0 +1,153 @@
...
+	if tb.tokens > tb.maxTokens {
+		tb.tokens = tb.maxTokens
+	}
+	tb.lastRefillTime = now
+
+	// Check if we have tokens available
+	if tb.tokens >= 1.0 {
+		tb.tokens -= 1.0
```
AkramBitar: Why -1.0? Could you please explain the algorithm of Allow?

token/services/selector/simple/inmemory/ratelimiter.go
```diff
@@ -0,0 +1,153 @@
...
+	refillRate float64 // tokens per second
+}
+
+// NewRateLimiter creates a new rate limiter with the specified parameters
+// requestsPerSecond: average number of requests allowed per second per identity
+// burstSize: maximum burst of requests allowed (should be >= requestsPerSecond)
+func NewRateLimiter(requestsPerSecond float64, burstSize float64) *RateLimiter {
+	if burstSize < requestsPerSecond {
+		burstSize = requestsPerSecond
+	}
```
AkramBitar: What happens in else? which value burstSize  will have?

token/services/selector/simple/inmemory/ratelimiter.go
```diff
@@ -0,0 +1,153 @@
...
+// Returns true if allowed, false if rate limit exceeded
+func (rl *RateLimiter) Allow(identity string) error {
+	if identity == "" {
+		return errors.New("identity cannot be empty")
+	}
+
+	// Fast path: check if bucket exists
+	rl.mu.RLock()
+	bucket, exists := rl.buckets[identity]
+	rl.mu.RUnlock()
```
AkramBitar: Maybe to add the logic in a method xyz for example, then take the lock and defere it before calling the method.

rl.mu.Lock()
defer rl.mu.Unlock()
xyz (...)

token/services/selector/simple/inmemory/ratelimiter.go
```diff
@@ -0,0 +1,153 @@
...
+
+	// Check if request is allowed
+	if !bucket.allow() {
+		return simple.ErrRateLimitExceeded
+	}
+
+	return nil
+}
+
+// Reset removes the rate limit state for a specific identity
```
AkramBitar: What is the criteria to cleanup and reset? 
When the Rest, or RestAll called?

I want to make sure that in long stability tests this map does not grow with no limit as well.

token/services/selector/simple/inmemory/ratelimiter.go
```diff
@@ -0,0 +1,153 @@
...
+// ResetAll clears all rate limit state
+func (rl *RateLimiter) ResetAll() {
+	rl.mu.Lock()
+	defer rl.mu.Unlock()
+	rl.buckets = make(map[string]*tokenBucket)
+}
+
+// GetStats returns current statistics for an identity (for monitoring/debugging)
+func (rl *RateLimiter) GetStats(identity string) (availableTokens float64, exists bool) {
+	rl.mu.RLock()
```
AkramBitar: When this method is called?

token/services/selector/simple/inmemory/ratelimiter_test.go
```diff
@@ -0,0 +1,265 @@
...
+	"testing"
+	"time"
+
+	"github.com/hyperledger-labs/fabric-token-sdk/token/services/selector/simple"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestRateLimiter_Allow(t *testing.T) {
+	t.Run("allows requests within rate limit", func(t *testing.T) {
```
AkramBitar: Could you please make sure that we have the folloing tests as well:
1) Concurrency 
2) Clean up the data from the map.

token/services/selector/simple/selector.go
```diff
@@ -113,8 +122,16 @@ func (s *selector) selectByID(ctx context.Context, ownerFilter token.OwnerFilter
 				return nil, nil, errors.Wrap(err, "failed to convert quantity")
 			}
 
-			// lock the token
-			if _, lockErr := s.locker.Lock(ctx, &t.Id, s.txID, reclaim); lockErr != nil {
+			// lock the token with identity tracking
+			if _, lockErr := s.locker.LockWithIdentity(ctx, &t.Id, s.txID, id, reclaim); lockErr != nil {
+				// Check if this is a quota or rate limit error - these should not be retried
+				if errors.HasType(lockErr, ErrQuotaExceeded) || errors.HasType(lockErr, ErrRateLimitExceeded) {
+					s.locker.UnlockIDs(ctx, toBeSpent...)
```
AkramBitar: When these locks where taken and why we need to unlock them?

token/services/selector/simple/selector.go
```diff
@@ -113,8 +122,16 @@ func (s *selector) selectByID(ctx context.Context, ownerFilter token.OwnerFilter
...
+					s.locker.UnlockIDs(ctx, toBeCertified...)
+
+					return nil, nil, lockErr
+				}
```
AkramBitar, line 130: Do we have a test that check this logic?

AkramBitar (changes requested):

docs/security/selector_resource_limits.md
HayimShaul: the application should implement some random backoff mechanism. where it waits a period of time and tries again

token/services/selector/simple/inmemory/locker.go
HayimShaul: we are under lock:
	d.lock.Lock()

token/services/selector/simple/inmemory/locker.go
HayimShaul: we are under lock (line 154)

token/services/selector/simple/inmemory/locker.go
HayimShaul: we are under lock (line 208)
	d.lock.Lock()

token/services/selector/simple/inmemory/locker.go
HayimShaul: we are under lock (line 238)
	d.lock.Lock()

token/services/selector/simple/inmemory/ratelimiter.go
HayimShaul: Reset and ResetAll are called when testing.
I added a cleanup to the ratelimiter. a bucket is removed when the identity did not make any transactions for 10 minutes (configurable)

token/services/selector/simple/inmemory/ratelimiter.go
HayimShaul: only the tests call this

token/services/selector/simple/inmemory/ratelimiter_test.go
HayimShaul: cleanup was missing - I added it now

HayimShaul: added concurrency tests

token/services/selector/simple/selector.go
HayimShaul: locks are locked in line 126 in the loop.
if something goes wrong we unlock all those we have locked

token/services/selector/simple/selector.go
HayimShaul: added now more tests

adecaro:
if I understand correctly the concerns of the issue this PR is trying to address, the selector service needs to have a definition of a `Locker` whose lock function takes in input also the wallet id the tokens are selected for. Then, we need to enforce that both `simple` and `sherdlock` selectors use this function. At that point the job of the token-sdk is done. We can update the documentation and tell the developer how to provide a new implementation of the Locker interface that integrate rate limiting. The application that uses the token-sdk can already have an overall infrastructure for rate limiting (using redis, for example) and they might want to reuse it as well.

HayimShaul:
> if I understand correctly the concerns of the issue this PR is trying to address, the selector service needs to have a definition of a `Locker` whose lock function takes in input also the wallet id the tokens are selected for. Then, we need to enforce that both `simple` and `sherdlock` selectors use this function. At that point the job of the token-sdk is done. We can update the documentation and tell the developer how to provide a new implementation of the Locker interface that integrate rate limiting. The application that uses the token-sdk can already have an overall infrastructure for rate limiting (using redis, for example) and they might want to reuse it as well.

I'm not sure I understand. Are you suggesting the rate limit be implemented by the application calling tokensdk?
Then why do we need a new Locker?

docs/security/selector_resource_limits.md
```diff
@@ -0,0 +1,301 @@
+# Selector Resource Limits
+
+## Overview
+
+The Token SDK selector service implements per-identity resource limits to prevent resource exhaustion attacks and ensure fair resource allocation. These limits are enforced at the service layer and apply regardless of the storage backend in use.
```
AkramBitar: Please do not use Token SDK now we need to replace it with Panurus

token/services/selector/simple/inmemory/locker.go
```diff
@@ -86,14 +117,28 @@ func (d *locker) Stop() error {
...
 }
 
 func (d *locker) Lock(ctx context.Context, id *token2.ID, txID string, reclaim bool) (string, error) {
+	return d.LockWithIdentity(ctx, id, txID, "", reclaim)
+}
+
+func (d *locker) LockWithIdentity(ctx context.Context, id *token2.ID, txID string, identity string, reclaim bool) (string, error) {
 	k := *id
 
+	// Rate limiting can be checked before acquiring the write lock.
```
AkramBitar: simple/selector.go calls this in a loop over candidate tokens. With default rateLimitBurst: 20, a legitimate transfer needing >20 inputs fails with ErrRateLimitExceeded. Doc says "requests per second" but it's actually per-token-lock. Fix: check rate limit once per Select, not per lock.

token/services/selector/config/driver.go
```diff
@@ -101,6 +107,30 @@ func (c *Config) GetFetcherCacheRefresh() time.Duration {
 	return c.FetcherCacheRefresh
 }
 
+func (c *Config) GetMaxLocksPerIdentity() int {
+	if c.MaxLocksPerIdentity > 0 {
+		return c.MaxLocksPerIdentity
```
AkramBitar: doesn't disable, contradicts docs

func (c *Config) GetMaxLocksPerIdentity() int {
    if c.MaxLocksPerIdentity > 0 { return c.MaxLocksPerIdentity }
    return defaultMaxLocksPerIdentity // 1000 — so 0 → 1000, not "disabled"
}
Same for GetRateLimit/GetRateLimitBurst. Docs claim "Set to 0 to disable" but config 0 silently becomes the default → limits can't be disabled via YAML. Distinguish unset from explicit 0 (use -1 sentinel or *int), or fix the docs.

token/services/selector/sherdlock/selector.go
```diff
@@ -155,6 +158,12 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
 	if s.isClosed() {
 		return nil, nil, 0, errors.Errorf("selector is already closed")
 	}
+	if s.enforcer != nil {
+		// CheckRateLimit already returns a wrapped simple.ErrRateLimitExceeded.
+		if err := s.enforcer.CheckRateLimit(owner.ID()); err != nil {
```
AkramBitar: Quota unenforced on the default driver

if s.enforcer != nil {
    if err := s.enforcer.CheckRateLimit(owner.ID()); err != nil { // only rate limit; no CheckQuota
        return nil, nil, 0, err
    }
}
sherdlock is the default ("": newSherdlockSelectorService). maxLocksPerIdentity is silently ignored here, so the PR's headline "lock quota" does nothing on default deployments. Implement quota for sherdlock or flag prominently.

docs/security/selector_resource_limits.md
```diff
@@ -0,0 +1,144 @@
...
+gives you two things:
+
+1. A **wallet-id-aware lock function**. Both selector drivers (simple and sherdlock)
+   pass the wallet id the tokens are being selected for into the `Locker`'s lock
+   function, so a custom `Locker` can apply per-wallet policies.
+2. A **fail-fast contract**, `token.SelectorRateLimited`. When a `Locker` denies a
+   lock by returning an error that wraps this sentinel, the selector aborts the
+   selection immediately and returns the error to the caller instead of retrying.
+
+This keeps the token-sdk minimal and lets applications reuse whatever rate-limiting
```
AkramBitar: Please remove all the appearance of token sdk it is Panurus

token/services/selector/sherdlock/selector_test.go
```diff
@@ -11,6 +11,7 @@ import (
 	"testing"
 	"time"
```
AkramBitar, line 13: Add a sherdlock release-on-abort test — lock a couple tokens, then return SelectorRateLimited, and assert UnlockAll runs (behavior is already correct; just untested for sherdlock).

docs/security/selector_resource_limits.md
```diff
@@ -0,0 +1,144 @@
...
+delegating. Below, a Redis-backed limiter throttles per wallet; the same shape works
+for an in-process limiter, a quota table, etc.
+
+### Simple selector
+
+```go
+import (
+    "context"
+
+    "github.com/hyperledger-labs/fabric-token-sdk/token"
```
AkramBitar: use github.com/hyperledger-labs/fabric-token-sdk/...; should be github.com/LFDT-Panurus/panurus/....

AkramBitar:
@HayimShaul 
Does the PR now to add hooks for per-wallet rate limiting" rather than "implement … rate limit," since no limiter ships by default? If so then we need to update the PR description and maybe the title

docs/security/selector_resource_limits.md
```diff
@@ -0,0 +1,144 @@
...
+## Overview
+
+Token selection acquires a short-lived *lock* on each candidate token so that two
+concurrent transactions do not try to spend the same token. Under load, a single
+wallet can drive a large number of selection/lock requests. Applications that need to
+throttle this — to protect the lock store, to enforce fairness between wallets, or to
+integrate with an existing quota system — can do so by supplying their own `Locker`
+implementation.
+
+The Token SDK deliberately ships **no built-in rate limiter or quota**. Instead it
```
AkramBitar: Remove token sdk, it is panurus

AkramBitar (changes requested):

token/selector.go
```diff
@@ -25,6 +25,14 @@ var (
 	// SelectorSufficientFundsButConcurrencyIssue is returned when funds are sufficient to cover the request, but
 	// concurrency issues does not make some of the selected tokens available.
 	SelectorSufficientFundsButConcurrencyIssue = errors.New("sufficient funds but concurrency issue")
+	// SelectorRateLimited is the contract error a Locker implementation returns (directly
+	// or wrapped) to deny a lock for policy reasons such as rate limiting or quota.
+	// Both the simple and sherdlock selectors detect it via errors.Is and abort the
+	// selection immediately, returning the error to the caller instead of retrying.
+	// The token-sdk ships no built-in limiter: applications integrate their own
```
AkramBitar: Change token-sdk to panurus also I saw sdk word alone maybe we need to replace this too.

AkramBitar (changes requested):
LGTM. 
Just to replace all the apperance of Token SDK/SDK with Panurus

token/selector.go
HayimShaul: done

AkramBitar (approved):
LGTM

PR #1796, opened 2026-06-22
fix(tokendb): make StoreToken idempotent with ON CONFLICT DO NOTHING

### LINKED ISSUE #1795: StoreToken is not idempotent: retried store hits primary-key violation (SQLSTATE 23505)
### Description

`TokenTransaction.StoreToken` (`token/services/storage/db/sql/common/tokens.go`) inserts the token row (PK `(tx_id, idx)`, line 1113) and the ownership rows (PK `(tx_id, idx, wallet_id)`, line 1125) with no conflict handling. Re-storing the same output therefore fails with a primary-key unique violation (SQLSTATE 23505).

This scenario is already known to the code — `AppendToken` swallows the violation at the Go level:

```go
// token/services/tokens/storage.go:202
err = t.Tx.StoreToken(ctx, ...)
if err != nil && !errors.HasCause(err, driver.UniqueKeyViolation) {
    return errors.Wrapf(err, "cannot store token in db")
}
```

On PostgreSQL this catch does not actually work. `AppendValid` appends every output of a transaction on a single shared `dbdriver.Transaction` (loop at `token/services/tokens/tokens.go:145`, followed by `DeleteTokens` at :153). Once the first INSERT raises 23505, the whole transaction enters the *aborted* state, so:

- the swallowed 23505 leaves the transaction poisoned;
- the next statement on that transaction (the next output's INSERT, the spent-token deletes, or the final `COMMIT`) fails with `25P02 current transaction is aborted, commands ignored until end of transaction block` — which is **not** a `UniqueKeyViolation`, so the Go-level catch no longer applies;
- the whole append rolls back, dropping even the legitimately-new outputs.

So a single duplicated output fails the entire `AppendValid` for that transaction. Duplicated stores are reachable via retries after a partial/failed append, recovery re-processing, and concurrent re-delivery of the same `tx_id`.

### Expected

Re-storing an already-recorded output should be idempotent. A token output is content-addressed by `(tx_id, idx)`, so the stored row cannot legitimately differ across attempts.

### Proposed fix

Add `ON CONFLICT DO NOTHING` to the token and ownership inserts in `StoreToken`. This suppresses the conflict at the SQL level, so the transaction is never poisoned — achieving what the `storage.go:202` catch intends, in the one place that works under Postgres. The existing Go-level catch then becomes redundant but harmless.

### Version

`main`

EvanYan1024:
Fixes #1795.

`StoreToken` commits the token and ownership rows in its own transaction. When a caller retries the surrounding flow after a downstream failure (e.g. a later audit or finality step that runs after the token rows are already committed), `StoreToken` re-inserts the same `(tx_id, idx)` and hits the primary-key unique violation (SQLSTATE 23505), aborting the caller.

This adds `ON CONFLICT DO NOTHING` to the token and ownership inserts so re-storing an already-recorded output is a no-op and the retry succeeds. A token output is content-addressed by `(tx_id, idx)`, so keeping the first stored row is safe.

The `OnConflictDoNothing()` builder is already used elsewhere in this package (e.g. the atomic upsert in `WalletStore.StoreIdentity`, #1662), so this follows the existing pattern.

### DIFF
--- a/token/services/storage/db/sql/common/tokens.go
+++ b/token/services/storage/db/sql/common/tokens.go
@@ -1317,6 +1317,7 @@ func (t *TokenTransaction) StoreToken(ctx context.Context, tr driver.TokenRecord
 	query, args := q.InsertInto(t.table.Tokens).
 		Fields("tx_id", "idx", "issuer_raw", "owner_raw", "owner_type", "owner_identity", "owner_wallet_id", "ledger", "ledger_type", "ledger_metadata", "token_type", "quantity", "amount", "stored_at", "owner", "auditor", "issuer").
 		Row(tr.TxID, tr.Index, tr.IssuerRaw, tr.OwnerRaw, tr.OwnerType, tr.OwnerIdentity, tr.OwnerWalletID, tr.Ledger, tr.LedgerFormat, tr.LedgerMetadata, tr.Type, tr.Quantity, tr.Amount, time.Now().UTC(), tr.Owner, tr.Auditor, tr.Issuer).
+		OnConflictDoNothing().
 		Format()
 	logging.Debug(logger, query, args)
 	if _, err := t.tx.ExecContext(ctx, query, args...); err != nil {
@@ -1339,6 +1340,7 @@ func (t *TokenTransaction) StoreToken(ctx context.Context, tr driver.TokenRecord
 	query, args = q.InsertInto(t.table.Ownership).
 		Fields("tx_id", "idx", "wallet_id").
 		Rows(rows).
+		OnConflictDoNothing().
 		Format()
 	logging.Debug(logger, query, args)
 


adecaro:
Hi @EvanYan1024 , thanks for submitting this 🙏 
Question: Did you identify already a case where a token gets re-inserted? I'm wondering if the other operations should be included in the same SQL transaction to avoid these collisions.

EvanYan1024:
Thanks @adecaro! 🙏

**Re: a concrete case** — yes, we hit this under high-concurrency stress testing. Our client retries a transport-failed audit round-trip once (in a fresh view context) when the first attempt times out or its reply is lost on a dead stream; if that first attempt had already reached the auditor, the retry re-runs the audit responder for the same txID. The responder registers a finality listener unconditionally on each run (`Append` → `AddFinalityListener`), so the same txID ends up with two listeners.

Why the two run concurrently: `AddFinalityListener` wraps each registration in a fresh `OnlyOnceListener` (`network/fabricx/finality/finality.go`), whose `sync.Once` dedups the two triggers *within a single registration* (the speculative `TxCheck` at registration and the later finality push) but **not across registrations**. The finality worker pool is a single shared queue with **no per-txID affinity**, so the two listeners' `OnStatus` run concurrently → two `Commit` → two `AppendValid`.

`AppendValid`'s re-processing guard, `TransactionExists` (`token/services/tokens/tokens.go:110`), runs on `readDB` *before and outside* the write transaction opened at line 129, and is just `SELECT tx_id FROM tokens WHERE tx_id = ? LIMIT 1` (`tokens.go:923`). Both concurrent `AppendValid`s read `exists=false` before the write tx, both call `StoreToken`, and the second hits the `(tx_id, idx)` PK. Recovery re-processing reaches the same path.

**Re: the same SQL transaction** — the other writes are *already* in one transaction: within a single `AppendValid`, the per-output `StoreToken` loop (line 145) and `DeleteTokens` (line 153) run on the same `ts` from `ContinueTransaction(tx)` and commit atomically with `SetStatus(Confirmed)`. The collision isn't between operations inside one append — it's between two *separate, concurrent* transactions racing on the same txID, which wrapping each in a transaction can't serialize. To prevent it that way you'd have to serialize concurrent appends of the same txID (move the existence check inside the write tx with `SELECT … FOR UPDATE`, or a per-txID advisory lock), which adds contention on the hot path.

`ON CONFLICT DO NOTHING` is the lighter, equally-correct option because a token output is content-addressed by `(tx_id, idx)` and immutable — re-inserting the same row is a genuine no-op. It's also the intent `AppendToken` already encodes by swallowing `UniqueKeyViolation` (`token/services/tokens/storage.go:202`); that Go-level catch just doesn't hold on Postgres, where the failed INSERT aborts the surrounding tx and the next statement/COMMIT fails with `25P02`. This PR moves the same intent to the SQL level, where it actually works.

adecaro:
Hi @EvanYan1024 , thanks for the clear explanation and sorry for the delay of this reply.
I'll approve this but I want to explore how to better control these cases. In a replicated setting probably the scenario is even worse because we cannot deduplicate on the single machine. Interesting case.

Thanks a lot for this effort, really appreciate 🙏

adecaro (approved):
LGTM

PR #1801, opened 2026-06-23
Implement Keystore Cleanup Service

### LINKED ISSUE #1800: Implement Keystore Cleanup Service for Token SDK
# Implement Keystore Cleanup Service for Token SDK

## Problem Statement

The Fabric Token SDK currently lacks an automated mechanism to clean up cryptographic key material (SKI - Subject Key Identifier) from the identity keystore after tokens are deleted. This creates several issues:

### Security Concerns
- **Key Material Accumulation**: Deleted tokens leave behind orphaned cryptographic keys in the keystore indefinitely
- **Compliance Risk**: Regulatory requirements (GDPR, data retention policies) may mandate timely deletion of cryptographic material
- **Attack Surface**: Unnecessary key material increases the potential attack surface for key compromise

### Operational Issues
- **Storage Growth**: Keystore grows unbounded over time as keys accumulate
- **Performance Degradation**: Large keystores can impact identity service performance
- **Resource Waste**: Storage resources consumed by obsolete key material

### Current Workaround Limitations
Manual keystore cleanup is:
- Error-prone and requires deep system knowledge
- Difficult to coordinate in multi-replica deployments
- Not scalable for high-volume token operations
- Lacks safety guarantees (e.g., premature deletion)

## Proposed Solution

Implement an automated **Keystore Cleanup Service** that:

1. **Safely Identifies Eligible Keys**: Only cleans keys for tokens that are:
   - Marked as deleted in the token database
   - Past a configurable Time-To-Live (TTL) threshold to ensure finality
   - Not already cleaned

2. **Supports Multi-Replica Deployments**: 
   - Uses PostgreSQL advisory locks for leader election
   - Ensures only one replica performs cleanup at a time
   - Handles node failures and restarts gracefully

3. **Provides Flexible Configuration**:
   - Disabled by default for backward compatibility
   - Configurable TTL, scan intervals, batch sizes, and worker counts
   - Tunable for different deployment scenarios (high-volume, resource-constrained, security-sensitive)

4. **Enables Identity-Type-Specific SKI Extraction**:
   - Pluggable provider architecture for different identity types
   - Idemix: Extracts SKI from NymPublicKey
   - IdemixNym: Extracts SKI via audit info lookup
   - X.509: No SKI extraction needed (certificate-based)
   - Fallback: SHA256 hash for unknown types

adecaro:
# Implement Keystore Cleanup Service for Automated Key Material Management

## Overview

This PR implements a comprehensive **Keystore Cleanup Service** that automatically removes cryptographic key material (SKI) from the identity keystore after tokens are deleted. The service addresses security, compliance, and operational concerns related to key material accumulation.

#### New Files
- **`manager.go`**: Orchestrates periodic cleanup sweeps with configurable intervals
  - Leader election via PostgreSQL advisory locks for multi-replica coordination
  - Batch processing with configurable worker pools
  - Graceful shutdown and error handling
  - Comprehensive logging and metrics integration

- **`manager_test.go`**: Extensive test suite (15+ test cases)
  - Unit tests for configuration validation
  - Integration tests for cleanup workflows
  - Concurrency and race condition testing
  - Mock-based testing for all dependencies

- **`config.go`**: Configuration structure with validation
  - TTL, scan interval, batch size, worker count
  - Advisory lock ID and instance ID
  - Sensible defaults with backward compatibility

- **`identity.go`**: Pluggable SKI extraction architecture
  - `TypedSKIProvider` interface for identity-type-specific extraction
  - `SKIExtractor` with provider registry and delegation logic
  - `FallbackSKIProvider` for unknown identity types (SHA256)
  - Thread-safe provider registration

#### Idemix Provider (`token/services/identity/idemix/crypto/skiprovider.go`)
- Extracts SKI from NymPublicKey in serialized Idemix identity
- Uses existing `SKIFromIdentity()` function for proper extraction
- 10+ test cases covering all scenarios

#### IdemixNym Provider (`token/services/identity/idemixnym/skiprovider.go`)
- Complex extraction via audit info lookup from identity store
- Follows same logic as `DeserializeSigner` in key manager
- Requires `IdentityStoreService` dependency
- 12+ test cases including error scenarios

#### X.509 Provider (`token/services/identity/x509/skiprovider.go`)
- Returns empty SKI (certificates don't require cleanup)
- Simple implementation with 8+ test cases
- Automatically registered in DI configuration

#### Token Store Interface (`token/services/storage/db/driver/token.go`)
Added three new methods:
```go
// Query deleted tokens eligible for cleanup
GetDeletedTokensPendingSKICleanup(ctx context.Context, ttl time.Duration, limit int) ([]DeletedToken, error)

// Mark token as cleaned to prevent duplicate processing
MarkTokenCleaned(ctx context.Context, txID string, index uint64, instanceID string) error

// Acquire cleanup leadership (PostgreSQL advisory lock)
AcquireCleanupLeadership(ctx context.Context, lockID int64) (CleanupLeadership, bool, error)
```

#### PostgreSQL Implementation (`token/services/storage/db/sql/postgres/`)
- **`tokens.go`**: Implemented all three cleanup methods
  - Efficient SQL queries with proper indexing
  - Advisory lock acquisition and release
  - Atomic cleanup state updates

- **`recovery.go`**: Extended for cleanup state tracking
  - New `cleaned_at` and `cleaned_by` columns
  - Migration-compatible schema changes

#### SQLite Implementation (`token/services/storage/db/sql/common/`)
- **`tokens.go`**: Implemented query and mark methods
  - Single-node deployment support
  - No advisory lock support (returns false for leadership)

- **`keystore.go`**: Identity keystore cleanup operations
  - Batch SKI deletion with error handling
  - Transaction support for atomicity

#### Configuration (`docs/configuration.md`)
Added comprehensive cleanup service configuration section:
- Parameter descriptions and default values
- Tuning recommendations for different scenarios:
  - High-volume environments
  - Resource-constrained systems
  - Security-sensitive deployments
  - Multi-instance vs single-node deployments
- Performance considerations and relationships

#### Service Documentation (`docs/services/cleanup.md`)
New documentation covering:
- Architecture and design patterns
- Configuration options and tuning
- Multi-replica coordination
- Identity-type-specific SKI extraction
- Monitoring and troubleshooting
- Deployment scenarios and best practices

#### Storage Documentation (`docs/services/storage.md`)
Updated with cleanup service integration details

#### Evolution Summary (`docs/evolution_summary.md`)
Added cleanup service to project evolution timeline

### 5. Dependency Injection (`token/sdk/dig/providers.go`)
- Registered X.509 SKI provider automatically
- Documentation for registering custom providers
- Backward-compatible initialization

### 6. Generated Mocks
- **`token/services/storage/db/driver/fake_token_store.go`**: Updated with new cleanup methods
- All mock methods properly implemented with counterfeiter


docs/services/cleanup.md
```diff
@@ -0,0 +1,232 @@
+# Keystore Cleanup Service
+
```
AkramBitar: I recommend to change the name of the file to keystore_cleanup.md

token/services/identity/idemix/crypto/ski.go
```diff
@@ -0,0 +1,55 @@
...
+//
+// Example usage:
+//
+//	ski, err := SKIFromIdentity(identity)
+//	if err != nil {
+//	    return err
+//	}
+//	// Use ski for identity lookups or key management
+func SKIFromIdentity(id view.Identity) ([]byte, error) {
+	serialized := &SerializedIdemixIdentity{}
```
AkramBitar, line 42: I am wondering if we need to check if id is not empty?
if len(id) == 0 {
    return nil, errors.New("empty identity")
}

token/services/storage/db/kvs/tracker.go
```diff
@@ -89,6 +89,15 @@ func (f *TrackedKVS) Get(id string, entry any) error {
 	return err
 }
 
+func (f *TrackedKVS) Delete(id string) error {
+	f.mutex.Lock()
```
AkramBitar, line 94: Minor comment. If the delete method does not do anything for now, why then we take the mutex lock?

token/services/storage/db/sql/common/tokens.go
```diff
@@ -1069,6 +1090,107 @@ func (db *TokenStore) GetCertifications(ctx context.Context, ids []*token.ID) ([
...
+			tokenTable.Field("spent_at"),
+		).
+		From(tokenTable.Join(cleanupTable, cond.And(
+			cond.Cmp(tokenTable.Field("tx_id"), "=", cleanupTable.Field("tx_id")),
+			cond.Cmp(tokenTable.Field("idx"), "=", cleanupTable.Field("idx")),
+		))).
+		Where(cond.And(
+			cond.CmpVal(tokenTable.Field("is_deleted"), "=", true),
+			cond.CmpVal(tokenTable.Field("spent_at"), "<", cutoffTime),
+			cond.CmpVal(cleanupTable.Field("tx_id"), "=", nil), // No cleanup record exists (LEFT JOIN returns NULL)
```
AkramBitar: Does CmpVal(..., "=", nil) generate IS NULL, not = NULL, since SQL null comparisons require IS NULL.

token/services/storage/db/sql/common/tokens.go
```diff
@@ -1069,6 +1090,107 @@ func (db *TokenStore) GetCertifications(ctx context.Context, ids []*token.ID) ([
...
+		)).
+		OrderBy(q.Asc(tokenTable.Field("spent_at"))).
+		Limit(limit).
```
AkramBitar, line 1124: I am wondering if we need to validate that limit is not <=0.

token/services/storage/services/cleanup/manager.go
```diff
@@ -0,0 +1,358 @@
...
+		}
+	}
+}
+
+// cleanupToken cleans up keys for a single deleted token
+func (m *Manager) cleanupToken(ctx context.Context, token DeletedToken) error {
+	m.logger.Debugf("cleaning up keys for deleted token [%s:%d]", token.TxID, token.Index)
+
+	// Get keystore
+	keystore, err := m.keystoreProvider.Keystore(m.tmsID)
```
AkramBitar, line 311: Do need to add defer keystore.Close()?

token/services/storage/services/cleanup/manager.go
```diff
@@ -0,0 +1,358 @@
...
+		}
+	}
+
+	// If all deletions failed, return error without marking as cleaned
+	if len(deleteErrors) == len(skis) {
+		return errors.Errorf("failed to delete all keys for token [%s:%d]: %v", token.TxID, token.Index, deleteErrors)
+	}
+
+	// Mark token as cleaned (even if some keys failed to delete)
+	if err := m.storage.MarkTokenCleaned(ctx, token.TxID, token.Index, m.config.InstanceID); err != nil {
```
AkramBitar, line 351: In case if len(deleteErrors) != len(skis), this means partial cleanup, then in this case we mark all of them as cleaned. Is that ok?

token/services/storage/services/cleanup/manager.go
```diff
@@ -0,0 +1,358 @@
...
+
+	// Start workers
+	for range m.config.WorkerCount {
+		workerWG.Add(1)
+		go m.worker(ctx, &workerWG, work, errCh)
+	}
+
+	// Fan out work to workers
+	for _, token := range tokens {
+		work <- token
```
AkramBitar: What about case <-ctx.Done():?

token/services/storage/services/cleanup/manager.go
```diff
@@ -0,0 +1,358 @@
...
+	var firstErr error
+	failures := 0
+	for err := range errCh {
+		if err == nil {
+			continue
+		}
+		failures++
+		m.logger.Warnf("cleanup failure: %v", err)
+		if firstErr == nil {
+			firstErr = err
```
AkramBitar: Only first error is retuned in this case? (do we need to aggregate these errors?)

token/services/storage/services/cleanup/manager.go
```diff
@@ -0,0 +1,358 @@
...
+
+	if m.started {
+		return errors.Errorf("cleanup manager already started")
+	}
+
+	if err := m.validateConfig(); err != nil {
+		return err
+	}
+
+	if m.config.InstanceID == "" {
```
AkramBitar, line 128: Why this can be empty?

token/services/storage/services/cleanup/manager_test.go
```diff
@@ -0,0 +1,394 @@
...
+		TTL:          100 * time.Millisecond,
+		ScanInterval: 100 * time.Millisecond,
+		BatchSize:    100,
+		WorkerCount:  1,
+	}
+
+	mockLeadership := &mock.Leadership{}
+	mockLeadership.CloseReturns(nil)
+	mockStorage.AcquireCleanupLeadershipReturns(mockLeadership, true, nil)
+	mockStorage.GetDeletedTokensPendingSKICleanupReturns([]cleanup.DeletedToken{}, nil)
```
AkramBitar, line 65: Can we have concurrency test that adds tokens to the table in one side and remove them in the otherside?

AkramBitar (changes requested):

token/services/identity/idemix/crypto/ski.go
adecaro: proto.Unmarshal will return an error anyway unless you mean that the caller of this function has an interest in knowing when id was passed empty

token/services/storage/db/kvs/tracker.go
adecaro: actually, the backend must have the Delete function. I'll update the code

token/services/storage/db/sql/common/tokens.go
adecaro: I would say that the caller is responsible for it, but if you think we need to enforce this then I guess we need to enforce also other checks on the other fields?

token/services/storage/services/cleanup/manager.go
adecaro: because it might not be specified in the configuration, no?

token/services/storage/services/cleanup/manager.go
adecaro: good catch. I'll update the code. We should check the rest of the codebase to see if we have a similar case elsewhere.

token/services/storage/services/cleanup/manager.go
adecaro: I don't know if we need but we can. I'll aggregate them.

token/services/storage/services/cleanup/manager.go
adecaro: no, Close is only during the shutdown of the system

token/services/storage/services/cleanup/manager_test.go
adecaro: this should be covered by the integration tests, right? Here we are using mocks only. Unless, you want a more scoped integration tests that uses a given DB.

token/services/storage/services/cleanup/manager.go
AkramBitar: Do we need to enforce it?

token/services/storage/db/sql/common/tokens.go
adecaro: good catch, I'll update

token/services/storage/services/cleanup/manager.go
adecaro: no, we don't. It will selected by the code.

token/services/storage/services/cleanup/manager.go
AkramBitar: @adecaro did you have the chance to look at that one?

adecaro: missed this one, sorry. Let me have a look

docs/imgs/storage_db.puml
```diff
@@ -115,9 +115,18 @@ package "Token Store (TokenDB)" {
         created_at : TIMESTAMP
     }
 
+    entity "TokenSKICleanups" as tkn_ski_cleanups {
+        * tx_id : TEXT <<PK, NOT NULL, FK>>
```
AkramBitar, line 119: Usually we create a picture for this. Do we have one?

token/services/storage/services/cleanup/manager.go
adecaro: indeed, let's fail even with one error so that the operation will be tried.

AkramBitar (approved):
LGTM
One last comment: 
https://github.com/hyperledger-labs/fabric-token-sdk/pull/1801#discussion_r3467127420

docs/imgs/storage_db.puml
adecaro: yes, there is. It is storage_db.png

PR #1802, opened 2026-06-24
Recovery never confirms transactions on non-owner nodes

### LINKED ISSUE #1773: Recovery never confirms transactions on non-owner nodes: NewFullRequestFromBytes fails on the bare TokenRequest (proto3 "string field contains invalid UTF-8")
### Summary

On a Fabric-X deployment, an FSC node that participates in transactions as a
**non-owner** (an endorser node, in our case) never marks those transactions
`Confirmed` — they stay `Pending` indefinitely. The recovery service repeatedly
claims and fails them:

    recoverTransactions WARN completed recovery sweep: claimed=100, succeeded=0, failed=100
    recovery failure: ... failed to process token request [<txid>]:
      failed to create token request from bytes:
      failed to unmarshal request: string field contains invalid UTF-8

Owner nodes (initiators) confirm normally. The non-owner node's `Pending` set
therefore grows by one per transaction it endorses, and the recovery sweep
becomes a permanent zero-progress loop.

### Environment
- fabric-token-sdk: recent `main`
- Network: Fabric-X
- Driver: zkatdlog (confidential)

### Root cause (our analysis)

Recovery reconstructs the token request from stored bytes via `NewFullRequestFromBytes`:

- `token/services/ttx/finality/recovery.go` — `TTXRecoveryHandler.applyFinalityLogic`:
  on `network.Valid`, loads the stored token request and calls `hasher.ProcessTokenRequest`.
- `token/services/ttx/finality/hasher.go` — `ProcessTokenRequest` →
  `tms.NewFullRequestFromBytes(tokenRequestRaw)`.
- `token/request.go` — `Request.FromBytes` → `proto.Unmarshal(raw, &request.TokenRequestWithMetadata{})`.

`NewFullRequestFromBytes` assumes the bytes are a full `TokenRequestWithMetadata`
(`version`, **`anchor`**, `request`, `metadata`). A non-owner node never has that
form: during endorsement collection the initiator distributes only `RequestToBytes()`
— the bare `TokenRequest` (actions/signatures/auditing); the metadata is owner-private
and is not sent. The non-owner persists that bare form.

Unmarshaling a bare `TokenRequest` as a `TokenRequestWithMetadata` collides on field #2:

| field # | `TokenRequest`                       | `TokenRequestWithMetadata` |
|---------|--------------------------------------|----------------------------|
| 2       | `actions` (repeated message, binary) | `anchor` (**string**)      |

proto3 enforces UTF-8 on the `anchor` string field, so the binary action bytes are
rejected → `string field contains invalid UTF-8`.

Owner nodes are unaffected: they persist the full `TokenRequestWithMetadata`, and the
finality listener usually resolves the request from the in-memory cache
(`GetCachedTokenRequest`), rarely hitting the DB reconstruction. A non-owner node has
no such cache after a restart, so every recovery attempt hits the failing path.

(`ProcessTokenRequest` is shared by both the recovery sweep `recovery.go` and the push
finality listener `listener.go`; `NewFullRequestFromBytes` is also used by
`ttx/info.go` `loadTransient`.)

### Questions

1. Is this TTX recovery / finality expected to run on non-owner nodes at all? Should
   listener registration / recovery be scoped to owner (and auditor) roles?

2. If non-owners are meant to recover, would a deserialization fallback be acceptable —
   try `NewFullRequestFromBytes`, and on failure fall back to
   `NewRequestFromBytes(anchor, raw, nil)` (parse as bare actions)? The confirm path
   appears to need only the actions (`MarshalToSign` excludes metadata, and the
   resulting hash matches the on-ledger reference). Is that reasoning correct, or is the
   metadata required somewhere in the confirm path?

3. A deserialization failure is currently treated like a transient error: the
   transaction is neither confirmed nor retired, so it is re-claimed on every sweep
   indefinitely (and, depending on claim ordering, can pin the head of the queue).
   Should recovery classify non-retryable failures (e.g. unmarshal errors) and move them
   to a terminal state instead of looping forever?

### Impact

A non-owner node accumulates unbounded `Pending` rows and runs a continuous failing
recovery sweep; under load the resulting churn contributes to instability.

adecaro:
This PR addresses #1773 by introducing a new db in the storage services dedicated to store the audit traces of the endorsers.


token/services/storage/db/sql/common/transactions.go
```diff
@@ -682,11 +682,23 @@ func (w *TransactionStoreTransaction) AddValidationRecord(ctx context.Context, t
...
+		Row(txID, tokenRequest, dbdriver.Pending, "", []byte("{}"), []byte("{}"), ppHash, now).
+		Format()
+	logging.Debug(logger, query, txID, fmt.Sprintf("(%d bytes)", len(tokenRequest)), len(ppHash), now)
+
+	if _, err = w.txn.ExecContext(ctx, query, args...); err != nil {
+		return ttxDBError(err)
+	}
+
+	// Then insert the validation record as a self-contained entity (no foreign key to Requests)
+	query, args = q.InsertInto(w.table.Validations).
```
AkramBitar: What happens in the case that we succussed to insert to Requests table and failed to insert to Validations table?  
In this case we have inconsistent data. 
Does this covered under transaction section that will be rollbacked?

AkramBitar (changes requested):

token/services/storage/db/sql/common/transactions.go
adecaro: indeed, the transaction db should not carry any reference to validation records anymore. I have moved everything to a new db called endorserdb

docs/imgs/storage_db.puml
```diff
@@ -56,10 +49,20 @@ package "Transaction & Audit Store (TTXDB / AuditDB)" {
...
     movements }o--|| requests : "tx_id"
-    req_vals ||--|| requests : "tx_id"
     tx_ends }o..|| requests : "tx_id (no FK constraint)"
 }
 
+package "Endorser Store (EndorserDB)" {
+    entity "Validations" as req_vals {
+        * tx_id : TEXT <<PK, NOT NULL>>
+        --
+        request : BYTEA
```
AkramBitar, line 59: Do we need to have foreign keys that points to tx_id and request in other stores?

docs/services/storage/endorserdb.md
```diff
@@ -0,0 +1,219 @@
+# Endorser Database (endorserdb)
+
+The `endorserdb` package provides storage services for validation records in the Fabric Token SDK. It manages the persistence and querying of token transaction validation records that are created during the endorsement process.
+
+## Overview
+
+The endorser database stores validation records that track token requests validated by endorsers. Each validation record contains:
+- Transaction ID
+- Token request data
```
AkramBitar, line 9: Does this data is duplicated? Do we have it in the Request table?

token/services/storage/db/sql/common/endorser.go
```diff
@@ -0,0 +1,214 @@
...
+// Close closes the database connections
+func (db *EndorserStore) Close() error {
+	return nil // Connections are managed externally
+}
+
+// GetSchema returns the SQL schema for creating the endorser store tables
+func (db *EndorserStore) GetSchema() string {
+	return fmt.Sprintf(`
+		CREATE TABLE IF NOT EXISTS %s (
+			tx_id TEXT NOT NULL PRIMARY KEY,
```
AkramBitar, line 64: Related to my above comment in foreign keys.  Maybe we need t consider adding relationship between this table and other tables, this will preserve consistsency.

token/services/storage/endorserdb/store.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	// Store the token request directly in the validation record
+	if err := w.AddValidationRecord(ctx, txID, tokenRequest, meta, ppHash); err != nil {
+		w.Rollback()
+
+		return errors.WithMessagef(err, "append validation record for txid [%s] failed", txID)
+	}
+	if err := w.Commit(); err != nil {
+		return errors.WithMessagef(err, "append validation record commit for txid [%s] failed", txID)
+	}
```
AkramBitar, line 137: In case of error in the commit do we need to call rollback?

See set SetStatus method below

token/services/storage/db/sql/sqlite/endorser_test.go
```diff
@@ -0,0 +1,276 @@
...
+	record := driver.ValidationRecord{
+		TxID:         "1234",
+		TokenRequest: []byte("some request"),
+		Timestamp:    timeFrom,
+	}
+	output := []driver2.Value{
+		record.TxID, record.TokenRequest, nil, record.Timestamp,
+	}
+
+	query := "SELECT VALIDATIONS.tx_id, VALIDATIONS.request, VALIDATIONS.metadata, VALIDATIONS.stored_at " +
```
AkramBitar, line 51: I am wondering why we do not have common endorser tests for both databases?

AkramBitar (changes requested):

docs/imgs/storage_db.puml
adecaro: no, this table is designed to be self sufficient. Only endorsers will have it with data in it.

docs/services/storage/endorserdb.md
adecaro: no, no duplication. The endorserdb is used only by endorsers that don't use the other dbs

token/services/storage/db/sql/common/endorser.go
adecaro: they are independent cause the endorserdb is used only by the endorsers who do not use the other tables.

token/services/storage/db/sql/sqlite/endorser_test.go
adecaro: we have, but this part is about the generation of the sql queries that are db specific

token/services/storage/endorserdb/store.go
adecaro: no need to rollback explicitly, it is not required. SetStatus is not used yet. We will understand later if we need to update this flag in case the endorser must run statefull validation

AkramBitar (approved):
LGTM

PR #1805, opened 2026-06-25
Implements token lock cleanup for orphan transactions

### LINKED ISSUE #1797: Clean Token Locks for Orphan Transactions
Orphan transactions never reached the ledger, so their locks are safe to clean immediately, just like Deleted transactions (when tokens spent). The recovery system already marks these transactions as Orphan, but the cleanup doesn't check for it.

That means:
DELETE FROM token_locks WHERE
  -- Dead transactions (immediate cleanup)
  consumer_tx_id IN (
    SELECT tx_id FROM requests 
    WHERE status IN (Deleted, Orphan)
  )

AkramBitar:
Addresses issue: https://github.com/hyperledger-labs/fabric-token-sdk/issues/1797

### DIFF
--- a/token/services/storage/db/sql/postgres/tokenlock.go
+++ b/token/services/storage/db/sql/postgres/tokenlock.go
@@ -67,19 +67,24 @@ func (db *TokenLockStore) Cleanup(ctx context.Context, leaseExpiry time.Duration
 	if err := db.logStaleLocks(ctx, leaseExpiry); err != nil {
 		db.Logger.Warnf("Could not log stale locks: %v", err)
 	}
-	tokenLocks, _ := q.Table(db.Table.TokenLocks), q.Table(db.Table.Requests)
+	tokenLocks, tokenRequests := q.Table(db.Table.TokenLocks), q.Table(db.Table.Requests)
 
-	query, args := common3.NewBuilder().
-		WriteString("DELETE FROM ").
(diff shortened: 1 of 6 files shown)


token/services/storage/db/sql/postgres/tokenlock.go
```diff
@@ -76,8 +76,8 @@ func (db *TokenLockStore) Cleanup(ctx context.Context, leaseExpiry time.Duration
 		WriteConditionSerializable(cond.OlderThan(tokenLocks.Field("created_at"), leaseExpiry), db.ci).
 		WriteString(" OR ").
 		WriteString(
-			fmt.Sprintf("EXISTS (SELECT 1 FROM %s WHERE %s.tx_id = %s.consumer_tx_id AND %s.status IN (%d))",
-				db.Table.Requests, db.Table.Requests, db.Table.TokenLocks, db.Table.Requests, driver.Deleted,
+			fmt.Sprintf("EXISTS (SELECT 1 FROM %s WHERE %s.tx_id = %s.consumer_tx_id AND %s.status IN (%d, %d))",
```
adecaro: could we use the query builder like in the the rest of the functions (logStaleLocks, for example)?

adecaro (changes requested):
LGTM overall. Just a comment to make the code more robust.

adecaro (approved):
LGTM

PR #1811, opened 2026-06-25
perf(ttxdb): batch transaction records into a single multi-row INSERT

### LINKED ISSUE #1810: perf(ttxdb): AppendTransactionRecord issues one INSERT per record instead of a single multi-row INSERT
### Description

`StoreService.AppendTransactionRecord` (`token/services/storage/ttxdb/store.go:227`) persists the per-action transaction records parsed by `TransactionRecords`. It writes them with a per-record loop:

```go
// token/services/storage/ttxdb/store.go:268
for _, tx := range txs {
    if err := w.AddTransaction(ctx, tx); err != nil {
        ...
    }
}
```

`AddTransaction` is variadic (`token/services/storage/db/driver/ttx.go:42`), and its SQL implementation already builds a single multi-row `INSERT ... VALUES (..),(..)` from every record passed in (`token/services/storage/db/sql/common/transactions.go:580`, `q.InsertInto(...).Rows(rows)` at :597). Calling it once per record defeats that batching: N records become N single-row INSERTs (N `ExecContext` round-trips) inside the same store transaction.

A typical transfer parses into two records — the transfer leg plus the change-back-to-sender leg — so the common path issues two INSERTs where one would do; multi-output / multi-action transactions issue more. This sits on the per-transaction store path, so it also affects the auditor, which runs it for every transaction.

### Expected

All records for a transaction are inserted in a single multi-row INSERT — one round-trip — matching what the variadic `AddTransaction` builder already supports.

### Proposed fix

Pass the whole slice in one call: `w.AddTransaction(ctx, txs...)`, guarded by `len(txs) > 0` so a transaction that parses into zero records does not emit an empty INSERT.

### Version

`main`

EvanYan1024:
Fixes #1810.

`AppendTransactionRecord` looped over the parsed transaction records and called `AddTransaction` once per record, issuing N single-row INSERTs inside the store transaction. `AddTransaction` is variadic and its SQL implementation already builds a single multi-row `INSERT ... VALUES (..),(..)`, so this passes all records in one call to collapse the per-tx writes (typically the transfer leg plus the change-back-to-sender leg) into one round-trip.

Guarded with `len(txs) > 0` so a transaction that parses into zero records does not emit an empty INSERT.

No behavior change: the same rows are written in the same store transaction, just in one statement instead of N.

### DIFF
--- a/token/services/storage/ttxdb/store.go
+++ b/token/services/storage/ttxdb/store.go
@@ -271,8 +271,8 @@ func (d *StoreService) AppendTransactionRecord(ctx context.Context, req *token.R
 
 		return errors.WithMessagef(err, "append token request for txid [%s] failed", record.Anchor)
 	}
-	for _, tx := range txs {
-		if err := w.AddTransaction(ctx, tx); err != nil {
+	if len(txs) > 0 {
+		if err := w.AddTransaction(ctx, txs...); err != nil {
 			w.Rollback()
 
 			return errors.WithMessagef(err, "append transactions for txid [%s] failed", record.Anchor)


adecaro (approved):
LGTM

PR #1477, opened 2026-04-05
Add benchmarks for Quantity operations

theycallmeaabie:
## Summary                                                                                                                                    
- Add benchmarks for `BigQuantity` and `UInt64Quantity` operations (`Add`, `Sub`, `Cmp`)                                                      
- Add benchmark for `ToQuantity` parsing  

Resolves #1467

PR #1818, opened 2026-06-25
Add benchmarks for Quantity operations

### LINKED ISSUE #1467: chore(token/token/quantity.go): improvements

### LINKED ISSUE #1819: Add benchmarks (token/token/quantity.go)

AkramBitar:
## Summary
- Add benchmarks for `BigQuantity` and `UInt64Quantity` operations (`Add`, `Sub`, `Cmp`)
- Add benchmark for `ToQuantity` parsing

Continues the work from #1477 (closed), rebased on latest `main` with review comments addressed:
- Assert on errors instead of ignoring them (`require.NoError`)
- Move setup out of the timed loop and call `b.ResetTimer()` so only the target op is measured
- Store side-effect-free results in package-level sinks to prevent dead-code elimination

Resolves #1819

### DIFF
--- a/docs/drivers/benchmark/benchmark.md
+++ b/docs/drivers/benchmark/benchmark.md
@@ -24,3 +24,6 @@
 
 ### Node Level Benchmarks
 - [Token Validation Service Benchmark](token_validation_service_benchmark.md)
+
+### Foundational Types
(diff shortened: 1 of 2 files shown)


adecaro (approved):
LGTM

PR #1821, opened 2026-06-25
Add inbound message size validation

### LINKED ISSUE #1629: Validate and Bound Incoming Message Sizes [MED]
**Description**
Before processing any inbound message  --  signature replies, auditor payloads, or distribution acks  --  enforce a maximum message-size limit and reject oversized or malformed payloads immediately. This prevents a responder from forcing the view to allocate unbounded memory for deserialization or signature verification. Pair size checks with schema/format validation so that structurally invalid messages are dropped at the earliest possible point in the pipeline.

**Short Summary**
Add max message size limits before deserializing signature replies, auditor payloads, distribution acks

AkramBitar:
Add a max message-size check at the one place all inbound messages pass through (ReceiveRawWithTimeout), so signature replies, auditor payloads, and distribution acks are rejected before deserialization. Oversized payloads fail fast with ErrMessageTooLarge; malformed ones are still caught by the existing envelope validation. Default limit 100 MiB, configurable.

### DIFF
--- a/token/services/utils/json/session/envelope.go
+++ b/token/services/utils/json/session/envelope.go
@@ -169,6 +169,11 @@ func ReceiveTypedWithTimeout(s *session.S, expectedType string, dst any, d time.
 func ReceiveTypedWithTimeoutAndMetrics(s *session.S, expectedType string, dst any, d time.Duration, m *EnvelopeMetrics) error {
 	raw, err := s.ReceiveRawWithTimeout(d)
 	if err != nil {
+		// Count oversized-payload rejections; other transport errors are left as-is.
+		if errors.Is(err, session.ErrMessageTooLarge) {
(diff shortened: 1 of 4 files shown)


adecaro (commented):
I think we should have here a more robust solution directly at the FSC level so that we can be sure these limits are enforced everywhere. What do you think?

AkramBitar:
> I think we should have here a more robust solution directly at the FSC level so that we can be sure these limits are enforced everywhere. What do you think?

@adecaro I agree. 

@mbrandenburger 
What do you think about @adecaro sugesstion?

Thanks a lot,
Akram

adecaro:
@AkramBitar , I'll close this PR in the meantime. Please, move the discussion directly on the smart client by creating an issue. Thank you.

PR #1823, opened 2026-06-26
feat(identity/x509): make PKCS11 SessionCacheSize configurable

### LINKED ISSUE #1822: feat(identity/x509): PKCS11 session cache size is not configurable (pinned to fabric-lib-go default of 10)
### Description

The x509 identity PKCS11 config (`token/services/identity/x509/crypto/config.go`, type `PKCS11`) exposes the standard PKCS11 options (Library, Label, Pin, KeyIDs, ...) but not the session cache size. `ToPKCS11OptsOpts` builds a `pkcs11.PKCS11Opts` without setting it, so the PKCS11 session pool is always pinned to fabric-lib-go's hardcoded default of 10 (`defaultSessionCacheSize` in `bccsp/pkcs11`).

On a node that drives a large, concurrent PKCS11/HSM signing workload, 10 cached sessions cap the pool and serialize signing under load — there is no way to raise it from configuration.

The currently pinned fabric-lib-go v1.1.3 keeps `sessionCacheSize` unexported, so even adding a config field would have no API to set it. v1.1.4 exposes `SessionCacheSize` on `bccsp/pkcs11.PKCS11Opts` (#49), making it settable.

### Expected

The PKCS11 session cache size is configurable from the x509 identity config, falling back to the existing default (10) when unset.

### Proposed fix

Bump fabric-lib-go to v1.1.4 and plumb a new `SessionCacheSize` field from the x509 crypto `PKCS11` config into `pkcs11.PKCS11Opts` via `ToPKCS11OptsOpts`. Add the field to the `!pkcs11` build-tag stub as well so the default build compiles. Omitting it keeps current behavior (zero → fabric-lib-go's default of 10).

### Version

`main`

EvanYan1024:
Fixes #1822.

The x509 identity PKCS11 config exposed the standard PKCS11 options but not the session cache size, so `ToPKCS11OptsOpts` always left it at fabric-lib-go's hardcoded default of 10. On nodes that drive a large, concurrent PKCS11/HSM signing workload, that pins the session pool and serializes signing under load, with no config knob to raise it.

fabric-lib-go v1.1.4 exposes `SessionCacheSize` on `bccsp/pkcs11.PKCS11Opts` (#49), so this:

- bumps fabric-lib-go v1.1.3 → v1.1.4 (pulls miekg/pkcs11 v1.1.1 → v1.1.2 as an indirect);
- adds a `SessionCacheSize` field to the x509 crypto `PKCS11` config and to the `!pkcs11` build-tag stub so both build variants compile;
- passes it into `PKCS11Opts` in `ToPKCS11OptsOpts`.

Backward compatible: omitting the field leaves it zero, which fabric-lib-go maps to its default cache size of 10. Verified `go build ./token/...` and `go build -tags pkcs11 ./token/...` both pass.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/hashicorp/go-uuid v1.0.3
 	github.com/hyperledger-labs/fabric-smart-client v0.12.1
 	github.com/hyperledger/fabric-chaincode-go/v2 v2.3.0
-	github.com/hyperledger/fabric-lib-go v1.1.3
+	github.com/hyperledger/fabric-lib-go v1.1.4
 	github.com/hyperledger/fabric-protos-go-apiv2 v0.3.7
 	github.com/hyperledger/fabric-x-common v0.2.4
 	github.com/jackc/pgx/v5 v5.9.2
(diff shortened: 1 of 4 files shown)


adecaro (approved):
LGTM. Thanks @EvanYan1024 for this effort 🙏

PR #1826, opened 2026-06-29
perf(identity): cache default identifier to avoid O(n) scan

### LINKED ISSUE #1825: perf(identity): LocalMembership.getDefaultIdentifier does an O(n) scan of localIdentities on every resolution
### Description

`LocalMembership.getDefaultIdentifier` (`token/services/identity/membership/lm.go`) resolves the default identity by linearly scanning the entire `localIdentities` slice for the entry whose `Default` flag is set:

```go
func (l *LocalMembership) getDefaultIdentifier() string {
	for _, li := range l.localIdentities {
		if l.anonymous && !li.Anonymous {
			continue
		}
		if li.Default {
			return li.Name
		}
	}
	return ""
}
```

It runs on the identity-resolution hot path: `Role.MapToIdentity` (via `mapStringToID` / `mapIdentityToID`) calls `GetDefaultIdentifier()` unconditionally at the top, and `MapToIdentity` is invoked on every wallet lookup. The default identity is set once (the first registered identity when none exists, or the Load fallback) and changes only when an identity is explicitly made default — so the scanned result is effectively constant between mutations, yet it is recomputed from scratch on every lookup.

For a node holding a large number of owner identities — e.g. a custodial node that registers hundreds of thousands of wallets into a single `LocalMembership` — this O(n) scan dominates CPU under load (~26% of CPU in our deployment's profiling). Bulk paths that resolve every wallet (one lookup per wallet) make it effectively O(n²).

### Expected

Resolving the default identifier is O(1); the linear scan does not run on every lookup.

### Proposed fix

Cache the default identity name on `LocalMembership` and maintain it only where the default actually changes — Load reset, the Load fallback, and `addLocalIdentity` — honoring the anonymity-mode selectability filter (empty when no selectable default exists). `getDefaultIdentifier` then returns the cached value. Behavior is unchanged.

### Version

`main`

EvanYan1024:
Fixes #1825.

`getDefaultIdentifier` resolved the default identity by linearly scanning the entire `localIdentities` slice on every call. It runs on the identity-resolution hot path — `Role.MapToIdentity` computes it unconditionally — so on a node holding a large number of owner identities (a custodial node can register hundreds of thousands of wallets into one `LocalMembership`) the scan dominates: ~26% of CPU in our deployment's profiling, and effectively O(n²) on bulk paths that resolve every wallet.

The default identity is set once and changes only when an identity is explicitly made default, so the result is constant between mutations. This caches the default identity name on `LocalMembership` and maintains it at the three points where it can change:

- Load reset — cleared;
- the Load fallback — set to the chosen default (`firstDefaultIdentifier` already honors the anonymity mode);
- `addLocalIdentity` — set/cleared per the anonymity-mode selectability filter.

`getDefaultIdentifier` then returns the cached value, so the lookup is O(1).

Behavior is unchanged, including the anonymity-mode filter (empty when no selectable default exists); all sites that set `Default` are covered by the cache maintenance.


adecaro (approved):
LGTM. Simple and effective, thanks @EvanYan1024 🙏

PR #1829, opened 2026-06-29
validator robustness

### LINKED ISSUE #1830: Validator robustness
This Issue is a placeholder for a PR that will contain fixes to the validation code in the zkat dlog driver.

adecaro:
This PR hardens the ZKATDLog (zkatdlog/nogh/v1) token engine across several security dimensions: deterministic public-parameter generation, validator robustness against malformed inputs, and a new proof-type safety gate that prevents an attacker from selecting a verifier whose required `PublicParams` sub-struct is absent.

### 1. Deterministic Pedersen and Range-Proof Generators (`setup/setup.go`)
Pedersen generators and all range-proof generators (IPA, CSP) are now derived using `curve.HashToG1` with a **domain-separated, deterministic label** (`lfdt-panurus.<driver>.<version>.<role>.<index>`) instead of multiplying `GenG1` by a random scalar. This makes public parameters reproducible from the stored `DriverName`/`DriverVersion` fields, removes the dependency on a PRNG at setup time, and improves the existing labels for range-proof generators (previously bare integers `"0"`, `"1"`, `"RangeProof.2"`, etc.).

### 2. `SupportsRangeProofType` + Proof-Type Mismatch Guard (`setup/setup.go`, `transfer/transfer.go`, `issue/issue.go`)
A new `PublicParams.SupportsRangeProofType(proofType)` helper gates verifier construction. Both `transfer.NewVerifier` and `issue.NewVerifier` now return an `(Verifier, error)` instead of `Verifier`, and both return `ErrProofTypeMismatch` when the required sub-struct (`RangeProofParams` or `CSPRangeProofParams`) is `nil`. This closes the attack vector where an adversary could embed a proof type in an action whose matching params sub-struct is absent, causing a nil-pointer dereference inside the verifier.

### 3. CSP: Reject Points at Infinity (`crypto/rp/csp/validation.go`)
The CSP verifier's `validateG1Slice` and `validateCSPProof` functions now reject **identity elements (points at infinity)** in generators and in proof folding-commitment arrays (`proof.Left`, `proof.Right`). Previously, infinity was explicitly permitted with a comment noting "edge cases"; this is corrected because an infinity point in a generator collapses the commitment scheme and in a proof element neutralises the corresponding round constraint, weakening range-proof soundness. The per-element loops are replaced with a call to `math.CheckElements` (which already checks for nil, wrong curve, and infinity).

### 4. Validator Robustness – Nil & Out-of-Bounds Guards
Multiple validator functions are hardened against nil or inconsistent inputs to prevent panics in custom or adversarial validation pipelines:

| Location | Guard added |
|---|---|
| `validator_transfer.go` – `TransferSignatureValidate` | Rejects nil `ActionInput` or nil `ActionInput.Token` |
| `validator_transfer.go` – `TransferZKProofValidate` | Rejects nil `ctx.InputTokens[i]` |
| `validator_transfer.go` – `TransferHTLCValidate` | Rejects nil `ctx.InputTokens[i]`; bounds-checks `ctx.Signatures` before indexing |
| `validator_transfer.go` – `TransferUpgradeWitnessValidate` | Rejects empty/nil owner on either side (closes free-claim of ownerless tokens, FR-2) |
| `transfer/action.go` – `GetInputs` | Nil input entries produce `nil` IDs instead of panicking |
| `transfer/action.go` – `GetSerializedInputs` | Returns error for nil `input.Token` |
| `transfer/action.go` – `IsRedeemAt` / `SerializeOutputAt` | Bounds-check index before accessing `t.Outputs` |
| `transfer/action.go` – `GetSerializedOutputs` / `GetOutputCommitments` | Skip or error on nil output entries |
| `transfer/action.go` – `Deserialize` | Nil output entries during deserialization now return an error instead of silently continuing |

### 5. `TypeAndSumProof` Validation Fix (`transfer/typeandsum.go`)
- `Validate`: the length check for `InputBlindingFactors` was incorrectly using `len(p.InputBlindingFactors)` as the expected length, making the check trivially pass. Corrected to `len(p.InputValues)`.
- `TypeAndSumVerifier.Verify`: added a guard to reject proofs where `len(stp.InputBlindingFactors) != len(v.Inputs)`, preventing an out-of-bounds access in the verifier loop.

### 6. `ValidateForDeployment` (`setup/setup.go`)
Adds a new `PublicParams.ValidateForDeployment()` method that warns when `IssuerIDs` is empty (open-policy mode). Existing `Validate()` continues to accept empty `IssuerIDs`; this is a separate, optional deployment-time check. Docstrings in `IssueValidate` and `TransferSignatureValidate` document the open-policy behaviour explicitly.

### 7. Documentation and Test Coverage
- `docs/drivers/dlogwogh.md` updated to reflect the new deterministic generator derivation and security improvements.
- Added comprehensive validator security tests in `validator/validator_security_test.go` and `validator/validator_security_extra_test.go` covering the nil/infinity/mismatch attack vectors described above.
- Extended tests in `setup/setup_test.go`, `transfer/action_test.go`, `transfer/bftransfer_test.go`, `transfer/typeandsum_test.go`, `issue/issuer_test.go`, `issue/prover_test.go`, and `crypto/rp/csp/`.

## Testing
All existing unit tests pass. New security-focused tests verify:
- Nil/infinity element rejection in CSP validators
- `ErrProofTypeMismatch` when no matching params sub-struct is present
- Nil-input panics converted to proper errors in transfer and issue validators
- `TypeAndSumProof` length invariant enforcement
- Ownerless-token free-claim prevention in upgrade witness validation

### DIFF
--- a/token/core/zkatdlog/nogh/v1/issue/action.go
+++ b/token/core/zkatdlog/nogh/v1/issue/action.go
@@ -302,7 +302,7 @@ func (i *Action) Deserialize(raw []byte) error {
 	i.Outputs = make([]*token.Token, len(issueAction.Outputs))
 	for j, output := range issueAction.Outputs {
 		if output == nil || output.Token == nil {
-			continue
+			return errors.Errorf("invalid issue action: output at index [%d] is nil", j)
 		}
 		data, err := utils.FromG1Proto(output.Token.Data)
 		if err != nil {
--- a/token/core/zkatdlog/nogh/v1/issue/action.go
+++ b/token/core/zkatdlog/nogh/v1/issue/action.go
@@ -301,8 +301,11 @@ func (i *Action) Deserialize(raw []byte) error {
 	// outputs
 	i.Outputs = make([]*token.Token, len(issueAction.Outputs))
 	for j, output := range issueAction.Outputs {
-		if output == nil || output.Token == nil {
+		if output == nil {
 			return errors.Errorf("invalid issue action: output at index [%d] is nil", j)
+		}
+		if output.Token == nil {
+			return errors.Errorf("invalid issue action: token of output at index [%d] is nil", j)
 		}
 		data, err := utils.FromG1Proto(output.Token.Data)
 		if err != nil {
--- a/token/core/zkatdlog/nogh/v1/transfer/action.go
+++ b/token/core/zkatdlog/nogh/v1/transfer/action.go
@@ -266,6 +266,9 @@ func (t *Action) IsRedeemAt(index int) bool {
 	if index < 0 || index >= len(t.Outputs) {
 		return false
 	}
+	if t.Outputs[index] == nil {
+		return false
+	}
 
 	return t.Outputs[index].IsRedeem()
 }
@@ -472,8 +475,11 @@ func (t *Action) Deserialize(raw []byte) error {
 	// outputs
 	t.Outputs = make([]*token.Token, len(action.Outputs))
 	for j, output := range action.Outputs {
-		if output == nil || output.Token == nil {
+		if output == nil {
 			return errors.Errorf("invalid transfer action: output at index [%d] is nil", j)
+		}
+		if output.Token == nil {
+			return errors.Errorf("invalid transfer action: token of output at index [%d] is nil", j)
 		}
 		data, err := utils.FromG1Proto(output.Token.Data)
 		if err != nil {
--- a/token/core/zkatdlog/nogh/v1/transfer/action.go
+++ b/token/core/zkatdlog/nogh/v1/transfer/action.go
@@ -191,20 +191,29 @@ func (t *Action) NumInputs() int {
 	return len(t.Inputs)
 }
 
-// GetInputs returns the identifiers of the tokens spent in the Action
+// GetInputs returns the identifiers of the tokens spent in the Action.
+// A nil ActionInput entry produces a nil *token2.ID at the corresponding index,
+// which is consistent with the slice-length contract used by callers.
 func (t *Action) GetInputs() []*token2.ID {
 	res := make([]*token2.ID, len(t.Inputs))
 	for i, input := range t.Inputs {
+		if input == nil {
+			res[i] = nil
+
+			continue
+		}
 		res[i] = input.ID
 	}
 
 	return res
 }
 
-// GetSerializedInputs returns the serialized tokens spent in the Action
+// GetSerializedInputs returns the serialized tokens spent in the Action.
+// Returns an error instead of panicking when an input's Token field is nil
+// (which Validate() guards against, but some code paths skip Validate).
 func (t *Action) GetSerializedInputs() ([][]byte, error) {
 	var res [][]byte
-	for _, input := range t.Inputs {
+	for i, input := range t.Inputs {
 		if input == nil {
 			res = append(res, nil)
 
@@ -218,6 +227,9 @@ func (t *Action) GetSerializedInputs() ([][]byte, error) {
 			res = append(res, ser)
 
 			continue
+		}
+		if input.Token == nil {
+			return nil, errors.Errorf("nil token in input at index [%d]", i)
 		}
 		r, err := input.Token.Serialize()
 		if err != nil {
@@ -251,6 +263,10 @@ func (t *Action) GetOutputs() []driver.Output {
 
 // IsRedeemAt checks if output in the Action at the passed index is redeemed
 func (t *Action) IsRedeemAt(index int) bool {
+	if index < 0 || index >= len(t.Outputs) {
+		return false
+	}
+
 	return t.Outputs[index].IsRedeem()
 }
 
@@ -265,8 +281,17 @@ func (t *Action) IsRedeem() bool {
 	return false
 }
 
-// SerializeOutputAt marshals the output in the Action at the passed index
+// SerializeOutputAt marshals the output in the Action at the passed index.
+// Returns an error when index is out of bounds or the output entry is nil
+// instead of panicking.
 func (t *Action) SerializeOutputAt(index int) ([]byte, error) {
+	if index < 0 || index >= len(t.Outputs) {
+		return nil, errors.Errorf("SerializeOutputAt: index [%d] out of bounds (len=%d)", index, len(t.Outputs))
+	}
+	if t.Outputs[index] == nil {
+		return nil, errors.Errorf("SerializeOutputAt: nil output at index [%d]", index)
+	}
+
 	return t.Outputs[index].Serialize()
 }
 
@@ -275,6 +300,9 @@ func (t *Action) GetSerializedOutputs() ([][]byte, error) {
 	res := make([][]byte, len(t.Outputs))
 	var err error
 	for i, token := range t.Outputs {
+		if token == nil {
+			return nil, errors.Errorf("nil output entry at index %d", i)
+		}
 		res[i], err = token.Serialize()
 		if err != nil {
 			return nil, err
@@ -445,7 +473,7 @@ func (t *Action) Deserialize(raw []byte) error {
 	t.Outputs = make([]*token.Token, len(action.Outputs))
 	for j, output := range action.Outputs {
 		if output == nil || output.Token == nil {
-			continue
+			return errors.Errorf("invalid transfer action: output at index [%d] is nil", j)
 		}
 		data, err := utils.FromG1Proto(output.Token.Data)
 		if err != nil {
@@ -485,11 +513,16 @@ func (t *Action) GetProof() []byte {
 	return t.Proof
 }
 
-// GetOutputCommitments returns the cryptographic commitments of the outputs
+// GetOutputCommitments returns the cryptographic commitments of the outputs.
+// It skips nil output entries defensively; callers (validators) ensure all outputs
+// are non-nil via Action.Validate() before this method is reached.
 func (t *Action) GetOutputCommitments() []*math.G1 {
-	com := make([]*math.G1, len(t.Outputs))
-	for i := range com {
-		com[i] = t.Outputs[i].Data
+	com := make([]*math.G1, 0, len(t.Outputs))
+	for _, out := range t.Outputs {
+		if out == nil {
+			continue
+		}
+		com = append(com, out.Data)
 	}
 
 	return com
--- a/token/core/zkatdlog/nogh/v1/transfer/action_test.go
+++ b/token/core/zkatdlog/nogh/v1/transfer/action_test.go
@@ -729,9 +729,10 @@ func TestAction_Deserialize_ErrorPaths(t *testing.T) {
 }
 
 func TestAction_Deserialize_ExtraBranches(t *testing.T) {
-	// A nil output element is now rejected during deserialization (security fix FA3):
-	// before the fix, nil outputs were silently left as nil in t.Outputs, creating
-	// an inconsistent struct that could cause nil-pointer panics further down the pipeline.
+	// A nil output element is now rejected during deserialization.
+	// This is the FA3 (False Acceptance #3) attack vector; see
+	// TestSecurityFalseAcceptanceTransferDeserializeNilOutput in
+	// validator/validator_security_test.go for the full threat description.
 	protoNilOutput := &actions.TransferAction{
 		Version: 1, // ProtocolV1
 		Outputs: []*actions.TransferActionOutput{
@@ -757,7 +758,7 @@ func TestAction_Deserialize_ExtraBranches(t *testing.T) {
 	action = &transfer.Action{}
 	err = action.Deserialize(raw)
 	require.Error(t, err)
-	assert.Contains(t, err.Error(), "output at index [0] is nil")
+	assert.Contains(t, err.Error(), "token of output at index [0] is nil")
 }
 
 func TestAction_SerializeOutputAt(t *testing.T) {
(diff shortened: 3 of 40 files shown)


token/core/zkatdlog/nogh/v1/issue/action.go
```diff
@@ -302,7 +302,7 @@ func (i *Action) Deserialize(raw []byte) error {
 	i.Outputs = make([]*token.Token, len(issueAction.Outputs))
 	for j, output := range issueAction.Outputs {
 		if output == nil || output.Token == nil {
-			continue
+			return errors.Errorf("invalid issue action: output at index [%d] is nil", j)
```
AkramBitar, line 305: or outpToken == nil, Maybe we need from the log message to be able to know which error we had.

token/core/zkatdlog/nogh/v1/transfer/action.go
```diff
@@ -191,20 +191,29 @@ func (t *Action) NumInputs() int {
...
 }
 
-// GetInputs returns the identifiers of the tokens spent in the Action
+// GetInputs returns the identifiers of the tokens spent in the Action.
+// A nil ActionInput entry produces a nil *token2.ID at the corresponding index,
+// which is consistent with the slice-length contract used by callers.
 func (t *Action) GetInputs() []*token2.ID {
 	res := make([]*token2.ID, len(t.Inputs))
 	for i, input := range t.Inputs {
+		if input == nil {
```
AkramBitar, line 200: What if all inputs are nill?

token/core/zkatdlog/nogh/v1/transfer/action.go
```diff
@@ -191,20 +191,29 @@ func (t *Action) NumInputs() int {
...
-// GetSerializedInputs returns the serialized tokens spent in the Action
+// GetSerializedInputs returns the serialized tokens spent in the Action.
+// Returns an error instead of panicking when an input's Token field is nil
+// (which Validate() guards against, but some code paths skip Validate).
 func (t *Action) GetSerializedInputs() ([][]byte, error) {
 	var res [][]byte
-	for _, input := range t.Inputs {
+	for i, input := range t.Inputs {
 		if input == nil {
 			res = append(res, nil)
```
AkramBitar, line 218: What if all inputs are nil?

token/core/zkatdlog/nogh/v1/transfer/action.go
```diff
@@ -251,6 +263,10 @@ func (t *Action) GetOutputs() []driver.Output {
 
 // IsRedeemAt checks if output in the Action at the passed index is redeemed
 func (t *Action) IsRedeemAt(index int) bool {
+	if index < 0 || index >= len(t.Outputs) {
+		return false
+	}
+
 	return t.Outputs[index].IsRedeem()
```
AkramBitar, line 273: I am wondering if  t.Outputs[index] can be nil.

token/core/zkatdlog/nogh/v1/transfer/action.go
```diff
@@ -445,7 +473,7 @@ func (t *Action) Deserialize(raw []byte) error {
 	t.Outputs = make([]*token.Token, len(action.Outputs))
 	for j, output := range action.Outputs {
 		if output == nil || output.Token == nil {
-			continue
+			return errors.Errorf("invalid transfer action: output at index [%d] is nil", j)
```
AkramBitar, line 479: or output.Token = nil. Maybe we need from the log message to be able to know which error we had.

token/core/zkatdlog/nogh/v1/transfer/action.go
```diff
@@ -485,11 +513,16 @@ func (t *Action) GetProof() []byte {
...
+// GetOutputCommitments returns the cryptographic commitments of the outputs.
+// It skips nil output entries defensively; callers (validators) ensure all outputs
+// are non-nil via Action.Validate() before this method is reached.
 func (t *Action) GetOutputCommitments() []*math.G1 {
-	com := make([]*math.G1, len(t.Outputs))
-	for i := range com {
-		com[i] = t.Outputs[i].Data
+	com := make([]*math.G1, 0, len(t.Outputs))
+	for _, out := range t.Outputs {
+		if out == nil {
```
AkramBitar, line 528: what happens if all out are nil. Is that valid?

token/core/zkatdlog/nogh/v1/transfer/action.go
```diff
@@ -485,11 +513,16 @@ func (t *Action) GetProof() []byte {
...
 
-// GetOutputCommitments returns the cryptographic commitments of the outputs
+// GetOutputCommitments returns the cryptographic commitments of the outputs.
+// It skips nil output entries defensively; callers (validators) ensure all outputs
+// are non-nil via Action.Validate() before this method is reached.
 func (t *Action) GetOutputCommitments() []*math.G1 {
-	com := make([]*math.G1, len(t.Outputs))
-	for i := range com {
-		com[i] = t.Outputs[i].Data
+	com := make([]*math.G1, 0, len(t.Outputs))
```
AkramBitar, line 526: In case len(t.Outputs), is that valid and do we need to return empty slice?

token/core/zkatdlog/nogh/v1/transfer/action_test.go
```diff
@@ -729,24 +729,35 @@ func TestAction_Deserialize_ErrorPaths(t *testing.T) {
 }
 
 func TestAction_Deserialize_ExtraBranches(t *testing.T) {
-	// Test with nil output and nil output.Token
-	protoAction := &actions.TransferAction{
+	// A nil output element is now rejected during deserialization (security fix FA3):
```
AkramBitar: (security fix FA3) what is FA3?

token/core/zkatdlog/nogh/v1/transfer/typeandsum.go
```diff
@@ -105,7 +105,7 @@ func (p *TypeAndSumProof) Validate(curveID math.CurveID) error {
 	if err := math2.CheckElement(p.CommitmentToType, curveID); err != nil {
 		return errors.Join(err, ErrInvalidCommitmentToType, ErrInvalidSumAndTypeProof)
 	}
-	if err := math2.CheckZrElements(p.InputBlindingFactors, curveID, uint64(len(p.InputBlindingFactors))); err != nil {
+	if err := math2.CheckZrElements(p.InputBlindingFactors, curveID, uint64(len(p.InputValues))); err != nil {
 		return errors.Join(err, ErrInvalidInputBlindingFactors, ErrInvalidSumAndTypeProof)
```
AkramBitar, line 109: Yo changed uint64(len(p.InputValues) in the math2.CheckZrElements
However in the return error you return ErrInvalidInputBlindingFactors, ErrInvalidSumAndTypeProof)

AkramBitar (changes requested):

token/core/zkatdlog/nogh/v1/transfer/typeandsum.go
adecaro: this is correct, because we must have as many InputBlindingFactors as InputValue. At verification time, then, we check that InputValues has a certain length

token/core/zkatdlog/nogh/v1/transfer/action.go
adecaro: `GetInputs` is intentionally not the enforcement point here. `Action.Validate()` already rejects any nil `ActionInput` entry with `ErrEmptyInput`, and also returns `ErrInvalidInputs` when the slice is entirely empty. `Deserialize` also rejects nil entries during unmarshalling. Every pipeline path that does something meaningful with `GetInputs` calls `Validate()` first.

The nil pass-through preserves the slice-length contract (a nil entry produces a nil ID at the same index), which callers rely on when pairing inputs with outputs by position. Happy to add a hard-fail here too if you prefer, but it would be redundant with the existing gate.

token/core/zkatdlog/nogh/v1/transfer/action.go
adecaro: Same reasoning as the `GetInputs` thread above. `Validate()` is the enforcement gate and rejects nil entries before `GetSerializedInputs` is ever reached in any valid pipeline. The nil append preserves index alignment so callers can pair the byte-slices with outputs by position. A zero-length inputs slice is caught by `Validate()` via `ErrInvalidInputs`.

token/core/zkatdlog/nogh/v1/issue/action.go
adecaro: Good catch. I've split the combined condition into two separate checks with distinct messages — pushed in the latest commit — so the log tells you exactly which field was nil:

```go
if output == nil {
    return errors.Errorf("invalid issue action: output at index [%d] is nil", j)
}
if output.Token == nil {
    return errors.Errorf("invalid issue action: token of output at index [%d] is nil", j)
}
```

Same split applied to the parallel block in `transfer/action.go` (comment below).

token/core/zkatdlog/nogh/v1/transfer/action.go
adecaro: Yes, it can be nil in adversarial or partially-constructed structs — exactly the scenario this PR targets. Fixed in the commit just pushed: a nil-element guard is added after the bounds check:

```go
if t.Outputs[index] == nil {
    return false
}
```

This extends the existing "out of range → false" contract to nil elements.

token/core/zkatdlog/nogh/v1/transfer/action.go
adecaro: Agreed. Split into two distinct checks in the commit just pushed, consistent with the same fix in `issue/action.go`:

```go
if output == nil {
    return errors.Errorf("invalid transfer action: output at index [%d] is nil", j)
}
if output.Token == nil {
    return errors.Errorf("invalid transfer action: token of output at index [%d] is nil", j)
}
```

token/core/zkatdlog/nogh/v1/transfer/action_test.go
adecaro: `FA3` stands for **False Acceptance attack #3** — it is the codename used consistently across the security test suite for the nil-output-in-deserialization vector. The full definition lives in `validator/validator_security_test.go` in the docstring of `TestSecurityFalseAcceptanceTransferDeserializeNilOutput`:

> Attack vector (FA3): before the fix, nil output proto fields were silently left as nil in t.Outputs. Between Deserialize() and Validate(), the struct was in an inconsistent state that could cause nil-pointer dereferences in any code that accessed t.Outputs[i].Data directly (e.g., GetOutputCommitments).

The cross-reference in this test is intentional — it ties the lower-level unit test back to the named security scenario. I've updated the comment in the commit just pushed to make this self-explanatory:

```go
// A nil output element is now rejected during deserialization.
// This is the FA3 (False Acceptance #3) attack vector; see
// TestSecurityFalseAcceptanceTransferDeserializeNilOutput in
// validator/validator_security_test.go for the full threat description.
```

token/core/zkatdlog/nogh/v1/transfer/typeandsum.go
adecaro: Thanks for confirming — that's exactly the intent. The two-phase validation is:
1. Here in `Validate()`: assert `|InputBlindingFactors| == |InputValues|` and both slices contain well-formed Zr elements.
2. In `TypeAndSumVerifier.Verify()`: assert `|InputValues| == len(v.Inputs)` (the circuit width), checked against the public parameters.

The original self-referential check collapsed both phases into one that could never fail for a short blinding-factor slice. The fix restores the intended invariant.

adecaro: The error codes are correct and intentional. `CheckZrElements(p.InputBlindingFactors, curveID, uint64(len(p.InputValues)))` checks that `InputBlindingFactors` is a well-formed slice of *exactly* `len(p.InputValues)` valid Zr elements. When it fails, the invariant violated is specifically about `InputBlindingFactors` (wrong count or a malformed element), so `ErrInvalidInputBlindingFactors` is the right sentinel.

The previous code passed `uint64(len(p.InputBlindingFactors))` as the expected count — a self-referential check that trivially passes for any slice length, including an adversarially short one. Switching to `len(p.InputValues)` as the authoritative expected count is the bug fix; the error codes are unchanged. See also adecaro's confirmation in this thread.

token/core/zkatdlog/nogh/v1/transfer/action.go
adecaro: `len(t.Outputs) == 0` is rejected upstream by `Validate()` (`ErrInvalidOutputs`), so that path is never legitimately reachable. If somehow reached, the code returns a non-nil empty slice because `make([]*math.G1, 0, 0)` always allocates an empty slice — no nil-pointer issue here.

The switch from `make([]*math.G1, len(t.Outputs))` to `make([]*math.G1, 0, len(t.Outputs))` is intentional: the old fixed-length allocation left nil holes for nil entries; the new compact allocation avoids passing nil `*math.G1` pointers into the verifier, which would otherwise panic.

token/core/zkatdlog/nogh/v1/transfer/action.go
adecaro: No, all-nil outputs is not valid, and the system already enforces this: `Action.Validate()` returns `ErrInvalidOutputs` if `len(t.Outputs) == 0` and `ErrEmptyOutputToken` for any nil element. All validator paths call `Validate()` before reaching `GetOutputCommitments`.

The nil-skip here is a belt-and-suspenders guard for any code path that bypasses `Validate()`: it prevents a panic but produces a shorter-than-expected commitment slice that the ZK verifier then rejects on a length mismatch — so the outcome is still a clean error, not a silent pass.

AkramBitar (approved):
LGTM.

PR #1834, opened 2026-07-01
remove left over token-sdk strings

adecaro:
This PR removes left over `token-sdk` string from panurus.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,4 +1,4 @@
-# Panurs Configuration Example
+# Panurus Configuration Example
 
 The following example provides descriptions for the various keys required by Panurus.
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -1,4 +1,4 @@
-# Token-SDK Configuration Example
+# Panurs Configuration Example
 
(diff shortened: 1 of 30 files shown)


docs/configuration.md
```diff
@@ -1,4 +1,4 @@
-# Token-SDK Configuration Example
+# Panurs Configuration Example
```
AkramBitar: Typo should be Panurus (missing u before the s)

token/services/logging/logger.go
```diff
@@ -18,14 +18,21 @@ import (
 const loggerNameSeparator = "."
 
 // Logger provides logging API
-type Logger = logging.Logger
+type (
+	Logger = logging.Logger
+	Conifg = logging.Config
```
AkramBitar: typo with should be Config

AkramBitar (changes requested):

docs/configuration.md
adecaro: yes

token/services/logging/logger.go
adecaro: yes

AkramBitar (approved):

PR #1835, opened 2026-07-01
Check withdrawal requester owns its recipient identity before registering it

### LINKED ISSUE #1836: Confirm that withdrawal requester owns its recipient identity before registering it

AkramBitar:
Problem
When an issuer received a withdrawal request, it registered the recipient identity sent by the requester without checking that the sender actually owns it. Any member who could open a session to an issuer could register an identity it doesn't control, causing issued tokens to be misdirected or stranded and corrupting auditor records.

Fix
The requester now signs a proof with its recipient key and includes it (plus a nonce) in the withdrawal request. The issuer rebuilds the same message and verifies the signature before registering the identity — rejecting anything empty or invalid. It reuses the existing proof helpers, adds no extra network round-trips, and binds the TMS, nonce, session, and context to prevent replay.

### DIFF
--- a/token/services/ttx/withdrawal.go
+++ b/token/services/ttx/withdrawal.go
@@ -23,6 +23,13 @@ type WithdrawalRequest struct {
 	TokenType     token2.Type
 	Amount        uint64
 	NotAnonymous  bool
+
+	// Nonce keeps each proof unique.
+	Nonce []byte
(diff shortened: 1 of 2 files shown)


AkramBitar:
Following a chat with @adecaro closing this one.

PR #1837, opened 2026-07-01
cmd: ski cleanup cli

### LINKED ISSUE #1844: cmd: skicleanup
This issue asks to introduce a new cmd: `skicleanup`. A `signers` command checks that entries in the id_signers table are assigned to a token in the tokens table. If not, the cmd reports its corresponding SKIs to signal that they can be removed. At the end of the execution, the tool reports some stats as well.

This issue calls also for an extension of the storage service in the following way:
- possibility to change the table names
- possibility to remove the fsc prefix

adecaro:
This PR introduces a new `cmd`: `skicleanup`. The `signers` command checks that entries in the `id_signers` table are assigned to a token in the `tokens` table. If not, the cmd reports its corresponding SKIs to signal that they can be removed.
At the end of the execution, the tool reports some stats as well. 

As a side effect of the tool, this PR extends the storage service in the following way:
- possibility to change the table names
- possibility to remove the `fsc` prefix


integration/ports.go
```diff
@@ -60,8 +60,8 @@ var (
 
 	AllTestTypes = []*InfrastructureType{
 		WebSocketNoReplication,
-		LibP2PNoReplication,
-		WebSocketWithReplication,
+		// LibP2PNoReplication,
```
AkramBitar: Is that ok to disable these?

cmd/skicleanup/cobra/signers/runner.go
```diff
@@ -0,0 +1,144 @@
...
+				continue
+			}
+			if result.notInToken {
+				s.notInToken++
+			} else {
+				s.skipped++
+			}
+		}
+
+		if len(entries) < batchSize {
```
AkramBitar, line 64: In case batchSize is set to 0 then the if will not break, maybe this will cause infinite loop.

Maybe we need to add at the begining og the run a check that 
 if batchSize <= 0 {
              return error ...
      }

cmd/skicleanup/cobra/config/cmd.go
```diff
@@ -0,0 +1,104 @@
...
+#   dataSource: /var/lib/panurus/node/data.db
+dataSource: "postgres://user:pass@localhost:5432/panurus?sslmode=disable"
+
+# tablePrefix is the optional prefix that was used when the Panurus node created
+# its database tables. Leave empty if the node was configured without a prefix.
+tablePrefix: ""
+
+# skipPrefix controls whether the FSC-generated prefix is omitted from all table
+# names. Set this to true when the Panurus node was started with
+# token.storage.skipPrefix: true. Default is false.
```
AkramBitar, line 41: Just wondering where token.storage.skipPrefix actually gets used on the code? The only place I see reading it is LoadStorageConfig, and I can't find anything that calls it? (maybe I miss something here)

AkramBitar (changes requested):

token/services/storage/db/sql/common/tokens.go
```diff
@@ -839,6 +839,28 @@ func (db *TokenStore) GetTokens(ctx context.Context, inputs ...*token.ID) ([]*to
...
 }
 
+// HasTokenForIdentity returns true if any token in the Tokens table has the given owner identity
+// and owner type, regardless of whether the token has been spent or not.
+func (db *TokenStore) HasTokenForIdentity(ctx context.Context, ownerIdentity []byte, ownerType string) (bool, error) {
+	query, args := q.Select().
+		FieldsByName("tx_id").
+		From(q.Table(db.table.Tokens)).
+		Where(cond.And(
+			cond.Eq("owner_identity", ownerIdentity),
```
AkramBitar, line 849: If we have a lot of rows, maybe we need to consider indexing (owner_identity, owner_type)

AkramBitar (changes requested):

token/services/storage/db/sql/common/tokens.go
adecaro: good point

integration/ports.go
adecaro: good catch

cmd/skicleanup/cobra/config/cmd.go
adecaro: it is used in `token/services/storage/db/sql/common/init.go`:

```go
func GetTableNamesWithConfig(prefix string, cfg StorageConfig, params ...string) (TableNames, error) {
	if cfg.SkipPrefix {
		return GetTableNamesWithOverridesSkipPrefix(prefix, cfg.TableNames, params...)
	}

	return GetTableNamesWithOverrides(prefix, cfg.TableNames, params...)
}
```
No?

token/services/storage/db/sql/common/tokens.go
adecaro: indeed

cmd/skicleanup/cobra/signers/runner.go
adecaro: good point

AkramBitar (approved):

PR #1840, opened 2026-07-01
fix(boolpolicy): bound recursive-descent depth and input length to prevent stack exhaustion

### LINKED ISSUE #1839: Unbounded recursive descent in `boolpolicy` parser causes goroutine stack exhaustion
`token/services/identity/boolpolicy/parser.go` implements a recursive-descent parser for boolean
policy expressions with **no depth limit and no input-length cap**. An attacker who can set a
transfer output's `Owner` field to a `boolpolicy`-typed identity can supply a deeply-nested
parenthesis string that drives the parser to goroutine stack exhaustion. Go's stack overflow is a
**fatal runtime abort** (`runtime.throw`), not a recoverable panic — `defer/recover` cannot catch
it, so the chaincode container (and any FSC node deserializing the same field) **crashes
unconditionally**.

## Affected code

| File | Symbol | Issue |
|------|--------|-------|
| `token/services/identity/boolpolicy/parser.go:237` | `parsePrimary` | calls `parseOr` on every `(` with no depth counter |
| `token/services/identity/boolpolicy/parser.go:172` | `Parse` | no length check before tokenising |
| `token/core/fabtoken/v1/driver/deserializer.go` | `AddTypedVerifierDeserializer` | registers `boolpolicy` parser on the `Owner` field path |
| `token/core/zkatdlog/nogh/v1/driver/deserializer.go` | `AddTypedVerifierDeserializer` | same, for the zkatdlog driver |

## Exploit scenario

1. Attacker submits a transfer whose output `Owner` is a `TypedIdentity{Type: boolpolicy}` with a
   policy string of `((((...))))` nested ~100 000 levels deep (~200 KB payload, well within a
   Fabric transaction size limit).
2. During `TransferActionValidate` the on-chain validator calls
   `TypedIdentityDeserializer.DeserializeVerifier`, which calls `Parse(pi.Policy)`.
3. `parsePrimary` → `parseOr` → `parseAnd` → `parsePrimary` recurs ~3× per nesting level,
   producing ~300 000 stack frames.
4. The Go runtime emits `runtime: goroutine stack exceeds 1000000000-byte limit` and calls
   `runtime.throw` — a **non-recoverable** abort.
5. The chaincode container exits; endorsement halts until the peer restarts it.

## Impact

- **Availability**: chaincode container crash; endorsement halted network-wide until the peer
  restarts the container.
- **Affected parties**: any channel member submitting a transfer can trigger this with a single
  transaction.
- **`recover()` does NOT help**: `tcc.go`'s `defer recover()` only catches `panic`, not
  `runtime.throw`.

## Preconditions

- Attacker holds a channel identity capable of submitting a transfer (i.e. any enrolled member).
- The `boolpolicy` type is registered in the deserializer multiplex of either shipped driver
  (confirmed for both `fabtoken` and `zkatdlog`).

## Suggested fix

Two independent guards, both required:

1. **Input-length cap** — reject any policy string longer than a fixed limit (e.g. 4 KB) at the
   `Parse` entry point, before any tokenisation.
2. **Depth counter** — thread a `depth int` parameter through `parseOr` / `parseAnd` /
   `parsePrimary`; return an error when `depth >= maxDepth` (e.g. 64) instead of recursing.

## Severity

**Medium** — unauthenticated availability impact against a core chaincode component; not caught by
any existing recovery mechanism.

adecaro:
## Problem

`boolpolicy/parser.go` had no defence against attacker-supplied input:

- **No input-length cap** — `Parse` passed arbitrarily long strings directly to the lexer.
- **No depth limit** — `parsePrimary` called `parseOr` on every `(` token with no counter,
  allowing unbounded mutual recursion (`parsePrimary → parseOr → parseAnd → parsePrimary → …`).

A policy string with N nested parentheses produces ~3 N Go stack frames. At ~100 000 levels the
Go runtime emits `runtime: goroutine stack exceeds 1000000000-byte limit` and aborts the process
via `runtime.throw` — **not** a `panic`, so no `defer/recover` wrapper can intercept it. Both
production drivers (`fabtoken` and `zkatdlog`) register the `boolpolicy` deserializer on the
`Owner` field path, meaning a single malicious transfer transaction is sufficient to crash the
chaincode container.

## Solution

Two complementary guards added entirely within `parser.go`; no public API or wire-format change.

### 1 — Input-length cap (`maxPolicyLen = 4096`)

```go
// Parse, first line of the function body
if len(input) > maxPolicyLen {
    return nil, fmt.Errorf("policy expression exceeds maximum length of %d bytes (got %d)", maxPolicyLen, len(input))
}
```

Stops the attack before the lexer is created. 4 KB is generous for any legitimate policy.

### 2 — Nesting depth cap (`maxParseDepth = 64`)

`parseOr`, `parseAnd`, and `parsePrimary` each receive a `depth int` parameter.  
`depth` is passed **unchanged** through `OR`/`AND` chaining (no new call per infix token) and
incremented by exactly **one** when descending into a parenthesised sub-expression:

```go
case tokLParen:
    if depth >= maxParseDepth {
        p.err = fmt.Errorf("policy expression exceeds maximum nesting depth of %d", maxParseDepth)
        return nil
    }
    p.advance()
    node := p.parseOr(depth + 1)
```

At the limit the Go call-stack depth is bounded to ~3 × 64 = ~192 frames — negligible.

## Files changed

| File | Change |
|------|--------|
| `token/services/identity/boolpolicy/parser.go` | Add `maxPolicyLen`/`maxParseDepth` constants; guard in `Parse`; thread `depth` through recursive methods |
| `token/services/identity/boolpolicy/parser_test.go` | Three new security regression tests |

## Testing

Three new tests in `parser_test.go` (all in the same `boolpolicy` package so the unexported
constants are accessible):

| Test | Assertion |
|------|-----------|
| `TestErrorInputTooLong` | A `maxPolicyLen+1`-byte string is rejected with "exceeds maximum length" |
| `TestErrorNestingTooDeep` | `maxParseDepth+1` nested parentheses are rejected with "exceeds maximum nesting depth" |
| `TestNestingAtLimitIsAllowed` | Exactly `maxParseDepth` nested parentheses are accepted (fence-post check) |

Full package test run — all 74 tests pass:

```
ok  github.com/LFDT-Panurus/panurus/token/services/identity/boolpolicy  2.171s
```

## Backwards compatibility

- **Wire format**: unchanged — `PolicyIdentity` serialization is not touched.
- **Public API**: `Parse(string) (Node, error)` signature is unchanged; callers that previously
  passed oversized or over-nested input will now receive an error rather than crashing.
- **Limits**: 4 KB / depth 64 are deliberately conservative. Real-world policies in the existing
  test suite are well under 100 bytes and 3 levels deep.

## Checklist

- [x] Signed-off (`git commit -s`)
- [x] `make checks` passes
- [x] `make unit-tests` passes
- [x] No new exported symbols; no wire-format change
- [x] Security regression tests added

### DIFF
--- a/token/services/identity/boolpolicy/parser.go
+++ b/token/services/identity/boolpolicy/parser.go
@@ -25,6 +25,19 @@ import (
 	"unicode"
 )
 
+const (
+	// maxPolicyLen is the maximum byte length accepted by Parse.
+	// Policies longer than 4 KB cannot represent a sensible real-world
+	// expression and are almost certainly an attack or a bug.
+	maxPolicyLen = 4 * 1024
+
+	// maxParseDepth is the maximum parenthesis nesting depth allowed by the
+	// recursive-descent parser.  Each open parenthesis adds one frame to the
+	// Go call stack; capping at 64 prevents goroutine stack exhaustion from
+	// attacker-supplied deeply-nested expressions.
+	maxParseDepth = 64
+)
+
 // ---------------------------------------------------------------------------
 // AST nodes
 // ---------------------------------------------------------------------------
@@ -169,14 +182,24 @@ type parser struct {
 
 // Parse parses a boolean expression string and returns the root AST node.
 // It returns an error for any lexical or syntactic problems.
+//
+// Parse enforces two hard limits to prevent resource exhaustion:
+//   - input longer than maxPolicyLen bytes is rejected immediately.
+//   - parenthesis nesting deeper than maxParseDepth levels is rejected;
+//     this bounds the Go call-stack depth of the recursive descent and
+//     prevents goroutine stack exhaustion from attacker-supplied input.
 func Parse(input string) (Node, error) {
+	if len(input) > maxPolicyLen {
+		return nil, fmt.Errorf("policy expression exceeds maximum length of %d bytes (got %d)", maxPolicyLen, len(input))
+	}
+
 	p := &parser{lex: newLexer(input)}
 	p.advance() // prime the lookahead
 	if p.err != nil {
 		return nil, p.err
 	}
 
-	node := p.parseOr()
+	node := p.parseOr(0)
 	if p.err != nil {
 		return nil, p.err
 	}
@@ -196,11 +219,11 @@ func (p *parser) advance() {
 
 // parseOr handles: and_expr ( 'OR' and_expr )*
 // OR is left-associative and lower precedence than AND.
-func (p *parser) parseOr() Node {
-	left := p.parseAnd()
+func (p *parser) parseOr(depth int) Node {
+	left := p.parseAnd(depth)
 	for p.err == nil && p.current.kind == tokOr {
 		p.advance()
-		right := p.parseAnd()
+		right := p.parseAnd(depth)
 		left = &OrNode{Left: left, Right: right}
 	}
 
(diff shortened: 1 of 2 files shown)


AkramBitar (approved):
LGTM

PR #1843, opened 2026-07-02
Confirm that withdrawal requester owns its recipient identity before registering it  #1836

### LINKED ISSUE #1836: Confirm that withdrawal requester owns its recipient identity before registering it

adecaro:


docs/services/ttx.md
```diff
@@ -11,28 +11,24 @@ The lifecycle of a token transaction typically involves the following stages, co
 ```mermaid
 sequenceDiagram
     autonumber
-    participant Initiator
```
AkramBitar, line 14: When I view this file I see the following error (see picture)

Image

token/services/ttx/withdrawal.go
```diff
@@ -80,20 +107,36 @@ func (r *RequestWithdrawalView) Call(context view.Context) (any, error) {
...
+	if len(challenge.Nonce) == 0 {
+		return nil, errors.New("withdrawal challenge missing nonce")
+	}
+
+	// Sign the issuer-controlled nonce to prove key ownership.
+	message, err := buildAttestationMessage(*tmsID, nil, recipientData.Identity, false, "", challenge.Nonce, s.Info().ID, context.ID())
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to build attestation message")
+	}
+	sig, err := signRecipientAttestation(context.Context(), w, message, recipientData.Identity, true)
```
AkramBitar: When freshPath is true and we have a remote wallet then this will fail always with error.


func signRecipientAttestation(ctx context.Context, w *token.OwnerWallet, message []byte, identity view.Identity, freshPath bool) ([]byte, error) {
	if w.Remote() {
		if freshPath {
			return nil, errors.Errorf("remote wallet [%s] cannot attest fresh recipient identity locally", w.ID())
		}

		return nil, nil
	}

AkramBitar (changes requested):

docs/services/ttx.md
adecaro: it was from before. Could you submit a PR to fix that, please 🙏

token/services/ttx/withdrawal.go
adecaro: fixed

AkramBitar (approved):
LGTM

Issue #1553, opened 2026-04-15
Add benchmarks to track performance over time


AkramBitar:

Soumya8898:
Hi @AkramBitar , can you please describe the issue a bit and assign it to me.

adecaro:
@AkramBitar , can you help here, please? 🙏

AkramBitar:
> [@AkramBitar](https://github.com/AkramBitar) , can you help here, please? 🙏

Hi @adecaro and @Soumya8898, 
@Effi-S already started to work on that.

Soumya8898:
Hi @Effi-S 
Are you currently working on this? 

I’ve been exploring how to create realistic test fixtures (e.g., fake token databases, ZK parameters, wallet state), and I think those could be useful here for testing.
If you are working on this and are open to suggestions, I’d be happy to share them. Otherwise, if you’re not working on it, I can ask @AkramBitar or @adecaro  to assign it to me.

AkramBitar:
Closing dup of Over Night benchmarks
https://github.com/LFDT-Panurus/panurus/issues/1973

PR #1849, opened 2026-07-03
perf(fabricx/finality): batch finality via shared poller

### LINKED ISSUE #1848: perf: FabricX finality resolution serializes on per-transaction status queries
## Problem

`NSListenerManager.AddFinalityListener` (`token/services/network/fabricx/finality/finality.go`) resolves finality with one remote committer query per transaction:

1. At registration it enqueues a `TxCheck` that immediately calls `GetTransactionStatus` — for a freshly submitted tx this almost always returns not-yet-committed, so the round-trip is wasted.
2. Status delivery then relies on per-tx work items, each performing its own remote query, so every queue worker blocks on its own network round-trip.

Under sustained load this serializes the worker pool on network I/O. In our deployment (Fabric-X backend, sustained submission from 4000 concurrent clients), finality resolution settled ~92 tx/s while the chain committed ~1400 tx/s, accumulating >100k inflight finality watchers.

## Proposal

Replace the per-tx remote check with a single shared poller:

- registrations go into a pending set; a lazily started poller sweeps it on an interval
- statuses are resolved with one batched committer query per sweep chunk, using the batched `GetTransactionStatuses` recently added to the FabricX query service (hyperledger-labs/fabric-smart-client#1553), with a per-tx fallback when the query service doesn't support batching — so no FSC version bump is required
- token-request hashes for valid txs are fetched with one batched `GetStates` call, grouped by namespace
- terminal txs are handed to the existing worker pool as pre-resolved notification events with no network I/O on that path

This also lines up with the discussion on hyperledger-labs/fabric-smart-client#1553 about using the batched status query for finality.

We have been running this in our test environments under sustained load; PR follows.

EvanYan1024:
Fixes #1848.

## Motivation

`NSListenerManager` resolves FabricX finality with one remote committer query per transaction: a speculative `GetTransactionStatus` at registration (almost always seeing a not-yet-committed tx) plus per-tx work items that each block a queue worker on its own network round-trip. In our deployment (Fabric-X backend, sustained submission from 4000 concurrent clients), finality resolution settled ~92 tx/s while the chain committed ~1400 tx/s, accumulating >100k inflight finality watchers.

## Changes

- `AddFinalityListener` now adds the tx to a pending set instead of enqueuing a per-tx `TxCheck`; a single shared poller, started lazily on first registration, sweeps the pending set every poll interval.
- Each sweep resolves statuses in chunks with one committer query per chunk: it uses the batched `GetTransactionStatuses` (hyperledger-labs/fabric-smart-client#1553) when the query service provides it, and falls back to per-tx `GetTransactionStatus` otherwise. No FSC dependency bump is required; the batch path activates once the FSC dependency includes #1553. In the per-tx path a failing query only skips that tx (a not-yet-committed tx returns an error rather than a status), so one fresh tx cannot fail the whole chunk.
- A tx can have several finality waiters: the pending set keeps a waiter list per txID and notifies all of them on resolution, matching the semantics of the other listener managers.
- Token-request hashes for valid txs are fetched with one batched `GetStates` query, grouped by namespace.
- Terminal txs are handed to the existing worker pool as pre-resolved notification events that perform no network I/O; if the queue is full, the tx stays pending and is retried on the next sweep.
- Pending entries are reclaimed after a TTL so permanently dropped txs cannot grow the pending set; callers' own finality timeouts settle those waiters.
- Poll interval, batch size, and pending TTL are read from configuration (`token.finality.poller.interval` / `.batchSize` / `.pendingTTL`), following the existing `token.finality.notification.*` and `token.fabricx.lookup.*` config patterns; unset values fall back to defaults (1s / 2000 / 10m).
- `docs/configuration.md`: documented the new `token.finality.poller.*` keys and updated the now-stale note about the immediate status query at subscription time.

## Results

With this change our deployment's finality resolution improved ~5x immediately and stopped being network-bound (remaining limits in our runs were local, e.g. DB connection pool sizing). It has been running in our test environments under sustained load since.

## Notes for reviewers

- The wrapped notification-service `ListenerManager` is still constructed and held, but no longer used for status delivery. Happy to either drop it or re-wire it so push notifications complement the poller (poller as sweep/fallback) if that's the preferred direction.
- The poller goroutine starts lazily and runs for the lifetime of the process. If a shutdown hook is preferred, point me at the lifecycle to hang it on.


adecaro:
super intresting, @EvanYan1024. Thanks for this effort 🙏

adecaro (approved):
LGTM

adecaro:
@EvanYan1024 , if you can fix the current issues with the CI, I'll proceed with the merge. Thanks 🙏

PR #1850, opened 2026-07-03
feat(evm): initial foundation - design, crypto primitives and address/hash types

atharrva01:
First slice of the EVM network driver (Approach 2). This locks in the design we hammered out on #1701 and adds the basic crypto layer everything else will build on. No actual driver logic yet.

## What's included
- Updated design doc with the decisions we settled on: single-list StateDelta with a contract-side graphHiding flag, tokenID-only spend refs (since the EVM contract is stateful, we don't need to bind the token hash into the key like Fabric does), SHA-256 for request/public-params hashes, and quorum governance once the deployer seeds the initial state.
- Crypto helpers: keccak256 (via x/crypto/sha3) and SHA-256. The SHA-256 one matches the SDK's token-request hash so finality lines up later.
- Local Address (20B) and Hash (32B) types with hex/text/JSON encoding. No go-ethereum dependency, plus a build test that fails if it ever sneaks in (license issue for us).
- Basic package scaffolding under x/token/services/network/evm.
- Small go.sum fix in the integration module (missing badger entry) so the build works again.

## Not in this PR
StateDelta types, key derivation, EIP-712, and driver wiring are coming in follow-up PRs.

### DIFF
--- a/integration/go.mod
+++ b/integration/go.mod
@@ -59,6 +59,7 @@ require (
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
 	github.com/davidlazar/go-crypto v0.0.0-20200604182044-b73af7476f6c // indirect
 	github.com/decred/dcrd/dcrec/secp256k1/v4 v4.4.1 // indirect
+	github.com/dgraph-io/badger/v4 v4.9.2 // indirect
 	github.com/dgraph-io/ristretto/v2 v2.4.0 // indirect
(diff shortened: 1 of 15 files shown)


AkramBitar:
@atharrva01 

Thanks a lot for submitting this PR. Could you please open corresponding issue for it and link it to this PR? 

Regards,
Akram

atharrva01:
Hi @AkramBitar , this is LFX project , and this is the foundation pr of the etherum driver I m building , so how can I open a issue for it ? For each pr ?

Issue #1183, opened 2025-08-03
Introduce SQL Prepared Statements


adecaro:
For all the queries that are executed often, introduce a prepared statement.

AkramBitar:
@adecaro I can implement this. Do you know which quires we need to start with. Is that support should be in FSC?

sid200727:
Hi @adecaro! I've been working extensively in the SQL storage layer, recently merged the squirrel migration for SignerInfoStore, SimpleKeyDataStore, KeyValueStore, MetadataStore, EndorseTxStore, and EnvelopeStore (PRs #1199 and #1235 in fabric-smart-client). I'd like to take this on.
Based on your suggestion, I'll start by identifying the most frequently called queries in app/services/storage/db/sql/common and benchmarking the effect of prepared statements on those. Does starting with the token selection queries sound right to you?

adecaro:
Hi @sid200727 , please, go ahead. First, have an assessment if it makes sense to use prepared statements in our case and design also benchmarks to evaluate the improvements. Thanks much 🙏

sid200727:
Hi @adecaro! After exploring the codebase, here's my initial assessment:
Codebase Analysis ->
--> tokens.go - 22 query calls (highest in the package): heavy JOIN queries for unspent token lookups, soft-deletes on spend, ownership inserts - primary target
--> transactions.go - 13 query calls: movement/request queries with ORDER BY and LIMIT
--> identity.go - 8 query calls: configuration lookups by type and URL

Why tokens.go first :
These queries run on every token operation (select, spend, list). Preparing them once and reusing the compiled plan should reduce per-call overhead, especially under concurrent load.
Benchmark approach :
The existing token/services/benchmark framework (via GoBenchmarkParallel + RunBenchmark) supports exactly what's needed. 
My plan:
1. Write baseline benchmarks for the most frequent queries in tokens.go (without prepared statements)
2. Introduce prepared statements
3. Compare P50–P99 latency and allocations/op before and after
SQLite vs Postgres consideration :
SQLite caches query plans internally so the gain there may be minimal. Postgres should show a clearer improvement. I'll benchmark both to confirm whether the trade-off is worth the added complexity.

Does this approach sound right before I proceed?

adecaro:
Hi @sid200727 , apologies for this very late reply. The plan sounds great. Please, go ahead also in the light of the PR that introduces squirrel. Thanks much 🙏

sid200727:
[...]

PR #1855, opened 2026-07-06
test(tokens): add UnspentTokensIteratorBy prepared-statement comparison benchmark

sid200727:
## Summary

Adds a benchmark comparing `UnspentTokensIteratorBy` executed via the current dynamic query-building path against the same query executed via a statement prepared once and reused, to get concrete numbers for the discussion in #1183.

Closes to #1861

## Changes

-> Extracted the query-building logic from `UnspentTokensIteratorBy` into a standalone `buildUnspentTokensIteratorByQuery` function (pure refactor, no behavior change — verified via existing `TestTokens` suite, all 28 subtests pass including `ListUnspentTokensByWallets`)
-> Added `RunUnspentTokensIteratorByPreparedComparison` benchmark comparing:
   - `UnspentTokensIteratorBy_Dynamic`: current production path (query built + executed fresh each call)
   - `UnspentTokensIteratorBy_Prepared`: same query, prepared once via `readDB.PrepareContext`, reused via `stmt.QueryContext`
-> Wired into the existing Postgres benchmark suite as `BenchmarkTokenStorePreparedComparison`

## Results (median of 3 runs, same seed as existing baseline: 1000 tokens, wallet0/GOLD, 4 workers, 5s window)

|---|---|---|---|
| Throughput | 3,334 ops/s | 3,879 ops/s | +11% (range +6.5% to +17% across runs) |
| P99 latency | ~8.3ms | ~2.1ms | -73% (consistent across all runs) |
| Allocs/op | 131-132 | 27 | -80% (deterministic, zero variance) |

## Testing

Verified with `-count=3`, all runs consistent. Existing `TestTokens` suite passes unchanged, confirming the query-building extraction preserves behavior.

## Notes for the Reviewer

Related to #1183. This PR is the measurement step only, no production behavior changes. Happy to proceed with the actual prepared-statement implementation (pre-prepare fixed variants + fallback to dynamic building, per the design discussed on the issue) if these numbers make the case.

### DIFF
--- a/token/core/fabtoken/protos-go/v1/pp/ftpp.pb.go
+++ b/token/core/fabtoken/protos-go/v1/pp/ftpp.pb.go
@@ -12,12 +12,13 @@
 package pp
 
 import (
-	v1 "github.com/LFDT-Panurus/panurus/token/driver/protos-go/v1"
-	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
-	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	reflect "reflect"
 	sync "sync"
 	unsafe "unsafe"
+
+	v1 "github.com/LFDT-Panurus/panurus/token/driver/protos-go/v1"
+	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
+	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 )
 
 const (
--- a/token/core/fabtoken/protos-go/v1/pp/ftpp.pb.go
+++ b/token/core/fabtoken/protos-go/v1/pp/ftpp.pb.go
@@ -12,13 +12,12 @@
 package pp
 
 import (
-	reflect "reflect"
-	sync "sync"
-	unsafe "unsafe"
-
 	v1 "github.com/LFDT-Panurus/panurus/token/driver/protos-go/v1"
 	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
 	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
+	reflect "reflect"
+	sync "sync"
+	unsafe "unsafe"
 )
 
 const (
--- a/token/services/storage/db/sql/common/tokens_prepared_bench_util.go
+++ b/token/services/storage/db/sql/common/tokens_prepared_bench_util.go
@@ -53,6 +53,7 @@ func RunUnspentTokensIteratorByPreparedComparison(b *testing.B, store *TokenStor
 						break
 					}
 				}
+
 				return nil
 			},
 		)
@@ -91,6 +92,7 @@ func RunUnspentTokensIteratorByPreparedComparison(b *testing.B, store *TokenStor
 						break
 					}
 				}
+
 				return nil
 			},
 		)
--- a/token/services/storage/db/sql/common/tokens_prepared_bench_util.go
+++ b/token/services/storage/db/sql/common/tokens_prepared_bench_util.go
@@ -1,0 +1,99 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package common
+
+import (
+	"context"
+	"database/sql"
+	"testing"
+	"time"
+
+	"github.com/LFDT-Panurus/panurus/token/services/benchmark"
+	tokentype "github.com/LFDT-Panurus/panurus/token/token"
+)
+
+// RunUnspentTokensIteratorByPreparedComparison benchmarks UnspentTokensIteratorBy
+// against a version of the same query executed via a statement prepared once
+// outside the timed loop, using the same seed data, worker count, and duration
+// as RunTokenStoreBenchmarks so the numbers are directly comparable.
+//
+// This exists to answer: "how much would moving to prepared statements help
+// for our primary token-selection query?" (see #1183). The query built by
+// UnspentTokensIteratorBy is dynamic (its shape depends on whether walletID /
+// tokenType are empty), so this comparison fixes both parameters to non-empty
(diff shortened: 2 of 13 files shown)


AkramBitar:
Hello @sid200727 ,
Thanks a lot for your effort and work on that PR.
This looks promising 

See my comment in https://github.com/LFDT-Panurus/panurus/issues/1183#issuecomment-4892042750

If the query is dynamic, then I am wondering how do we determine which parameter combinations are actually common enough to justify pre-prepared statements? Wouldn’t identifying these combinations require runtime tracking/analysis, and could that add additional overhead or complexity?

Regards,
Akram

sid200727:
@AkramBitar, ready for re-review!
Thank You ☺️

token/core/fabtoken/protos-go/v1/pp/ftpp.pb.go
```diff
@@ -12,12 +12,13 @@
 package pp
 
 import (
-	v1 "github.com/LFDT-Panurus/panurus/token/driver/protos-go/v1"
-	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
-	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	reflect "reflect"
 	sync "sync"
 	unsafe "unsafe"
+
```
AkramBitar: Why we have changes in these protos files? Seems to me only a new line is added. 
Could you please revert if this is the case?

token/services/storage/db/sql/common/tokens_prepared_bench_util.go
```diff
@@ -0,0 +1,101 @@
+/*
```
AkramBitar, line 1: I'm wondering why we need to add a utility file. Wouldn't it be more convenient to move this code into the test file instead (token/services/storage/db/sql/postgres/tokens_bench_test.go)?

What do you think?

token/services/storage/db/sql/common/tokens_prepared_bench_util.go
```diff
@@ -0,0 +1,101 @@
...
+				return nil
+			},
+		)
+		result.Print()
+	})
+
+	b.Run("UnspentTokensIteratorBy_Prepared", func(b *testing.B) {
+		SeedBenchTokens(b, store, 1000)
+
+		query, args := buildUnspentTokensIteratorByQuery(store, walletID, tokenType)
```
AkramBitar, line 66: My understanding is that the idea behind a prepared statement is to create the query once and execute it multiple times with different arguments to improve performance.

In our case, though, we're passing the same arguments every time:

const walletID = "wallet0"
tokenType := tokentype.Type("GOLD")

Is my understanding correct?

AkramBitar (changes requested):

AkramBitar:
Hello @sid200727 

Thanks a lot. I reviewed the PR. 

Could you please open a new issue if not exists for this benchmark and attach it to this PR (since #1183 is for adding prepared statement not for that benchmark)?

Regards,
Akram

token/services/storage/db/sql/common/tokens_prepared_bench_util.go
sid200727: Following existing pattern, SeedBenchTokens/RunTokenStoreBenchmarks already live in tokens_bench_test_util.go (non-_test.go, same package) for the same reason: this needs unexported common internals (readDB, buildUnspentTokensIteratorByQuery), and _test.go files can't be imported cross-package, so it can't live in postgres's test file.

token/services/storage/db/sql/common/tokens_prepared_bench_util.go
sid200727: Fixed args isolate per-call query-build/parse cost vs reused stmt, that's the overhead prepared statements remove regardless of arg variance. Can add a randomized-args run too if you want it confirmed under variance.

sid200727:
Yes, @AkramBitar i have opened a new issue - Opened #1861, linked to this PR.

token/core/fabtoken/protos-go/v1/pp/ftpp.pb.go
AkramBitar: @sid200727 this issue is in all the changed protos-go. 
If this is nit needed then could you please revert these files?

token/services/storage/db/sql/common/tokens_prepared_bench_util.go
AkramBitar, line 1: I'm wondering why we need to add a utility file. Wouldn't it be more convenient to move this code into the test file instead (token/services/storage/db/sql/postgres/tokens_bench_test.go)?

What do you think?

AkramBitar (changes requested):

AkramBitar:
@sid200727 thanks a lot, all looks good except one comment https://github.com/LFDT-Panurus/panurus/pull/1855#discussion_r3536130308 could you have look?

sid200727:
Hi @AkramBitar, that link points to your earlier comment asking me to open a new issue for the benchmark, I did that (#1861, linked to this PR). 
Let me know if there's something else you meant, or if this is good to go!

AkramBitar:
> Hi @AkramBitar, that link points to your earlier comment asking me to open a new issue for the benchmark, I did that (#1861, linked to this PR). Let me know if there's something else you meant, or if this is good to go!

@sid200727 

I am sorry by mistake I meant that one: https://github.com/LFDT-Panurus/panurus/pull/1855#discussion_r3536130308

"@sid200727 this issue is in all the changed protos-go.
If this is nit needed then could you please revert these files?"

All good expect to revert these files if no changes made to them.

Thanks a lot,
Akram

AkramBitar (approved):
LGTM

sid200727:
Thank you! 🙏

PR #1859, opened 2026-07-06
fix(multisig): reject zero-member identity/verifier to close anyone-can-spend bypass

### LINKED ISSUE #1858: Zero-length MultiIdentity yields vacuously-true `Verify()` — anyone-can-spend output
## Summary

A multisig-typed owner identity with zero constituent identities can be placed on a token output.
The resulting `Verifier` has zero sub-verifiers; its `Verify()` length-equality guard passes when
both sides are zero, the loop body never runs, and the function returns `nil`. Anyone can then
spend that output by supplying an empty `MultiSignature{}`, with no real signature required.

**Severity:** Low  
**Component:** `token/services/identity/multisig`  
**Affected files:** `sig.go`, `deserializer.go`

---

### 1. `sig.go` — vacuously-true `Verify()`

```go
// sig.go:67 (before fix)
if len(v.Verifiers) != len(sig.Signatures) {  // 0 != 0 → false, guard passes
    return errors.Errorf(...)
}
for k, ver := range v.Verifiers {  // iterates zero times
    ...
}
return nil  // ← returns success with no signature checked
```

When `len(v.Verifiers) == 0` and the attacker supplies `MultiSignature{Signatures: []}`, the
length check `0 != 0` is `false`, the loop body never executes, and `Verify` returns `nil`.

### 2. `deserializer.go` — no rejection of empty `MultiIdentity` at the trust boundary

```go
// deserializer.go:119 (before fix)
verifier.Verifiers = make([]driver.Verifier, len(multisigIdentity.Identities))
// len == 0 → empty slice, returned without error
return verifier, nil
```

`DeserializeVerifier` builds `verifier.Verifiers` from `multisigIdentity.Identities` without
checking whether that slice is non-empty, producing a zero-verifier `Verifier`.

### 3. `WrapIdentities` guard is constructor-side only

`identity.go:47` has a `len(ids) == 0` guard, but this is only reached via the honest-client
constructor. An attacker can directly `asn1.Marshal` a `MultiIdentity{Identities: []}`, wrap it
in a `TypedIdentity{Type: Multisig}`, and submit it on-chain — bypassing the constructor
entirely. The on-chain transfer validators check only that the owner byte slice is non-empty,
not its content.

---

## Exploit Scenario

1. Attacker constructs a transfer output whose `Owner` bytes encode
   `TypedIdentity{Type: Multisig, Identity: asn1(MultiIdentity{Identities: []})}`.
2. The output is committed on-chain (validators only require `len(owner) != 0`).
3. Anyone later submits a transfer spending that output with the signature slot set to
   `asn1(MultiSignature{Signatures: []})`.
4. `HasBeenSignedBy` draws that signature, calls `multisig.Verifier.Verify`, which returns
   `nil`. The spend is authorized with no real signature.

The attacker can only do this to outputs they themselves create (not direct theft from others),
but it creates tokens spendable by **anyone** — a deliberate "anyone-can-spend" output that
may interact badly with auditor holding-limit logic or be used to launder provenance.

---

## Preconditions

- The multisig identity type is registered in the deserializer multiplex (it is, in both shipped drivers).
- The attacker can author a transfer output's `Owner` bytes (any sender can).

---

## Proposed Fix

Two guards, both required:

**`sig.go` — fail closed on degenerate verifier state (defence-in-depth):**
```go
func (v *Verifier) Verify(msg, raw []byte) error {
    if len(v.Verifiers) == 0 {
        return errors.New("multisig verifier has no members")
    }
    ...
}
```

**`deserializer.go` — reject at the trust boundary:**
```go
if len(multisigIdentity.Identities) == 0 {
    return nil, errors.New("multisig identity has no members")
}
```

---

## References

- `token/services/identity/multisig/sig.go`
- `token/services/identity/multisig/deserializer.go`
- `token/services/identity/multisig/identity.go`

adecaro:
Closes the "anyone-can-spend" bypass described in the linked security issue: a
`MultiIdentity` with zero constituent identities could be serialised directly (bypassing the
honest-client `WrapIdentities` guard), committed on-chain as a token output owner, and later
spent by anyone supplying an empty `MultiSignature{}` — no real signature required.

Two minimal, targeted guards close the two independent paths that allowed this:

### `token/services/identity/multisig/sig.go` — fail closed on degenerate verifier state

```go
func (v *Verifier) Verify(msg, raw []byte) error {
+   if len(v.Verifiers) == 0 {
+       return errors.New("multisig verifier has no members")
+   }
    sig := &MultiSignature{}
    ...
```

The previous `len(v.Verifiers) != len(sig.Signatures)` guard passed vacuously when both
lengths were zero (`0 != 0` → `false`). The loop then iterated zero times and returned `nil`.
The new check fires before any ASN.1 parsing, making the function fail closed regardless of
how a zero-verifier `Verifier` was constructed.

### `token/services/identity/multisig/deserializer.go` — reject at the trust boundary

```go
 func (d *TypedIdentityDeserializer) DeserializeVerifier(...) (driver.Verifier, error) {
     multisigIdentity := &MultiIdentity{}
     err := multisigIdentity.Deserialize(raw)
     if err != nil { ... }
+    if len(multisigIdentity.Identities) == 0 {
+        return nil, errors.New("multisig identity has no members")
+    }
     verifier := &Verifier{}
     ...
```

This is the primary enforcement point: attacker-controlled bytes enter here. The guard ensures
a zero-identity `MultiIdentity` is rejected before a `Verifier` object is ever allocated.

---

## Why two guards

- **Deserializer guard** — primary: blocks the attacker path at the point where raw bytes
  become a typed Go value.
- **`Verify` guard** — defence-in-depth: ensures the function fails closed even if a
  zero-verifier `Verifier` is constructed through any future programmatic path (e.g. a test
  helper, a different deserializer, or a future refactor) that does not go through
  `DeserializeVerifier`.

The two guards are independent and complement each other; neither alone is sufficient for
robust defence-in-depth.

---

## Test coverage

Five new tests across the three test files pin the expected behaviour and would catch any
regression:

| File | Test | Covers |
|---|---|---|
| `sig_test.go` | `TestVerifier_Verify_ZeroVerifiers` | Zero-verifier `Verifier` + empty `MultiSignature{}` bytes → error |
| `deserializer_test.go` | `TestDeserializeVerifier_EmptyIdentities` | Direct `asn1.Marshal(MultiIdentity{})` attacker path → rejected at deserialization |
| `identity_test.go` | `TestWrapIdentities_SingleIdentity` | Boundary: one identity is the minimum valid input — happy path at the lower bound |
| `identity_test.go` | `TestUnwrap_EmptyMultiIdentity` | Documents that `Unwrap` itself does not reject empty identity (not the enforcement point); empty slice is returned, no error |
| `identity_test.go` | `TestInfoMatcher_Match_ZeroMatchers` | Documents the analogous vacuous-true behaviour in `InfoMatcher.Match`; pins expectation that the mitigation lives at the deserializer, not inside `Match` |

---

## Checklist

- [x] Existing unit tests pass (`go test ./token/services/identity/multisig/...`)
- [x] New tests added for both fixed paths
- [x] No behaviour change for valid (non-empty) multisig identities
- [x] No new dependencies introduced
- [x] Both changed files have DCO sign-off commit

### DIFF
--- a/token/services/identity/multisig/deserializer.go
+++ b/token/services/identity/multisig/deserializer.go
@@ -115,6 +115,9 @@ func (d *TypedIdentityDeserializer) DeserializeVerifier(ctx context.Context, typ
 	if err != nil {
 		return nil, errors.New("failed to unmarshal multisig identity")
 	}
+	if len(multisigIdentity.Identities) == 0 {
+		return nil, errors.New("multisig identity has no members")
+	}
 	verifier := &Verifier{}
 	verifier.Verifiers = make([]driver.Verifier, len(multisigIdentity.Identities))
 	for k, i := range multisigIdentity.Identities {
--- a/token/services/identity/multisig/deserializer_test.go
+++ b/token/services/identity/multisig/deserializer_test.go
@@ -46,6 +46,27 @@ func TestDeserializeVerifier_Success(t *testing.T) {
 	assert.Equal(t, 1, verifierDES.DeserializeVerifierCallCount())
 }
 
+// DeserializeVerifier must reject a zero-identity MultiIdentity at the trust boundary.
+// This is the attacker path: directly asn1.Marshal a MultiIdentity{Identities: []}
+// and wrap it in a TypedIdentity{Type: Multisig}, bypassing the WrapIdentities guard.
+func TestDeserializeVerifier_EmptyIdentities(t *testing.T) {
+	verifierDES := &mock.VerifierDES{}
+	deserializer := NewTypedIdentityDeserializer(verifierDES, nil)
+
+	// Attacker serializes MultiIdentity{Identities: []} directly
+	emptyID := &MultiIdentity{Identities: []token.Identity{}}
+	emptyRaw, err := emptyID.Bytes()
+	require.NoError(t, err)
+
+	verifier, err := deserializer.DeserializeVerifier(context.Background(), Multisig, emptyRaw)
+	require.Error(t, err)
+	assert.Nil(t, verifier)
+	assert.Contains(t, err.Error(), "multisig identity has no members")
+
+	// No sub-verifier deserialization should have been attempted
+	assert.Equal(t, 0, verifierDES.DeserializeVerifierCallCount())
+}
+
 // Create an invalid raw multi-id and fail to deserialize a verifier from it
 func TestDeserializeVerifier_InvalidMultisigIdentity(t *testing.T) {
 	verifierDES := &mock.VerifierDES{}
--- a/token/services/identity/multisig/identity_test.go
+++ b/token/services/identity/multisig/identity_test.go
@@ -138,6 +138,56 @@ func TestWrapIdentities_Error(t *testing.T) {
 	require.Error(t, err)
 }
 
+// Boundary: a single identity is the minimum valid case; WrapIdentities must succeed.
+func TestWrapIdentities_SingleIdentity(t *testing.T) {
+	id := identities(t, "only-id")
+	wrapped, err := WrapIdentities(id...)
+	require.NoError(t, err)
+
+	unwrapped, isMultisig, err := Unwrap(wrapped)
+	require.NoError(t, err)
+	assert.True(t, isMultisig)
+	assert.Equal(t, id, unwrapped)
+}
+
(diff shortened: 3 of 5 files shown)


AkramBitar (commented):
LGTM

AkramBitar (commented):
LGTM

AkramBitar (approved):

PR #1863, opened 2026-07-08
issuer wallet extensions

### LINKED ISSUE #872: issuer wallet: how much this wallet issued?

adecaro:
This PR addresses #872


token/driver/wallet.go
```diff
@@ -151,6 +163,22 @@ type IssuerWallet interface {
 
 	// HistoryTokens returns the list of tokens issued by this wallet filtered using the passed options.
 	HistoryTokens(ctx context.Context, opts *ListTokensOptions) (*token.IssuedTokens, error)
+
```
AkramBitar, line 166: I am wondering if we need to add this to the documenation?

token/services/storage/db/sql/common/tokens.go
```diff
@@ -1265,6 +1325,7 @@ func (db *TokenStore) GetSchema() string {
 			owner BOOL NOT NULL DEFAULT false,
 			auditor BOOL NOT NULL DEFAULT false,
 			issuer BOOL NOT NULL DEFAULT false,
+			redeemed BOOL NOT NULL DEFAULT false,
```
AkramBitar, line 1328: The new redeemed column is added in CREATE TABLE IF NOT EXISTS, but there isn't a corresponding ALTER TABLE for existing databases. Otherwise, existing installations may fail because the column won't exist.

AkramBitar (changes requested):

AkramBitar (approved):
LGTM

Added some comments

token/services/storage/db/sql/common/tokens.go
adecaro: we are still allowing breaking changes but we could investigate libraries that do this job of porting db schemas automatically

token/driver/wallet.go
adecaro: done

token/services/storage/db/sql/common/tokens.go
adecaro: mentioned this in the documentation

PR #1868, opened 2026-07-09
perf(identity): point-query the identity store on LocalMembership cache miss

### LINKED ISSUE #1619: Auditor nodes call IsMine() on every output identity, causing LocalMembership lock contention under high load

EvanYan1024:
Refs #1619. Follow-up to the discussion in #1827 (closed in favor of this approach, see https://github.com/LFDT-Panurus/panurus/pull/1827#issuecomment-4921186034).

## Problem

`LocalMembership.refreshAndGet` reloads **every** stored identity configuration under the global write lock whenever a label misses the in-memory maps. On nodes that resolve many foreign identities — an auditor checking `IsMine` for every transaction is the worst case (#1619) — every lookup pays a full store scan serialized on one mutex, and a pending writer also blocks all readers. In our test environments under sustained load, this lock dominates mutex contention on auditor nodes.

#1827 skipped the lookup for pure auditors behind a cached flag; as the discussion there concluded, any "cached answer + event-driven invalidation" design is fragile under replication — the flag is only as correct as the delivery of the event that invalidates it.

## Change

Keep the miss, make it cheap:

- **`refreshAndGet` point-queries the store for the missed label only**, via a new `IdentityStoreService.ConfigurationsByID(id, type)`. A hit loads just that configuration (the write lock is held only for the in-memory registration); a miss returns without taking the write lock at all. The SQL implementation is answered by the existing `PRIMARY KEY(id, type, url)` prefix; the kvs implementation scans the type and filters, since its composite key encodes id and url together.
- Labels that are not valid UTF-8 (raw identity bytes) cannot match a stored configuration id (a TEXT column) and skip the store entirely.
- **`handleConfig` reads the store outside the write lock.** Notifications fire for every stored configuration write, including the node's own, so each one used to stall all lookups for the duration of a store round trip.

Correctness under replication is structural rather than event-driven: the shared store is consulted on every miss — the same semantics as today's full reload — so replicas need no flags and no notifications to answer correctly. Notifications remain a warm-cache optimization; a failed notification is repaired by the next miss for that label.

Lookups **by raw identity bytes** for wallets created on another replica also resolve through shared state: the wallet store binding maps the identity to its wallet id (`Registry.Lookup`'s existing fallback), and the by-name point query then loads the configuration. This chain is pinned by a test.

## Tests

- Point query loads a configuration registered by another replica; unknown labels do not trigger a full scan; non-UTF-8 labels do not touch the store (`lm_test.go`).
- Discovery tests moved to the point-query channel; the concurrent-discovery test asserts a single registration under concurrency (`lm_discovery_test.go`).
- Lock discipline: the notification handler re-enters a read path from within the store read — this deadlocks if the handler held the lock across the read (verified red-green) (`lm_discovery_test.go`).
- Cross-replica lookup by raw identity bytes resolves via the wallet-store binding plus the point query, with no full scan and no notification (`registry_test.go`).

`go build ./token/...`, `go vet`, `go test -race` on the identity and storage trees, and golangci-lint v2.12.2 with the repo config are green.

## Notes for reviewers

- Both `IdentityStoreService` interfaces (driver-level and the membership-local one) gain `ConfigurationsByID`; both counterfeiter fakes regenerated.
- While reworking `refreshAndGet` I hit a latent quirk worth knowing about: `registerIdentityConfiguration` rewrites `config.URL` via `TranslatePath` before `addLocalIdentity` keys `localIdentitiesByConfig`, so a `configKey` computed before registration never matches the stored one. Duplicate registration is actually prevented by the label double-check under the write lock (as in the current code), which the new code keeps.

### DIFF
--- a/token/services/identity/driver/mock/iss.go
+++ b/token/services/identity/driver/mock/iss.go
@@ -48,6 +48,21 @@ type IdentityStoreService struct {
 		result1 bool
 		result2 error
 	}
+	ConfigurationsByIDStub        func(context.Context, string, string) ([]driver.IdentityConfiguration, error)
+	configurationsByIDMutex       sync.RWMutex
+	configurationsByIDArgsForCall []struct {
+		arg1 context.Context
+		arg2 string
+		arg3 string
+	}
+	configurationsByIDReturns struct {
+		result1 []driver.IdentityConfiguration
+		result2 error
+	}
+	configurationsByIDReturnsOnCall map[int]struct {
+		result1 []driver.IdentityConfiguration
+		result2 error
+	}
 	GetAuditInfoStub        func(context.Context, []byte) ([]byte, error)
 	getAuditInfoMutex       sync.RWMutex
 	getAuditInfoArgsForCall []struct {
@@ -121,6 +136,21 @@ type IdentityStoreService struct {
 		result1 []byte
 		result2 []byte
 		result3 error
+	}
+	IterateSignersStub        func(context.Context, int, int) ([]driver.SignerEntry, error)
+	iterateSignersMutex       sync.RWMutex
+	iterateSignersArgsForCall []struct {
+		arg1 context.Context
+		arg2 int
+		arg3 int
+	}
+	iterateSignersReturns struct {
+		result1 []driver.SignerEntry
+		result2 error
+	}
+	iterateSignersReturnsOnCall map[int]struct {
+		result1 []driver.SignerEntry
+		result2 error
 	}
 	IteratorConfigurationsStub        func(context.Context, string) (driver.IdentityConfigurationIterator, error)
 	iteratorConfigurationsMutex       sync.RWMutex
@@ -202,21 +232,6 @@ type IdentityStoreService struct {
 	}
 	storeSignerInfoReturnsOnCall map[int]struct {
 		result1 error
-	}
-	IterateSignersStub        func(context.Context, int, int) ([]driver.SignerEntry, error)
-	iterateSignersMutex       sync.RWMutex
-	iterateSignersArgsForCall []struct {
-		arg1 context.Context
-		arg2 int
-		arg3 int
-	}
-	iterateSignersReturns struct {
-		result1 []driver.SignerEntry
-		result2 error
-	}
-	iterateSignersReturnsOnCall map[int]struct {
-		result1 []driver.SignerEntry
-		result2 error
 	}
 	invocations      map[string][][]interface{}
 	invocationsMutex sync.RWMutex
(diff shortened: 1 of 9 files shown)


adecaro (approved):
beautiful effort. Thanks a lot @EvanYan1024 . I think we are in a better shape here 🙏

PR #1869, opened 2026-07-09
ttx recipient callback to choose the wallet

### LINKED ISSUE #1870: chore(ttx): recipient callback for wallet selection
Currently, it is not possible to specify a callback when responding to a request for a recipient identity. This is a useful feature to have to be able to delegate to the application further business decisions.

adecaro:
This PR adds the possibility to specify a callback to select the wallet during the recipient-related flows.

### DIFF
--- a/token/services/ttx/opts.go
+++ b/token/services/ttx/opts.go
@@ -243,3 +243,65 @@ func getRecipientWalletID(opts *token.ServiceOptions) string {
 
 	return wBoxed.(string)
 }
+
+// WalletSelector lets the responder to a RecipientRequest choose which wallet to use
+// to answer the request. defaultWallet is the wallet that would be used absent a
+// selector (the responder's configured wallet, or the request's WalletID as fallback).
+// Returning defaultWallet unchanged preserves the default behavior.
+type WalletSelector func(request *RecipientRequest, defaultWallet string) (string, error)
+
+// ExchangeWalletSelector is the WalletSelector equivalent for ExchangeRecipientRequest.
+type ExchangeWalletSelector func(request *ExchangeRecipientRequest, defaultWallet string) (string, error)
+
+// WithWalletSelector sets a callback used by the responder to a RecipientRequest to
+// choose which wallet to use, overriding the default wallet resolution.
+func WithWalletSelector(selector WalletSelector) token.ServiceOption {
+	return func(options *token.ServiceOptions) error {
+		if selector == nil {
+			return nil
+		}
+		if options.Params == nil {
+			options.Params = map[string]any{}
+		}
+		options.Params["WalletSelector"] = selector
+
+		return nil
+	}
+}
+
+// getWalletSelector extracts the WalletSelector from service options, if set.
+func getWalletSelector(opts *token.ServiceOptions) WalletSelector {
+	sBoxed, ok := opts.Params["WalletSelector"]
+	if !ok {
+		return nil
+	}
+
+	return sBoxed.(WalletSelector)
+}
+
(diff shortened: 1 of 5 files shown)


token/services/ttx/opts.go
```diff
@@ -243,3 +243,65 @@ func getRecipientWalletID(opts *token.ServiceOptions) string {
...
+// choose which wallet to use, overriding the default wallet resolution.
+func WithWalletSelector(selector WalletSelector) token.ServiceOption {
+	return func(options *token.ServiceOptions) error {
+		if selector == nil {
+			return nil
+		}
+		if options.Params == nil {
+			options.Params = map[string]any{}
+		}
+		options.Params["WalletSelector"] = selector
```
AkramBitar: I am wondering if to have constant for this string or not?

token/services/ttx/opts.go
```diff
@@ -243,3 +243,65 @@ func getRecipientWalletID(opts *token.ServiceOptions) string {
...
+type WalletSelector func(request *RecipientRequest, defaultWallet string) (string, error)
+
+// ExchangeWalletSelector is the WalletSelector equivalent for ExchangeRecipientRequest.
+type ExchangeWalletSelector func(request *ExchangeRecipientRequest, defaultWallet string) (string, error)
+
+// WithWalletSelector sets a callback used by the responder to a RecipientRequest to
+// choose which wallet to use, overriding the default wallet resolution.
+func WithWalletSelector(selector WalletSelector) token.ServiceOption {
+	return func(options *token.ServiceOptions) error {
+		if selector == nil {
```
AkramBitar, line 270: This method return error = nil always. Do we need to return error at all?

token/services/ttx/opts.go
```diff
@@ -243,3 +243,65 @@ func getRecipientWalletID(opts *token.ServiceOptions) string {
...
+		}
+		options.Params["WalletSelector"] = selector
+
+		return nil
+	}
+}
+
+// getWalletSelector extracts the WalletSelector from service options, if set.
+func getWalletSelector(opts *token.ServiceOptions) WalletSelector {
+	sBoxed, ok := opts.Params["WalletSelector"]
```
AkramBitar: Do we need to check if sBoxed is nil? or does this already covered above in the insert to the map?

token/services/ttx/opts.go
```diff
@@ -243,3 +243,65 @@ func getRecipientWalletID(opts *token.ServiceOptions) string {
...
+// WithExchangeWalletSelector sets a callback used by the responder to an
+// ExchangeRecipientRequest to choose which wallet to use, overriding the default wallet
+// resolution.
+func WithExchangeWalletSelector(selector ExchangeWalletSelector) token.ServiceOption {
+	return func(options *token.ServiceOptions) error {
+		if selector == nil {
+			return nil
+		}
+		if options.Params == nil {
+			options.Params = map[string]any{}
```
AkramBitar, line 308: Same comments as above

AkramBitar (approved):
LGTM
Left some comments.

token/services/ttx/opts.go
adecaro: okay

token/services/ttx/opts.go
adecaro: okay

token/services/ttx/opts.go
adecaro: yes, because it is the expected signature

token/services/ttx/opts.go
adecaro: yes

PR #1878, opened 2026-07-09
feat(auditdb): add HoldingsFilter.SumByEnrollmentID

### LINKED ISSUE #1877: auditdb: HoldingsFilter has no per-enrollment-id aggregation
`HoldingsFilter.ByEnrollmentId` can be chained to load the movements of several enrollment ids with a single query, but the only aggregation on the executed filter is `Sum()`, which returns one grand total. A caller that needs the individual balances of N parties (e.g. an auditor checking balance limits for the sender and every recipient of a transaction) has to build and execute N single-EID filters — N database round trips per transaction on a hot path.

Proposal: add a `SumByEnrollmentID()` accessor that partitions the loaded records by enrollment id and returns `map[string]*big.Int`, making the multi-EID form of the filter usable for per-party balances: one query, N sums.

Happy to open a PR.

EvanYan1024:
Fixes #1877

`ByEnrollmentId` already accepts multiple enrollment ids in one filter (single `WHERE enrollment_id IN (...)` query), but the executed filter only exposes the grand total via `Sum()`. This adds `SumByEnrollmentID()`, a pure accessor over the records loaded by `Execute` that partitions them by enrollment id.

In our deployment the auditor checks per-party balance limits for every transaction (sender + each recipient); this collapses one filter execution per enrollment id into a single query for all parties, which removed N-1 database round trips per audited transaction in our test environments under sustained load.

### DIFF
--- a/token/services/storage/auditdb/filter.go
+++ b/token/services/storage/auditdb/filter.go
@@ -104,3 +104,19 @@ func (f *HoldingsFilter) Sum() *big.Int {
 
 	return sum
 }
+
+// SumByEnrollmentID returns the sum of the loaded records grouped by enrollment id.
+// Enrollment ids with no records are absent from the returned map.
+func (f *HoldingsFilter) SumByEnrollmentID() map[string]*big.Int {
+	sums := make(map[string]*big.Int)
+	for _, record := range f.records {
+		sum, ok := sums[record.EnrollmentID]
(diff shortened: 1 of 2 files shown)


adecaro (approved):
LGTM. Thanks @EvanYan1024 🙏

PR #1879, opened 2026-07-09
feat(evm): week 2 contracts (EIP-712 + EndorsementVerifier)

atharrva01:
First half of week 2. Adds the Solidity contracts under x/token/services/network/evm/contracts (Foundry), kept out of the Go module. TokenState.sol and the integration tests come in PR 2b.

What's here:
- StateDelta.sol and EIP712.sol: the frozen structs and EIP-712 encoding, mirroring the Go side byte for byte
- EndorsementVerifier.sol: quorum signature check (ecrecover, low-s, unique signers, threshold)

21 forge tests pass. The Solidity digest matches the Go fixture for both the transfer and setup deltas (0xc9326b72 and 0xdca9a011), also checked against ethers v6.

On the Go side, StateDelta.Validate now rejects a few malformed deltas, and there's a second golden vector for the setup case.

The docs commit resyncs the plan and design (Besu as the backend, working rules, module isolation).

Note: forge-std is a git submodule, so run `git submodule update --init`.


adecaro (approved):
LGTM. Please, rebase and run `make checks` and `make lint` to make sure everything is fine.

PR #1880, opened 2026-07-09
fix(ci): restore integration coverage of main module after integration/ split

### LINKED ISSUE #1852: Reported coverage drops even when all tests pass

SaidAltury-ibm:
# Background: 
integration tests measure coverage by running cover-instrumented FSC node binaries. Since integration/ became its own module (integration/go.mod), those node binaries build in the `integration` module, and `go build -cover` defaults -coverpkg to "packages in the main Go module" only. So the main-module `token/...` packages (now an external dependency via `replace`) stop being instrumented and their end-to-end coverage disappears from the report. We restore it by widening -coverpkg to the main module.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,19 @@ FAB_BINS ?= $(FABRIC_BINARY_BASE)/bin
 GINKGO_TEST_OPTS ?=
 GINKGO_TEST_OPTS += --keep-going -cover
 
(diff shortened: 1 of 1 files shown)


adecaro (approved):
Thanks much @SaidAltury-ibm 🙏

PR #1882, opened 2026-07-10
perf(auditor): reuse the audit record computed by Audit in Append

### LINKED ISSUE #1881: Auditor computes each transaction's audit record twice (Audit + Append)
## Description

On the auditor node, each audited transaction computes its `token.AuditRecord` twice:

1. `auditor.Service.Audit()` calls `request.AuditRecord()` to extract the inputs/outputs and acquire the enrollment-ID locks.
2. `auditor.Service.Append()` then persists the request through `auditDB.Append(newRequestWrapper(...))`, whose `AuditRecord()` calls `request.AuditRecord()` again.

`request.AuditRecord()` deserializes the request's actions and issues a vault query (`ListAuditTokens`) for the inputs, so the second computation doubles the per-transaction audit-record cost and the vault read traffic on the auditor hot path. We observed this in our test environments under sustained load.

## Proposal

Cache a snapshot of the audit record computed by `Audit()`, keyed by request anchor, and let `Append()` reuse it instead of recomputing; drop the entry in `Release()`. The gap-filling step (`completeInputsWithEmptyEID`) must still run on cache hits since it depends on current vault state.

I have a patch ready and will open a PR shortly.

EvanYan1024:
## Description

`Audit()` already computes the transaction's `token.AuditRecord` to extract the inputs/outputs and acquire the enrollment-ID locks, but `Append()` recomputed the same record — request deserialization plus a `ListAuditTokens` vault query — when persisting the request to the audit DB. In our test environments under sustained load this doubles the per-transaction audit-record cost and the vault read traffic on the auditor hot path.

This PR caches, per request anchor, a snapshot of the record computed by `Audit()` and lets `Append()` reuse it:

- The cached snapshot is a deep copy: the streams `Audit()` returns to business code stay independent of what `Append()` persists, and gap filling cannot mutate them.
- Each cache entry is bound to the `*token.Request` it was computed from; a lookup from a different request reusing the same anchor misses and safely falls back to recomputation. Duplicate anchors are outside the platform's operating assumptions (the audit DB locker keys locks by anchor as well), so on collision the cache only degrades to the previous behavior.
- Entries are dropped by `Release()`, which `Append()` already defers, so the cache stays bounded by in-flight audits. The map is lazily initialized, keeping the `Service` zero value safe.
- Gap filling (`completeInputsWithEmptyEID`) still runs on cache hits, since it depends on current vault state.

New unit tests cover the cache lifecycle (stash/lookup/drop, same-anchor isolation), snapshot isolation in both directions (caller-side mutations, including through inner pointers, never reach the cached record; gap filling never reaches the caller's streams), and a service-level Audit → Append chain verifying the record is computed exactly once on the cached path and recomputed on a cache miss.

Fixes #1881

### DIFF
--- a/token/services/auditor/auditor.go
+++ b/token/services/auditor/auditor.go
@@ -8,6 +8,8 @@ package auditor
 
 import (
 	"context"
+	"slices"
+	"sync"
 	"time"
 
 	"github.com/LFDT-Panurus/panurus/token"
@@ -83,6 +85,18 @@ type Service struct {
 	metrics         *Metrics
 	checkService    CheckService
 	lockConfig      *LockConfig
+
+	// auditRecords caches, per request anchor, a snapshot of the audit record
+	// computed by Audit for reuse by Append; entries are dropped by Release.
+	recordsMu    sync.Mutex
+	auditRecords map[token.RequestAnchor]auditRecordEntry
+}
+
+// auditRecordEntry pairs a cached audit record with the request it was
+// computed from.
+type auditRecordEntry struct {
+	request *token.Request
+	record  *token.AuditRecord
 }
 
 // NewService creates a new auditor Service with the provided dependencies.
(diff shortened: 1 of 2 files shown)


adecaro (approved):
LGTM. Thanks @EvanYan1024

PR #1884, opened 2026-07-10
shared token selector: first check locally

### LINKED ISSUE #886: shared token selector: first check locally

adecaro:
Problem: sherdlock.Selector.selectInternal calls Locker.TryLock → SQL INSERT INTO TOKEN_LOCKS for every candidate token, even when this process already knows — from an earlier lock
  attempt in the same process's lifetime — that the token is locked (by itself or by another in-process transaction). That's a wasted network round trip on the selection hot path.

  Prior attempt: PR #1483 tried a manager-level shared map guarded by one sync.RWMutex, checked unconditionally on every token examined. Its own benchmarks showed a regression (+56%
  single-threaded, +882% under contention) because (a) the lookup ran even on the already-cheap no-conflict path, and (b) a single writer-preferring RWMutex became a worse bottleneck than
  Postgres's own row locking under real contention. This PR is a fresh design, not a rebase of that branch.

  Design

  - sherdlock.localLockCache (sherdlock/localcache.go): a sync.Map-backed, process-local, best-effort cache of token.ID → (consumerTxID, lockedAt). Lock-free reads for existing entries,
  no global mutex.
  - One cache instance lives per Manager, shared by every Selector/StubbornSelector it creates (manager.go), so it dedups both across candidates within one selection and across different
  transactions in the same replica.
  - The locker adapter (selector.go) consults the cache before TryLock and records the outcome after — on success, on a genuine lock conflict, and it's cleared for a transaction's tokens
  on UnlockAll.
  - Correctness / staleness: cache entries are never authoritative. An entry older than leaseExpiry is treated as stale and falls through to the DB; Manager.cleaner also sweeps expired
  entries after each Cleanup tick. Cross-replica unlocks stay invisible to this process by design — the lease-expiry bound is the same correctness model the DB-side cleanup already relies
  on.
  - Distinguishing a genuine conflict from a transient DB error (needed so a flaky connection doesn't poison the cache): reused the existing fabric-smart-client
  SQLErrorWrapper/ErrorMapper infrastructure already wired into KeystoreStore, rather than writing bespoke error-code detection. TokenLockStore.Lock() (storage/db/sql/common/tokenlock.go,
  wired for both postgres/sqlite) now classifies unique-key violations via errors.HasCause(errorWrapper.WrapError(err), UniqueKeyViolation) and returns a new driver.ErrTokenLockConflict
  sentinel.

  Testing

  - sherdlock/locker_test.go: unit tests for same-token-twice-in-one-selection dedup, cross-selector-same-process dedup, invalidation-after-unlock, conflict-caching, and a no-cache
  regression guard, plus direct localLockCache tests (staleness, releaseOwned, sweepExpired).
  - benchmark_test.go: added sherdlock+localcache settings as a mock-locker regression guard for the no-conflict path.
  - sherdlock/postgres_bench_test.go: new real-Postgres (testcontainers) benchmark comparing nocache vs localcache under intra-replica contention over a small seeded token pool — the mock
  on.
  - Distinguishing a genuine conflict from a transient DB error (needed so a flaky connection doesn't poison the cache): reused the existing fabric-smart-client
  SQLErrorWrapper/ErrorMapper infrastructure already wired into KeystoreStore, rather than writing bespoke error-code detection. TokenLockStore.Lock() (storage/db/sql/common/tokenlock.go,
  wired for both postgres/sqlite) now classifies unique-key violations via errors.HasCause(errorWrapper.WrapError(err), UniqueKeyViolation) and returns a new driver.ErrTokenLockConflict
  sentinel.

  Testing

  - sherdlock/locker_test.go: unit tests for same-token-twice-in-one-selection dedup, cross-selector-same-process dedup, invalidation-after-unlock, conflict-caching, and a no-cache
  regression guard, plus direct localLockCache tests (staleness, releaseOwned, sweepExpired).
  - benchmark_test.go: added sherdlock+localcache settings as a mock-locker regression guard for the no-conflict path.
  - sherdlock/postgres_bench_test.go: new real-Postgres (testcontainers) benchmark comparing nocache vs localcache under intra-replica contention over a small seeded token pool — the mock
  locker can't show a real network round-trip saving, so this is the track that actually validates the win.
  - go test -race clean across token/services/selector/... and token/services/storage/db/sql/....

  Docs

  Merged token/services/selector/README.md into docs/services/selector.md (Available Implementations + Benchmarking sections) and removed the now-redundant README.

  Benchmark results

  Mock-locker no-conflict path shows no regression (within noise of a single-iteration run). Real-Postgres contention benchmark (8 workers, 8-token pool) landed within noise between
  nocache/localcache for this particular contention shape, since most attempts genuinely fail at the DB; the local cache's benefit is targeted at avoiding repeated re-checks of an
  already-known-locked token within one selection or across in-process transactions — exactly what the unit tests exercise directly. No regression observed anywhere.


adecaro:
I'll drop this cause the benefits are not clear. The local cache introduces backoff issues. The backoff becomes too fast and the number of retries are exhausted too fast with the respect to delay introduced by the db query.

PR #1887, opened 2026-07-10
fix(storage): drop duplicate cache loop in GetExistingSignerInfo

### LINKED ISSUE #1874: bug: GetExistingSignerInfo runs the same cache-lookup loop twice
## Problem

`IdentityStore.GetExistingSignerInfo` (`token/services/storage/db/sql/common/identity.go`) scans `idHashes` against `db.signerInfoCache`, collecting hits into `result` and misses into `notFound`. It then immediately repeats the exact same loop over the resulting `notFound` list (reassigned to `idHashes`), checking the same cache again and rebuilding `result`/`notFound` a second time before the query against the `Signers` table.

Nothing writes to `signerInfoCache` between the two loops, so the second pass can only reproduce the first pass's outcome: every id that was a miss in the first loop is guaranteed to still be a miss in the second, and every "found" entry was already collected once. The second loop is dead code that adds an extra O(n) pass with no effect on correctness.

## Suggested fix

Delete the duplicated loop; keep a single cache-lookup pass before falling through to the DB query for the remaining misses.

## Test plan

- Existing `GetExistingSignerInfo`/`SignerInfoExists` unit tests should still pass with identical results.

AkramBitar:
## Problem

The second cache-lookup loop in `GetExistingSignerInfo` was once the write-locked re-check of a double-checked-locking pattern: pass 1 ran under an external `signerCacheLock.RLock()`, then the lock was upgraded to `Lock()` and pass 2 re-checked the cache (introduced in 9fcfd3d9, PR https://github.com/LFDT-Panurus/panurus/pull/825). The July 2025 rework (3d818810, PR: https://github.com/LFDT-Panurus/panurus/pull/1168) removed `signerCacheLock` — moving synchronization inside the cache — but left the second loop behind.

With no lock between the two passes, and `secondcache.Get` being a pure read (loading lives in the separate, unused `GetOrLoad`), the second pass can no longer differ from the first. It is now dead code.

## Fix

Remove the stale re-check loop; keep a single cache-lookup pass before the DB query for the remaining misses.

## Tests

Added `TGetExistingSignerInfo` to the shared `dbtest` suite covering the multi-id found/not-found path. Passes on memory and sqlite backends.

Fixes #1874

### DIFF
--- a/token/services/storage/db/dbtest/identity.go
+++ b/token/services/storage/db/dbtest/identity.go
@@ -13,6 +13,7 @@ import (
 	"testing"
 	"time"
 
+	tdriver "github.com/LFDT-Panurus/panurus/token/driver"
 	"github.com/LFDT-Panurus/panurus/token/driver/mock"
 	idriver "github.com/LFDT-Panurus/panurus/token/services/identity/driver"
 	"github.com/LFDT-Panurus/panurus/token/services/logging"
@@ -62,6 +63,7 @@ var IdentityCases = []struct {
 	{"Configurations", TConfigurations},
 	{"GetConfiguration", TGetConfiguration},
 	{"SignerInfoConcurrent", TSignerInfoConcurrent},
+	{"GetExistingSignerInfo", TGetExistingSignerInfo},
 	{"RegisterIdentityDescriptor", TRegisterIdentityDescriptor},
 }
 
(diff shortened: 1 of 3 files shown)


adecaro (approved):
LGTM

PR #1888, opened 2026-07-10
VerifyTransfer in fabtoken is a local debug verification function, not the on-chain validator  #1768

### LINKED ISSUE #1768: VerifyTransfer in fabtoken is a local debug verification function, not the on-chain validator
The VerifyTransfer function located at token/core/fabtoken/v1/transfer.go:184-188 was initially flagged as a critical
  security issue because it returned nil without performing any verification. However, upon clarification from the
  maintainer, this understanding was incorrect.

  Key Points:

  1. VerifyTransfer is NOT the validator: This function is NOT executed by the token validator during transaction
  processing. It is a local verification function that parties can optionally use to verify if a transfer action is
  valid locally - primarily for debugging purposes.
  2. Actual validation happens elsewhere: The real on-chain validation is performed by these properly implemented
  validator functions in token/core/fabtoken/v1/validator/:
    - TransferActionValidate - Validates the transfer action structure
    - TransferSignatureValidate - Validates signatures for inputs spent by an action
    - TransferBalanceValidate - Checks that sum of inputs equals sum of outputs
    - TransferHTLCValidate - Validates HTLC scripts if present
  3. Implementation status: The VerifyTransfer function has now been implemented , but this is a
  "nice to have" feature for local verification - not a security-critical gap since the actual validation is handled by
  the separate validator functions.

adecaro:
TransferService.VerifyTransfer and IssueService.VerifyIssue were empty // TODO: stubs that unconditionally returned nil. Both are meant as sanity checks confirming that the action +
  metadata produced by Transfer()/Issue() (or received over the wire from a counterparty) are internally consistent — not merely a local debug helper, since Request.IsValid also invokes
  them when validating a TransferAction/IssueAction received from another party.

  This PR implements both, following the same balance/type rules already enforced by the on-chain validator (TransferBalanceValidate) and the metadata-matching pattern used by the
  zkatdlog driver's real implementations.

  Changes

  - VerifyTransfer (token/core/fabtoken/v1/transfer.go)
    - Rejects a nil or wrong-concrete-type action.
    - Runs action.Validate() for structural correctness.
    - Verifies Σ(input quantities) == Σ(output quantities) and that all inputs/outputs share a single token type — mirrors the authoritative on-chain TransferBalanceValidate.
    - Checks that the number of output metadata entries matches the number of outputs.
    - For each non-redeem output, verifies the receivers' audit info genuinely corresponds to their identity via Deserializer.MatchIdentity; redeem outputs must carry no receivers.
    - Requires a non-empty issuer identity whenever the transfer redeems tokens.
  - VerifyIssue (token/core/fabtoken/v1/issue.go)
    - Same nil/type and Validate() checks.
    - Confirms output metadata count matches outputs, each output has non-zero quantity, the metadata's recorded issuer matches the action's issuer, and each receiver's audit info matches
  its identity via MatchIdentity.
  - Tests (transfer_test.go, issue_test.go) — replaced the placeholder Verify*(ctx, nil, nil) → NoError tests with real cases: nil action, happy path, redeem path, unbalanced sums, type
  mismatch, metadata-count mismatch, missing/invalid receiver metadata, and audit-info mismatch.

### DIFF
--- a/token/core/fabtoken/v1/issue.go
+++ b/token/core/fabtoken/v1/issue.go
@@ -136,9 +136,14 @@ func (s *IssueService) VerifyIssue(ctx context.Context, ia driver.IssueAction, m
 			return errors.Errorf("output [%d] has zero quantity", i)
 		}
 
+		// Metadata for an output can be legitimately absent here: a TokenRequest received
+		// over the wire may have been filtered by enrollment ID before reaching us (see
+		// Metadata.FilterBy), in which case outputs we are not a receiver of carry a nil
+		// entry. Treat absent metadata as "nothing to verify from our vantage point" rather
+		// than as a validation failure, mirroring the analogous check in VerifyTransfer.
 		om := metadata[i] // #nosec G602 -- lengths already checked equal above
 		if om == nil || len(om.OutputMetadata) == 0 {
-			return errors.Errorf("missing metadata for output [%d]", i)
+			continue
 		}
 		outputMetadata := &v1.OutputMetadata{}
 		if err := outputMetadata.Deserialize(om.OutputMetadata); err != nil {
--- a/token/core/fabtoken/v1/issue.go
+++ b/token/core/fabtoken/v1/issue.go
@@ -111,7 +111,53 @@ func (s *IssueService) Issue(ctx context.Context, issuerIdentity driver.Identity
 
 // VerifyIssue checks if the outputs of an IssueAction match the passed tokenInfos
 func (s *IssueService) VerifyIssue(ctx context.Context, ia driver.IssueAction, metadata []*driver.IssueOutputMetadata) error {
-	// TODO:
+	if ia == nil {
+		return errors.Errorf("nil issue action")
+	}
+	action, ok := ia.(*v1.IssueAction)
+	if !ok {
+		return errors.Errorf("expected *actions.IssueAction, got [%T]", ia)
+	}
+	if err := action.Validate(); err != nil {
+		return errors.Wrap(err, "invalid action")
+	}
+	if len(action.Outputs) != len(metadata) {
+		return errors.Errorf("number of outputs [%d] does not match number of metadata entries [%d]", len(action.Outputs), len(metadata))
+	}
+
+	precision := s.PublicParamsManager.PublicParameters().Precision()
+	zero := token2.NewZeroQuantity(precision)
+	for i, out := range action.Outputs {
+		q, err := token2.ToQuantity(out.Quantity, precision)
+		if err != nil {
+			return errors.Wrapf(err, "failed parsing output quantity [%s]", out.Quantity)
+		}
+		if q.Cmp(zero) == 0 {
+			return errors.Errorf("output [%d] has zero quantity", i)
+		}
+
+		om := metadata[i] // #nosec G602 -- lengths already checked equal above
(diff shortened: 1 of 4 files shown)


token/core/fabtoken/v1/issue.go
```diff
@@ -111,7 +111,58 @@ func (s *IssueService) Issue(ctx context.Context, issuerIdentity driver.Identity
...
+			return errors.Errorf("output [%d] has zero quantity", i)
+		}
+
+		// Metadata for an output can be legitimately absent here: a TokenRequest received
+		// over the wire may have been filtered by enrollment ID before reaching us (see
+		// Metadata.FilterBy), in which case outputs we are not a receiver of carry a nil
+		// entry. Treat absent metadata as "nothing to verify from our vantage point" rather
+		// than as a validation failure, mirroring the analogous check in VerifyTransfer.
+		om := metadata[i] // #nosec G602 -- lengths already checked equal above
+		if om == nil || len(om.OutputMetadata) == 0 {
```
AkramBitar, line 145: om.OutputMetadata can be nil?

token/core/fabtoken/v1/issue.go
```diff
@@ -111,7 +111,58 @@ func (s *IssueService) Issue(ctx context.Context, issuerIdentity driver.Identity
...
+		}
+		outputMetadata := &v1.OutputMetadata{}
+		if err := outputMetadata.Deserialize(om.OutputMetadata); err != nil {
+			return errors.Wrapf(err, "failed unmarshalling metadata for output [%d]", i)
+		}
+		if !driver.Identity(outputMetadata.Issuer).Equal(action.Issuer) {
+			return errors.Errorf("issuer in metadata for output [%d] does not match issuer in action", i)
+		}
+
+		if len(om.Receivers) == 0 {
```
AkramBitar, line 156: Do we need to check that om.Receivers != nil?

token/core/fabtoken/v1/transfer.go
```diff
@@ -184,7 +184,77 @@ func (s *TransferService) Transfer(ctx context.Context, anchor driver.TokenReque
...
+	}
+	if len(action.Outputs) != len(outputMetadata) {
+		return errors.Errorf("number of outputs [%d] does not match number of metadata entries [%d]", len(action.Outputs), len(outputMetadata))
+	}
+
+	precision := s.PublicParametersManager.PublicParameters().Precision()
+
+	// check that the sum of the inputs is equal to the sum of the outputs, and that
+	// all inputs and outputs share the same token type
+	typ := action.Inputs[0].Input.Type
```
AkramBitar, line 205: Do we need to check?
if len(action.Inputs) == 0 {
    return errors.Errorf("transfer action has no inputs")
}

AkramBitar (approved):
LGTM

Left some comments

token/core/fabtoken/v1/transfer.go
adecaro: action.Validate() will fail in that case

token/core/fabtoken/v1/issue.go
adecaro: yes, it can

token/core/fabtoken/v1/issue.go
adecaro: if it is nil, then its length is zero. No?

PR #1889, opened 2026-07-10
feat(tokens): prepare UnspentTokensIteratorBy statements by argument shape

### LINKED ISSUE #1183: Introduce SQL Prepared Statements

sid200727:
## Summary
Implements prepared statements for `UnspentTokensIteratorBy` per the design discussed on #1183, following the benchmark work in #1855 that measured the win (+11% throughput, -73% P99 latency, -80% allocs/op).

## Changes
-> Added on-demand prepared-statement caching to `TokenStore`, keyed by argument shape (`walletID` present/absent, `tokenType` present/absent)
-> The 3 shapes that occur in current production call sites (documented on #1183) are each prepared once on first use and reused for all subsequent calls of that shape
-> Any call falling outside those 3 shapes (shouldn't happen given today's callers, but handled defensively) falls back to the existing dynamic query-building path, same for any prepare/execute error, so this degrades gracefully rather than failing
-> `Close()` now closes all cached prepared statements before closing the DB connections

## Design notes
- Statements are prepared lazily (on first use per shape) rather than eagerly at construction, since `newTokenStore` has no context/error return today and eager preparation would ripple that signature change through every constructor caller. Happy to revisit if eager preparation is preferred.
- Cache key is a cheap 2-byte shape encoding (`walletID present?`, `tokenType present?`), not the argument values themselves, statements are shared across different wallet/type values of the same shape.

## Testing
-> New unit test (`tokens_prepared_test.go` in `common`) covers the shape-key logic directly
-> New integration test (`tokens_prepared_test.go` in `postgres`) verifies statement reuse across repeated/varied calls of the same shape, and distinct caching per shape, against real Postgres
-> Existing `TestTokens` suite passes unchanged on both sqlite and postgres

Closes #1183


token/services/storage/db/sql/common/tokens.go
```diff
@@ -268,9 +276,78 @@ func buildUnspentTokensIteratorByQuery(db *TokenStore, walletID string, tokenTyp
...
+	}
+
+	return string(key[:])
+}
+
+// getOrPrepareUnspentTokensStmt returns a prepared statement for the given
+// argument shape, preparing and caching it on first use. The query args are
+// returned alongside since buildUnspentTokensIteratorByQuery computes both.
+func (db *TokenStore) getOrPrepareUnspentTokensStmt(ctx context.Context, walletID string, tokenType token.Type) (*sql.Stmt, []any, error) {
 	query, args := buildUnspentTokensIteratorByQuery(db, walletID, tokenType)
```
adecaro: this should be called right before PrepareContext

adecaro:
@sid200727 , thanks for this effort 🙏 
Please, could you create a helper struct that would encapsulate the logic of maintaining these statements?
Something like a statements holder struct that uses a `key` function, with signature (...any), and a `buildQuery` function, with a similar signature. Then, the TokenStore would instantiate as many holders as needed for the different queries. In UnspentTokensIteratorBy, at the end, we need to return

```go
	return &dedupedTokenRowsIterator{
		rows: rows,
		seen: make(map[string]struct{}),
	}, nil
```

no matter where rows comes from. So, the helper can also performs the query to the db with the prepared statement first or directly as needed.

What do you think?
I would like to isolate this logic in a struct external to the store itself so we can use it everywhere.

Thanks 🙏

sid200727:
Thank you @adecaro! 
Makes sense on both counts. 
One thing worth confirming before I implement: buildUnspentTokensIteratorByQuery currently builds the SQL text and collects the bound args together in one pass, so on a cache hit we still need bound args for QueryContext even though the text itself is already cached. I'd split it into a cheap buildArgs(...any) []any (always called) and keep the full text-building path gated to the miss branch only, wired through a generic preparedStmtHolder[K] type as you described (key/buildQuery/execute). 
Does that split match what you had in mind, or would you rather the holder always re-derive both together and just accept the small waste on hits?

adecaro:
> Thank you @adecaro! Makes sense on both counts. One thing worth confirming before I implement: buildUnspentTokensIteratorByQuery currently builds the SQL text and collects the bound args together in one pass, so on a cache hit we still need bound args for QueryContext even though the text itself is already cached. I'd split it into a cheap buildArgs(...any) []any (always called) and keep the full text-building path gated to the miss branch only, wired through a generic preparedStmtHolder[K] type as you described (key/buildQuery/execute). Does that split match what you had in mind, or would you rather the holder always re-derive both together and just accept the small waste on hits?

What fit best to have a generic statements holder 👍

sid200727:
Thank you @adecaro! 
Pushed the generalized preparedStmtHolder[K] in a new file (prepared_stmt_holder.go), TokenStore now instantiates one per query that needs this treatment. Went with buildQuery called on every execute (not just misses) since I confirmed UnspentTokensIteratorBy's bound args aren't cheaply separable from building the query text without risking a correctness bug, caught exactly that in testing before pushing (an earlier version derived args from placeholder values on cache hits and silently returned wrong results). 
Full TestTokens suite + the statement-reuse test are green on both sqlite and postgres. 
Ready for another look whenever you get a chance!

token/services/storage/db/sql/common/prepared_stmt_holder.go
```diff
@@ -0,0 +1,85 @@
...
+	"database/sql"
+	"sync"
+)
+
+// preparedStmtHolder caches prepared statements keyed by K. It centralizes
+// the get-or-prepare-and-execute logic shared by any query that runs
+// repeatedly with the same SQL shape but different bound argument values
+// (see #1183). TokenStore instantiates one holder per query that benefits
+// from this treatment.
+type preparedStmtHolder[K comparable] struct {
```
adecaro, line 46: please, introduce also an interface `PreparedStmtHolder` that can be used directly in the stores that require this holder. This will make unit-testing simpler.

adecaro: please, unit-test also this struct directly.

adecaro: investigate also the use of sync.Map that would guarantee lock-free reads. We are not updating any existing key once created, right?

token/services/storage/db/sql/common/tokens.go
```diff
@@ -268,11 +276,50 @@ func buildUnspentTokensIteratorByQuery(db *TokenStore, walletID string, tokenTyp
...
+		return &dedupedTokenRowsIterator{
+			rows: rows,
+			seen: make(map[string]struct{}),
+		}, nil
+	}
+
+	// Fall back to the dynamic path if preparing or executing the prepared
+	// statement fails (e.g. the driver does not support prepared statements,
+	// or the cached statement was invalidated). See #1183.
+	logging.Debug(logger, "prepared statement path failed, falling back to dynamic query", err)
```
adecaro: can we embed this fallback directly in execute?

token/services/storage/db/sql/common/tokens.go
```diff
@@ -268,11 +276,50 @@ func buildUnspentTokensIteratorByQuery(db *TokenStore, walletID string, tokenTyp
...
+}
+
+// PreparedStmtCount returns the number of cached prepared statements for
+// UnspentTokensIteratorBy. Intended for tests verifying statement reuse
+// across argument shapes.
+func (db *TokenStore) PreparedStmtCount() int {
+	return db.unspentTokensStmts.count()
+}
+
 func (db *TokenStore) UnspentTokensIteratorBy(ctx context.Context, walletID string, tokenType token.Type) (tdriver.UnspentTokensIterator, error) {
```
adecaro, line 302: please, try to unit-test this also in isolation without instantiating any actual DB.

sid200727:
Hi @adecaro!
pushed all the feedback: extracted PreparedStmtHolder[K] as an exported interface, switched to sync.Map per your write-once/read-many observation, folded the fallback into Execute, and added unit tests for the holder in isolation using go-sqlmock plus a DB-free test for PreparedStmtCount. Also merged in latest main. Full test suite is green on both sqlite and postgres. 
Ready for another look!
Thank you 🙏

token/services/storage/db/sql/common/prepared_stmt_holder.go
sid200727: Done! extracted as an exported interface.

sid200727: Added, see prepared_stmt_holder_test.go, using go-sqlmock (already a repo dep).

sid200727: yes! keys are write-once (double-checked on prepare), never overwritten, so this fits sync.Map exactly.

token/services/storage/db/sql/common/tokens.go
sid200727: Done, fallback now lives inside Execute, single call site returns rows either way.

token/services/storage/db/sql/common/tokens.go
sid200727: Added, TestTokenStore_PreparedStmtCount_NoDB, no DB involved.

token/services/storage/db/sql/common/prepared_stmt_holder.go
```diff
@@ -0,0 +1,116 @@
...
+	query, args, err := buildQuery()
+	if err != nil {
+		return nil, err
+	}
+
+	if stmt, err := h.getOrPrepare(ctx, db, key, query); err == nil {
+		if rows, qErr := stmt.QueryContext(ctx, args...); qErr == nil {
+			return rows, nil
+		}
+	}
```
adecaro, line 69: please, let's log a warning here otherwise an error here goes unnoticed

adecaro:
Great, @sid200727 . One last comment to address and we are good to merge. In another PR, it would be nice to attach also metrics here. But we will see. Thanks 🙏

token/services/storage/db/sql/postgres/tokens_prepared_test.go
```diff
@@ -0,0 +1,65 @@
...
+	tokentype "github.com/LFDT-Panurus/panurus/token/token"
+	fscpostgres "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/sql/postgres"
+	_ "github.com/jackc/pgx/v5/stdlib"
+	"github.com/stretchr/testify/require"
+)
+
+// TestUnspentTokensIteratorByPreparedReuse verifies that repeated calls with
+// the same argument shape reuse one prepared statement, and different shapes
+// prepare distinct statements (see #1183).
+func TestUnspentTokensIteratorByPreparedReuse(t *testing.T) {
```
adecaro: can this be transformed to a unit-test in `common` that does require the initialization of the actual db?

adecaro (approved):
LGTM. Thanks @sid200727, great job 🙏

adecaro (approved):
LGTM

PR #1890, opened 2026-07-10
[BUG]: RequestSpendView in multisig blocks forever — WithTimeout() exists but is never applied to the channel wait  #1671

### LINKED ISSUE #1671: [BUG]: RequestSpendView in multisig blocks forever — WithTimeout() exists but is never applied to the channel wait

adecaro:
Fixes #1671: RequestSpendView.Call() in the multisig and boolpolicy identity services could block forever waiting for a co-signer's answer, since the timeout field/WithTimeout() setter
  on multisig.RequestSpendView were dead code (never applied to the wait loop), and boolpolicy.RequestSpendView had no timeout knob at all.

  - Extracted the duplicated fan-out/collect pattern into a new generic utils.AnswersCollector[K, T] (token/services/utils/collector.go), which bounds the wait with a per-call timeout
  (defaulting to 30s if unset) and honors context cancellation, returning a descriptive wrapped error either way.
  - Rewired multisig/spend.go to actually apply its timeout/WithTimeout() via the new collector, and added the same timeout field + WithTimeout() setter to boolpolicy/spend.go, which
  previously had none.
  - Added regression tests for both packages proving Call now returns promptly (instead of hanging) when a co-signer/co-owner never responds, plus a happy-path test, and unit tests for
  the collector itself (all-answers, worker error, timeout, context cancellation, late-send-after-return).

  Additional changes

  - Enforced a new coding rule: never use fmt.Errorf/fmt to construct or wrap errors — always use github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors (New, Errorf, Wrap,
  Wrapf, WithMessage, WithMessagef, Join, etc.). Documented this in AGENTS.md and fixed the one violation found in docs/upgradability.md, plus rewrote collector.go's two error sites to
  use errors.Join/errors.Wrapf instead of fmt.Errorf.
  - Added a "Documentation Updates" workflow rule to AGENTS.md requiring docs/ to be updated (when affected) before any task is reported as complete.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -158,6 +158,7 @@ All PRs and pushes to `main` trigger these workflows:
 
 ### Coding Standards
 - **Error Handling**: Handle errors explicitly; avoid blank identifier for errors
+- **Error Construction**: Never use `fmt.Errorf` (or `fmt` at all) to build or wrap errors. Always use `github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors` instead (`errors.New`, `errors.Errorf`, `errors.Wrap`, `errors.Wrapf`, `errors.WithMessage`, `errors.WithMessagef`, `errors.Join`, etc.). This applies to source code, tests, and code samples in `docs/`.
 - **Interfaces**: Define small, focused interfaces on consumer side; favor composition
 - **Concurrency**: Use goroutines and channels; avoid shared state; validate with race detector
 - **Globals**: Avoid global variables for testability
@@ -178,7 +179,7 @@ Before implementing any task:
 3. Log blockers/decisions under `## Notes & Decisions`
 4. Mark plan as `✅ COMPLETE` when finished
 
-### Documentation Rule
+### Documentation Updates (Workflow Rule)
 Before marking a task complete, update or create the relevant documentation under `docs/`:
 - If the task changes a public API, protocol, or user-facing behaviour, update the corresponding `docs/` page (or create one if it does not exist).
 - Keep docs consistent with code: function names, message fields, flow diagrams, and examples must match the implementation.
--- a/docs/upgradability.md
+++ b/docs/upgradability.md
@@ -172,7 +172,7 @@ ppm := tms.PublicParametersManager()
 // Retrieve the current public parameters
 currentPP := ppm.PublicParameters()
 if currentPP == nil {
-    return fmt.Errorf("public parameters not available")
+    return errors.New("public parameters not available")
 }
 
 // Verify the parameters match the expected values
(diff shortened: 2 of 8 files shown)


KElkhiyaoui (approved):
Looks good to me.

PR #1893, opened 2026-07-10
fix(ttx): close endorsement sessions on all return paths

### LINKED ISSUE #1637: Implement Structured Failure Handling and Resource Cleanup [MED]
**Description**
The multi-phase orchestration of `CollectEndorsementsView`  --  signature collection, auditing, endorsement request, and distribution  --  must treat every phase as a potential failure point and guarantee that a failure in any one phase triggers an orderly teardown of all resources acquired in prior phases. Concretely, this means: open sessions are closed, database locks (such as the audit-DB enrollment-ID locks acquired during `Audit`) are released, in-progress storage writes are rolled back or marked invalid, and all spawned goroutines are cancelled via context propagation. Without this, a partial failure leaves dangling sessions, held locks, and orphaned records that accumulate over time and themselves become a resource-exhaustion vector. A `defer`-based cleanup chain (or equivalent RAII pattern) at each phase boundary, combined with explicit error-path tests, is the canonical way to enforce this invariant.

AkramBitar:
Fixes #1637.

`cleanupAudit` only ran after distribution succeeded, so a failure in a later phase (`requestApproval` or `distributeTxToParties`) returned from `CollectEndorsementsView.Call` early and left the auditor session open — a repeatable leak of the FSC communication session on every error path (hence the `security` label).

Note: this does **not** hold any audit-DB enrollment-ID lock. That lock lives on the auditor node and is released by `Append`'s `defer Release()` before the auditor signature is sent back — i.e. before `requestAudit` returns — so it is already released by the time any later phase can fail. The bug is strictly a leaked session handle.

Fix: close all endorsement sessions in `CollectEndorsementsView.Call` via a deferred, idempotent `cleanupSessions` (safe on repeated calls and on any return path); add TTX unit tests.

### DIFF
--- a/token/services/ttx/cleanupsessions_test.go
+++ b/token/services/ttx/cleanupsessions_test.go
@@ -1,0 +1,77 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// This white-box (package ttx) file covers cleanupSessions, which needs access to
+// the unexported sessions field. It is kept separate from the black-box
+// collectendorsements_test.go so the latter can keep importing dep/mock, which
+// cannot be imported from package ttx without creating an import cycle.
+package ttx
+
+import (
+	"context"
+	"testing"
+
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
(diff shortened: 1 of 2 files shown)


adecaro (approved):
LGTM

PR #1896, opened 2026-07-13
perf(ttx): batch concurrent finality status lookups; remove dead transaction filter

### LINKED ISSUE #1872: perf: concurrent finality views issue unbatched single-tx status queries; remove dead commit-pipeline filter
## Problem

Originally this issue targeted `AcceptTxInDBsFilter.Accept`, which issues up to two serial DB round trips per transaction. However, that filter turned out to be dead code: its registration with FSC's committer (`AddTransactionFilter`) was intentionally removed in cd7e7a44 (#1200) because the FSC fabric commit pipeline was too slow, and the filter has not been invoked since. The `filterProvider` plumbing remained threaded through the fabric/fabricx network drivers and the dig container without any consumer.

The real occurrence of the pattern is in the ttx finality path (`token/services/ttx/finality.go`): every finality view queries ttxdb and auditdb for transaction status — an initial known-check against both stores, plus a periodic poll — each a separate synchronous single-tx `GetStatus` round trip. Under concurrent load (many transactions awaiting finality at once), N concurrent views issue O(N) individual queries against the same stores.

## Fix (implemented in #1896)

- Remove the dead code: `AcceptTxInDBsFilter`, `AcceptTxInDBFilterProvider`, `TransactionFilterProvider`, and the unused `filterProvider` plumbing.
- Add a batch status lookup `GetStatuses(ctx, txIDs) (map[string]TxStatus, error)` to the ttxdb/auditdb store services.
- Wrap each store service in a per-TMS batching decorator behind the `dep.GetTransactionDB`/`GetAuditDB` providers: single-tx `GetStatus` lookups landing within a 300µs window are coalesced into one batched query per store. `finality.go` itself is unchanged — the batching is transparent at the provider layer.

## Test plan

- Batcher unit tests: coalescing, per-batch deduplication, missing-tx→Unknown semantics, error propagation, context cancellation.
- Decorator/provider tests: concurrent `GetStatus` calls issue a single batched query; listeners pass through; one decorator (and one batcher) cached per TMS ID.
- Existing finality view tests pass unchanged.

hstarorg:
Reworked per review discussion (https://github.com/LFDT-Panurus/panurus/pull/1896#issuecomment-4955122265 and below): the FSC commit-pipeline transaction filter path was intentionally removed in cd7e7a44 (#1200) and should not be restored, so this PR now does two things instead of optimizing the filter.

**1. Remove the dead transaction-filter code**

`AcceptTxInDBsFilter`, `AcceptTxInDBFilterProvider`, `TransactionFilterProvider`, and the inert `filterProvider` plumbing threaded through the fabric/fabricx network drivers and the dig container are deleted. Nothing consumed them since the `AddTransactionFilter` registration was removed in cd7e7a44.

**2. Batch concurrent status lookups in the ttx finality path**

Each finality view queries ttxdb and auditdb for transaction status (initial known-check plus periodic polling), so under concurrent load many single-tx `GetStatus` queries hit the same stores. This PR adds `GetStatuses(ctx, txIDs)` to the ttxdb/auditdb store services and wraps each store in a per-TMS batching decorator behind the `dep.GetTransactionDB`/`GetAuditDB` providers: single-tx lookups landing within a 300µs window are coalesced into one batched query per store. `finality.go` itself is unchanged — the batching is transparent at the provider layer.

Notes:
- Missing tx ids are absent from the batched result and map to `Unknown` with no error; all finality call sites treat that identically to the old per-tx error path.
- The status message is not part of the batched query and is returned empty; the interface's only consumer (the finality view) discards it.
- The batched query runs on `context.Background()` since it serves multiple callers; each waiter still honors its own context while waiting.

Refs LFDT-Panurus/panurus#1872


adecaro:
Please, next time, get first the issue assigned and then start with the implementation. Many thanks 🙏

adecaro:
Hi @hstarorg , what's the performance improvement that you have seen with this patch?

hstarorg:
## Heads-up: this filter currently isn't reachable from FSC's committer

Before merging — I want to flag something I found while re-reviewing, since it changes what this PR actually delivers at runtime.

`AcceptTxInDBsFilter` (the filter this PR optimizes) is not currently registered with FSC's committer anywhere in this repo. The two calls that used to do that —

```go
transactionFilter, err := n.filterProvider.New(tmsID)
...
committer := n.ch.Committer()
if err := committer.AddTransactionFilter(transactionFilter); err != nil {
```

— were removed from `token/services/network/fabric/network.go`'s `connect()` in cd7e7a44 (#1200, "remove RWSetProcessor", Aug 12 2025). The `filterProvider` field/parameter itself was left in place and is still threaded through `NewNetwork` (both `fabric` and `fabricx` variants) and the `dig` container, but nothing reads it anymore — it's just passed along and stored, never consumed. A repo-wide search (excluding the vendored fabric-smart-client module) turns up zero calls to `AddTransactionFilter`.

Practical effect: `AcceptTxInDBsFilter.Accept` is dead code on `main` today. This PR's batching is correct and well-tested (verified independently — build/vet/gofmt/golangci-lint clean, `-race` clean, CI green), but it currently has zero runtime effect, because FSC never calls `Accept` in the first place. Issue #1872's premise ("FSC calls it once per unknown transaction while committing a block") no longer holds on `main`.

This predates this PR and isn't something it needs to fix by itself, but it seems worth deciding explicitly before merge:
- merge as prep work and open a follow-up to restore the `filterProvider.New` + `AddTransactionFilter` registration (removed in cd7e7a44) — that's the only way this optimization actually takes effect, or
- fold the re-registration into this PR, or
- hold off if there's a reason the filter registration was intentionally dropped that I'm missing (worth confirming with whoever wrote cd7e7a44/#1200).

Happy to send the follow-up PR restoring the registration if that's the preferred path.

hstarorg:
> Hi @hstarorg , what's the performance improvement that you have seen with this patch?
What it does: once that registration exists, concurrent `Accept` calls
(currently up to 2 serial DB queries each) get merged into one batched
query per store within a 300us window — O(1) per window instead of O(N).

Note: the current caller path has that registration removed
(`AddTransactionFilter` was dropped in cd7e7a44/#1200), so this filter
isn't invoked today.

hstarorg:
> Hi @hstarorg , thanks for submitting this 🙏 Please, next time, get first the issue assigned and then start with the implementation. Many thanks 🙏

Understood, will do 🙏

adecaro:
Hi @hstarorg , super interesting. So, the reliance on the FSC's fabric's commit pipeline was removed because that path was too slow. We should not restore therefore it and I would suggest to remove the unused code.

Nevertheless, it might be that your patch might still be helpful in `token/services/ttx/finality.go`, where in addition to a finality listener, the code queries the db for transaction status. What do you think?

adecaro:
Hi @hstarorg , any follow up on this? Thanks 🙏

hstarorg:
> Hi @hstarorg , any follow up on this? Thanks 🙏

Thanks for the pointer — that makes sense. I'll rework this PR in that direction: drop the unused `AcceptTxInDBsFilter` / `filterProvider` plumbing, and wire the status batching into `token/services/ttx/finality.go` instead, where the concurrent `GetStatus` calls actually happen.

Will update the PR shortly.

hstarorg:
@adecaro PR updated as discussed:

- Removed the dead code: `AcceptTxInDBsFilter`, its provider, and the unused `filterProvider` plumbing in the fabric/fabricx network drivers and the dig container.
- Re-targeted the batching at the ttx finality path: the providers behind `dep.GetTransactionDB`/`GetAuditDB` now wrap each store service in a per-TMS batching decorator, so single-tx `GetStatus` lookups from concurrent finality views get coalesced into one batched `GetStatuses` query per 300µs window. `finality.go` itself is unchanged.

One note: the batched query runs on `context.Background()` because it serves multiple callers; each waiter still honors its own context while waiting.

adecaro:
Hi @hstarorg , thanks for this effort. I'll review ASAP 🙏 

Do you have already information about the impact of these changes on the performance of your deployment?

hstarorg:
Not yet — we still need some time to validate the improvement in a real deployment, so for now the benefit is theoretical: N concurrent finality views issuing single-tx `GetStatus` queries collapse into one batched query per store per 300µs window, so DB round trips go from O(N) to O(1) per window under concurrent load. I'll follow up with measured numbers once we have them.

adecaro:
> Not yet — we still need some time to validate the improvement in a real deployment, so for now the benefit is theoretical: N concurrent finality views issuing single-tx `GetStatus` queries collapse into one batched query per store per 300µs window, so DB round trips go from O(N) to O(1) per window under concurrent load. I'll follow up with measured numbers once we have them.

Very good, thanks @hstarorg . Merging for now this. Let us know once you have numbers.

adecaro (approved):
LGTM

adecaro:
@hstarorg , please, could you update the Issue to better reflect the problem addressed by this PR? 🙏 

Please, can you also rebase? I'm not able to do that form the web ui.

hstarorg:
@adecaro Both done:

- Rebased onto latest main — the branch is now 3 linear commits, no merge commit.
- Updated the issue (#1872) to reflect what this PR actually addresses: the dead commit-pipeline filter removal, and the batching of concurrent finality status lookups.

adecaro:
@hstarorg , another rebase, sorry. I merged another PR from @EvanYan1024 that was ready. Next in the pipeline is this one. Thanks for the patience 🙏

hstarorg:
@adecaro I've added a simulated performance benchmark to the PR (`token/services/ttx/dep/wrapper/status_batcher_bench_test.go`), so the numbers below are reproducible with `go test -bench BenchmarkStatusLookups -benchtime 5000x ./token/services/ttx/dep/wrapper/`.

What it simulates: N workers model concurrent finality views doing single-tx status lookups against a real sqlite-backed store, comparing direct `GetStatus` calls vs the batching path. A second profile adds a simulated 1ms round trip with a 16-connection pool to approximate a remote postgres. Not a real deployment measurement yet — those numbers will follow.

Results:

- **DB queries per lookup** (backend-independent): 8 workers → 0.125 (8x fewer queries), 32 → 0.031 (32x), 128 → 0.0084 (119x). Coalescing is essentially perfect — DB load drops by the concurrency factor.
- **Wall time per lookup** (remote profile, direct vs batched): 32 workers: 69µs vs 50µs (1.4x faster), 128 workers: 69µs vs 14µs (4.8x faster). At low concurrency (1–8 workers) direct is faster, since every batched call pays the 300µs batch window as a latency tax — that's the documented tradeoff and the reason the window is kept small.

Net: the main benefit is DB load reduction under concurrent finality traffic, plus throughput gains once concurrency exceeds the connection pool.

PR #1898, opened 2026-07-13
respect context on wallet operations

### LINKED ISSUE #1634: Context-Propagated Deadlines and Timeouts [MED]
**Description**
Every wallet operation that touches the database, the cryptographic backend, or the identity cache must respect a short, caller-supplied `context.Context` deadline. Operations that exceed their deadline must be cancelled and their resources released immediately. This prevents a slow or stalled operation (e.g., a database query holding a lock, or a cryptographic derivation consuming CPU) from blocking the service for subsequent legitimate callers. Timeouts should be enforced at the service boundary before the call is dispatched to the registry or identity provider, not only inside individual sub-calls.

HayimShaul:
respect context timeout and cancelation on wallet operations
Addresses issue #1634


adecaro:
okay, I think we don't need this PR, @HayimShaul . The functions are already taking in input the context. The caller can set the deadline as necessary. We don't need to add an additional configuration there. 
@AkramBitar , what do you think?

HayimShaul:
The timeouts and deadline are already respected

PR #1900, opened 2026-07-13
Remove redundant indexes that duplicate primary keys

AkramBitar:
Follow-up to #1891 (indexing work for #1873).

Removes six indexes that duplicated — or were a leading prefix of — their table's `PRIMARY KEY`. A non-unique index that is an exact prefix of the PK's btree provides no read benefit (the PK index already serves those prefix lookups on both PostgreSQL and SQLite), but every one of them is maintained on each `INSERT`/`UPDATE`. Dropping them is a pure write-path win.

| File | Index removed | Duplicates |
|------|---------------|-----------|
| `identity.go` | `idx_ic_id_type` | `PRIMARY KEY(id, type, url)` (exact) |
| `identity.go` | `idx_audits` | `PRIMARY KEY(identity_hash)` (exact) |
| `identity.go` | `idx_signers` | `PRIMARY KEY(identity_hash)` (exact) |
| `wallet.go` | `idx_identity_hash_and_wallet_and_role` | `PRIMARY KEY(identity_hash, wallet_id, role_id)` (exact) |
| `wallet.go` | `idx_identity_hash` | leading prefix of the same PK |
| `tokens.go` | `idx_tx_id` | leading prefix of `PRIMARY KEY(tx_id, idx)` |

### DIFF
--- a/token/services/storage/db/sql/common/identity.go
+++ b/token/services/storage/db/sql/common/identity.go
@@ -535,7 +535,6 @@ func (db *IdentityStore) GetSchema() string {
 			PRIMARY KEY(id, type, url)
 		);
 		CREATE INDEX IF NOT EXISTS idx_ic_type_%s ON %s ( type );
-		CREATE INDEX IF NOT EXISTS idx_ic_id_type_%s ON %s ( id, type, url );
 
 		-- IdentityInfo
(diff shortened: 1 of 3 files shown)


adecaro (approved):
LGTM

PR #1905, opened 2026-07-14
fix(token): preserve auditor signatures in Request.SetSignatures

### LINKED ISSUE #1897: Request.SetSignatures silently drops auditor signatures attached beforehand (since #1717)
### Context

PR #1717 (the protobuf refactor for #1622) changed how signatures are stored on a token request. Before, action signatures and auditor signatures lived in two separate fields:

```go
// before #1717
func (r *Request) AddAuditorSignature(identity Identity, sigma []byte) {
	r.Actions.AuditorSignatures = append(r.Actions.AuditorSignatures, ...)
}

func (r *Request) SetSignatures(sigmas map[string][]byte) bool {
	...
	r.Actions.Signatures = signatures // action signatures only
}
```

Since #1717 both kinds share the single unified slice `Actions.Signatures` (`[]*driver.RequestSignature` with either `.Action` or `.Auditor` set), but `SetSignatures` kept its whole-field assignment:

```go
// since #1717 (token/request.go)
r.Actions.Signatures = signatures // only action signatures — pre-attached auditor entries are discarded
```

### Behavior change

Before #1717, `AddAuditorSignature` and `SetSignatures` were order-independent. Since #1717, calling `AddAuditorSignature` before `SetSignatures` silently drops the auditor signature; endorsement later fails in `common.AuditingSignaturesValidate` with `auditor signatures missing`.

We hit this in our test environments with a flow that requests auditing before collecting endorsements (and passes `WithSkipAuditing()` to `CollectEndorsementsView`). The SDK's own default flow is unaffected because `CollectEndorsementsView` happens to call `SetSignatures` before `requestAudit`.

### Minimal reproduction

```go
r := &token.Request{
	Actions:  &driver.TokenRequest{},
	Metadata: &driver.TokenRequestMetadata{},
}
r.AddAuditorSignature(token.Identity("auditor1"), []byte("sigma"))
r.SetSignatures(map[string][]byte{})
// r.Actions.Signatures is now empty — the auditor signature is gone
```

### Question

Was this an intentional contract change (i.e., `SetSignatures` must now always run before `AddAuditorSignature`), or an unintended side effect of merging the two signature containers in #1717?

Neither #1622 nor #1717 mentions the signature-ordering semantics, so we suspect the latter. If so, we are happy to submit a small fix: have `SetSignatures` preserve existing `.Auditor` entries when assigning the action signatures (restoring the pre-#1717 contract), with a regression test.

Refs #1622, #1717

YanYuanFE:
## Motivation

Since #1717 action signatures and auditor signatures share the unified `Actions.Signatures` slice, but `Request.SetSignatures` kept its whole-field assignment. Any auditor signature attached beforehand via `AddAuditorSignature` is silently discarded, and endorsement later fails in `common.AuditingSignaturesValidate` with `auditor signatures missing`.

We hit this in our test environments with a flow that requests auditing before collecting endorsements (passing `WithSkipAuditing()` to `CollectEndorsementsView`). The SDK's default flow is unaffected because it happens to call `SetSignatures` before `requestAudit`. See #1897 for the full analysis; maintainers confirmed there that this is an unintended side effect of #1717.

## Changes

- `SetSignatures` now preserves existing `.Auditor` entries when assigning action signatures, restoring the pre-#1717 order-independence of `AddAuditorSignature` and `SetSignatures`. The resulting slice order (action entries first, auditor entries after) matches what the default flow produces.
- Regression test `TestRequest_SetSignatures_PreservesAuditorSignatures`: an auditor signature attached before `SetSignatures` survives the call, and a second `SetSignatures` call does not duplicate it.

Fixes #1897

### DIFF
--- a/token/request.go
+++ b/token/request.go
@@ -1168,9 +1168,11 @@ func (r *Request) AddAuditorSignature(identity Identity, sigma []byte) {
 	})
 }
 
-// SetSignatures assigns signatures to all signers. Returns true if all signatures are present.
-// This method now correctly preserves the action context by using SignerWithAction information.
-func (r *Request) SetSignatures(sigmas map[string][]byte) bool {
+// AppendSignatures appends the action signatures for all issue and transfer
+// signers from sigmas, keyed by signer unique ID. Signatures already attached
+// to the request, e.g. via AddAuditorSignature, are left in place. Returns
+// true if every signer has a signature in sigmas.
+func (r *Request) AppendSignatures(sigmas map[string][]byte) bool {
 	issueSignersWithActions := r.IssueSignersWithActions()
 	transferSignersWithActions := r.TransferSignersWithActions()
 
@@ -1198,7 +1200,7 @@ func (r *Request) SetSignatures(sigmas map[string][]byte) bool {
 		signatures = append(signatures, signature)
 	}
 
-	r.Actions.Signatures = signatures
+	r.Actions.Signatures = append(r.Actions.Signatures, signatures...)
 
 	return all
 }
--- a/token/request_test.go
+++ b/token/request_test.go
@@ -890,8 +890,8 @@ func TestRequest_IssueSigners(t *testing.T) {
 	assert.Equal(t, Identity("issuer2"), signers[3])
 }
 
-// TestRequest_SetSignatures tests SetSignatures method
-func TestRequest_SetSignatures(t *testing.T) {
+// TestRequest_AppendSignatures tests AppendSignatures method
+func TestRequest_AppendSignatures(t *testing.T) {
 	r := &Request{
 		Actions: &driver.TokenRequest{
 			Actions: []*driver.TypedAction{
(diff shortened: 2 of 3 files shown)


token/request.go
```diff
@@ -1170,6 +1170,7 @@ func (r *Request) AddAuditorSignature(identity Identity, sigma []byte) {
 
 // SetSignatures assigns signatures to all signers. Returns true if all signatures are present.
 // This method now correctly preserves the action context by using SignerWithAction information.
+// Auditor signatures already attached to the request are preserved.
```
adecaro: at this point, Let's rename this function and call it AppendSignatures. The contract of the function will be that any existing signature is left in place. What do you think?

adecaro (commented):
@YanYuanFE , left a comment. Please, have a look.

token/request.go
YanYuanFE: Good point, done

token/request.go
```diff
@@ -1198,6 +1200,13 @@ func (r *Request) SetSignatures(sigmas map[string][]byte) bool {
...
 	}
 
+	// preserve auditor signatures attached before this call, e.g., via AddAuditorSignature
+	for _, signature := range r.Actions.Signatures {
+		if signature != nil && signature.Auditor != nil {
+			signatures = append(signatures, signature)
+		}
+	}
+
 	r.Actions.Signatures = signatures
```
adecaro: let's just append the signatures here, we don't need the above loop, no?

adecaro:
@YanYuanFE , thanks for the prompt update. Another small thing and we are good to merge 🙏

adecaro (approved):
LGTM

token/request.go
YanYuanFE: okay

PR #1914, opened 2026-07-15
fix(ttx): merge appendPayload instead of overwriting TokenRequest/Transient

### LINKED ISSUE #1913: ttx: Transaction.appendPayload overwrites TokenRequest/Transient instead of merging
## Summary

`Transaction.appendPayload` in `token/services/ttx/transaction.go` merges a
received `Payload` into the current transaction by wholesale overwrite instead
of an actual merge:

```go
func (t *Transaction) appendPayload(payload *Payload) error {
	// TODO: change this
	t.TokenRequest = payload.TokenRequest
	t.Transient = payload.Transient

	return nil

	// ~25 lines of dead, commented-out merge code (Issues/Transfers/TokenInfos/ValidationRecords)
}
```

Both the function-level TODO ("This implementation is incomplete and needs to
be enhanced to properly merge all payload components") and the inline TODO
("change this") flag this as a known, long-standing gap.

## Where it's used

`appendPayload` has exactly one call site: `collectactions.go`, inside
`collectRemote`, invoked once per action from `CollectActionsView.Call` when a
transfer action is collected from a remote counterparty. Each iteration
deserializes a fresh `Payload` from the counterparty's response and calls
`appendPayload` to fold it into the initiator's in-progress transaction.

## Problem

- **`TokenRequest` overwrite**: `t.TokenRequest = payload.TokenRequest` replaces
  the entire token request rather than merging in only the new content. In the
  current collect-actions protocol this happens to be safe *in practice*
  because the responder round-trips the initiator's full transaction state and
  appends its own action before replying — so the incoming request is usually
  a superset of what's already on `t`. But nothing verifies this invariant. If
  the protocol changes, if a responder ever returns a partial/stale request,
  or if `CollectActionsView` is ever driven with multiple concurrent/interleaved
  remote actions, this overwrite can silently drop previously-collected actions
  with no error surfaced.

- **`Transient` overwrite**: `t.Transient = payload.Transient` unconditionally
  replaces the initiator's transient map with whatever the remote party's
  (typically near-empty, freshly-reconstructed) transient map contains. Any
  transient data the initiator had set before or during the collect loop is
  silently discarded. This is not currently triggered by any in-tree caller
  (no production code writes to `Transaction.Transient` yet), but the bug is
  real and will surface silently the moment something does.

- **Dead code**: the commented-out "intended" merge implementation below the
  early return references fields (`payload.Request.Issues`, `TokenInfos`,
  `ValidationRecords`, `t.Set(key, v)`) that no longer exist on the current
  `Payload`/`token.Request` model — it predates a refactor of the request type
  and cannot be reinstated as-is.

- **No test coverage**: there are no unit or integration tests exercising
  `appendPayload`, `collectRemote`, or `CollectActionsView` with more than one
  action (mixed local+remote, or multiple remote actions). All existing
  integration call sites (`integration/token/dvp`, `integration/token/fungible`
  swap) only ever collect a single action, so this code path is effectively
  untested.

## Impact

Silent, unflagged data loss in transaction assembly if the "incoming request
is always a superset" assumption is ever violated, and in the transient-data
path as soon as any caller starts using it. Because failures here are silent
(no error, no panic), this is the kind of gap that's easy to miss until an
audit, reconciliation step, or a future protocol change exposes it.

## Suggested fix direction

Replace the overwrite with an explicit, checked merge:
- Verify the incoming `TokenRequest`'s actions are an extension of the actions
  already on the transaction (same prefix, same order); error out rather than
  silently accept a shorter or diverging request.
- Merge `Transient` entries without letting the incoming payload clobber keys
  already set locally.
- Remove the stale dead code.
- Add unit tests covering non-empty pre-existing state (today's tests only
  exercise the empty-state case).

adecaro:
## Summary

Fixes #1913.

`Transaction.appendPayload` (`token/services/ttx/transaction.go`) previously merged
a received `Payload` into the current transaction by wholesale overwrite:

```go
func (t *Transaction) appendPayload(payload *Payload) error {
	// TODO: change this
	t.TokenRequest = payload.TokenRequest
	t.Transient = payload.Transient
	return nil
	// ~25 lines of dead, commented-out merge code
}
```

This is the single call site used by `collectRemote` (`collectactions.go`), invoked
once per action when `CollectActionsView` collects a transfer action from a remote
counterparty.

## Problem

- **`TokenRequest` overwrite**: replaced the entire token request instead of merging
  in only new content. Safe *today* only because the collect-actions responder
  happens to round-trip the initiator's full transaction state before replying — an
  unchecked assumption. A protocol change, a partial/stale responder reply, or
  multiple concurrent/interleaved remote actions could silently drop previously
  collected actions with no error.
- **`Transient` overwrite**: unconditionally replaced the initiator's transient map
  with the (typically near-empty) remote one, silently discarding any transient data
  set locally. Currently latent (no in-tree writer populates `Transaction.Transient`
  yet) but a real bug waiting for its first caller.
- **Dead code**: the commented-out block referenced `payload.Request.Issues`,
  `TokenInfos`, `ValidationRecords`, and `t.Set(...)` — none of which exist on the
  current `Payload`/`token.Request` model. It predates a request-type refactor and
  could not be reinstated as-is.
- **No test coverage** for `appendPayload`, or for collecting more than one action.

## Fix

Replaced the overwrite with an explicit, checked merge:

- **`TokenRequest`**: if the transaction has no actions yet, adopt the incoming
  request unconditionally (today's normal first-write path, unchanged behavior).
  Otherwise, verify the incoming request's actions are a superset-extension of the
  existing ones (same type + raw bytes, same order, for every existing action) before
  adopting it. A violation now returns an explicit error instead of silently
  corrupting/losing state.
- **`Transient`**: entries from the incoming payload are merged in only for keys not
  already present locally — existing values win, so nothing set on the initiator can
  be clobbered by a remote round-trip.
- Deleted the stale, unusable commented-out code.
- Added white-box unit tests (`transaction_internal_test.go`) covering: first-write
  adoption, superset-extension success, shorter/divergent-prefix violations (both
  error), transient existing-wins merge, transient merge into a nil map, and a
  combined `appendPayload` merge.
- Documented the merge semantics in `docs/services/ttx.md`'s Collect Actions Flow
  section.

No behavior change for existing single-action integration call sites
(`integration/token/dvp`, `integration/token/fungible` swap) — they only ever
exercise the first-write path.

## Test plan

- [x] `go test ./token/services/ttx/...` — new unit tests pass
- [x] `make unit-tests-race` — no races, ttx package green
- [x] `make checks` — license/fmt/vet/staticcheck clean

### DIFF
--- a/token/services/ttx/transaction.go
+++ b/token/services/ttx/transaction.go
@@ -429,13 +429,16 @@ func (t *Transaction) TMSID() token.TMSID {
 // request must be a superset-extension of the current one (i.e. its actions start with
 // exactly the actions already present in this transaction, in the same order) since that
 // is what the collect-actions protocol produces; violating this invariant is treated as
-// an error rather than silently accepted. Transient entries from the payload are merged in
-// without overwriting any key already set on this transaction.
+// an error rather than silently accepted. Transient entries from the payload are merged in;
+// a key already set on this transaction with a different value is treated as an error
+// rather than silently resolved.
 func (t *Transaction) appendPayload(payload *Payload) error {
 	if err := t.mergeTokenRequest(payload.TokenRequest); err != nil {
 		return errors.Wrap(err, "failed merging token request")
 	}
-	t.mergeTransient(payload.Transient)
+	if err := t.mergeTransient(payload.Transient); err != nil {
+		return errors.Wrap(err, "failed merging transient")
+	}
 
 	return nil
 }
@@ -471,15 +474,23 @@ func (t *Transaction) mergeTokenRequest(incoming *token.Request) error {
 	return nil
 }
 
-// mergeTransient copies entries from incoming into this transaction's transient map,
-// without overwriting any key already present.
-func (t *Transaction) mergeTransient(incoming network.TransientMap) {
+// mergeTransient copies entries from incoming into this transaction's transient map.
+// If a key is already present with a different value, it returns an error rather than
+// silently picking one side.
+func (t *Transaction) mergeTransient(incoming network.TransientMap) error {
 	if t.Transient == nil {
 		t.Transient = network.TransientMap{}
 	}
 	for k, v := range incoming {
-		if !t.Transient.Exists(k) {
-			t.Transient[k] = v
+		if existing, ok := t.Transient[k]; ok {
+			if !bytes.Equal(existing, v) {
+				return errors.Errorf("incoming transient value for key [%s] diverges from the existing one", k)
+			}
+
+			continue
 		}
+		t.Transient[k] = v
 	}
+
+	return nil
 }
--- a/token/services/ttx/transaction.go
+++ b/token/services/ttx/transaction.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package ttx
 
 import (
+	"bytes"
 	"context"
 
 	"github.com/LFDT-Panurus/panurus/token"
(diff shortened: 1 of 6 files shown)


KElkhiyaoui (approved):

token/services/ttx/transaction.go
```diff
@@ -424,38 +425,72 @@ func (t *Transaction) TMSID() token.TMSID {
...
-	//	}
-	// }
-	// return nil
+// mergeTransient copies entries from incoming into this transaction's transient map.
+// If a key is already present with a different value, it returns an error rather than
+// silently picking one side.
+func (t *Transaction) mergeTransient(incoming network.TransientMap) error {
+	if t.Transient == nil {
+		t.Transient = network.TransientMap{}
+	}
```
AkramBitar, line 483: What if incoming is nil?

token/services/ttx/transaction.go
```diff
@@ -424,38 +425,72 @@ func (t *Transaction) TMSID() token.TMSID {
...
+// exactly the actions already present in this transaction, in the same order) since that
+// is what the collect-actions protocol produces; violating this invariant is treated as
+// an error rather than silently accepted. Transient entries from the payload are merged in;
+// a key already set on this transaction with a different value is treated as an error
+// rather than silently resolved.
 func (t *Transaction) appendPayload(payload *Payload) error {
-	// TODO: change this
-	t.TokenRequest = payload.TokenRequest
-	t.Transient = payload.Transient
+	if err := t.mergeTokenRequest(payload.TokenRequest); err != nil {
```
AkramBitar, line 436: If incoming is nil, no error is return and this considered ok. 
Is that ok to continue with mergeTransient?

AkramBitar (changes requested):
LGTM.

I added to comments. Please have a look.

token/services/ttx/transaction.go
adecaro: nothing, the loop does not run

token/services/ttx/transaction.go
adecaro: do you have in mind a case where we should not append the transient if the token request is nil?

AkramBitar: No I do not have one

AkramBitar (approved):
LGTM

PR #1918, opened 2026-07-15
LongTermOwnerWallet.RegisterRecipient was an unconditional no-op, bypassing recipient validation  #1917

### LINKED ISSUE #1917: LongTermOwnerWallet.RegisterRecipient was an unconditional no-op, bypassing recipient validation
## Summary

`LongTermOwnerWallet.RegisterRecipient` (`token/services/identity/role/wallets.go`) had an unconditional no-op implementation that accepted *any* `driver.RecipientData` and always returned `nil`, without checking that the supplied identity/audit-info actually belonged to the wallet.

```go
// Before
func (w *LongTermOwnerWallet) RegisterRecipient(ctx context.Context, data *driver.RecipientData) error {
	// TODO: if identity is equal to the one this wallet is bound to, then we are good. Otherwise return an error
	return nil
}
```

## Impact

This is a validation bypass on a wallet type used for non-anonymous, long-term-identity flows. A caller could register recipient data pointing at an arbitrary/incorrect identity against this wallet with no error, silently accepting mismatched data.

Contrast with the sibling `AnonymousOwnerWallet.RegisterRecipient`, which validates `data != nil`, matches identity against audit info via `Deserializer.MatchIdentity`, registers audit info with the identity provider, and binds the identity to the wallet — `LongTermOwnerWallet` did none of this.

This method is reachable from production code via `token/request.go`'s `Transfer`/`prepareTransfer` path, when a caller supplies `WithRestRecipientIdentity(...)` to redirect leftover transfer balance to caller-supplied `RecipientData`. Since the identity/audit-info were never checked against the wallet's own bound identity, mismatched recipient data would be silently accepted instead of rejected.

## Fix

`RegisterRecipient` on `LongTermOwnerWallet` now validates the passed `RecipientData` against the identity this wallet is bound to:

- Rejects `nil` data (`wallet.ErrNilRecipientData`).
- Rejects data whose `Identity` doesn't match `w.OwnerIdentity` (via the existing `w.Contains` check).
- Rejects data whose `AuditInfo` doesn't match `w.OwnerAuditInfo`.
- Otherwise returns `nil`.

No side effects (registration/binding) are performed, since the wallet's identity is already resolved and bound at construction time in `NewLongTermOwnerWallet` — unlike `AnonymousOwnerWallet`, there's nothing left to register.

Test coverage was added under `TestLongTermOwnerWallet` in `token/services/identity/role/wallets_test.go` covering nil data, matching data, mismatched identity, and mismatched audit info.

adecaro:
Fix #1917

## Description

 - Rejects nil data (wallet.ErrNilRecipientData).
  - Rejects data whose Identity doesn't match w.OwnerIdentity (reuses the existing w.Contains check).
  - Rejects data whose AuditInfo doesn't match w.OwnerAuditInfo.
  - Otherwise returns nil.

No registration/binding side effects are added (unlike AnonymousOwnerWallet.RegisterRecipient), since the wallet's identity is already resolved and bound at construction time in NewLongTermOwnerWallet — this method is purely a guard.

### DIFF
--- a/token/services/identity/role/wallets.go
+++ b/token/services/identity/role/wallets.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package role
 
 import (
+	"bytes"
 	"context"
 	"math/big"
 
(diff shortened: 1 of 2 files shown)


AkramBitar (approved):
LGTM

PR #1923, opened 2026-07-15
fix(ttx): enforce recipient data registration on external withdrawal path

### LINKED ISSUE #1922: Unregistered external RecipientData accepted during withdrawal
## Summary

`RequestWithdrawalView.getRecipientIdentity` (`token/services/ttx/withdrawal.go:204`) accepts caller-supplied `RecipientData` for the external-wallet withdrawal path as-is, with no verification that this recipient data was ever registered/validated against an identity/audit-info pair.

```go
func (r *RequestWithdrawalView) getRecipientIdentity(context view.Context) (*token.TMSID, *RecipientData, *token.OwnerWallet, error) {
	if r.RecipientData != nil {
		tms, err := token.GetManagementService(context, token.WithTMSID(r.TMSID))
		...
		// TODO: check that RecipientData is registered
		r.ExternalWallet = true
		return new(tms.ID()), r.RecipientData, nil, nil
	}
	...
```

## Impact

The wallet-backed path just below (`w.GetRecipientData(...)`) is backed by a wallet that did go through registration. The external path skips this entirely: `r.RecipientData` is used directly to drive the withdrawal handshake (signing the issuer's challenge nonce) without confirming the identity/audit-info pair was ever matched or bound to anything.

The one concrete caller (`integration/token/fungible/views/withdraw.go`) currently works around this by manually calling `w.RegisterRecipient(ctx, i.RecipientData)` before invoking `RequestWithdrawalForRecipient` — but that's caller discipline, not something the library enforces. Any other caller of `RequestWithdrawalForRecipient` that omits this manual call would silently skip validation, letting a withdrawal proceed for an identity/audit-info pair that was never matched.

Note: even with this closed, full protection is limited to callers whose local wallet is `AnonymousOwnerWallet`-backed (see #1917 for the sibling `LongTermOwnerWallet` gap).

adecaro:
## Summary
- `RequestWithdrawalView.getRecipientIdentity` accepted caller-supplied `RecipientData` for the external-wallet withdrawal path as-is, with no check that the identity/audit-info pair was ever registered against the local wallet.
- Resolve the caller's local wallet and call `w.RegisterRecipient(ctx, r.RecipientData)` before using the data to drive the withdrawal handshake, so a mismatched identity/audit-info pair is rejected instead of silently accepted (`AnonymousOwnerWallet.RegisterRecipient` runs `Deserializer.MatchIdentity` before registering/binding).
- Removed the now-redundant manual `RegisterRecipient` call in `integration/token/fungible/views/withdraw.go`, since the library now enforces it.
- Updated `docs/services/ttx.md` to document the new enforcement point and its scope (full protection requires an `AnonymousOwnerWallet`-backed local wallet; `LongTermOwnerWallet` validation was addressed separately in #1917).

Fixes #1922

## Test plan
- [x] `go build ./...` and `(cd integration && go build ./...)` succeed
- [x] New unit tests in `token/services/ttx/withdrawal_test.go` cover: registration succeeds (external flag set, nil wallet returned), registration fails (error bubbles up, external flag not set), and the unchanged local-wallet fallback path (no `RecipientData` supplied)
- [x] `go test ./token/services/ttx/...` passes
- [x] `make checks` passes

### DIFF
--- a/integration/token/fungible/views/withdraw.go
+++ b/integration/token/fungible/views/withdraw.go
@@ -47,16 +47,8 @@ func (i *WithdrawalInitiatorView) Call(context view.Context) (any, error) {
 	var session view.Session
 	var err error
 	if i.RecipientData != nil {
-		// Use the passed RecipientData.
-		// First register it locally
-		var tms *token.ManagementService
-		tms, err = token.GetManagementService(context, token.WithTMSID(i.TMSID))
-		assert.NoError(err, "failed getting management service")
-		var w *token.OwnerWallet
-		w, err = tms.WalletManager().OwnerWallet(context.Context(), i.Wallet)
-		assert.NoError(err, "cannot find wallet [%s:%s]", i.TMSID, i.Wallet)
-		assert.NoError(w.RegisterRecipient(context.Context(), i.RecipientData), "failed to register remote recipient")
-
+		// Use the passed RecipientData. RequestWithdrawalForRecipient validates and
+		// registers it locally against the wallet before using it.
 		// We have an external wallet here.
 		// We need to provide access to it to RequestWithdrawalForRecipientView.
 		// Indeed, a signature is needed to prove to the issuer that the caller owns the recipient data
--- a/token/services/ttx/withdrawal_test.go
+++ b/token/services/ttx/withdrawal_test.go
@@ -1,0 +1,219 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// This white-box (package ttx) file covers getRecipientIdentity, which is
+// unexported. It is kept separate from black-box tests so it can avoid
+// importing dep/mock, which cannot be imported from package ttx without
+// creating an import cycle (see cleanupsessions_test.go).
+package ttx
+
+import (
+	"context"
+	"testing"
+
+	"github.com/LFDT-Panurus/panurus/token"
+	"github.com/LFDT-Panurus/panurus/token/driver"
+	driver_mock "github.com/LFDT-Panurus/panurus/token/driver/mock"
+	token_mock "github.com/LFDT-Panurus/panurus/token/mock"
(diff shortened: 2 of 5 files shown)


KElkhiyaoui (approved):

integration/token/fungible/views/withdraw.go
```diff
@@ -47,16 +47,8 @@ func (i *WithdrawalInitiatorView) Call(context view.Context) (any, error) {
 	var session view.Session
 	var err error
 	if i.RecipientData != nil {
-		// Use the passed RecipientData.
-		// First register it locally
```
AkramBitar, line 51: I am wondering why we do not want to register it locally (for my education only), this since the library now enforces it?

AkramBitar (commented):
LGTM

token/services/ttx/withdrawal_test.go
```diff
@@ -0,0 +1,219 @@
...
+		Identity:  driver.Identity("local-identity"),
+		AuditInfo: []byte("local-audit-info"),
+	}
+
+	mockOwnerWallet := &driver_mock.OwnerWallet{}
+	mockOwnerWallet.GetRecipientDataReturns(localRecipientData, nil)
+
+	mockWalletService := &driver_mock.WalletService{}
+	mockWalletService.OwnerWalletReturns(mockOwnerWallet, nil)
+
```
AkramBitar, line 209: What do you think do we need to check the case where OwnerWalletReturns(nil, err)?

integration/token/fungible/views/withdraw.go
adecaro: so, here we are assuming that the recipient data has been generated by an external wallet for which this node has only a read-only reference. Somehow we need to register this information in the db. Now, we let the view take care of it directly.

AkramBitar (approved):

token/services/ttx/withdrawal_test.go
adecaro: Test coverage increased 👍

PR #1925, opened 2026-07-15
feat(storage): service-wide, stackable guard layer for API limits (#1630)

### LINKED ISSUE #1630: API-level payload size limits and query result caps [MED]
**Description**
Enforce strict upper bounds on every write and read path exposed by the View API before data reaches the database layer. Write operations (`AddTokenRequest`, `AddTransaction`, `AddMovement`) should reject any call whose serialised payload exceeds a configurable maximum size. Query operations (`QueryTransactions`, `QueryMovements`, `QueryTokenRequests`) should require a bounded `Pagination` argument  --  the `Pagination` type is already present in the codebase, so enforcing its use rather than treating it as optional is a low-friction change. This prevents a single oversized write or unbounded table scan from monopolising database resources.

**Short Summary**
- Enforce max serialized payload size on [`AddTokenRequest`](token/services/ttx/), [`AddTransaction`](token/services/ttx/), [`AddMovement`](token/services/ttx/)
- Make `Pagination` parameter mandatory (not optional) in query operations

AkramBitar:
Add size limits to the storage service so one huge write or one unlimited query can't overload the database. Fixes the [MED] issue #1630 (denial of service from unlimited resource use, CWE-400 / CWE-770).

## What changed since the first revision

The first revision implemented the limits **only for the transaction store**, via per-constructor `WithMax*` options + inline checks. Per review feedback (that approach doesn't scale beyond one store), this has been reworked into a single, **stackable guard decorator layer** applied across the whole storage service. The transaction store was migrated onto it, so there is now one mechanism.

## How it works

- New package `token/services/storage/db/guard/`: a `Policy` (loaded once from config), payload-size checks, a row-capping `LimitIterator`, and interface-embedding decorators for the transaction, token, endorser, identity and wallet stores. Each decorator embeds the store interface (delegating everything) and overrides only the methods that need a check, plus the nested write-transactions and returned iterators.
- Applied once at the `multiplexed.Driver` seam: every `NewXxx` result is wrapped, so all backing drivers (SQLite, PostgreSQL) and all stores are covered uniformly. Additional layers (metrics/tracing) can stack the same way, with the concrete SQL store at the bottom.

## What is guarded

**Write payload size** — `maxPayloadSize` (default 4 MiB; `0` disables), rejected before reaching the DB:
- transaction: `AddTokenRequest`, `AddTransaction`, `AddMovement`, `AddTransactionEndorsementAck`
- token: `StoreToken`, `StorePublicParams`, `StoreCertifications`
- endorser: `AddValidationRecord`
- identity: `StoreIdentityData`, `StoreSignerInfo`, `RegisterIdentityDescriptor`, `AddConfiguration`
- wallet: `StoreIdentity`

**Read caps** — `maxPageSize` (default 1000):
- paginated `QueryTransactions` — `nil` / `pagination.None()` / over-max pages are rejected
- streaming iterators — token unspent/spendable/unsupported iterators, `QueryValidations`, `QueryTokenRequests`, `IteratorConfigurations` — capped by a `LimitIterator` that errors (rather than silently truncating) once the cap is exceeded

**Intentionally uncapped:** `QueryMovements` (feeds balance totals; dropping rows would silently return wrong balances).

Also closes two gaps in `pagination.ValidateLimited`: keyset page size is now capped against `maxPageSize`, and a typed-nil offset is guarded.

## Configuration / upgrade note

Limits are **on by default** (4 MiB / 1000), so existing deployments get them automatically. Override via `token.storage.maxPayloadSize` / `token.storage.maxPageSize`; an explicit `0` disables the respective check. Paging on `QueryTransactions` cannot be turned off — `nil`/`None` are rejected — so any caller that read everything at once now pages (see `collectAllTransactions` / `checks.go`).

## Not yet addressed (tracked follow-up)

These vectors can't be closed by a wrapper alone, because the SQL has already materialised the full result before the decorator sees it — they need query-level changes and will be handled in a follow-up:

- **Materialized slice/map reads**: `ListUnspentTokens`, `ListUnspentTokensByWallets`, `ListHistoryIssuedTokens`, `ListAuditTokens`, `QueryTokenDetails`, `ConfigurationsByID`, `GetWalletIDs`, `GetExistingSignerInfo` — need a SQL-level `LIMIT` (with a defined order + documented truncation) or conversion to iterators.
- **Variadic input-size caps**: `DeleteTokens`, `GetTokens`, `GetTokenRequests` — a huge id list expands into a huge `IN (...)` clause.
- **`Keystore.Put`**: opaque value serialised inside the store, so its size is only known at marshal time.

Closes #1630.

### DIFF
--- a/integration/token/fungible/views/history.go
+++ b/integration/token/fungible/views/history.go
@@ -17,10 +17,49 @@ import (
 	"github.com/LFDT-Panurus/panurus/token/services/ttx"
 	token2 "github.com/LFDT-Panurus/panurus/token/token"
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
+	driver2 "github.com/hyperledger-labs/fabric-smart-client/platform/common/driver"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/assert"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils/collections/iterators"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/view"
 )
+
+// historyPageSize is the number of rows fetched per page when listing all
+// transactions. It must stay <= the store's max page size.
+const historyPageSize = 100
+
+// collectAllTransactions pages through queryPage (one page per call) and returns
+// every record. queryPage runs the underlying query for the given pagination.
+//
+// The storage layer now rejects unlimited queries, so these trusted views can no
+// longer fetch everything in one call. This helper is only that adaptation: it
+// still accumulates the full result set in memory, so it is not itself a memory
+// safeguard. The actual DoS protection (rejecting unlimited scans, hard LIMITs)
+// lives in the storage layer; these views legitimately need the complete list.
+func collectAllTransactions(queryPage func(driver2.Pagination) (iterators.Iterator[*ttxdb.TransactionRecord], error)) ([]*ttxdb.TransactionRecord, error) {
+	var page driver2.Pagination
+	page, err := pagination.Offset(0, historyPageSize)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create pagination")
+	}
+	var all []*ttxdb.TransactionRecord
+	for {
+		items, err := queryPage(page)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed querying transactions")
+		}
+		records, err := iterators.ReadAllPointers(items)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed reading transactions")
+		}
+		all = append(all, records...)
+		if len(records) < historyPageSize {
+			return all, nil
+		}
+		if page, err = page.Next(); err != nil {
+			return nil, errors.Wrapf(err, "failed advancing pagination")
+		}
+	}
+}
 
 // ListIssuedTokens contains the input to query the list of issued tokens
 type ListIssuedTokens struct {
@@ -143,12 +182,14 @@ func (p *ListAuditedTransactionsView) Call(context view.Context) (any, error) {
 		return nil, errors.Wrapf(err, "failed to get auditor instance")
 	}
 
-	it, err := auditor.Transactions(context.Context(), ttxdb.QueryTransactionsParams{From: p.From, To: p.To, SearchDirection: p.SearchDirection}, pagination.None())
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed querying transactions")
-	}
+	return collectAllTransactions(func(page driver2.Pagination) (iterators.Iterator[*ttxdb.TransactionRecord], error) {
+		it, err := auditor.Transactions(context.Context(), ttxdb.QueryTransactionsParams{From: p.From, To: p.To, SearchDirection: p.SearchDirection}, page)
+		if err != nil {
+			return nil, err
+		}
 
-	return iterators.ReadAllPointers(it.Items)
+		return it.Items, nil
+	})
 }
 
 type ListAuditedTransactionsViewFactory struct{}
@@ -184,21 +225,24 @@ func (p *ListAcceptedTransactionsView) Call(context view.Context) (any, error) {
 	tms, err := token.GetManagementService(context, ServiceOpts(p.TMSID)...)
 	assert.NoError(err, "failed getting management service")
 	owner := ttx.NewOwner(context, tms)
-	it, err := owner.Transactions(context.Context(), ttxdb.QueryTransactionsParams{
-		SenderWallet:    p.SenderWallet,
-		RecipientWallet: p.RecipientWallet,
-		From:            p.From,
-		To:              p.To,
-		ActionTypes:     p.ActionTypes,
-		Statuses:        p.Statuses,
-		IDs:             p.IDs,
-		SearchDirection: p.SearchDirection,
-	}, pagination.None())
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed querying transactions")
-	}
 
-	return iterators.ReadAllPointers(it.Items)
+	return collectAllTransactions(func(page driver2.Pagination) (iterators.Iterator[*ttxdb.TransactionRecord], error) {
+		it, err := owner.Transactions(context.Context(), ttxdb.QueryTransactionsParams{
+			SenderWallet:    p.SenderWallet,
+			RecipientWallet: p.RecipientWallet,
+			From:            p.From,
+			To:              p.To,
+			ActionTypes:     p.ActionTypes,
+			Statuses:        p.Statuses,
+			IDs:             p.IDs,
+			SearchDirection: p.SearchDirection,
+		}, page)
+		if err != nil {
+			return nil, err
+		}
+
+		return it.Items, nil
+	})
 }
 
 type ListAcceptedTransactionsViewFactory struct{}
--- a/token/sdk/dig/providers.go
+++ b/token/sdk/dig/providers.go
@@ -10,6 +10,7 @@ import (
 	"github.com/LFDT-Panurus/panurus/token/core"
 	"github.com/LFDT-Panurus/panurus/token/driver"
 	dbdriver "github.com/LFDT-Panurus/panurus/token/services/storage/db/driver"
+	"github.com/LFDT-Panurus/panurus/token/services/storage/db/guard"
 	"github.com/LFDT-Panurus/panurus/token/services/storage/db/multiplexed"
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	driver2 "github.com/hyperledger-labs/fabric-smart-client/platform/common/driver"
@@ -25,6 +26,23 @@ func newMultiplexedDriver(in struct {
 },
 ) multiplexed.Driver {
 	return multiplexed.NewDriver(in.ConfigProvider, in.Drivers...)
+}
+
+// newStoragePolicy loads the storage guard policy from configuration. It is the
+// single source of the storage resource limits (max payload / max page size) so
+// consumers such as the integrity checkers page within the same cap the guard
+// layer enforces. Falls back to the built-in defaults if the config is unreadable.
+func newStoragePolicy(in struct {
+	dig.In
+	ConfigProvider driver2.ConfigService
+},
+) guard.Policy {
+	policy, err := guard.LoadPolicy(in.ConfigProvider)
+	if err != nil {
+		return guard.DefaultPolicy()
+	}
+
+	return policy
 }
 
 // newTokenDriverService creates a token driver service from registered token drivers.
--- a/token/services/storage/db/common/checks.go
+++ b/token/services/storage/db/common/checks.go
@@ -15,6 +15,7 @@ import (
 	"github.com/LFDT-Panurus/panurus/token/services/logging"
 	"github.com/LFDT-Panurus/panurus/token/services/network"
 	"github.com/LFDT-Panurus/panurus/token/services/storage/db/driver"
+	"github.com/LFDT-Panurus/panurus/token/services/storage/db/sql/query/pagination"
 	"github.com/LFDT-Panurus/panurus/token/services/tokens"
 	"github.com/LFDT-Panurus/panurus/token/services/utils"
 	token2 "github.com/LFDT-Panurus/panurus/token/token"
@@ -26,6 +27,13 @@ import (
 var (
 	logger = logging.MustGetLogger()
 )
+
+// transactionsCheckPageSize is the fallback page size used to iterate all
+// transactions during integrity checks when the storage layer's max page size
+// is disabled (0). When a cap is configured, that cap is used as the page size
+// instead (see DefaultCheckers.maxPageSize), so the checks always stay within
+// the bound the guard layer enforces.
+const transactionsCheckPageSize = 100
 
 type TokenTransactionDB interface {
 	GetTokenRequest(ctx context.Context, txID string) ([]byte, error)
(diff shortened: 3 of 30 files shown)


adecaro:
Hi @AkramBitar , thanks for submitting this.

My concern is the following: The storage service has many queries that would require similar constraints. Right now, this opts-based approach works only for the Transactions db.

I would suggest to perform an analysis of the entire storage service to understand which limits need to be put in place and then design a flexible solution to address the issues. A wrapper-based solution could also be interested so that we can stack different layers on top of each other each perform a different function and the bottom one running the actual SQL query.

What do you think?

Thanks 🙏

AkramBitar:
> Hi @AkramBitar , thanks for submitting this.
> 
> My concern is the following: The storage service has many queries that would require similar constraints. Right now, this opts-based approach works only for the Transactions db.
> 
> I would suggest to perform an analysis of the entire storage service to understand which limits need to be put in place and then design a flexible solution to address the issues. A wrapper-based solution could also be interested so that we can stack different layers on top of each other each perform a different function and the bottom one running the actual SQL query.
> 
> What do you think?
> 
> Thanks 🙏

Hi @adecaro,

Thanks a lot for the review. Good idea. I implemented it. 

Thanks,
Akram

token/services/storage/db/common/checks.go
```diff
@@ -27,6 +28,10 @@ var (
 	logger = logging.MustGetLogger()
 )
 
+// transactionsCheckPageSize is the page size used to iterate all transactions
+// during integrity checks; kept below the storage layer's max page size.
+const transactionsCheckPageSize = 100
```
adecaro, line 36: shouldn't this be configurable as well?

token/services/storage/db/guard/token.go
```diff
@@ -0,0 +1,135 @@
...
+func (g *guardedTokenStore) UnspentTokensIteratorBy(ctx context.Context, walletID string, tokenType token.Type) (tokendriver.UnspentTokensIterator, error) {
+	it, err := g.TokenStore.UnspentTokensIteratorBy(ctx, walletID, tokenType)
+	if err != nil {
+		return nil, err
+	}
+
+	return LimitIterator(it, g.policy.MaxPageSize, "UnspentTokensIteratorBy"), nil
+}
+
+func (g *guardedTokenStore) SpendableTokensIteratorBy(ctx context.Context, walletID string, typ token.Type) (tokendriver.SpendableTokensIterator, error) {
```
adecaro: I think this function is used by the token selectors and I think these token selectors work on the assumption that they can load all the tokens that belong to a wallet. If this is not the case anymore, we selector might report unexpected errors. No?

token/services/storage/db/sql/query/pagination/validate.go
```diff
@@ -0,0 +1,69 @@
...
+	// so the checks below don't panic dereferencing it.
+	if rv := reflect.ValueOf(p); rv.Kind() == reflect.Pointer && rv.IsNil() {
+		return errors.New("pagination is required: a page-size limit must be provided")
+	}
+
+	switch p.(type) {
+	case *none:
+		return errors.New("unlimited pagination (None) is not allowed")
+	case *empty:
+		return nil
```
adecaro, line 47: if maxPageSize > 0, empty should not be allowed, no?

token/services/storage/db/guard/transactions.go
```diff
@@ -0,0 +1,160 @@
...
+
+func (g *guardedAuditTx) QueryTransactions(ctx context.Context, params driver.QueryTransactionsParams, p driver2.Pagination) (*driver2.PageIterator[*driver.TransactionRecord], error) {
+	if err := pagination.ValidateLimited(p, g.policy.MaxPageSize); err != nil {
+		return nil, err
+	}
+
+	return g.AuditTransactionStore.QueryTransactions(ctx, params, p)
+}
+
+func (g *guardedAuditTx) QueryTokenRequests(ctx context.Context, params driver.QueryTokenRequestsParams) (driver.TokenRequestIterator, error) {
```
adecaro: so, in QueryTransactions the caller can extract all the transactions even though in a paginated format. Here instead can only extract up to a certain total number of token requests. I think we need to introduce a standard here. We can say that all queries need to be paginated and we can even say that despite the pagination only a maximum number of rows can be extracted. What do you think?

token/services/storage/db/common/checks.go
AkramBitar: Good point. Rather than adding a separate config key, I've made the checker page at the storage layer's existing token.storage.maxPageSize ca. So it reads within the same bound the guard layer already enforces, with no new knob to keep in sync. The 100 constant is now just the fallback for when the cap is disabled (maxPageSize = 0), since an unbounded read gets rejected and we still need some batch size to page by.

token/services/storage/db/guard/token.go
AkramBitar: Good catch, you're right. LimitIterator errors instead of truncating, so a wallet with more than MaxPageSize tokens would fail selection even though the funds exist. UnspentTokensIteratorBy / SpendableTokensIteratorBy are a legitimate "load all tokens of this wallet" operation the selectors rely on, so I'll drop the cap on them and keep it only on the unbounded, non-wallet-scoped reads. Write payload limits are unaffected.

token/services/storage/db/sql/query/pagination/validate.go
AkramBitar: Empty compiles to LIMIT 0, it returns zero rows, so it can never exceed maxPageSize. The unbounded case is None (no LIMIT), which we already reject. Allowing empty is safe; rejecting it would only break harmless zero-row queries.

token/services/storage/db/guard/transactions.go
AkramBitar: Good point. I think a consistent standard makes sense. Looking through the query layer, there are really two patterns today:

1) QueryTransactions is paginated, so it enforces a per-page limit.
2) QueryTokenRequests, QueryValidations, and IteratorConfigurations are streaming queries, so they enforce a hard total limit with LimitIterator.

So QueryTokenRequests isn't doing anything unique here, it's following the same pattern as the other streaming queries.

I agree that making all query operations paginated would be a cleaner design. But that would require changes to both the driver interface and the public store API, and for consistency we'd want to do it for all three streaming queries, not just this one. That feels like a separate breaking change, so I'd rather keep this PR focused on the limits work and open a follow-up issue/PR to standardize pagination across all query operations.

What do you think?

AkramBitar:
@adecaro thanks a lot for the comments.
I have answered all your comments. Please let me know if you have additional comments.

token/services/storage/db/guard/wallet.go
```diff
@@ -0,0 +1,38 @@
...
+
+	return &guardedWalletStore{WalletStore: s, policy: p}
+}
+
+type guardedWalletStore struct {
+	driver.WalletStore
+	policy Policy
+}
+
+func (g *guardedWalletStore) StoreIdentity(ctx context.Context, identity token2.Identity, eID string, wID idriver.WalletID, roleID int, meta []byte, confID string) error {
```
adecaro, line 31: **Should fix: `guardedWalletStore.StoreIdentity`'s size check has 0% test coverage.**

Same is true for `guardedIdentityStore`'s four write methods (`identity.go:38-64`), `guardedEndorserStoreTx.AddValidationRecord` (`endorser.go:49`), `guardedAuditTx.QueryTransactions`/`NewTransactionStoreTransaction` (`transactions.go:80-95`), and `guardedTokenStore`'s three write methods (`token.go:37-80`) — only the owner-transaction-store paths (`AddTransaction`/`AddMovement`/`AddTokenRequest`/`QueryTransactions`) are exercised by `decorator_test.go`.

Each repeats the field-summing pattern the tested paths use (sum a set of lengths, call `CheckPayload`) — exactly the arithmetic where a future struct-field change ships silently uncaught, since nothing runs this code path in CI.

```
// add, mirroring TestGuardRejectsOversizeWrite:
func TestGuardRejectsOversizeStoreIdentity(t *testing.T) { ... }
```

adecaro (commented):
## Review: does this fully address #1630?

@AkramBitar

**This is solid — the decorator design is right, the `pageSized` fix closes a real unbounded-scan bug, and the second commit is a genuine self-correction rather than scope creep. One non-blocking gap: five of the seven guard decorators have zero direct test coverage.**

Checked out `feature/1630-api-payload-and-query-bounds` at `edb52d95c` and verified:

```
go build ./...                                                     # clean
go vet ./token/services/storage/db/{guard,sql/query/pagination,common,multiplexed}/... ./token/sdk/{dig,db}/...   # clean
go test -count=1 (same packages)                                   # ok, all packages
go test -race -count=1 (guard, pagination, common)                 # ok
golangci-lint run (same packages)                                  # 0 issues
gofmt -l -s (all changed .go files)                                # clean
go tool cover -func (guard, pagination)                            # guard 34.8%, pagination 61.7%
```

I also grepped for every remaining caller of `pagination.None()`/`nil` across both modules — none
remain against a guarded store; `integration/token/fungible/views/history.go`'s
`collectAllTransactions` now pages internally instead.

To confirm the `keyset[I, V]` regression test actually pins the fix rather than just decorating the
diff, I reverted `ValidateLimited`'s type switch to the pre-fix shape locally (explicit `*offset`
case, everything else falling to `default: return nil`) and reran `TestValidateLimited`:
`keyset[any]_over_max_is_rejected`, `keyset[concrete]_over_max_is_rejected`, and
`unsupported_pagination_type_is_rejected` all fail with "expected an error, got nil" — so the test
genuinely catches the bug it's named for. Restored and reran clean afterward.

CI is fully green: `checks`, both `utest` legs (race + regression), the full `itest` matrix,
`cgo-check`, `bench-check`, and DCO on both signed-off commits.

### What it fixes

| # | Finding | Verified |
|---|---|---|
| 1630.1 | Writes across wallet/identity/endorser/token/transaction stores need a payload-size cap | ✅ `guard.CheckPayload` wired into every write path via the `multiplexed` seam |
| 1630.2 | Unbounded page reads (`pagination.None()`/no cap) need rejecting | ✅ `ValidateLimited`, closes the additional `keyset[I, concreteType]` gap the old type-switch missed |
| 1630.3 | Reads that can't be paged shouldn't be capped with an error the caller can't act on | ✅ second commit removes the cap from `QueryValidations`/`QueryTokenRequests`/`IteratorConfigurations`, with `IteratorConfigurations` specifically traced as a real functional regression on the identity-load path before the fix |

All three rows check out under the verification above; no gaps in the core mechanism itself.

### Non-blocking

- **Test coverage gap:** `guardedWalletStore.StoreIdentity`, `guardedIdentityStore`'s four write
  methods, `guardedEndorserStoreTx.AddValidationRecord`, `guardedAuditTx.QueryTransactions`, and
  `guardedTokenStore`'s three write methods are all at 0% coverage — see the inline comment on
  `wallet.go`.
- **Already-discussed:** the `QueryTokenRequests`/`QueryMovements` pagination-shape question from
  the earlier thread is a deferred follow-up per your reply there — not re-raising it, just noting
  the in-code comment at `transactions.go:51-56` matches what was agreed.
- **Pre-existing, out of scope:** `offset.go:24,27`'s `fmt.Errorf` typos ("shoud be grater") predate
  this PR and aren't touched by it.

### Recommendation

Add a `sqlmock` case per untested decorator (same shape as the existing
`TestGuardRejectsOversizeWrite`/`TestGuardRejectsOversizeAddTransaction`) so the size-check arithmetic
in wallet/identity/endorser/audit-tx/token-write guards gets the same regression protection the
owner-transaction path already has. Everything else here is good to merge as-is.

token/services/storage/db/guard/wallet.go
AkramBitar: Fixed in `6b776f7cc` — added `decorator_writes_test.go` covering all of them: `StoreIdentity`, the four `guardedIdentityStore` writes, `AddValidationRecord`, `guardedAuditTx.QueryTransactions`/`NewTransactionStoreTransaction`, and the three `guardedTokenStore` writes (through both `NewTokenDBTransaction` and `ContinueTokenDBTransaction`). Also covered three gaps not in your list: `AddTransactionEndorsementAck`, the multi-record `AddTransaction`/`AddMovement` sums, and `amountLen`'s nil-amount branch. Guard package statement coverage: 34.8% -> 100%.

One deviation from your suggestion: I used fakes that embed the driver store interface rather than `sqlmock`. The guards reject before any SQL runs, so `sqlmock` can only show that no query happened, whereas a fake also lets the test assert whether the call *was* delegated on the allowed path.

Each write is tested as a boundary pair — rejected at `MaxPayloadSize = size-1`, delegated at exactly `size` — with the expected size stated independently of the implementation, so the arithmetic is pinned in both directions: adding a field to a sum breaks the at-the-limit case, dropping one breaks the over-the-limit case. Verified by mutation (dropped `len(tr.Type)` from `StoreToken`, double-counted `wp.ID` in `AddConfiguration`, deleted the `StoreIdentity` guard entirely) — each one fails the suite.

AkramBitar:
@adecaro 

Thanks for the review. 
I addressed all your comments please let me know if you are ok with this PR or do you still need more changes. 

Regards,
Akram

integration/token/fungible/views/history.go
```diff
@@ -17,11 +17,50 @@ import (
...
+// every record. queryPage runs the underlying query for the given pagination.
+//
+// The storage layer now rejects unlimited queries, so these trusted views can no
+// longer fetch everything in one call. This helper is only that adaptation: it
+// still accumulates the full result set in memory, so it is not itself a memory
+// safeguard. The actual DoS protection (rejecting unlimited scans, hard LIMITs)
+// lives in the storage layer; these views legitimately need the complete list.
+func collectAllTransactions(queryPage func(driver2.Pagination) (iterators.Iterator[*ttxdb.TransactionRecord], error)) ([]*ttxdb.TransactionRecord, error) {
+	var page driver2.Pagination
+	page, err := pagination.Offset(0, historyPageSize)
```
Effi-S, line 40: `historyPageSize` is hardcoded as 100. Should be `maxPageSize`, no? 
If not, then a comment should explain why it's hardcoded.

token/services/storage/db/common/checks.go
```diff
@@ -115,52 +132,83 @@ func (a *DefaultCheckers) CheckTransactions(ctx context.Context) ([]string, erro
...
-		}
-
-		// check the ledger
-		lVC, _, err := l.Status(transactionRecord.TxID)
-		if err != nil {
-			lVC = network.Unknown
-		}
-		switch {
-		case transactionRecord.Status == driver.Confirmed && lVC != network.Valid:
+		count, err := func() (int, error) {
```
Effi-S, line 150: integrity check runs on a live node; a concurrent delete/insert shifts the offset window; a genuine vault/ledger inconsistency at a page boundary is silently missed

Offset pagination is not stable under concurrent mutation: if a row is deleted (recovery/cleanup) between page N and N+1, every later row shifts up one offset slot, so the row that would have landed at the page boundary is skipped and never validated. A concurrent insert conversely causes a boundary row to be checked twice.

**Suggested fix**: use keyset/cursor pagination (stable ordering on a monotonic key) instead of offset for this full-table scan, so page boundaries don't move under concurrent writes.

token/sdk/dig/providers.go
```diff
@@ -27,6 +28,23 @@ func newMultiplexedDriver(in struct {
...
+// consumers such as the integrity checkers page within the same cap the guard
+// layer enforces. Falls back to the built-in defaults if the config is unreadable.
+func newStoragePolicy(in struct {
+	dig.In
+	ConfigProvider driver2.ConfigService
+},
+) guard.Policy {
+	policy, err := guard.LoadPolicy(in.ConfigProvider)
+	if err != nil {
+		return guard.DefaultPolicy()
```
Effi-S, line 42: Please log the error here  (or other solution just don't silently discard it)

token/services/storage/db/multiplexed/driver.go
```diff
@@ -7,30 +7,45 @@ SPDX-License-Identifier: Apache-2.0
...
 func NewDriver(config driver4.Config, ds ...driver4.NamedDriver) Driver {
 	drivers := make(map[driver3.PersistenceType]driver4.Driver, len(ds))
 	for _, d := range ds {
 		drivers[d.Name] = d.Driver
 	}
 
+	// The storage guard policy is cross-cutting and applies to every backing
+	// driver, so it is loaded once here and applied as each store is created.
+	policy, err := guard.LoadPolicy(config)
+	if err != nil {
```
Effi-S, line 33: Here and `token/services/storage/db/multiplexed/providers.go` both load the policy. 

Suggested fix: provide the guard.Policy once via DI and inject the same value into both the multiplexed driver and the checkers, so there is genuinely one load.

Effi-S (dismissed):

adecaro (approved):
LGTM

PR #1929, opened 2026-07-16
test(tokens): add SpendableTokensIteratorBy prepared-statement comparison benchmark

### LINKED ISSUE #1919: Introduce SQL Prepared Statements for SpendableTokensIteratorBy
Follow-up to #1183/#1889. This issue targets `SpendableTokensIteratorBy(ctx, walletID, typ)` in `token/services/storage/db/sql/common/tokens.go`, following the same prepared-statement approach.

## Why this query
Same shape-based pattern as `UnspentTokensIteratorBy`: the SQL depends only on which of `walletID`/`typ` are present, not their values. It's simpler than `UnspentTokensIteratorBy` (single `Select().Where(HasTokenDetails(...))`, no UNION).

It's called from `token/services/selector/sherdlock/fetcher.go`, on the selector's spend-fetch path — every spend attempt goes through this, likely making it hotter than `UnspentTokensIteratorBy`.

One thing to verify before implementing: `HasTokenDetails` params also include `LedgerTokenFormats: db.getSupportedTokenFormats()`. Need to confirm this is set once at startup (via `SetSupportedTokenFormats`) and not mutated at runtime, since the caching approach assumes a stable shape per key.

## Plan
Same approach as #1889:
1. Benchmark current dynamic-query path vs prepared, to get concrete numbers first (per Akram's precedent on #1183)
2. Reuse the existing `PreparedStmtHolder[K]` from #1889 rather than duplicating the caching logic
3. Fall back to the dynamic path on any prepare/execute error

## Related
- #1183 (original design discussion)
- #1889 (UnspentTokensIteratorBy implementation, holder + benchmark this will reuse)

sid200727:
## Summary
Adds a benchmark comparing `SpendableTokensIteratorBy` executed via the current dynamic query-building path against the same query executed via a statement prepared once and reused, to get concrete numbers for the discussion in #1919.

## Changes
-> Extracted the query-building logic from `SpendableTokensIteratorBy` into a standalone `buildSpendableTokensIteratorByQuery` function (pure refactor, no behaviour change, verified via existing `TestTokens` suite)
-> Added `RunSpendableTokensIteratorByPreparedComparison` benchmark comparing:
   - `SpendableTokensIteratorBy_Dynamic`: current production path (query built + executed fresh each call)
   - `SpendableTokensIteratorBy_Prepared`: same query, prepared once via `readDB.PrepareContext`, reused via `stmt.QueryContext`
-> Wired into the existing Postgres benchmark suite as `BenchmarkSpendableTokensIteratorByPreparedComparison`

## Results (median of 2 runs, 1000 tokens, wallet0/GOLD, 4 workers, 5s window)

|---|---|---|---|
| Throughput | ~5,105 ops/s | ~6,231 ops/s | +22% |
| P50 latency | ~627µs | ~577µs | -8% |
| P99 latency | ~5.42ms | ~1.67ms | -69% |
| Allocs/op | ~84.5 | 24 | -72% |

## Testing
Verified against existing `TestTokens` suite (unchanged). Benchmark run on laptop with Docker Desktop, not an isolated bench, tool flagged high variance on both runs, and one `Prepared` run had a single 37ms outlier (system noise, not a regression given otherwise consistent allocs/op and throughput across both runs).

## Notes for the Reviewer
Related to #1919. This PR is the measurement step only, no production behavior changes. This query only appears in 2 production shapes today (`walletID`+`typ` present, and `("", "")` for the full-vault refresh scan in `sherdlock/fetcher.go`) a third shape (`walletID` empty, `typ` present) is exercised only in `dbtest/tokens.go` tests, not production. Happy to proceed with the implementation PR (reusing the existing `PreparedStmtHolder[K]` from #1889) if these numbers make the case.

### DIFF
--- a/token/services/storage/db/sql/common/tokens.go
+++ b/token/services/storage/db/sql/common/tokens.go
@@ -56,6 +56,10 @@ type TokenStore struct {
 	// keyed by argument shape (walletID present/absent, tokenType
 	// present/absent). See #1183.
 	unspentTokensStmts PreparedStmtHolder[string]
+
+	// spendableTokensStmts caches prepared statements for
+	// SpendableTokensIteratorBy, keyed the same way. See #1919.
+	spendableTokensStmts PreparedStmtHolder[string]
 }
 
 func newTokenStore(readDB, writeDB *sql.DB, tables tokenTables, ci common3.CondInterpreter, notifier driver.TokenNotifier, cleanupLeaderFactory func(context.Context, *sql.DB, int64) (driver.CleanupLeadership, bool, error)) *TokenStore {
@@ -68,6 +72,7 @@ func newTokenStore(readDB, writeDB *sql.DB, tables tokenTables, ci common3.CondI
 		cleanupLeaderFactory: cleanupLeaderFactory,
 	}
 	ts.unspentTokensStmts = newPreparedStmtHolder[string]()
+	ts.spendableTokensStmts = newPreparedStmtHolder[string]()
 
 	return ts
 }
@@ -383,10 +388,12 @@ func buildSpendableTokensIteratorByQuery(db *TokenStore, walletID string, typ to
 }
 
 func (db *TokenStore) SpendableTokensIteratorBy(ctx context.Context, walletID string, typ token.Type) (tdriver.SpendableTokensIterator, error) {
-	query, args := buildSpendableTokensIteratorByQuery(db, walletID, typ)
+	key := unspentTokensStmtKey(walletID, typ)
+	rows, err := db.spendableTokensStmts.Execute(ctx, db.readDB, key, func() (string, []any, error) {
+		query, args := buildSpendableTokensIteratorByQuery(db, walletID, typ)
 
-	logging.Debug(logger, query, args)
-	rows, err := db.readDB.QueryContext(ctx, query, args...)
+		return query, args, nil
+	})
 	if err != nil {
 		return nil, errors.Wrapf(err, "error querying db")
 	}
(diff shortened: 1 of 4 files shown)


AkramBitar:
@sid200727 could you please open an issue for this PR and attach it to https://github.com/LFDT-Panurus/panurus/issues/1919.

AkramBitar (approved):
LGTM

adecaro:
Hi @sid200727 , proceed directly in this PR with the addition of the prepared statement. Thanks a lot for this effort 🙏

sid200727:
Hi @adecaro @AkramBitar !
I have added the implementation directly here as requested: SpendableTokensIteratorBy now uses the existing PreparedStmtHolder[K] from #1889, same shape-based caching, falls back to dynamic on error. One correction worth flagging: unlike UnspentTokensIteratorBy, this query's HasTokenDetails does exercise InTuple (for LedgerTokenFormats) caught via a real test panic, fixed by implementing a correct stub CondInterpreter rather than assuming it wasn't used. Full test suite green on sqlite + postgres. 
Ready for another look!

AkramBitar:
> Hi @adecaro @AkramBitar ! I have added the implementation directly here as requested: SpendableTokensIteratorBy now uses the existing PreparedStmtHolder[K] from #1889, same shape-based caching, falls back to dynamic on error. One correction worth flagging: unlike UnspentTokensIteratorBy, this query's HasTokenDetails does exercise InTuple (for LedgerTokenFormats) caught via a real test panic, fixed by implementing a correct stub CondInterpreter rather than assuming it wasn't used. Full test suite green on sqlite + postgres. Ready for another look!

Hi @sid200727 thanks a lot. 
Could you please change the title and the description of the PR accordengly?

AkramBitar (approved):
LGTM

PR #1932, opened 2026-07-16
validator: enforce signed action order and signature envelopes (#1927)

### LINKED ISSUE #1927: validator: signed action order and signature envelopes are not enforced
## Summary

Token-request validation does not preserve the signed mixed-action sequence when returning actions, and the raw validation path discards the ActionID association carried by action signatures. Submitted requests can also contain surplus action signatures or no actions.

## Impact

For mixed issue/transfer requests, validation can return actions in a different order from the signed request, which can change downstream output indexing and execution semantics. Ignoring ActionID and unconsumed signatures makes the signature envelope malleable and prevents validation from detecting signatures assigned to the wrong action. Empty submissions can pass as no-op token requests.

The affected logic is shared by the FabToken and ZKAT-DLOG validators.

adecaro:
Fixes #1927

## Summary

- preserve the signed mixed-action sequence through validation, unmarshalling, and translator output indexing
- bind signatures to their declared action IDs and reject missing, misplaced, out-of-range, and unused envelopes
- reject empty submitted requests and return errors instead of panics for invalid public validator inputs
- document the stateless validator/state-transition boundary and add focused, fuzz, and regression coverage

## Validation

- env -u GOROOT go test ./token/core/common ./token/core/fabtoken/v1/validator ./token/core/zkatdlog/nogh/v1/validator ./token/services/network/common/rws/translator ./token


KElkhiyaoui (approved):

token/validator.go
```diff
@@ -66,5 +86,22 @@ func NewLedgerFromGetter(f driver.GetStateFnc) *stateGetter {
 }
 
 func (g *stateGetter) GetState(id token.ID) ([]byte, error) {
+	if g == nil || g.f == nil {
+		return nil, errors.New("ledger not available")
+	}
+
 	return g.f(id)
 }
+
```
AkramBitar: Is that same method as token/core/common/validator.go?


// IsNil returns true for nil values and interfaces containing typed nil values.
func IsNil(value any) bool {
	if value == nil {
		return true
	}
	reflected := reflect.ValueOf(value)
	switch reflected.Kind() {
	case reflect.Chan, reflect.Func, reflect.Interface, reflect.Map, reflect.Pointer, reflect.Slice:
		return reflected.IsNil()
	default:
		return false
	}
}

AkramBitar (commented):
LGTM
Left comments

token/validator.go
adecaro: I noticed as well a duplication of that function. I'll get rid of it.

AkramBitar (approved):
LGTM
Left comments

PR #1935, opened 2026-07-17
perf(postgres): install notification triggers lazily on first Subscribe

### LINKED ISSUE #1934: Postgres notification triggers serialize every commit even with no subscribers
The postgres token/transaction/identity stores install their notification triggers (per-row `pg_notify`) at store creation. PostgreSQL serializes the commit of every notifying transaction on a global notification-queue lock, so once the triggers are installed, every write to those tables pays that lock — whether or not anything ever subscribes.

The token and transaction notifiers currently have no subscriber in the code base (the only production subscriber is the identity-configuration sync in `token/services/identity/membership/lm.go`), so on write-heavy tables the triggers broadcast to nobody while serializing all commits. In our test environments under sustained load, the database showed hundreds of connections blocked on `COMMIT` waiting for the notification-queue lock (`wait_event_type=Lock, wait_event=object`); dropping the triggers took the lock waits from peaks of ~800 to zero.

**Proposal**: install the notification schema lazily on the first `Subscribe` instead of at store creation. Tables nobody subscribes to then get no trigger and pay no NOTIFY cost, while actual subscribers are unaffected. Deployments using `SkipCreateTable` keep full control: no runtime DDL, the notification schema is expected to be pre-created like the rest of the schema.

EvanYan1024:
## Motivation

The postgres stores install their notification triggers (per-row `pg_notify`) at store creation. PostgreSQL serializes the commit of every notifying transaction on a global notification-queue lock, so once the triggers exist, every write to those tables pays that lock — whether or not anything ever subscribes. The token and transaction notifiers currently have no subscriber in the code base (the only production subscriber is the identity-configuration sync in `token/services/identity/membership/lm.go`), so on write-heavy tables the triggers broadcast to nobody while serializing all commits.

In our test environments under sustained load, the database showed hundreds of connections blocked on `COMMIT` waiting for the notification-queue lock (`wait_event_type=Lock, wait_event=object`); dropping the triggers took the lock-wait peaks from ~800 to zero.

## Changes

- `Notifier` installs the notification schema on the first `Subscribe` (inside the existing `startOnce`) instead of at store creation. A failed installation is final: it is recorded and returned by every subsequent `Subscribe`; closing the notifier while the schema is installing does not start the listener.
- The token/transaction/identity store providers no longer create the notification schema at startup. With `SkipCreateTable` the notifier runs no DDL at all (`skipSchemaManagement`) and the deployment pre-creates the notification schema like the rest of the schema.
- The identity store provider no longer builds a second, throwaway `IdentityNotifier` just for schema creation; the store keeps a reference to the one it already has.

Existing subscribers are unaffected: their first `Subscribe` installs the same schema that store creation used to.

## Tests

- Unit: the schema is installed exactly once across concurrent subscribers; an installation failure surfaces to every subscriber and is not retried; close-during-install starts no listener; skip mode runs no DDL and still listens.
- Against a real postgres: no trigger exists before the first subscription; subscribing installs it and notifications are delivered.
- Full `go test -race ./token/...` passes.

Fixes #1934

### DIFF
--- a/token/services/storage/db/sql/postgres/driver.go
+++ b/token/services/storage/db/sql/postgres/driver.go
@@ -106,6 +106,9 @@ func newTokenStoreProvider(dbProvider fscPostgres.DbProvider, tableNamesConfig c
 		if err != nil {
 			return nil, err
 		}
+		if o.SkipCreateTable {
+			notifier.skipSchemaManagement()
+		}
 
 		// db
 		p, err := NewTokenStoreWithNotifier(dbs, tableNames, notifier)
@@ -114,9 +117,6 @@ func newTokenStoreProvider(dbProvider fscPostgres.DbProvider, tableNamesConfig c
 		}
 		if !o.SkipCreateTable {
 			if err := p.CreateSchema(); err != nil {
-				return nil, err
-			}
-			if err := notifier.CreateSchema(); err != nil {
 				return nil, err
 			}
 		}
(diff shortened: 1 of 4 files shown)


adecaro:
lovely, thanks @EvanYan1024, great insight 🙏

adecaro (approved):
LGTM

PR #1936, opened 2026-07-17
feat(evm): week 3, StateDelta translator and secp256k1 signer

atharrva01:
Week 3: the Go side that produces and signs the deltas the week 2 contracts consume.

- eip712/signer.go: secp256k1 signer over decred (no go-ethereum). 65-byte {r,s,v}, low-s enforced on both sign and recover, address derivation pinned against the well-known test keys.
- statedelta/translator.go: validated token actions to StateDelta, same surface the fabric responder drives. Counter rules match the fabric translator, metadata is sorted so every endorser emits identical bytes, and spend refs are the content-bound markers built from GetInputs plus GetSerializedInputs.
- the fixture now carries real Go signatures over the frozen digest, and GoEndorsement.t.sol verifies that 2-of-2 quorum on chain. Until now the forge tests simulated endorsers with vm.sign; this closes that gap.

Tests include content-binding round trips with the real fabtoken and zkatdlog actions (creation marker equals the spend ref, forged content diverges), determinism under shuffled map iteration, and a full loop: translate, digest, sign, recover.

44 forge tests, full Go suite, make checks and make lint green.


adecaro (approved):
LGTM. Great job, continue like this 🙏

PR #1939, opened 2026-07-17
perf(identity): parallelise stored identity configuration loading

### LINKED ISSUE #1938: LocalMembership.Load registers stored identity configurations sequentially, dominating node startup
## Problem

`LocalMembership.Load` (`token/services/identity/membership/lm.go`) registers every identity configuration persisted in the identity store one at a time, on a single goroutine, while holding the membership write lock. For each stored configuration this runs the full KeyManager construction (x509/idemix credential parsing, crypto provider setup).

## Impact

On nodes with a large number of registered wallet identities the sequential loop dominates process startup. In our deployment, a node with ~300k stored wallet identities spends about 5 minutes in `Load` on a single core while the other cores stay idle, and the node cannot serve requests until the reload completes. The cost grows linearly with the number of stored identities, so every restart pays it in full.

Stored identity configurations are independent of each other — they are all non-default and their registration does not depend on one another's results — so there is no inherent reason for the reload to be single-threaded.

EvanYan1024:
Fixes #1938.

## Motivation

`LocalMembership.Load` registers every stored identity configuration sequentially while holding the membership write lock. In our deployment, a node with ~300k stored wallet identities spends about 5 minutes in `Load` on a single core — the x509/idemix KeyManager construction dominates — while the other cores stay idle, and the node cannot serve requests until the reload completes.

## What

- `registerLocalIdentity` is split into `resolveKeyManager` (probes the providers to construct a KeyManager; does not mutate the membership identity indices) and `commitLocalIdentity` (index writes + persistence). The existing callers keep the composed sequential behaviour.
- `Load` resolves stored configurations (always non-default) through an errgroup capped at `NumCPU`, then commits the results sequentially **in the original store iterator order**. Identity ordering is therefore unchanged: fallback default selection (`localIdentities[0]`) and same-name/same-priority tie-breaks in `localIdentitiesByName` behave exactly as the sequential code did. Configured identities keep the fully sequential path to preserve default-identity semantics.
- Path translation stays serial, so the `Config` contract is untouched. `KeyManagerProvider` now documents that `Get` may be called concurrently for distinct configurations and that each returned KeyManager must either be independently owned by the caller or safe for concurrent `EnrollmentID` calls — the only method the resolve phase invokes. Both in-repo providers (x509, idemix) construct fresh state per call.
- The directory-fallback path (`registerLocalIdentities`) stays on the sequential commit side.
- `docs/services/identity.md` documents the loading behaviour and the provider contract.

## Tests

- `TestLoad_ParallelStoredConfigurations`: 500 stored configurations; asserts all are registered and that KeyManager construction actually overlaps (max in-flight > 1).
- `TestLoad_FallbackDefaultKeepsIteratorOrder`: the first configuration finishes construction last; the fallback default must still be the first in iterator order.
- `TestLoad_SameNameSamePriorityKeepsIteratorOrder`: two configurations sharing name and priority complete in reverse order; lookups must still resolve to the first in iterator order.

The two ordering tests fail against a variant that commits in completion order (verified red-green), and the package is race-clean under `go test -race`.


adecaro:
Hi @EvanYan1024 , you are hitting the setting where we need to pay attention to resource allocation. If a node has so many wallets, lazy loading become more and more valuable. Any proposal on this side is more than welcome 🙏 

In the meantime, thanks for this PR 😄

adecaro (approved):
LGTM. Thanks a lot @EvanYan1024 🙏

PR #1941, opened 2026-07-17
build(node): add cmd/node build target and wire it into CI

### LINKED ISSUE #1940: cmd/node module excluded from build tooling and CodeQL analysis
## Problem

`cmd/node` is a separate Go module (has its own `go.mod`/`go.sum`), but it isn't
wired into the repo's build and CI tooling the way the other `cmd/*` submodules are:

- `GO_MODULES` in the `Makefile` does not list `cmd/node`, so `make tidy`,
  `make lint`, and other module-wide targets silently skip it.
- The CodeQL workflow (`.github/workflows/codeql-analysis.yml`) uses
  `build-mode: autobuild`, which only builds the root module. Since `cmd/node`
  is its own module, it was never actually compiled during CodeQL analysis —
  meaning its code has not been covered by security scanning.

## Impact

- Dependency and lint issues in `cmd/node` can go unnoticed because standard
  maintenance targets don't touch it.
- CodeQL's weekly/PR-triggered security analysis has a blind spot: any
  vulnerability introduced in `cmd/node` would not be detected.

adecaro:
## Summary
- Give `cmd/node` its own `go.mod`/`go.sum` and add it to the Makefile's `GO_MODULES` list, so it's picked up by `make tidy`, `make lint`, and other module-wide targets like the other `cmd/*` submodules.
- Switch CodeQL from `build-mode: autobuild` to `build-mode: manual` with an explicit build step that compiles both the root module and `cmd/node`, since autobuild only builds the root module and was silently skipping `cmd/node`.
- Bump `codeql-action/init` and `codeql-action/analyze` from v3 to v4.
- Handle the previously ignored error from `n.InstallSDK(...)` in `cmd/node/main.go`.

Fixes #1940

## Test plan
- [x] `go build -o /dev/null .` in `cmd/node` succeeds
- [x] `make fmt` run, no formatting diffs
- [x] CI (checks, CodeQL) passes on this PR


AkramBitar (approved):

AkramBitar:
LGTM

PR #1947, opened 2026-07-17
perf(ttx): parallelize remote round-trips in CollectEndorsementsView

### LINKED ISSUE #1871: perf: parallelize remote round-trips in CollectEndorsementsView
## Problem

`CollectEndorsementsView` collects signatures and distributes the signed envelope by looping over remote parties one at a time, each blocking on a network round-trip (up to 1 minute) before moving to the next. These parties are independent, so total latency is the sum of every party's round-trip instead of the slowest one. With enough parties (e.g. a multisig/policy identity with several remote co-owners, or many counterparties/auditors), this can also eat into the overall timeout budget and fail otherwise-healthy slow parties.

## Suggested fix

Fan out both loops concurrently (bounded worker pool, e.g. errgroup with a concurrency limit), keeping fail-fast behavior — first error cancels the rest. The multisig spend flow elsewhere in the codebase already does goroutine-per-party with an answer channel; reuse that pattern. 

## Test plan

- Existing multisig/policy integration tests should still pass.
- Add a test with an artificially slow responder to confirm latency tracks the slowest party, not the sum.

arunabha003:
## Addressed issue

Fixes #1871

## Summary

`CollectEndorsementsView` contacted remote parties one at a time, blocking up to
1 minute per party in both signature collection (`requestSignatures` →
`signRemote`) and envelope distribution (`distributeTxToParties`), so total
latency grew with the number of parties and slow-but-healthy parties queued
last could exhaust the overall timeout budget.

Both loops now fan out one goroutine per remote party, reusing the
`utils.AnswersCollector` pattern from the multisig spend flow (`ttx/multisig/spend.go`)
via a small `fanOut` helper. Design notes:

- Only the remote/network paths fan out; local, external-wallet, multisig and
  policy signing are unchanged and stay inline. Multisig/policy recursion is
  untouched — each nesting level parallelizes its own remote members.
- Workers only do network I/O and signature verification. All shared-state
  writes stay on the calling goroutine: payload marshalling (full vs
  enrollment-ID-filtered) happens before the fan-out, and the sigmas map /
  `appendTransactionEndorseAck` writes happen after it.
- The sessions cache is now mutex-guarded: on the fail-fast path, workers still
  in flight can read it while the deferred cleanup deletes from it.
- Remote signers are deduplicated by unique ID before the fan-out so no two
  workers share a party's session (the sigmas map only ever held one entry per
  identity, so behavior is unchanged).
- Fail-fast is preserved: the first error stops the wait; stragglers drain into
  the collector's buffered channel and exit without leaking.

## Validation done

- New unit tests: latency tracks the slowest party (10 parties × 200ms ≈ 0.2s
  vs 2s serial), first error stops the wait while slow parties are in flight,
  context cancellation unblocks the collection, and the zero-remote-parties
  case is a no-op.
- Full `token/services/ttx` suite passes under `-race`.
- `make checks` and golangci-lint clean.


adecaro:
@hstarorg , what do you think of the PR?

adecaro:
Hi @arunabha003 , very nice job. Thank you 🙏

adecaro (approved):
LGTM

PR #1948, opened 2026-07-18
validator: define consensus-safe resource limits for untrusted requests

### LINKED ISSUE #1928: validator: define consensus-safe resource limits for untrusted requests
## Summary

The reusable token-request validators do not define protocol-level limits for raw request size, action and signature counts, metadata sizes, input/output counts, or cryptographic proof sizes.

## Impact

A validator can allocate memory and enter expensive deserialization or cryptographic verification based on attacker-controlled sizes. Transport-specific limits do not protect every validator caller, while node-local limits could introduce inconsistent validation across peers. A consensus-safe policy is needed before enforcement or continuous fuzz campaigns are enabled.

The policy must cover the common validator as well as driver-specific proof and action bounds.

adecaro:
## Summary
- Add a common resource-limit gate (`token/core/common/limits.go`) enforced in `VerifyTokenRequestFromRaw`, rejecting oversized raw requests, and requests with too many actions/signatures or oversized actions/signatures, before any protobuf parsing or cryptographic work.
- Add driver-specific action limits for fabtoken (`token/core/fabtoken/v1/actions/limits.go`) and zkatdlog/nogh (`token/core/zkatdlog/nogh/v1/issue/limits.go`, `.../transfer/limits.go`) bounding input/output/metadata counts and sizes, plus proof size for zkatdlog, enforced during action `Deserialize`/`Validate` before allocation and cryptographic verification.
- All limits are fixed package constants (not `PublicParameters` fields) so every validating peer applies identical bounds deterministically — a binary-version-scoped, consensus-safe policy that needs no protocol/PP migration.
- Exact-boundary unit tests (`limit`/`limit+1`) for every new constant across all three packages.
- New fuzz targets (`FuzzRequestResourceLimits`, `FuzzActionResourceLimits` x2) with persisted seed corpora, wired into the nightly fuzz workflow matrix; existing fuzz harness caps now reference the production limit constants instead of local ad-hoc values.
- New doc page `docs/drivers/validation-resource-limits.md` describing the policy and its consensus-safety contract, linked from `docs/driverapi.md`.

## Test plan
- [x] `make checks`
- [x] `make lint-auto-fix`
- [x] `make unit-tests` (full suite, no failures)
- [x] Targeted package tests: `token/core/common/...`, `token/core/fabtoken/v1/...`, `token/core/zkatdlog/nogh/v1/...`
- [x] Live fuzz campaigns on all three new targets (clean, no crashes)
- [ ] Fabric integration tests — not run in this environment (no `FAB_BINS` configured); no currently-valid request exercised by the unit/regression suites trips any new limit.

Fixes #1928

### DIFF
--- a/docs/drivers/validation-resource-limits.md
+++ b/docs/drivers/validation-resource-limits.md
@@ -1,31 +1,81 @@
 # Validator Resource Limits
 
 This page describes the resource limits enforced on untrusted token requests and actions before
-they reach cryptographic verification, and the consensus-safety contract those limits carry.
+they reach cryptographic verification, the configuration mechanism that controls them, and the
+consensus-safety contract that mechanism carries.
 
 ## Why limits exist
 
 The token request validators (`token/core/common`, and the fabtoken/zkatdlog drivers built on top
-of it) accept raw, attacker-controlled bytes over the network. Before these limits were
-introduced, only the signing anchor had an explicit size bound (`driver.MaxAnchorSize`). Nothing
-else bounded the size of the raw request, the number of actions or signatures, the size of an
-individual action or signature, the number of inputs/outputs/metadata entries in an action, or the
-length of a zero-knowledge proof. An attacker could force unbounded allocations
+of it) accept raw, attacker-controlled bytes over the network. Aside from the signing anchor
+(`driver.MaxAnchorSize`), nothing else bounds the size of the raw request, the number of actions or
+signatures, the size of an individual action or signature, the number of inputs/outputs/metadata
+entries in an action, or the length of a zero-knowledge proof — unless these limits are enforced.
+Without them, an attacker could force unbounded allocations
 (`make([]..., len(attackerControlledCount))`) and expensive cryptographic work (proof
 deserialization, ZK verification) purely by shaping the wire bytes, without needing any valid
 signature.
 
+## Configuration mechanism
+
+Limits are held in a single struct, `driver.ResourceLimits` (`token/driver/limits.go`), injected
+into every validator at construction time — the validator itself never reads a package constant.
+`driver.DefaultResourceLimits()` returns the historical, always-safe values (see the tables below);
+`driver.ResourceLimits.WithDefaults()` overlays those defaults onto any zero-valued field, so a
+partially-specified override never silently disables a limit by leaving it at zero.
+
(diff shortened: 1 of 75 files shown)


token/driver/limits.go
```diff
@@ -0,0 +1,131 @@
...
+
+		MaxProofBytes: 128 << 10, // 128 KiB
+	}
+}
+
+// WithDefaults returns a copy of l where every zero-valued field is replaced by the
+// corresponding field from DefaultResourceLimits. It lets callers accept a partially-specified
+// ResourceLimits (e.g. parsed from a config file or environment variables where most fields are
+// left unset) without ever silently disabling a limit by leaving it at zero.
+func (l ResourceLimits) WithDefaults() ResourceLimits {
```
AkramBitar, line 76: WithDefaults() only considers 0 as an unset value. As a result, negative values (for example, from a configuration) are preserved instead of falling back to the default.

docs/drivers/validation-resource-limits.md
```diff
@@ -0,0 +1,181 @@
...
+  standalone Fabric chaincode process (`token/services/network/fabric/tcc/main/main.go`), which has
+  no configuration service wired. Reads `TOKEN_VALIDATION_MAX_*` environment variables (e.g.
+  `TOKEN_VALIDATION_MAX_ACTIONS`), applying the same unset-field-defaults overlay.
+
+A `driver.StaticResourceLimits` provider (a trivial wrapper returning a fixed value) is used by
+tests, tools, and any caller that only needs the defaults (e.g.
+`cmd/token_validation_service`, the zkatdlog regression suite).
+
+The resolved `driver.ResourceLimits` flows: composition root → `core.NewValidatorDriverService(limits, ...)`
+→ `driver.ValidatorDriver.NewValidator(pp, limits)` → the per-driver `common.NewValidator(..., limits, ...)`
```
AkramBitar, line 55: Do we need to update docs/drivers/extending_validator.md with the change of adding limits to the interface?

AkramBitar (changes requested):

token/driver/limits.go
adecaro: fixed

docs/drivers/validation-resource-limits.md
adecaro: fixed

AkramBitar (approved):
LGTM

PR #1950, opened 2026-07-18
identity: fix panics, lock-handling bug, info leak, and signer-duplication bypasses

### LINKED ISSUE #1949: identity: panics, lock-handling bug, info leak, and signer-duplication bypasses across idemix/idemixnym/membership/boolpolicy/multisig/marshal
## Summary

A security audit of `token/services/identity/` (x509, idemix, idemixnym, multisig, boolpolicy, marshal, membership) found 8 confirmed attacker-reachable defects: several panics/DoS on malformed or attacker-supplied input, a lock-handling bug that can bring down the whole process, an information leak in an error path, and two authorization-bypass patterns that let a single signer satisfy a policy meant to require multiple independent signers.

## Impact

- **Panics on malformed audit info** (idemix and idemixnym): deserializing attacker-supplied audit data with insufficient validation of its internal structure can panic downstream accessors instead of failing deserialization cleanly.
- **Unrecoverable process crash + info leak in the identity membership cache**: a lock-handling bug on an error path can corrupt process-wide lock state, taking down unrelated in-flight requests; a separate "not found" error path over-shares internal state that shouldn't be attacker-visible.
- **Authorization bypass via duplicate identities** (boolpolicy AND-policies and multisig thresholds): identity lists used to build multi-signer policies aren't checked for duplicates, letting one real signer be counted multiple times against a policy that should require independent signers.
- **Panic on a specific malformed component identity** in the boolpolicy/multisig audit-info-matcher path.
- **Encoding correctness gaps in the low-level identity wire-format decoder** (`token/services/identity/marshal`): a length-encoding edge case can silently corrupt data instead of erroring, and a separate accumulator width issue is panic-prone on 32-bit architectures.

## Scope

`token/services/identity/idemix/crypto`, `token/services/identity/idemixnym/nym`, `token/services/identity/membership`, `token/services/identity/boolpolicy`, `token/services/identity/multisig`, `token/services/identity/marshal`.

Fixes, regression tests, and new fuzz coverage for all of the above are ready and will follow in a linked PR.

adecaro:
Fixes #1949

## Summary

- idemix/idemixnym: bounds/nil-check audit-info attributes and EID/RH auditing data at deserialization time (`crypto.AuditInfo.Validate`) instead of panicking in later accessors; route `km.go` call sites through `DeserializeAuditInfo` so the validation can't be bypassed
- membership: `defer` the `RLock`/`RUnlock` pair around `refreshAndGet` so a panic can't leave the RWMutex unbalanced and crash the process; stop logging full identity-map values on the "not found" error path
- boolpolicy/multisig: reject empty/none and duplicate component identities at Wrap- and deserialize-time, closing both a nil-`Matcher` panic and a duplicate-identity signature bypass that let one signer satisfy a policy meant to require multiple independent signers
- marshal: fix DER long-form length encoding (previously only ever emitted the 2-byte form, silently corrupting declared lengths ≥ 65536) and widen the length accumulator to `uint64` to avoid a 32-bit-`int` overflow panic
- Added fuzz tests for every byte-parsing entry point in this tree that lacked one, and wired all of them (new and pre-existing) into `nightly-fuzz.yml`'s matrix so they run under extended `-fuzztime` in CI instead of only exercising their seed corpus
- Codified the "add fuzz tests + wire them into nightly-fuzz.yml" rule in `AGENTS.md` so it isn't lost after this PR

Every pre-existing regression test that previously proved a bug existed (`require.Panics`, asserting a bypass/leak) is flipped to prove the fix instead.

## Validation

- `go test ./token/services/identity/...` and `go test -race ./token/services/identity/...` — all packages pass
- `gofmt -l`, `go vet ./token/services/identity/...`, `golangci-lint run ./token/services/identity/...` — clean
- Each new/updated fuzz target run locally for a 20s campaign — no panics found, beyond passing seeded corpus
- `make checks` — clean across all modules
- New `nightly-fuzz.yml` matrix entries verified via `go test <pkg> -list='^<func>$'`, YAML parse, and `actionlint` (0 issues)


plan.md
```diff
@@ -0,0 +1,483 @@
+✅ COMPLETE (Analysis Phase) — ✅ COMPLETE (Fix Phase, see bottom of file)
```
AkramBitar: Do we need this file?

AkramBitar (changes requested):

AkramBitar (approved):
LGTM
Only one comment

plan.md
adecaro: removeed

PR #1955, opened 2026-07-19
fix(zkatdlog): close unauthenticated validator DoS in issue path (#1954)

### LINKED ISSUE #1954: zkatdlog issue path: unauthenticated validator DoS via nil-deref in proof verification
## Summary

Two unauthenticated denial-of-service vectors exist in the zkatdlog/nogh/v1 **issue** action validation path (`token/core/zkatdlog/nogh/v1/issue/`). A crafted issue action reaching `IssueValidate` (`validator/validator_issue.go`) can trigger a nil-pointer-dereference panic during zero-knowledge verification — **before** issuer authorization or signature checks run. Any party able to submit a token request to a validating peer can crash the goroutine handling validation, without holding any issuer credential or valid signature.

### Vector 1 — nil output commitment

- `utils.FromG1Proto` returns `(nil, nil)` when the proto `G1.Raw` field is empty (this is treated as "absent", not malformed).
- `issue.Action.Deserialize` stores that `nil` into `Outputs[j].Data` and returns no error.
- `issue.Action.Validate` only checked `output == nil`, not `output.Data == nil`.
- `issue.Action.GetCommitments` likewise only checked `Outputs[j] == nil`, returning a slice containing a `nil` `*mathlib.G1`.
- `issue.Verifier.Verify` calls `.Copy()` on that nil element → panic.

The **transfer path** already guards against this (`transfer.Action.Validate` calls `token.Token.Validate(false)`, which rejects `Data == nil`). The issue path never called that helper, so this is an asymmetry between the two paths rather than a fundamentally new class of bug.

### Vector 2 — truncated SameType sub-proof

- ASN.1 decoding (`asn1.NextG1`/`NextZr`) returns `(nil, nil)` — not an error — when an element index runs past the end of the encoded values.
- A `SameType` proof serialized with missing trailing elements deserializes successfully with `CommitmentToType`/`Type`/`BlindingFactor`/`Challenge` left `nil`.
- `SameType` had no `Validate` method (unlike its transfer-side sibling `TypeAndSumProof.Validate`, which nil-checks every field before use).
- `SameTypeVerifier.Verify` and `issue.Verifier.Verify` then call `.Mul()`/`.Sub()`/`.Copy()` on those nil fields → panic.

## Impact

- **Unauthenticated validator DoS.** The panic path in `IssueValidate` is reached before `ctx.PP.Issuers()` authorization and `ctx.SignatureProvider.HasBeenSignedBy` signature checks, so no valid issuer credential or signature is required to trigger it — only a syntactically-parseable issue action with a stripped/nil field.
- Repeated submission can repeatedly crash the validating goroutine/process on any node that validates the token request, which is a network availability concern for any peer running this driver.

## Suggested fix direction

- Reject nil output commitments at the earliest validation boundary in `issue.Action.Validate`/`GetCommitments`, mirroring the existing guard on the transfer path (`token.Token.Validate`).
- Add a `SameType.Validate` method (mirroring `transfer.TypeAndSumProof.Validate`) and invoke it right after deserialization in `issue.Verifier.Verify`, before the proof's fields are used.
- Add regression coverage at both the unit level (`issue` package) and the validator entry point (`Engine.VerifyTokenRequestFromRaw`), plus fuzz coverage over `issue.Action.Deserialize` → `Validate` → `GetCommitments` to catch sibling nil-derefs.

adecaro:
Fixes #1954

## Problem

Two unauthenticated denial-of-service vectors existed in the zkatdlog/nogh/v1 issue-action validation path. A crafted issue action reaching `IssueValidate` could trigger a nil-pointer-dereference panic during zero-knowledge verification, **before** issuer authorization or signature checks run — so no valid issuer credential or signature was required to crash the validating goroutine.

1. **Nil output commitment.** `utils.FromG1Proto` returns `(nil, nil)` for an empty proto `Raw` field, so an output with an empty commitment deserializes cleanly with `Outputs[j].Data == nil`. Neither `Action.Validate` nor `Action.GetCommitments` checked for this, and `issue.Verifier.Verify` panicked calling `.Copy()` on the nil element.
2. **Truncated SameType sub-proof.** ASN.1 decoding (`asn1.NextG1`/`NextZr`) returns `(nil, nil)` — not an error — for elements past the end of the encoded values. `SameType` had no `Validate` method (unlike its transfer-side sibling `TypeAndSumProof.Validate`), so a proof with missing trailing elements deserialized with nil fields that later panicked in `.Mul()`/`.Sub()`/`.Copy()` calls.

The transfer path was already protected against the first case (`transfer.Action.Validate` calls `token.Token.Validate(false)`); the issue path never called the equivalent guard.

## Fix

- `issue.Action.Validate`/`GetCommitments` now reject nil output commitments, mirroring the transfer path's guard.
- Added `issue.SameType.Validate`, mirroring `transfer.TypeAndSumProof.Validate`, invoked in `BulletProofVerifier.Verify` immediately after deserialization and before any field is used.

## Testing

- Unit tests for both vectors in the `issue` package (`action_test.go`, `sametype_test.go`), asserting an error is returned (not a panic).
- Validator-entry-point regression tests in `validator/validator_security_test.go` exercising `Engine.VerifyTokenRequestFromRaw` end-to-end for both vectors.
- New fuzz target `FuzzIssueActionValidateNoPanic` in `validator/validator_fuzz_test.go` driving `issue.Action.Deserialize` → `Validate` → `GetCommitments` under `require.NotPanics`, seeded with the nil-commitment wire shape. Ran 30s locally (677k executions, 0 crashes) with no new corpus entries.
- Promoted a scratch positive-regression test for the CSP range-verifier's out-of-range rejection into `crypto/rp/csp/rp_test.go` (unrelated to the DoS fix, but was sitting uncommitted in the same investigation).
- Full suite green with `-race`: `validator`, `issue`, `transfer`, `crypto/rp/{bulletproof,csp,executor}`.
- `make lint-auto-fix && make checks` clean.

## Investigated, no change needed

The CSP range-proof verifier's `expectedRounds` (in `validateRangeProof`) is derived from the proof's own `Left` slice length rather than from a verifier-side parameter. Traced this through `rangeVerifier.Verify`: the round count actually used for indexing (`cspRounds`) is computed independently from `NumberOfBits` (a public parameter), and the inner CSP verifier re-validates the proof's lengths against that correct value before any indexing occurs. A self-consistent-but-wrong-length proof is safely rejected there, so this is not exploitable — no fix applied.

### DIFF
--- a/token/core/zkatdlog/nogh/v1/crypto/rp/csp/rp_test.go
+++ b/token/core/zkatdlog/nogh/v1/crypto/rp/csp/rp_test.go
@@ -164,9 +164,21 @@ func TestRangeProofOutOfRange(t *testing.T) {
 			prover.WithTranscriptHeader([]byte("transcript-header"))
 
 			// the proof is still generated, the verification will fail though
-			_, err = prover.Prove()
+			proof, err := prover.Prove()
 			require.NoError(t, err)
-			// require.Contains(t, err.Error(), "does not fit")
+
+			verifier := &rangeVerifier{
+				VGenerators:  vGens,
+				AGenerators:  aGens,
+				BGenerators:  bGens,
+				VCommitment:  vComm,
+				NumberOfBits: n,
+				Curve:        curve,
+			}
+			verifier.WithTranscriptHeader([]byte("transcript-header"))
+
+			err = verifier.Verify(proof)
+			require.Error(t, err, "out-of-range value must be rejected by verifier")
 		})
 	}
 }
--- a/token/core/zkatdlog/nogh/v1/issue/action.go
+++ b/token/core/zkatdlog/nogh/v1/issue/action.go
@@ -204,6 +204,9 @@ func (i *Action) Validate() error {
 		if output == nil {
 			return ErrNilOutput
 		}
+		if err := output.Validate(false); err != nil {
+			return errors.Join(ErrInvalidOutput, err)
+		}
 	}
 	if i.ProofType != rp.RangeProofType && i.ProofType != rp.CSPRangeProofType {
 		return ErrInvalidProofType
@@ -343,6 +346,9 @@ func (i *Action) GetCommitments() ([]*math.G1, error) {
 	for j := range com {
 		if i.Outputs[j] == nil {
 			return nil, ErrNilOutput
+		}
+		if i.Outputs[j].Data == nil {
+			return nil, ErrNilOutputData
 		}
 		com[j] = i.Outputs[j].Data
 	}
--- a/token/core/zkatdlog/nogh/v1/issue/action_test.go
+++ b/token/core/zkatdlog/nogh/v1/issue/action_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/LFDT-Panurus/panurus/token/core/common/encoding/json"
 	"github.com/LFDT-Panurus/panurus/token/core/zkatdlog/nogh/protos-go/v1/actions"
 	"github.com/LFDT-Panurus/panurus/token/core/zkatdlog/nogh/v1/crypto/rp"
+	v1 "github.com/LFDT-Panurus/panurus/token/core/zkatdlog/nogh/v1/setup"
 	"github.com/LFDT-Panurus/panurus/token/core/zkatdlog/nogh/v1/token"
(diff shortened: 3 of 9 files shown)


AkramBitar (approved):
LGTM

PR #1957, opened 2026-07-20
perf(ttx/finality): push confirmed status events and replace per-waiter fallback polling with a shared batch poller

### LINKED ISSUE #1956: ttx finality: confirmed commit path never pushes a status event; per-waiter fallback polling dominates database load
### Context

`FinalityView` (`token/services/ttx/finality.go`) waits for a transaction to reach a terminal status in the ttxdb/auditdb. Detection is designed as push + fallback polling: the waiter registers a status listener channel on the store, and additionally re-queries `GetStatus` on its own `pollingTimeout` timer.

There are two problems with how this works today.

### 1. The confirmed commit path never pushes a status event

`finality.Commit` (`token/services/ttx/finality/listener.go`) writes the Confirmed status through the store transaction (`tx.SetStatus` inside `Commit`), which bypasses the store service and its in-process listener notification — only the Deleted path goes through `StoreService.SetStatus`, which is where `Notify` lives.

So waiters on a confirmed transaction (the overwhelmingly common case) are never woken by the push channel at all; they resolve exclusively through the fallback polling, adding up to one polling interval of latency to every finality resolution.

### 2. Fallback polling is per-waiter and dominates database load

Every waiter re-queries `GetStatus` on its own timer. With tens of thousands of in-flight waiters in our deployment, these per-waiter queries (~3.5k/s plus the setup checks at high concurrency) queued thousands of goroutines on the `database/sql` pool mutex. The batching decorator introduced in #1896 coalesces concurrent lookups into fewer round trips, but every waiter still wakes up each interval to issue one — the wake-up and query count stays proportional to the number of waiters.

Because of problem 1, this polling is not actually a fallback today: it is the only mechanism that resolves confirmed transactions, so its load is structural rather than a safety net.

### Proposal

1. Push the Confirmed event through the store service after the commit transaction completes, so waiters wake through the same path as a Deleted push.
2. With push working, replace the per-waiter fallback polling with a single shared poller per finality database that batch-fetches the statuses of all waited-on transactions through `GetStatuses` (#1896) and re-publishes terminal ones through the store's listener notification. The waiter then just registers, checks the status once (to cover the registration race), and blocks on the push channel.

This keeps `WithPollingTimeout` semantics (the shared poller sweeps at the smallest interval among active waiters) while making the wake-up latency push-driven and the fallback database load O(1) sweeps instead of O(waiters) queries.

I have a PR ready for this.

EvanYan1024:
Fixes #1956

### Motivation

`FinalityView` waiters are designed to be woken by push (the store's status listener) with polling as a fallback, but today the push channel never fires for confirmed transactions: the confirmed path writes its status through the store transaction (`tx.SetStatus` inside `finality.Commit`), bypassing `StoreService.SetStatus` and its listener notification — only the Deleted path pushes. Every confirmed transaction therefore resolves through the fallback polling, paying up to one polling interval of extra latency.

That also makes the per-waiter polling load structural: with tens of thousands of in-flight waiters in our test environments under sustained load, the per-waiter `GetStatus` queries (~3.5k/s plus setup checks at high concurrency) queued thousands of goroutines on the `database/sql` pool mutex. The batching decorator from #1896 coalesces concurrent lookups into fewer round trips, but every waiter still wakes up each interval to issue one.

### Changes

**fix(ttx/finality): notify status waiters on the confirmed commit path**

- ttxdb/auditdb `StoreService` gains `NotifyStatus`, which pushes a status event to the in-process listeners without touching the database; `SetStatus` routes through it, so status events now also carry the status message.
- `finality.Commit` pushes the Confirmed event through `NotifyStatus` after the transaction commits, waking waiters through the same path as a Deleted push. This covers both the live finality listener and the recovery handler, which share `Commit`.

**perf(ttx/finality): replace per-waiter fallback polling with a shared batch poller**

- `dbFinality` now registers the listener, checks the status once (covering the registration race), and blocks on the push channel — no per-transaction polling loop.
- A shared `statusPoller` per finality database batch-fetches the statuses of all waited-on transactions through `GetStatuses` (#1896), in chunks of 1000 (a failing chunk is skipped, not the whole sweep), and re-publishes terminal ones through the store's `NotifyStatus`, waking the registered waiters through the same path as a live push.
- The poller sweeps at the smallest polling interval among the active waiters, so `WithPollingTimeout` keeps its per-view meaning. Its loop is only woken when the sweep interval must shrink, when the smallest interval grew, or when the last waiter left — waking on every registration would keep resetting the timer and starve sweeps under sustained load. The goroutine stops when no waiter remains and restarts on the next registration.
- The batching decorator from #1896 keeps serving the remaining single-tx lookups (the known-check at view setup and the initial check after registration); the batch lookup, the listener registry and the notification surface pass through it untouched.

### Testing

- New unit tests for the poller (registration/unregistration, interval shrink/grow, stop/restart, sweep notification of terminal statuses) and for the Confirmed push on the commit path.
- Existing finality view tests adapted to the push-first flow.
- `go build ./token/...`, `go vet`, `go test -race` on the affected packages, and golangci-lint v2.12.2 with the repo configuration all pass locally.

### DIFF
--- a/docs/services/ttx.md
+++ b/docs/services/ttx.md
@@ -497,6 +497,8 @@ Once fully endorsed, the transaction metadata is distributed to all recipients s
 
 The `FinalityView` allows applications to wait for a transaction to be committed to the ledger. Internally, Panurus's **Network Service** listens for ledger events. When a transaction reaches finality, the Network Service notifies Panurus, which then updates the local `Transactions DB` and `Tokens DB` to reflect the new state.
 
+Waiting is push-first: each waiter registers a status listener on the local database (`TTXDB` or `AuditDB`), checks the status once to cover the registration race, and then blocks until a status event arrives. Status writes — including the confirmed commit path — push the event in-process. As a safety net for lost push events, a single fallback poller per database batch-fetches the statuses of all waited-on transactions (`GetStatuses`) and re-publishes terminal ones; it sweeps at the smallest polling interval among the active waiters (`WithPollingTimeout`, default 1s) and stops when no waiter remains. There is no per-transaction polling.
+
 ### Transaction Recovery
 
 Panurus includes an automatic recovery mechanism to handle pending transactions that may have lost their finality listeners due to node restarts, network interruptions, or other failures. 
--- a/token/services/network/fabric/network.go
+++ b/token/services/network/fabric/network.go
@@ -649,6 +649,7 @@ type transactionDB interface {
 	NewTransaction() (storage.TransactionStoreTransaction, error)
 	GetTokenRequest(ctx context.Context, txID string) ([]byte, error)
 	SetStatus(ctx context.Context, txID string, status storage.TxStatus, message string) error
+	NotifyStatus(ctx context.Context, txID string, status storage.TxStatus, message string)
 	AcquireRecoveryLeadership(ctx context.Context, lockID int64) (recovery2.Leadership, bool, error)
 	ClaimPendingTransactions(ctx context.Context, olderThan time.Duration, leaseDuration time.Duration, limit int, owner string) ([]*ttxdb.RecoveryClaim, error)
 	ReleaseRecoveryClaim(ctx context.Context, txID string, owner string, message string) error
--- a/token/services/storage/auditdb/store.go
+++ b/token/services/storage/auditdb/store.go
@@ -290,14 +290,17 @@ func (d *StoreService) SetStatus(ctx context.Context, txID string, status dbdriv
 	}
 
 	// notify the listeners
-	d.Notify(common.StatusEvent{
-		Ctx:            ctx,
-		TxID:           txID,
-		ValidationCode: status,
-	})
+	d.NotifyStatus(ctx, txID, status, message)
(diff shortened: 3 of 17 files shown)


adecaro:
Hi @EvanYan1024 , I see the point. With the postgres implementation, we have a listener attached to the db but we don't use it to trigger any event dispatched then to the application. This status listener support was there for sqlite. 

I guess, we don't really need the postgres listener because the replica that creates the transaction also listen to it. Moreover, if that replica dies, there is a leader replica that scans the transaction db for transactions whose status is not yet confirmed. 

Would you agree with this assessment? 

Anyway, I'll merge your PR now. We need to rework this postgres listeners thingy we have. We need some cleanup 😅

adecaro (approved):
LGTM

EvanYan1024:
Yes, I agree with this assessment. The in-process push covers the replica that created the transaction, and the shared poller introduced here reads the database directly, so a status written by a different replica is still picked up by the sweep and wakes local waiters — the cross-replica case doesn't need the postgres listener either. Together with the recovery leader scanning for unresolved transactions, I don't see a remaining role for it.

PR #1933, opened 2026-07-16
fsc endorsement: use endorsement policy of the namespace  #1908

adecaro:
## Summary

Closes #1908.

`token/services/network/fabric/endorsement/fsc` previously only supported `1outn` (contact one random configured endorser) and `all` (contact all configured endorsers) as FSC endorsement policies. This PR adds a third policy, `namespace`, which selects a random subset of the configured endorsers that satisfies the *real* endorsement policy of the target token namespace, instead of relying on a static configured policy.

- **Fabric**: `namespace` policy resolves the required MSPs via Fabric service discovery (`Channel.Chaincode(namespace).Discover()`), implemented in `token/services/network/fabric/endorsement/nspolicy.go` (`DiscoveryEndorserSelector`).
- **FabricX**: the equivalent selector fetches the namespace policy from the query service's `GetNamespacePolicies` RPC, implemented in `token/services/network/fabricx/endorsement/nspolicy.go` (`QueryServiceEndorserSelector`).
- Both selectors share `fsc.SelectEndorsersForMSPSets` to pick a random subset of the configured endorsers that covers the required MSP set(s); endorsement fails with an explicit error if the configured endorsers cannot satisfy the namespace's policy (no silent fallback to a weaker policy).

As part of implementing this, the FSC endorsement responder was also extended to support submitting/updating a namespace's public parameters (PP) through the same endorsement machinery, as an alternative to the chaincode `Init` lifecycle callback:
- `SetupPublicParamsView` (initiator) / `SetupPublicParamsResponderView` (responder) mirror the existing token-request `RequestApprovalView` / responder pair, distinguished by chaincode function (`invoke` vs `setup`) and a dedicated `public_params` transient key.
- The PP responder tolerates a missing TMS (first-time namespace setup) but otherwise reuses the same endorser-selection policy (`1outn`/`all`/`namespace`) and identity resolution as the token-request path.
- Endorser identity resolution (`fsc_endorsement.id`) now happens directly from configuration rather than through a fully-built TMS, so it works before any TMS/PP exists for the namespace.

See `docs/services/network-fabric.md` (FSC Endorsement / Public Parameters Setup/Update via FSC) and `docs/services/network-fabricx.md` (FSC Endorsement Service) for details, and `docs/configuration.md` for the updated `fsc_endorsement.policy.type` options.

## Testing

[...]

PR #1959, opened 2026-07-20
fix(ci): exclude Fabric Smart Client dependency code from coverage report

AkramBitar:
Coverage dropped **~82.6% → ~53%** after #1933 (https://coveralls.io/builds/80658631).

**Cause:** #1933 bumped `fabric-smart-client/integration` to v0.14.3, whose node builder now hardcodes `-coverpkg=github.com/hyperledger-labs/fabric-smart-client/...`. That overrides the `-coverpkg` we set via `GOFLAGS`, so integration tests instrument FSC's own `platform/*`/`pkg/*` (~17k lines, ~0.8% covered here) instead of our packages — flooding the report.

**Fix:** drop FSC-module lines in `filter-coverage.sh`.

Recovers the denominator (~71%). Full recovery to ~82% needs an upstream FSC change to make `-coverpkg` configurable.

### DIFF
--- a/ci/scripts/filter-coverage.sh
+++ b/ci/scripts/filter-coverage.sh
@@ -45,6 +45,9 @@ awk '
   # Exclude integration
   /\/integration\// { next }
 
+  # Exclude Fabric Smart Client dependency code (not part of this module)
+  /hyperledger-labs\/fabric-smart-client\// { next }
+
   # Exclude tools and docs
   /\/tools\// { next }
   /\/docs\// { next }


SaidAltury-ibm:
@AkramBitar are you sure this is needed? i see that the coverage at Panurus is back to normal without this
https://coveralls.io/github/LFDT-Panurus/panurus

AkramBitar:
After discussion with @SaidAltury-ibm we decided to close this PR for the following reasons:
1) The fix of FSC solved the issue
2) Adding more things to maintain without needing them can be overhead

PR #1960, opened 2026-07-20
fsc setup responder: enforce validation invariants on new public parameters

### LINKED ISSUE #1943: fsc setup responder: enforce validation invariants on new public parameters
`setupBehaviour.validate` in `token/services/network/fabric/endorsement/fsc/responder.go:508` currently only checks that the TMS ID matches (when a TMS already exists) and leaves the actual validation of the incoming public parameters unimplemented:

```go
// TODO: need to add validation logic...to be addressed in a specific PR
```

This means a `setup` request can currently install/replace public parameters for a namespace without any invariant checking beyond matching the TMS ID.

## Ask

Identify and enforce the basic invariants that a `setup` request's public parameters must satisfy before being translated/committed. In particular:

- **Signature by an existing issuer on re-setup**: if the namespace already has public parameters (`request.Tms != nil`), require that the new public parameters be signed/endorsed by an identity that is an issuer (`PublicParameters.Issuers()`) in the *current* public parameters, so that setup changes can't be unilaterally pushed by an unauthorized party.
- **Driver name/version compatibility**: decide and enforce whether `TokenDriverName()`/`TokenDriverVersion()` may change across a setup update, or must stay the same as the existing TMS's parameters.
- **Internal consistency**: call `PublicParameters.Validate()` on the parsed parameters (via the driver's `PPReader.PublicParametersFromBytes`) before accepting them.
- **Non-empty/sane issuer and auditor sets**: e.g. reject parameters with no issuers if the driver requires at least one, or duplicate/malformed identities in `Issuers()`/`Auditors()`.
- **Precision/MaxTokenValue sanity**: reject obviously invalid values (e.g. zero `MaxTokenValue`) if not already checked by `Validate()`.
- **Monotonicity / replay protection**: consider whether re-setup should require some form of versioning or freshness check to prevent replay of stale public parameters.
- **Certification driver compatibility**: if `CertificationDriver()` changes between old and new parameters, decide whether that transition is allowed and under what conditions.

Investigate the driver implementations (fabtoken, zkatdlog) and `token/driver/publicparams.go` for other invariants that should be enforced here, and land the resulting checks in `setupBehaviour.validate`.

Related: this responder is exercised by `setup_responder_test.go`; new invariant checks should come with corresponding test cases (both accept and reject paths).

adecaro:
## Summary
- Parses incoming public parameters and calls `Validate()` before accepting a setup/re-setup request.
- On re-setup of a namespace that already has a TMS, requires a detached signature over the raw public parameters bytes produced by a current issuer (from `PublicParameters.Issuers()`), carried via an optional `public_params_sig` transient, so setup changes can't be pushed by an unauthorized party.
- `fsc.EndorsementService.SetupPublicParams` looks up the existing TMS (if any) and signs on behalf of the initiator using a locally-held issuer wallet, so callers don't need to change.
- First-time setup (no existing TMS) is unaffected and requires no signature; empty issuer sets remain allowed (drivers support "anyone can issue" mode).
- Driver name/version and `CertificationDriver()` are allowed to change freely on re-setup — gated only by the issuer signature, per the driver implementations' existing design.

Fixes #1943

## Test plan
- [x] `go test ./token/services/network/fabric/endorsement/fsc/...` — accept/reject cases for validate/parse failures, missing signature, non-issuer signer, verifier lookup failure, signature verification failure, and the initiator-side signing paths (success, no current issuers, no local signer available, signer lookup failure, signing failure)
- [x] `signPublicParamsWithCurrentIssuer` at 100% coverage (previously untested)
- [x] `make checks` (license, gofmt, goimports, go vet, misspell, ineffassign, staticcheck, tidy)
- [x] `make unit-tests` (full suite, no failures)
- [x] `golangci-lint run` on the changed package — 0 issues
- [x] `docs/services/network-fabric.md` updated to describe the new invariants and the initiator-side signing flow

### DIFF
--- a/docs/services/network-fabric.md
+++ b/docs/services/network-fabric.md
@@ -178,7 +178,10 @@ initiator/responder pair handles this:
 The two initiator/responder pairs are distinguished by chaincode function name and
 transient key: the token-request flow uses the `invoke` function, while the PP flow
 uses a dedicated `setup` function and carries the raw parameters under a dedicated
-`public_params` transient key (in addition to the `tmsID` key used by both).
+`public_params` transient key (in addition to the `tmsID` key used by both). On
+re-setup of a namespace that already has a TMS, a third transient key,
+`public_params_sig`, carries a detached signature (JSON-encoded signer identity plus
+signature bytes) authorizing the change — see below.
 
 Both responders share a common `receive` step that only checks that the proposal's
 `tmsID` transient carries a non-empty network, channel, and namespace — it does not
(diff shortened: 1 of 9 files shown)


AkramBitar (commented):
LGTM

AkramBitar (approved):

PR #1962, opened 2026-07-20
ci: add nightly job to build against fabric-smart-client main

### LINKED ISSUE #345: Nightly upgrade FSC dep
We should create a nightly CI job that upgrades the FSC dep to the latest version.
If that build succeeds, this change can be automatically merged. 

TBD should this just also update all other deps?

### LINKED ISSUE #1961: Add nightly CI job to catch fabric-smart-client main breakage early
## Problem

Panurus pins `github.com/hyperledger-labs/fabric-smart-client` (FSC) to a specific
main-branch pseudo-version across every Go module in the repo. That pin is only
ever moved forward manually, on a best-effort basis, so upstream FSC changes on
`main` — API breakage, behavioral changes, new lint/vet findings — accumulate
silently between bumps.

## Impact

- By the time someone does bump the dependency, the amount of drift (and
  therefore the size/risk of the resulting change) can be large, since months
  of upstream commits land at once instead of being absorbed incrementally.
- A breaking upstream change has no chance to surface until a human happens to
  attempt the bump, so regressions are caught late and are harder to bisect to
  a specific upstream commit.
- There's no continuous signal on whether Panurus still builds, lints, and
  passes its unit/integration suites against FSC's current `main`.

This was previously tracked in #345 (closed as stale before being addressed).
The underlying problem — no automated nightly build against FSC `main` — is
still unresolved.

adecaro:
## Summary
- Adds `.github/actions/bump-fsc`, a reusable composite action that resolves the latest `fabric-smart-client` `main` commit and bumps it (plus its `/integration` submodule) across every Go module via `make update-dep`.
- Adds `.github/workflows/nightly-fsc.yml`, a nightly (03:23 UTC) + manually-dispatchable workflow that applies that bump and then runs `make checks`/`make lint`, unit tests (race + regression, with Coveralls), the full integration test matrix, a benchmark-compile sanity check, and cgo-free builds of `tokengen`/`artifactgen` — all against FSC's current `main` instead of the pinned pseudo-version.
- Adds `docs/development/update-fsc.md`, an agent-agnostic runbook for manually bumping the FSC dependency (also exposed to Claude Code as the `/update-fsc` skill via a symlink added in a prior commit on this branch).

## Why
Panurus pins FSC to a manually-bumped pseudo-version, so upstream breakage only surfaces whenever someone next attempts a bump — by then a lot of drift has accumulated. This closes that gap with a continuous nightly signal, per #1961 (a follow-up to the previously-closed #345).

Fixes #1961

## Test plan
- [x] `make checks` — clean
- [x] CI: checks / lint / unit / integration matrix on this PR
- [ ] First nightly run on `main` after merge (verify the composite action resolves FSC's SHA and the full matrix stays green)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -221,6 +221,18 @@ Before marking a task complete, update or create the relevant documentation unde
 - New `docs/` pages must follow the existing style (Markdown, same heading hierarchy as neighbouring files).
 - If no existing doc page covers the changed area, create `docs/<subsystem>/<topic>.md` and add a link from the nearest index or README.
 
+### Automation Runbooks (Workflow Rule)
+Reusable, agent-agnostic step-by-step procedures live under `docs/development/` and are
+readable by any agent that reads this file — not just Claude Code. When Claude Code also
+needs a `/slash-command` trigger for one, add a symlink at
+`.claude/skills/<name>/SKILL.md` pointing back at the doc, so there is one source of truth.
+
+- **Update `fabric-smart-client` to latest `main`**: [docs/development/update-fsc.md](docs/development/update-fsc.md)
+  (Claude Code: `/update-fsc`). Bumps the FSC dependency across every Go module, resolves
+  API/lint breakage until `make checks` and `make lint-auto-fix` are clean, then stops and
+  waits for the user's go-ahead before pushing a branch or opening the PR — the "never push
+  or open a PR without explicit go-ahead" rule above still applies to this runbook.
+
 ## 🔍 Debugging & Advanced Testing
 
 ### Log Locations
--- a/docs/development/update-fsc.md
+++ b/docs/development/update-fsc.md
@@ -1,0 +1,149 @@
+---
+name: update-fsc
+description: "Upgrade Panurus's fabric-smart-client dependency to the latest main commit, resolve API/lint breakage until `make checks` is clean, then open a PR. Trigger: /update-fsc"
+trigger: /update-fsc
+---
+
+# Update `fabric-smart-client` to latest `main`
+
+Panurus depends on `github.com/hyperledger-labs/fabric-smart-client` (FSC) across
+every Go module in the repo, pinned to a main-branch pseudo-version (e.g.
+`v0.14.3-0.20260716152906-60a85d843ab6`). This runbook is a reusable, agent-agnostic
+procedure — for a human or an AI agent — to bump that dependency to the latest FSC
+`main` commit, absorb any resulting API changes, and get the repo's checks green.
+
+This doc is the single source of truth. In Claude Code it is also exposed as the
+`/update-fsc` skill via a symlink at `.claude/skills/update-fsc/SKILL.md`.
+
+## Preconditions
+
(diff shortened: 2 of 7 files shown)


AGENTS.md
```diff
@@ -221,6 +221,18 @@ Before marking a task complete, update or create the relevant documentation unde
...
+### Automation Runbooks (Workflow Rule)
+Reusable, agent-agnostic step-by-step procedures live under `docs/development/` and are
+readable by any agent that reads this file — not just Claude Code. When Claude Code also
+needs a `/slash-command` trigger for one, add a symlink at
+`.claude/skills/<name>/SKILL.md` pointing back at the doc, so there is one source of truth.
+
+- **Update `fabric-smart-client` to latest `main`**: [docs/development/update-fsc.md](docs/development/update-fsc.md)
+  (Claude Code: `/update-fsc`). Bumps the FSC dependency across every Go module, resolves
+  API/lint breakage until `make checks` and `make lint-auto-fix` are clean, then stops and
+  waits for the user's go-ahead before pushing a branch or opening the PR — the "never push
```
AkramBitar, line 233: What is the use's go-ahead?
Is that manual explicit approval to create the branch and the PR?
How the user see the results and where?

docs/development/update-fsc.md
```diff
@@ -0,0 +1,149 @@
...
+
+Include the old → new pseudo-version in the commit body. Never introduce
+`fmt.Errorf`/`fmt` for error construction in any file you touch — use
+`github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors` per project
+convention.
+
+### 8. Stop and confirm before any remote action
+
+**Do not push or open a PR without the user's explicit go-ahead**, per `AGENTS.md`.
+Report what changed and that `make checks` / `make unit-tests` are green, and wait for
```
AkramBitar, line 126: What in case we have build error or failed tests, etc.?
What is the flow in that case?

AkramBitar (changes requested):
LGTM. 
Have some questions/clarifiaction?

docs/development/update-fsc.md
```diff
@@ -0,0 +1,149 @@
...
+confirmation.
+
+If `make checks` (or the build) cannot be made clean, **do not open a PR** — report the
```
AkramBitar, line 129: Recently we saw that the coverage fail down after integrating new FSC release. 
How we can catch this her, on the PR?

AGENTS.md
adecaro: this is for maintainers to use and the output depends on the agent one is using.

docs/development/update-fsc.md
adecaro: this is up to the agent how the errors will be addressed, if can be fixed or not.

AkramBitar (approved):

AkramBitar:
LGTM

docs/development/update-fsc.md
adecaro: this PR is about checking if there are incompatibilities. There is no PR involved.

Issue #1669, opened 2026-05-06
ethereum driver: endorser-based approval flow on top of TTX


malsomesh9:
## Summary
Implement the FSC endorser-based approval flow for the Ethereum/EVM network driver on top of the existing TTX validation path.

## Parent
Part of #1648.

## Scope
- integrate FSC endorsers with the Ethereum driver request flow
- have endorsers validate token actions off-chain
- collect signatures over the approved state update payload
- submit the endorsed payload to the contract layer for verification and execution
- preserve as much of the existing TTX validation model as possible

## Notes
This issue depends on the signing envelope and contract-layer decisions made in earlier slices.

PR #1967, opened 2026-07-21
network: reject already-processed or stale approval/setup requests

### LINKED ISSUE #1958: request approval: check if the request was already approved
Before starting evaluating a request, check if the request has been already processed.

adecaro:
## Summary
- Add a pluggable `replay.Guard` (`token/services/network/common/replay`) checked in `fsc.ResponderView.Call` right after `receive`, before any crypto verification runs — rejects a duplicate or replayed proposal before paying the cost of validating it.
- Dedup key = `(TxID, Creator, Nonce, Timestamp)` from the proposal itself; a symmetric freshness window also rejects requests whose timestamp has drifted too far from the guard's clock (`ErrOutOfWindow`, distinct from `ErrAlreadyProcessed`).
- Default backend is in-memory (TTL + bounded LRU, atomic check-and-set), selected via a `Config`/`factory.New` pair so a distributed backend can be added later without touching callers — reusable/driver-agnostic so the WIP Ethereum endorser flow (#1669) can adopt it too.
- Guard is constructed **per-TMS** (from `fsc_endorsement.replay` config, shared by the Fabric and FabricX loaders) rather than one instance per node, so tuning/freshness windows can differ across TMSs. Absent config falls back to `replay.DefaultConfig()` — fully backwards compatible.

Fixes #1958

## Test plan
- [x] `make checks` — license/fmt/vet/staticcheck/tidy all pass
- [x] `go test -race ./token/services/network/common/replay/... ./token/services/network/fabric/... ./token/services/network/fabricx/...` — all pass
- [x] `go build ./...`
- [x] Docs updated: `docs/configuration.md`, `docs/services/network-fabric.md`, cross-linked from `network-fabricx.md`/`network-ethereum.md`


docs/services/network-fabric.md
```diff
@@ -216,6 +216,83 @@ writing; the existing ledger setup-key listener (see
...
+
+```mermaid
+sequenceDiagram
+    participant FSC as FSC Endorsement Service
+    participant Guard as replay.Guard
+    participant Val as validateProposal / behaviour.validate
+
+    FSC->>FSC: receive(proposal)
+    FSC->>FSC: extract replay.Key (txID, creator, nonce, timestamp)
+    FSC->>Guard: Check(ctx, key)
```
AkramBitar: Based on my understanding, in Check we call g.cache.Add(k, ...), so we remember the request before we've validated it. Validation (signature / MSP / ACL) only runs afterwards, and nothing tells the guard to forget a request that turns out invalid.

A problem can be if validation fails (e.g. bad signature), we've already remembered the request. If the client retries the same one, Check rejects it as "already seen" for ~10 min — even though it was never actually approved.

Maybe we need validate first, remember after. What do you think?

AkramBitar (changes requested):

docs/services/network-fabric.md
adecaro: what if a malicious client tries to re-submit over and over again the same invalid transaction just to disrupt the service? If the client is honest, the client can regenerate the transaction.

AkramBitar: Ok, got it so honest client will generate a new one in this case.

AkramBitar (approved):
LGTM

adecaro:
@atharrva01 , would like to have a look at this one? It touches the endorsement part something that you also need for the EVM-related work stream you are handling 😄 🙏

atharrva01 (commented):
Went through this properly since it sits right next to the EVM endorser I'm building. The guard reads well and the tests are better than most, 11 cases in `memory_test.go` covering concurrency, eviction, TTL and a moving clock. Using `Contains` instead of `Get` is right too.

One thing I think needs a call before u merge.

**The cache gets written from unauthenticated input.** `ReceiveTx` just deserializes the session bytes, FSC's `builder.go` does no signature check at all. The guard runs at `responder.go:152` and the signature isn't verified until `responder.go:342`, so `cache.Add` happens on bytes nobody has checked.

I saw the thread with @AkramBitar and the answer that an honest client can regenerate. That holds while the party whose key gets burned is the one who submitted it. It stops holding when nothing is authenticated before the write, because then someone else can burn your key.

Concretely: the initiator sends the same signed proposal to every endorser. A malicious endorser, which a threshold policy already assumes exists, can forward it to the honest ones and burn the key before the initiator's own copy lands. Regenerating doesn't help, they can just do it again to the next one. That's a targeted denial of endorsement rather than a wasted slot.

Same cause, second effect: `MaxEntries` defaults to 100000 and the LRU drops the oldest on overflow (`expirable_lru.go:140`), so 100k unsigned junk proposals evict real keys and make them replayable again inside their window.

Would verifying the signature before `Add` work? It keeps the guard ahead of the expensive parts (MSP, ACL, translate, endorse), and one verify is cheap next to those and unavoidable anyway.

**Second, the EVM guidance in `network-ethereum.md` doesn't work yet.** It says build a `replay.Key` from txID/creator/nonce/timestamp, but my `EndorseRequest` only carries `TokenRequest`, `TMSID`, `Anchor` and `Metadata`. Only `Anchor` maps to `TxID`, so `Timestamp` would be the zero value, and I checked, that's outside the default 5 minute window. An EVM endorser following this doc would reject every request with `ErrOutOfWindow`.

Also, nonce and timestamp only mean anything if a signature covers them, otherwise a replayer just varies them and walks past the dedup. The EVM envelope isn't signed at that level today.

Happy to take the EVM side as follow-up, I'd just rather the doc not promise a shape the driver can't produce.

Two small things inline.

token/services/network/common/replay/config.go
```diff
@@ -0,0 +1,46 @@
...
+type Config struct {
+	// Backend selects the Guard implementation. Defaults to BackendMemory.
+	Backend Backend `yaml:"backend"`
+	// Window bounds how far a key's claimed Timestamp may lie from the guard's current time,
+	// in either direction, before it is rejected with ErrOutOfWindow. The window moves with
+	// the guard's clock. Window <= 0 disables the freshness check.
+	Window time.Duration `yaml:"window"`
+	// TTL is how long a seen key is remembered before it can be forgotten. Only meaningful
+	// for backends whose entries expire (e.g. BackendMemory). Must be at least 2*Window so an
+	// entry survives its entire potential-replay lifetime; backends enforce this floor.
```
atharrva01: This says backends enforce the floor, but `memory.New` doesn't, only `factory.New` clamps it (`factory.go:23`). Anyone calling `memory.New` directly, tests or a driver following the ethereum doc, silently gets `ttl < 2*window` and an entry that can be forgotten while it's still replayable. Either move the clamp into `memory.New` or reword this to say the factory enforces it.

token/services/network/fabric/endorsement/provider.go
```diff
@@ -122,13 +133,29 @@ func (l *loader) load(tmsID token2.TMSID) (Service, error) {
...
 	return tmsID.Network + tmsID.Channel + tmsID.Namespace
 }
 
+// NewReplayGuard builds the replay.Guard for tmsID from its fsc_endorsement.replay
+// configuration (see ReplayKey), falling back to replay.DefaultConfig() when the block is
+// absent or partially set. Shared by both the Fabric and FabricX endorsement loaders since
+// FabricX reuses the same fsc_endorsement config namespace.
+func NewReplayGuard(configuration tdriver.Configuration, tmsID token2.TMSID) (replay.Guard, error) {
+	replayCfg := replay.DefaultConfig()
+	_ = configuration.UnmarshalKey(ReplayKey, &replayCfg)
```
atharrva01: The error is dropped, so a malformed replay block silently falls back to defaults or a half-applied config. Someone who sets a 30 minute window and typos it runs 5 without knowing. AGENTS.md asks for errors to be handled rather than blanked, and for a security control I'd fail loudly.

atharrva01:
@adecaro gentle nudge on this one, mainly so the first point doesn't get lost since it's approved
and can go in any time.

The only thing I'd really like a call on is whether the proposal signature should be verified before
`cache.Add`. Right now the guard runs at `responder.go:152` and the signature isn't checked until
`responder.go:342`, so anything that reaches the session can write keys. The bit I'd want your read
on is that the initiator sends the same signed proposal to every endorser, so a malicious one can
forward it to the honest endorsers and burn the key before the real request lands, and regenerating
doesn't help because it can just do it again. Fine either way, I just didn't want it to merge without
someone having decided it.

The EVM doc paragraph I'll take as a follow-up PR, no need to hold this up for it. One thing that
would help me though: are you expecting the EVM envelope to carry a signed nonce and timestamp? That
changes the message format on my side, so it'd be good to know before week 6 rather than after.

The two inline ones are tiny, happy for them to be ignored or picked up whenever.

token/services/network/common/replay/config.go
adecaro: Fixed — `memory.New` now enforces the `TTL >= 2*Window` floor itself (raises `TTL` when it's set too low), instead of relying on `factory.New` to clamp it first. Updated the doc comment on `TTL` to say the floor holds however the backend is constructed, not just when going through `factory.New`. Added `TestNew_TTLFloorEnforcedDirectly` in `memory_test.go` to cover calling `memory.New` directly with a too-short TTL.

token/services/network/fabric/endorsement/provider.go
adecaro: Fixed — `NewReplayGuard` now checks the `UnmarshalKey` error and returns it wrapped instead of silently continuing with whatever `replay.DefaultConfig()` already had on `cfg`. Added `TestNewReplayGuard_MalformedBlockReturnsError` in `replay_guard_test.go` covering the decode-failure path.

adecaro:
Thanks for the thorough pass, @atharrva01 — the co-endorser-forwarding scenario is a sharper description of the risk than what we had.

**On the call you asked for: yes, verifying the signature before `Add` is the fix.** `ResponderView.Call` now runs a new `verifyProposalSignature` step — checks creator/proposal/signature are non-empty and that the signature verifies against the claimed creator via the MSP manager — *before* the replay guard is ever consulted. `validateProposal` (MSP `IsValid` + ACL) still runs after the guard, since those checks aren't needed to keep the guard safe from unauthenticated input. This directly closes both effects you described: an unsigned/invalid proposal can no longer occupy or evict a dedup slot, and the 100k-entry LRU can no longer be flooded with garbage that never had a valid signature.

One thing this does *not* fully close, and I want to be upfront about it rather than overstate the fix: the residual co-endorser race. A validly-forwarded proposal — one the malicious co-endorser received legitimately and didn't tamper with — still has a valid signature, so it can still beat the initiator's own copy to the guard. Fully closing that would mean binding a session to the identity that opened it at the transport layer, which is out of scope here. I've called this out explicitly in a code comment on `verifyProposalSignature` and in the Replay Detection section of `docs/services/network-fabric.md`, so it's a documented, accepted tradeoff rather than a silent gap.

Also fixed the two inline items (TTL floor now enforced in `memory.New` itself, and the dropped `UnmarshalKey` error in `NewReplayGuard`) — replied on each thread.

**On the EVM doc mismatch** — agreed, and since you said you'd take that as a follow-up I've left `network-ethereum.md` untouched here rather than guess at a shape that matches your `EndorseRequest`. Happy to coordinate once you're working on the EVM endorser if it'd help to align the `replay.Key` derivation with what's actually signed there.

AkramBitar (changes requested):
@adecaro 

Is that PR ready to merge?

Regards,
Akram

AkramBitar:
> @adecaro
> 
> Is that PR ready to merge?
> 
> Regards, Akram

Hello @adecaro 

Is there any update with PR?

Regards,
Akram

adecaro:
closing this. It needs to be reworked.

PR #1970, opened 2026-07-21
feat(tokens): prepare balance statements by argument shape

sid200727:
## Summary
Per @adecaro's request, benchmark and implementation together in one PR, following the same prepared-statement pattern as #1889 and #1929.

## Changes
-> Extracted `buildBalanceQuery` from `balance()` (pure refactor, no behavior change, verified via existing `TestTokens` suite)
-> Added prepared-statement caching to `balance()` (backing `Balance()`), reusing the existing `PreparedStmtHolder[K]` rather than duplicating the caching logic
-> Keyed by the same argument shape as `UnspentTokensIteratorBy`/`SpendableTokensIteratorBy` via the existing `unspentTokensStmtKey`, `balance()` has a single caller (`Balance()`) that always sets only `WalletID`/`TokenType`, so the shape encoding applies directly
-> Adapted from `QueryRowContext` to `QueryContext` + manual `rows.Next()`/`Scan()`, since `PreparedStmtHolder.Execute` returns `*sql.Rows`
-> Added `RunBalancePreparedComparison` benchmark, wired into the Postgres suite as `BenchmarkBalancePreparedComparison`
-> Added a DB-free reuse test (`go-sqlmock`) covering statement caching for this query

## Results (median of 2 runs, 1000 tokens, wallet0/GOLD, 4 workers, 5s window)

|---|---|---|---|
| Throughput | ~4,591 ops/s | ~6,720 ops/s | +46% |
| P50 latency | ~523µs | ~455µs | -13% |
| P99 latency | ~6.24ms | ~1.78ms | -71% |
| Allocs/op | ~120 | 19 | -84% |

Biggest allocs/op win of the three queries done so far likely because the `SUM(...)` aggregate + join means query-builder overhead is a larger relative share of total work.

## Testing
Full `TestTokens` suite (sqlite + postgres) unchanged and green. Benchmark run on laptop with Docker Desktop, not an isolated bench.

## Notes for the Reviewer
This completes the shape-based prepared-statement work across all three identified candidates (`UnspentTokensIteratorBy` #1889, `SpendableTokensIteratorBy` #1929, `balance` here).

### DIFF
--- a/token/services/storage/db/sql/common/tokens.go
+++ b/token/services/storage/db/sql/common/tokens.go
@@ -60,6 +60,12 @@ type TokenStore struct {
 	// spendableTokensStmts caches prepared statements for
 	// SpendableTokensIteratorBy, keyed the same way. See #1919.
 	spendableTokensStmts PreparedStmtHolder[string]
+
+	// balanceStmts caches prepared statements for balance, keyed the same
+	// way. balance() has a single caller (Balance()) that always sets only
+	// WalletID/TokenType, so unspentTokensStmtKey's shape encoding applies
+	// here too.
+	balanceStmts PreparedStmtHolder[string]
 }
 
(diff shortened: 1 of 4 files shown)


adecaro (approved):
LGTM

PR #1980, opened 2026-07-22
feat(transactions): prepare GetStatus and GetTokenRequest statements

sid200727:
## Summary
Per maintainer's request to assess remaining candidates, targeted the two most-used queries found: `GetStatus` and `GetTokenRequest`, both in `transactions.go`, both with 11 call sites, the widest of anything checked (`QueryTokenDetails` was ruled out, only 1 call site with too many optional-field shapes to cache cleanly).

## Why these two
Unlike the token-store queries (#1889/#1929/#1970), both have exactly **one query shape ever** (`tx_id = ?`, always present) no shape-key logic needed, just a single cached prepared statement per query. `GetStatus` specifically backs the finality-polling loop in `ttx/finality.go`.

## Changes
-> Added prepared-statement caching to `GetStatus` and `GetTokenRequest`, reusing the existing `PreparedStmtHolder[K]`
-> `GetTokenRequest`'s not-found behavior (`nil, nil`) preserved per the documented contract in `GetTokenRequests`' own comment
-> **Bug fix in `PreparedStmtHolder.Execute`**: when the context is already cancelled/expired, the fallback path was retrying via an unprepared query, which fails identically but wraps the error differently, breaking callers checking for the specific cancellation error via `errors.Is`. Now returns the original error directly when `ctx.Err() != nil`, without retrying.
-> Fixed two existing tests (`TestGetStatus`, `TestGetStatusContextCancelled`) that were silently passing via the fallback path without ever exercising the prepared path, now correctly assert `ExpectPrepare`.
-> Added benchmarks for both queries, plus DB-free reuse tests (`go-sqlmock`) covering statement caching.

## Results (median of 2 runs each, 4 workers, 5s window)

| Query | Throughput | P99 latency | Allocs/op |
|---|---|---|---|
| GetStatus | +15% | -66% | -47% |
| GetTokenRequest | +14% | -65% | -45% |

## Testing
Full test suite (common + sqlite + postgres) green.

## Notes for the Reviewer
The context-cancellation bug affects all four prepared queries done so far (`UnspentTokensIteratorBy`, `SpendableTokensIteratorBy`, `balance`, and now these two), since it's in the shared holder worth a careful look given it's a correctness fix, not just a new feature.

### DIFF
--- a/token/services/storage/db/sql/common/prepared_stmt_holder.go
+++ b/token/services/storage/db/sql/common/prepared_stmt_holder.go
@@ -61,9 +61,16 @@ func (h *preparedStmtHolder[K]) Execute(ctx context.Context, db *sql.DB, key K, 
 	if stmt, err := h.getOrPrepare(ctx, db, key, query); err == nil {
 		if rows, qErr := stmt.QueryContext(ctx, args...); qErr == nil {
 			return rows, nil
+		} else if ctx.Err() != nil {
+			// The context is already done - retrying via the fallback path
+			// would fail identically and only obscure the original
+			// cancellation/deadline error for callers checking its type.
+			return nil, qErr
 		} else {
 			logger.Warnf("prepared statement query failed for key [%v], falling back to unprepared query: %s", key, qErr)
 		}
+	} else if ctx.Err() != nil {
+		return nil, err
 	} else {
 		logger.Warnf("failed preparing statement for key [%v], falling back to unprepared query: %s", key, err)
 	}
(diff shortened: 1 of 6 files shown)


token/services/storage/db/sql/common/prepared_stmt_holder.go
```diff
@@ -61,9 +61,16 @@ func (h *preparedStmtHolder[K]) Execute(ctx context.Context, db *sql.DB, key K,
 	if stmt, err := h.getOrPrepare(ctx, db, key, query); err == nil {
 		if rows, qErr := stmt.QueryContext(ctx, args...); qErr == nil {
 			return rows, nil
+		} else if ctx.Err() != nil {
+			// The context is already done - retrying via the fallback path
```
adecaro, line 65: nice catch

adecaro (approved):
LGTM

token/services/storage/db/sql/common/prepared_stmt_holder.go
sid200727: Thank you! 🙏

PR #1981, opened 2026-07-22
perf(selector): shard simple selector's in-memory locker by owner and jitter retries

### LINKED ISSUE #1875: perf: simple selector serializes all token locking behind one global mutex, then retries in lockstep
## Problem

Two compounding issues in `token/services/selector/simple`:

1. **Single global lock for all tokens.** The in-memory `locker` (`token/services/selector/simple/inmemory/locker.go`) guards its entire `map[token2.ID]*lockEntry` with one `sync.RWMutex`. `Lock`, `UnlockIDs`, `UnlockByTxID`, `IsLocked`, and the periodic background `scan` all take this same mutex. Two transactions locking tokens belonging to two different, unrelated owners still serialize on this one mutex, even though they can never actually contend for the same token (a token has exactly one owner). Compare with `sherdlock`, whose `Lock` goes straight to a DB insert keyed by `(tx_id, idx)` (`token/services/storage/db/sql/common/tokenlock.go`) — concurrency is bounded by the database's own per-row locking, not a single process-wide mutex.
2. **Retry backoff has no jitter.** `selector.selectByID` (`token/services/selector/simple/selector.go`) retries on a fixed `time.Sleep(s.timeout)` (default 5s, `defaultRetryInterval` in `token/services/selector/config/driver.go`). When several transactions race for the same funds and lose, they all sleep for the same fixed duration and wake up at the same moment to retry — a thundering herd that hits the global lock from problem 1 all at once. `sherdlock`'s equivalent retry loop (`StubbornSelector.Select` in `token/services/selector/sherdlock/selector.go`) already randomizes the backoff with `rand.Int64N(backoffInterval)`; the simple selector has no equivalent.

Together, these mean contention gets worse than it needs to be: unrelated owners block each other on one lock, and retries pile back onto that same lock in sync.

## Suggested fix

1. Replace the single global mutex with per-owner locking. `selector.Select`/`selectByID` already operate within a single `ownerFilter` (`selector.go:46-51,70`), so the owner is known at the point `s.locker.Lock` is called (`selector.go:117`) and can be threaded through to the locker. Keep a dynamically-sized registry of locks keyed by owner (e.g. a small lazily-populated map guarded by its own mutex, not a fixed-size shard array) instead of one map-wide lock. This also means `UnlockByTxID` — which today scans the whole map under the global lock — only needs to touch the one owner's lock in the common case, since a given transaction's `Select` calls are scoped to a single owner.
2. Add randomized jitter to the retry backoff in `selector.go`, following the same pattern already used in `sherdlock/selector.go` (`rand.Int64N` over the configured interval instead of a fixed sleep).

## Test plan

- Existing `simple` selector/locker unit tests should still pass.
- Add a test with two concurrent `Select` calls for two different owners and assert they don't serialize on a shared lock (e.g. via timing or a lock-acquisition counter).
- Add a test asserting retry sleep durations are randomized rather than constant across repeated `SelectorSufficientButLockedFunds` retries.

hstarorg:
The simple selector's in-memory locker guarded all tokens with one global mutex, so transactions locking tokens of unrelated owners serialized on a single lock — worst when a reclaim held it across a status lookup, freezing every other owner's selection.

Changes:
- The locker keeps a lazily-created shard per owner; `Lock`/`UnlockIDs` take the owner (already in scope in `selectByID`). `UnlockByTxID`/`IsLocked` keep their signatures and visit each shard briefly. The empty owner is a valid shard key for callers without owner context.
- The background scan now snapshots entries per shard and resolves statuses without holding any lock, re-validating before deleting (TxID match + LastAccess re-check for Confirmed evictions), so status lookups no longer block an owner's operations.
- The fixed retry sleep is replaced with a random backoff in [0, timeout), the same jittering pattern as sherdlock's selector.

Breaking: `simple.Locker`'s `Lock`/`UnlockIDs` (and the mirrored interfaces in `sherdlock/inmemory` and `testutils`) gain an owner parameter; the single-lock-per-token guarantee is now scoped to the owner namespace. `inmemory.NewLocker`'s signature is unchanged.

Benchmarks (`locker_bench_test.go`, Apple M4 Pro): 8 workers reclaiming under a 100µs status lookup run 8.1x faster with per-owner shards (linear scaling); pure lock/unlock under 32 workers is 2.3–2.5x faster. Tests cover shard independence (deterministic, no timing), cross-shard UnlockByTxID, empty-owner fallback, same-owner race safety, scanner TOCTOU protection, and backoff jitter; all with `-race` clean.

Closes #1875

### DIFF
--- a/token/services/selector/benchmark_test.go
+++ b/token/services/selector/benchmark_test.go
@@ -27,8 +27,8 @@ import (
 type WalletIDByRawIdentityFunc func(rawIdentity []byte) string
 
 type Locker interface {
-	Lock(ctx context.Context, id *token2.ID, txID string, reclaim bool) (string, error)
-	UnlockIDs(ctx context.Context, ids ...*token2.ID) []*token2.ID
+	Lock(ctx context.Context, owner string, id *token2.ID, txID string, reclaim bool) (string, error)
+	UnlockIDs(ctx context.Context, owner string, ids ...*token2.ID) []*token2.ID
 	UnlockByTxID(ctx context.Context, txID string)
 	IsLocked(id *token2.ID) bool
 }
@@ -43,9 +43,9 @@ func (s *extendedSelector) Select(ctx context.Context, ownerFilter token.OwnerFi
 }
 func (s *extendedSelector) Close() error { return s.Selector.Close() }
 
-func (s *extendedSelector) Unselect(id ...*token2.ID) {
+func (s *extendedSelector) Unselect(owner string, id ...*token2.ID) {
 	if s.Lock != nil {
-		s.Lock.UnlockIDs(context.Background(), id...)
+		s.Lock.UnlockIDs(context.Background(), owner, id...)
 	}
 }
 
@@ -72,7 +72,7 @@ func BenchmarkSelectorSingle(b *testing.B) {
 				wg.Add(1)
 				go func(ids []*token2.ID) {
 					defer wg.Done()
-					s.selector.Unselect(ids...)
+					s.selector.Unselect(s.filter.ID(), ids...)
 				}(ids)
 			}
 		})
@@ -110,7 +110,7 @@ func BenchmarkSelectorParallel(b *testing.B) {
 					wg.Add(1)
 					go func(ids []*token2.ID) {
 						defer wg.Done()
-						s.selector.Unselect(ids...)
+						s.selector.Unselect(s.filter.ID(), ids...)
 					}(ids)
 				}
 			})
@@ -199,7 +199,7 @@ type LockerProviderFunction func() selector.Locker
 
 type ExtendedSelector interface {
 	token.Selector
-	Unselect(id ...*token2.ID)
+	Unselect(owner string, id ...*token2.ID)
 }
 
 type MockTokenIterator struct {
--- a/token/services/selector/sherdlock/inmemory/locker.go
+++ b/token/services/selector/sherdlock/inmemory/locker.go
@@ -15,7 +15,7 @@ import (
 )
 
 type Locker interface {
-	Lock(ctx context.Context, id *token.ID, txID string, reclaim bool) (string, error)
+	Lock(ctx context.Context, owner string, id *token.ID, txID string, reclaim bool) (string, error)
 	UnlockByTxID(ctx context.Context, txID string)
 }
 
@@ -32,7 +32,8 @@ func NewLocker(l Locker) *locker {
 }
 
 func (l *locker) Lock(ctx context.Context, tokenID *token.ID, consumerTxID transaction.ID) error {
-	_, err := l.Locker.Lock(ctx, tokenID, consumerTxID, false)
+	// this adapter has no owner context; the empty owner shares one default shard
+	_, err := l.Locker.Lock(ctx, "", tokenID, consumerTxID, false)
 
 	return err
 }
--- a/token/services/selector/simple/inmemory/locker.go
+++ b/token/services/selector/simple/inmemory/locker.go
@@ -47,10 +47,18 @@ func (l lockEntry) String() string {
 	return fmt.Sprintf("[[%s] since [%s], last access [%s]]", l.TxID, l.Created, l.LastAccess)
 }
 
+// shard holds the lock state for the tokens of a single owner. A token has
+// exactly one owner, so operations on different owners can never contend
+// for the same token and each shard can be guarded independently.
+type shard struct {
+	mu     sync.Mutex
+	locked map[token2.ID]*lockEntry
+}
+
 type locker struct {
 	ttxdb                  TXStatusProvider
-	lock                   *sync.RWMutex
-	locked                 map[token2.ID]*lockEntry
+	shardsMu               sync.RWMutex
+	shards                 map[string]*shard
 	sleepTimeout           time.Duration
 	validTxEvictionTimeout time.Duration
 	cancel                 context.CancelFunc
@@ -63,8 +71,7 @@ func NewLocker(ttxdb TXStatusProvider, timeout time.Duration, validTxEvictionTim
 	r := &locker{
 		ttxdb:                  ttxdb,
 		sleepTimeout:           timeout,
-		lock:                   &sync.RWMutex{},
-		locked:                 map[token2.ID]*lockEntry{},
+		shards:                 map[string]*shard{},
 		validTxEvictionTimeout: validTxEvictionTimeout,
 		cancel:                 cancel,
 		scanDone:               make(chan struct{}),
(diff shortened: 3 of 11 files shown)


adecaro:

I think this change makes sense for the `simple selector`. It does not in the case of `sherdlock selector` becasue we use the db there. Correct? 

I'm wondering if could be of help to store the `owner` in the tokens lock table as well. If needed, it will go into another PR.

On a side note, do you have feedback about additional requirements for token selectors in general? cc @EvanYan1024

AkramBitar:
Hi @hstarorg , thanks for submitting this 

One minor issue that I though about it, based on my understanding shards are created per owner but never pruned, so the shards map only grows. It's bounded by the number of wallet IDs, so this is fine in practice (~175 B/owner). Worth keeping in mind only if high-cardinality owners ever become possible — in that case we could drop empty shards in the scanner, which already visits every shard.
The scanner already loops over every shard each cycle (in scanShards, via allShards()) to snapshot entries and delete stale ones. So I think pruning is basically free . After a shard's delete phase, if its locked map is empty, remove it from shards.

adecaro:
good point @AkramBitar . In case of high cardinality wallets, the `simple` selector might not be the right one anyway because it loads all the tokens in memory. So, we should make sure to alert the developers that this selector should be used only when memory is not a constraint or when there is a limited number of wallets.

Anyhow, regarding the specific issue, I would suggest to investigate the use of `weak.Pointer`. Not sure it would work in this case but it might be worth a double check.

@hstarorg , I have restarted the unit-tests a few time. I think there is an issue there.

EvanYan1024:
Hi @adecaro, thanks for the cc. Yes, my understanding matches yours: this optimization only targets the simple selector, whose lock state is in-process and was guarded by one global mutex — sharding it by owner removes in-process contention between unrelated owners. The DB-backed sherdlock delegates locking to the database, where the `(tx_id, idx)` primary key already gives token-level mutual exclusion, so there is no process-wide lock to shard. The changes this PR makes under `sherdlock/inmemory` are signature adaptation only: the adapter passes an empty owner and behavior is unchanged on both paths.

On storing the owner in the token lock table: it could help later with owner-scoped diagnostics ("what does this wallet have locked right now"), owner-scoped cleanup, and owner-labelled metrics. It does not seem required for the locking path itself — the token ID already guarantees lock uniqueness and `consumer_tx_id` is already indexed.

As for additional requirements for token selectors in general — two things we actually found missing when running token selection in our test environments under sustained load:

1. **Amount-aware selection, orthogonal to the locking backend.** This applies to both selectors: candidates are consumed in iterator order, so the number of inputs a transaction ends up with is uncontrolled — and with the ZK driver every extra input is real prover cost. Amount-blind greedy selection also fragments the wallet over time (every payment leaves change, and dust is never preferentially swept), which in sherdlock's case additionally means more candidates to lock, one DB write each. sherdlock addresses where tokens are enumerated from, not how they are chosen. We experimented with a custom amount-bucketed best-fit index (configurable boundaries) in our dev environment; the same strategy could sit on top of either locker — e.g. as amount-range queries against the token store. Selection strategy feels like it should be a pluggable dimension, independent of which locker is used.

2. **Batch lock acquisition, and letting the lock table live on its own pool.** The `TokenLocker` interface is already narrow enough to swap implementations — that part is good. The friction is elsewhere: (a) there is no batch lock primitive, so locking a selection of N tokens costs N serial round trips whatever the backend — and multi-input transactions are routine: multi-recipient payouts and dust-consolidation sweeps easily reach tens of inputs. Unlocking is already batched (`UnlockByTxID`); a `TryLockBatch` (multi-row `INSERT .. ON CONFLICT` in SQL) would restore the symmetry; and (b) the SQL store pins the lock table to the same database that carries the vault/commit load: the lease cleanup joins `token_locks` against `token_requests` in one query, and the schema declares a foreign key from `token_locks` into the tokens table. In our stress tests that database and its connection pool were consistently the scarcest resource. Relaxing those two couplings — a status lookup plus a delete instead of the join, and making the FK optional — would let deployments point the lock table at its own persistence without forking.

hstarorg:
> Hi @hstarorg , thanks for submitting this 🙏
> 
> I think this change makes sense for the `simple selector`. It does not in the case of `sherdlock selector` becasue we use the db there. Correct?
> 
> I'm wondering if could be of help to store the `owner` in the tokens lock table as well. If needed, it will go into another PR.
> 
> On a side note, do you have feedback about additional requirements for token selectors in general? cc @EvanYan1024

Correct — this only optimizes the simple selector; sherdlock locks in the database and is not affected (the `sherdlock/inmemory` changes are signature adaptation only, behavior unchanged).

hstarorg:
> good point @AkramBitar . In case of high cardinality wallets, the `simple` selector might not be the right one anyway because it loads all the tokens in memory. So, we should make sure to alert the developers that this selector should be used only when memory is not a constraint or when there is a limited number of wallets.
> 
> Anyhow, regarding the specific issue, I would suggest to investigate the use of `weak.Pointer`. Not sure it would work in this case but it might be worth a double check.
> 
> @hstarorg , I have restarted the unit-tests a few time. I think there is an issue there.

Sorry, I updated the PR and fixed the CI issue — it was a deadlock in the new test itself, not in the implementation.

hstarorg:
> Hi @hstarorg , thanks for submitting this
> 
> One minor issue that I though about it, based on my understanding shards are created per owner but never pruned, so the shards map only grows. It's bounded by the number of wallet IDs, so this is fine in practice (~175 B/owner). Worth keeping in mind only if high-cardinality owners ever become possible — in that case we could drop empty shards in the scanner, which already visits every shard. The scanner already loops over every shard each cycle (in scanShards, via allShards()) to snapshot entries and delete stale ones. So I think pruning is basically free . After a shard's delete phase, if its locked map is empty, remove it from shards.

Agreed — pruning empty shards in the scanner's delete phase is basically free. Will implement shortly.

adecaro:
> Hi @adecaro, thanks for the cc. Yes, my understanding matches yours: this optimization only targets the simple selector, whose lock state is in-process and was guarded by one global mutex — sharding it by owner removes in-process contention between unrelated owners. The DB-backed sherdlock delegates locking to the database, where the `(tx_id, idx)` primary key already gives token-level mutual exclusion, so there is no process-wide lock to shard. The changes this PR makes under `sherdlock/inmemory` are signature adaptation only: the adapter passes an empty owner and behavior is unchanged on both paths.
> 
> On storing the owner in the token lock table: it could help later with owner-scoped diagnostics ("what does this wallet have locked right now"), owner-scoped cleanup, and owner-labelled metrics. It does not seem required for the locking path itself — the token ID already guarantees lock uniqueness and `consumer_tx_id` is already indexed.
> 
> As for additional requirements for token selectors in general — two things we actually found missing when running token selection in our test environments under sustained load:
> 
> 1. **Amount-aware selection, orthogonal to the locking backend.** This applies to both selectors: candidates are consumed in iterator order, so the number of inputs a transaction ends up with is uncontrolled — and with the ZK driver every extra input is real prover cost. Amount-blind greedy selection also fragments the wallet over time (every payment leaves change, and dust is never preferentially swept), which in sherdlock's case additionally means more candidates to lock, one DB write each. sherdlock addresses where tokens are enumerated from, not how they are chosen. We experimented with a custom amount-bucketed best-fit index (configurable boundaries) in our dev environment; the same strategy could sit on top of either locker — e.g. as amount-range queries against the token store. Selection strategy feels like it should be a pluggable dimension, independent of which locker is used.
> 2. **Batch lock acquisition, and letting the lock table live on its own pool.** The `TokenLocker` interface is already narrow enough to swap implementations — that part is good. The friction is elsewhere: (a) there is no batch lock primitive, so locking a selection of N tokens costs N serial round trips whatever the backend — and multi-input transactions are routine: multi-recipient payouts and dust-consolidation sweeps easily reach tens of inputs. Unlocking is already batched (`UnlockByTxID`); a `TryLockBatch` (multi-row `INSERT .. ON CONFLICT` in SQL) would restore the symmetry; and (b) the SQL store pins the lock table to the same database that carries the vault/commit load: the lease cleanup joins `token_locks` against `token_requests` in one query, and the schema declares a foreign key from `token_locks` into the tokens table. In our stress tests that database and its connection pool were consistently the scarcest resource. Relaxing those two couplings — a status lookup plus a delete instead of the join, and making the FK optional — would let deployments point the lock table at its own persistence without forking.

@EvanYan1024 thanks for jumping in with your assessment and your requirements. Really appreciate. 
Please, feel free to submit changes that might be of benefit for the entire community. It would be great 🙏

adecaro (approved):
LGTM 🙏

adecaro:
Hi @hstarorg , please, rebase your branch. I cannot do it from the UI 😞

hstarorg:
> Hi @hstarorg , please, rebase your branch. I cannot do it from the UI 😞

Rebased. Thank you.

PR #1982, opened 2026-07-22
feat(sherdlock): coordinate token-lock cleanup leadership across replicas

### LINKED ISSUE #1798: Prevent Concurrent TokenLocks Cleanup Across Replicas
Currently, each replica runs the TokenLocks cleanup function independently. We should investigate whether we can introduce a mechanism so that only a single replica performs this task at a time, in order to avoid contention

sid200727:
## Summary
Implements #1798: currently every replica runs the token-locks cleanup tick independently, causing wasted, overlapping work, all replicas race to delete the same expired rows at the same time. This adds a per-tick leadership check so only one replica performs cleanup, following the exact same pattern already established for keystore cleanup (`AcquireCleanupLeadership`).

## Changes
-> Added `AcquireCleanupLeadership` to both `driver.TokenLockStore` and `sherdlock.Locker` interfaces
-> Added a shared `driver.NoopCleanupLeadership` for non-distributed backends (sqlite, in-memory), which always grant leadership locally, no failover code needed since there's only ever one instance
-> Wired the real Postgres advisory-lock implementation via the existing `NewCleanupLeaderFactory()`, using a lock id independent of the schema-creation lock id already used by `TokenLockStore.lockID` (deliberately separate to avoid any collision between the two purposes)
-> `manager.go`'s `cleaner()` now acquires leadership before each `Cleanup` tick, skips the tick if not acquired, and releases leadership after, matching the design: per-tick election, no permanent leader, no failover code, advisory lock auto-releases if the winner crashes
-> Regenerated the `FakeLocker` mock; extended the hand-written `mockLocker` test double, defaulting to always-granted so all pre-existing tests are unaffected
-> Added two new subtests to `TestManager_Cleaner`: skip-on-no-leadership, and leadership-released-after-cleanup

## Testing
Full test suite green (`selector`, `sqlite`, `postgres`).

## Notes for the Reviewer
Followed the 4-step plan from the design discussion on #1798. Happy to add a Postgres integration test exercising two concurrent "replicas" if useful on top of the unit-level coverage already here, let me know if that's wanted before merge.

### DIFF
--- a/token/services/selector/sherdlock/manager.go
+++ b/token/services/selector/sherdlock/manager.go
@@ -91,15 +91,38 @@ func (m *Manager) cleaner(ctx context.Context) {
 	for {
 		select {
 		case <-ticker.C:
-			logger.DebugfContext(ctx, "release token locks older than [%s]", m.leaseExpiry)
-			if err := m.locker.Cleanup(ctx, m.leaseExpiry); err != nil {
-				logger.Errorf("failed to release token locks: [%s]", err)
-			}
+			m.runCleanupTick(ctx)
 		case <-ctx.Done():
 			logger.Debugf("cleaner stopping")
 
 			return
 		}
+	}
+}
+
+// runCleanupTick acquires cleanup leadership for this tick so only one
+// replica runs Cleanup at a time; other replicas skip the tick. See #1798.
+func (m *Manager) runCleanupTick(ctx context.Context) {
+	leadership, acquired, err := m.locker.AcquireCleanupLeadership(ctx)
+	if err != nil {
+		logger.Errorf("failed to acquire cleanup leadership: [%s]", err)
+
+		return
+	}
+	if !acquired {
+		logger.DebugfContext(ctx, "cleanup leadership not acquired, skipping tick")
+
+		return
+	}
+	defer func() {
+		if err := leadership.Close(); err != nil {
+			logger.Warnf("failed to release cleanup leadership: [%s]", err)
+		}
+	}()
+
+	logger.DebugfContext(ctx, "release token locks older than [%s]", m.leaseExpiry)
+	if err := m.locker.Cleanup(ctx, m.leaseExpiry); err != nil {
+		logger.Errorf("failed to release token locks: [%s]", err)
 	}
 }
 
--- a/token/services/storage/db/sql/postgres/tokenlock.go
+++ b/token/services/storage/db/sql/postgres/tokenlock.go
@@ -32,6 +32,16 @@ type TokenLockStore struct {
 	writeDB *sql.DB
 	ci      common3.CondInterpreter
 	lockID  int64
+
+	// cleanupLockID is derived from the fully-qualified table name (not the
+	// prefix alone, which is not unique per TMS - see review discussion on
+	// #1982), so it is unique per TMS - distinct from lockID (schema-creation
+	// lock) and from other TMSes' cleanup locks on the same node. A single global constant here
+	// was a real bug: it caused every TMS on a node to compete for the
+	// exact same advisory lock, so only one TMS across the whole fleet
+	// ever won cleanup on any tick. See #1798.
+	cleanupLockID        int64
+	cleanupLeaderFactory func(context.Context, *sql.DB, int64) (driver.CleanupLeadership, bool, error)
 }
 
 // GetSchema overrides the base GetSchema to prefix with advisory lock
@@ -55,11 +65,24 @@ func NewTokenLockStore(dbs *common2.RWDB, tableNames common5.TableNames) (*Token
 	}
 
 	return &TokenLockStore{
-		TokenLockStore: tldb,
-		writeDB:        dbs.WriteDB,
-		ci:             ci,
-		lockID:         createTableLockID("tokenlock"),
+		TokenLockStore:       tldb,
+		writeDB:              dbs.WriteDB,
+		ci:                   ci,
+		lockID:               createTableLockID(tableNames.TokenLocks),
+		cleanupLockID:        createTableLockID(tableNames.TokenLocks + "_cleanup"),
+		cleanupLeaderFactory: NewCleanupLeaderFactory(),
 	}, nil
+}
+
+// AcquireCleanupLeadership attempts to acquire a Postgres advisory lock so
+// only one replica runs Cleanup per tick for this TMS; others skip the tick
+// and release no held resources, so contention is limited to the acquire
+// attempt itself. The lock id and factory are fixed at construction. Note
+// the winner holds a dedicated connection off writeDB for the tick's
+// duration (see NewAdvisoryLock), so writeDB's connection pool needs at
+// least one spare connection beyond normal write traffic. See #1798.
+func (db *TokenLockStore) AcquireCleanupLeadership(ctx context.Context) (driver.CleanupLeadership, bool, error) {
+	return db.cleanupLeaderFactory(ctx, db.writeDB, db.cleanupLockID)
 }
 
 // Cleanup removes stale token locks that have expired.
--- a/token/services/storage/db/sql/postgres/tokenlock_leadership_test.go
+++ b/token/services/storage/db/sql/postgres/tokenlock_leadership_test.go
@@ -1,0 +1,165 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package postgres
+
+import (
+	"context"
+	"database/sql"
+	"sync"
+	"testing"
+
+	sqlcommon "github.com/LFDT-Panurus/panurus/token/services/storage/db/sql/common"
+	common2 "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/common"
+	fscpostgres "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/sql/postgres"
+	_ "github.com/jackc/pgx/v5/stdlib"
+	"github.com/stretchr/testify/require"
+)
+
+// TestTokenLockStore_AcquireCleanupLeadership_ConcurrentReplicas verifies
+// that when two "replicas" (independent TokenLockStore instances for the
+// same TMS, sharing the underlying connection pool but each acquiring its
+// own session via AcquireCleanupLeadership) race for the same cleanup lock,
+// exactly one wins. Uses the real public constructor throughout, so the
+// cleanup lock id is derived internally exactly as it is in production
+// (from the table prefix), not injected by the test. See #1798.
+func TestTokenLockStore_AcquireCleanupLeadership_ConcurrentReplicas(t *testing.T) {
+	cfg := fscpostgres.DefaultConfig(fscpostgres.WithDBName("test-cleanup-leadership"))
+	terminate, _, err := fscpostgres.StartPostgres(t.Context(), cfg, nil)
+	if err != nil {
+		t.Skipf("postgres not available: %v", err)
+	}
+	t.Cleanup(terminate)
+
+	db, err := sql.Open("pgx", cfg.DataSource())
+	require.NoError(t, err)
+	t.Cleanup(func() { _ = db.Close() })
+
+	tables, err := sqlcommon.GetTableNames("")
+	require.NoError(t, err)
+	rwdb := &common2.RWDB{ReadDB: db, WriteDB: db}
+
+	// TokenLockStore's schema has a foreign key to the Tokens table, so the
+	// base token store schema must exist first.
+	tokenStore, err := sqlcommon.NewTokenStoreWithNotifier(db, db, tables, NewConditionInterpreter(), nil)
+	require.NoError(t, err)
+	require.NoError(t, tokenStore.CreateSchema())
+
+	// Two independent TokenLockStore instances for the same TMS (same table
+	// prefix), simulating two replicas. They share the connection pool, but
+	// AcquireCleanupLeadership takes a dedicated connection (session)
+	// internally, so this exercises real pg_try_advisory_lock contention,
+	// not just two calls on one session.
+	replicaA, err := NewTokenLockStore(rwdb, tables)
+	require.NoError(t, err)
+	replicaB, err := NewTokenLockStore(rwdb, tables)
+	require.NoError(t, err)
+	require.NoError(t, replicaA.CreateSchema())
+
+	require.Equal(t, replicaA.cleanupLockID, replicaB.cleanupLockID,
+		"replicas for the same TMS must derive the same cleanup lock id")
+
+	var wg sync.WaitGroup
+	results := make([]bool, 2)
+	leaderships := make([]interface{ Close() error }, 2)
+	errs := make([]error, 2)
+
+	acquire := func(i int, store *TokenLockStore) {
+		defer wg.Done()
+		l, acquired, err := store.AcquireCleanupLeadership(context.Background())
+		results[i] = acquired
+		errs[i] = err
+		if acquired {
+			leaderships[i] = l
+		}
+	}
+
+	wg.Add(2)
+	go acquire(0, replicaA)
+	go acquire(1, replicaB)
+	wg.Wait()
+
+	require.NoError(t, errs[0])
+	require.NoError(t, errs[1])
+	require.NotEqual(t, results[0], results[1], "exactly one replica should acquire leadership, got: A=%v B=%v", results[0], results[1])
+
+	for _, l := range leaderships {
+		if l != nil {
+			require.NoError(t, l.Close())
+		}
+	}
+
+	l3, acquired3, err := replicaA.AcquireCleanupLeadership(context.Background())
+	require.NoError(t, err)
+	require.True(t, acquired3, "leadership should be acquirable again after release")
+	require.NoError(t, l3.Close())
+}
+
+// TestTokenLockStore_AcquireCleanupLeadership_DistinctTMS verifies that two
+// TokenLockStores for *different* TMSes (different table prefixes) derive
+// different cleanup lock ids, and so do not contend with each other at all
+// - this is the exact bug flagged in review: a single global lock id would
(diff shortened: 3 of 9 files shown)


token/services/selector/sherdlock/manager.go
```diff
@@ -103,6 +113,32 @@ func (m *Manager) cleaner(ctx context.Context) {
 	}
 }
 
+// runCleanupTick acquires cleanup leadership for this tick so only one
+// replica runs Cleanup at a time; other replicas skip the tick. See #1798.
```
AkramBitar, line 104: What happens if the leadership replica failed while deleting?

AkramBitar (commented):
LGTM.
See my comment

AkramBitar:
Hello @sid200727,
Thanks a lot for submitting this PR. I added one comment. 
Regarding the note in the description "Notes for the Reviewer"
It will be great to add such a test as described the question how much effort to add this test.

Regards,
Akram

token/services/selector/sherdlock/manager.go
sid200727: Hi @AkramBitar,
Good question. Two things guarantee this is safe:
1. **Leadership auto-releases on crash.** The advisory lock is session-scoped, held on a dedicated `*sql.Conn` (`AdvisoryLock.conn`) for the duration of the tick. If the replica holding leadership crashes, its TCP connection to Postgres dies too, Postgres itself detects the dead backend connection and automatically releases the advisory lock (this is the whole point of session-scoped locks, no lease/TTL needed on our side). The next replica's tick can then acquire it and proceed normally.

2. **The delete itself can't be left half-done.** `Cleanup()`'s deletion is a single `DELETE FROM ... WHERE ...` statement (see `tokenlock.go`), not a batch loop. Postgres executes a single statement atomically — if the connection drops mid-execution, Postgres rolls the statement back entirely rather than partially applying it. So a crash either means the delete fully happened before the crash, or didn't happen at all; there's no partial-delete state to clean up.

Net effect: worst case, one tick's cleanup is lost and picked up by the next replica on the next tick no locks are left stale, no partial deletes.

On the test, happy to add the concurrent-replica Postgres integration test I mentioned in the notes. Should be a small addition: spin up two `TokenLockStore` instances against the same DB, call `AcquireCleanupLeadership` on both concurrently, assert exactly one succeeds. 
I'd estimate ~30-45 min including verifying it actually exercises real contention rather than just happy-path. Want me to add it to this PR?

AkramBitar: @sid200727 
Thanks a lot, 
Yes, I think it is worth to have this test if it is possible.

sid200727: Added the test,  two independent TokenLockStore instances race for the same advisory lock concurrently. Passes consistently across repeated runs (-count=3): exactly one wins each time, the loser correctly fails to acquire, and after release a third attempt succeeds again cleanly.

atharrva01 (changes requested):
The per-tick election looks right to me, and reusing `NewCleanupLeaderFactory` instead of adding a second lock primitive is the right call. One thing I think is a real problem though.

`tokenLockCleanupLockID` is a single global constant, but there's one Manager per TMS (`SelectorService.managerLazyCache` is keyed by `*token.ManagementService`). So a node running N TMSs against the same postgres ends up with N cleaner goroutines all fighting over the same advisory lock, even though their `token_locks` tables are completely separate. One TMS gets cleaned per tick and the rest just log "leadership not acquired" and skip. Right now they all clean fine every tick, so this makes multi-TMS nodes worse off. interop's topology wires 5 TMSs, so it's not a corner case.

I think the fix is to derive the id per store instead of passing it down from the manager. `TokenLockStore` already does this for schema creation with `createTableLockID("tokenlock")`, so something like `createTableLockID(prefix + "tokenlock_cleanup")` inside the postgres `AcquireCleanupLeadership` would make it per-TMS. You could then also drop the `lockID` param from the `Locker` interface, since sqlite and in-memory ignore it anyway.

Related: the keystore cleanup you're following takes its id from config (`cleanup.Config.AdvisoryLockID`, used at `cleanup/manager.go:205`). Here it's hardcoded, so if two deployments share a postgres instance there's no knob to fix it.

On the tests:

- `leadership released after cleanup` only checks that `Close` was called, not that `Cleanup` ran, or ran first. It would pass if the tick acquired leadership and released without cleaning anything.
- Neither new subtest calls `m.Stop()`, so both cleaner goroutines keep ticking every 50ms for the rest of the run, and CI runs the race detector. The trailing `assert.NotNil(t, m)` isn't really asserting anything either.
- The postgres test passes `replicaA.lockID` (424242, the schema lock) as the cleanup id, so it's exercising the wrong constant. It can't reach the real one without an import cycle, which I think is another sign the id is in the wrong package.

Minor: the advisory lock holds a dedicated conn off `writeDB` for the whole tick, so a cleanup now needs two write conns where it used to need one. Fine at `MaxOpenConns: 10`, just worth a comment since it's operator configurable.

Happy to be wrong on the first point if TMSs are normally expected to have separate data sources.

AkramBitar:
@atharrva01

I think you're right, good catch.

Each TMS cleans its own token_locks table, but TMSs on the same node share one postgres DB, and the advisory lock is a single number covering the whole DB. So both cleaners ask for the same lock, and the one that loses skips its sweep even though nobody is touching its table. That has a cost: locking a token just inserts a row, and the selection path never looks at how old the row is, so an expired lock keeps its token unselectable until cleanup deletes it.

One tweak on the fix: you put createTableLockID(prefix + "tokenlock_cleanup") inside AcquireCleanupLeadership, but that runs on every tick and re-hashes each time (and re-creates NewCleanupLeaderFactory()). Better to compute it once in NewTokenLockStore and keep it on the struct next to the existing lockID.

On making the lock id configurable: I don't think we need to. The keystore needs that because its id is one fixed constant. Once the id comes from the table name it's already unique per TMS, so there's nothing left to configure.

@sid200727,

What do you think about this? 

Thanks a lot,
Akram

sid200727:
Hi @AkramBitar @atharrva01!
you're both right, this is a real bug, thank you for catching it. 
Agreed on the fix: derive the lock id from the table prefix once in NewTokenLockStore rather than a global constant in manager.go. I'll also drop the lockID parameter from AcquireCleanupLeadership entirely, since each backend can now own its own correct id internally removes the need for manager.go to know anything about lock-id uniqueness at all. Will also fix the two test issues you flagged (the weak "released after cleanup" assertion, and the postgres test reusing the schema-creation lock constant instead of a distinct one) and add Stop() calls so the tick goroutines don't keep running past each subtest. 
Give me a bit to implement and I'll push an update soon.

sid200727:
Fixed, pushed the per-TMS derivation as discussed. 
Summary:

->`AcquireCleanupLeadership` no longer takes a `lockID` param; each backend derives/owns it internally
-> Postgres now derives `cleanupLockID` from the table prefix at construction (`createTableLockID(prefix + "tokenlock_cleanup")`), so it's unique per TMS and distinct from the schema-creation lock the leader factory is also built once and cached, not rebuilt per call
-> Added `TestTokenLockStore_AcquireCleanupLeadership_DistinctTMS`, which directly proves the bug is fixed: two different TMSes now derive different lock ids and acquire leadership independently, rather than colliding on one global lock
-> Fixed the existing concurrent-replica test to go through the real public constructor throughout, so it exercises production code exactly, not a hand-built struct
-> Fixed the goroutine leak in "skipped when leadership not acquired" (blocking channel send on a per-tick callback `Stop()` couldn't unblock it)
-> "leadership released after cleanup" now verifies Cleanup actually ran and completed before Close, not just that Close was called

Full suite green under `-race`. Thanks both for catching this!

token/services/storage/db/sql/postgres/tokenlock.go
```diff
@@ -55,13 +64,26 @@ func NewTokenLockStore(dbs *common2.RWDB, tableNames common5.TableNames) (*Token
...
 
 	return &TokenLockStore{
-		TokenLockStore: tldb,
-		writeDB:        dbs.WriteDB,
-		ci:             ci,
-		lockID:         createTableLockID("tokenlock"),
+		TokenLockStore:       tldb,
+		writeDB:              dbs.WriteDB,
+		ci:                   ci,
+		lockID:               createTableLockID("tokenlock"),
```
adecaro: we should fix then also this one given that we are already touching this file.

adecaro:
@atharrva01 , interesting catch. Following the same reasoning, we have the same issue on the other locks as well (TransactionDB, for instance). Please, could you investigate and submit a fix? 
Many thanks 🙏

token/services/storage/db/sql/postgres/tokenlock.go
sid200727: Fixed, schema-creation lockID now also derives from the table prefix (createTableLockID(tableNames.Prefix + "tokenlock")), same reasoning as the cleanup lock. Existing tests already exercise this via CreateSchema() in both the concurrent-replica and distinct-TMS tests, and everything's green under -race.

sid200727:
Hi @atharrva01, just checking in - pushed the per-TMS lock derivation fix a few days back per our discussion above. Let me know if there is anything else to change, or if this is good to re-review. Thanks

atharrva01 (changes requested):
Thanks for turning this around so quickly @sid200727. The test changes are exactly right: the ordered `["cleanup", "closed"]` assertion is stronger than what I asked for, the `Stop()` calls are in, and building the factory once at construction rather than per tick is good. Documenting the spare-connection requirement on `AcquireCleanupLeadership` was a nice touch.

One problem though, I don't think the lock id is actually per-TMS yet.

The derivation uses `tableNames.Prefix`, but that is the prefix from the persistence config, not from the TMS. The TMS identity travels in the params: `db/manager.go:46` calls the store constructor with `tmsID.Network, tmsID.Channel, tmsID.Namespace` as the variadic params, and FSC's formatter builds the final name as `prefix + escaped(params) + "_" + shortcode` (`tablename.go:86`). So two TMSs sharing one persistence block get different tables but an identical `Prefix`.

I checked rather than guessing:

```
TMS A: prefix="tsdk"  token_locks="tsdk_netA__chA__nsA_tkn_locks"
TMS B: prefix="tsdk"  token_locks="tsdk_netB__chB__nsB_tkn_locks"
cleanup lock id A = -2281474731437264828
cleanup lock id B = -2281474731437264828
```

Same id, which is the original bug.

`TestTokenLockStore_AcquireCleanupLeadership_DistinctTMS` passes because it builds the two stores with different prefixes (`tms_a` and `tms_b`) and no params. That is a real deployment if someone configures separate persistence blocks, but it is not how the manager creates them, so the test does not cover the case that was broken.

Deriving from the fully qualified table name fixes it, since that name already carries the params:

```go
lockID:        createTableLockID(tableNames.TokenLocks),
cleanupLockID: createTableLockID(tableNames.TokenLocks + "_cleanup"),
```

I verified that gives distinct ids for two TMSs sharing a prefix, and keeps the cleanup id distinct from the schema one. Worth switching the test to same-prefix-different-params too, so it actually pins the regression.

The schema lock Angelo asked about has the same issue, it is keyed on the prefix today so it still serializes across TMSs on one persistence config.

thanks for the efforts....

AkramBitar:
> Thanks for turning this around so quickly @sid200727. The test changes are exactly right: the ordered `["cleanup", "closed"]` assertion is stronger than what I asked for, the `Stop()` calls are in, and building the factory once at construction rather than per tick is good. Documenting the spare-connection requirement on `AcquireCleanupLeadership` was a nice touch.
> 
> One problem though, I don't think the lock id is actually per-TMS yet.
> 
> The derivation uses `tableNames.Prefix`, but that is the prefix from the persistence config, not from the TMS. The TMS identity travels in the params: `db/manager.go:46` calls the store constructor with `tmsID.Network, tmsID.Channel, tmsID.Namespace` as the variadic params, and FSC's formatter builds the final name as `prefix + escaped(params) + "_" + shortcode` (`tablename.go:86`). So two TMSs sharing one persistence block get different tables but an identical `Prefix`.
> 
> I checked rather than guessing:
> 
> ```
> TMS A: prefix="tsdk"  token_locks="tsdk_netA__chA__nsA_tkn_locks"
> TMS B: prefix="tsdk"  token_locks="tsdk_netB__chB__nsB_tkn_locks"
> cleanup lock id A = -2281474731437264828
> cleanup lock id B = -2281474731437264828
> ```
> 
> Same id, which is the original bug.
> 
> `TestTokenLockStore_AcquireCleanupLeadership_DistinctTMS` passes because it builds the two stores with different prefixes (`tms_a` and `tms_b`) and no params. That is a real deployment if someone configures separate persistence blocks, but it is not how the manager creates them, so the test does not cover the case that was broken.
> 
> Deriving from the fully qualified table name fixes it, since that name already carries the params:
> 
> ```go
> lockID:        createTableLockID(tableNames.TokenLocks),
> cleanupLockID: createTableLockID(tableNames.TokenLocks + "_cleanup"),
> ```
> 
> I verified that gives distinct ids for two TMSs sharing a prefix, and keeps the cleanup id distinct from the schema one. Worth switching the test to same-prefix-different-params too, so it actually pins the regression.
> 
> The schema lock Angelo asked about has the same issue, it is keyed on the prefix today so it still serializes across TMSs on one persistence config.
> 
> thanks for the efforts....

@atharrva01 Thanks a lot for this comment. 

@sid200727 could you please have a look?

sid200727:
Fixed. 
Thank you for the precise investigation, @atharrva01. Now deriving both lockID and cleanupLockID from `tableNames.TokenLocks` (the fully-qualified table name) instead of `Prefix`, since that's what actually carries the per-TMS params. Also switched the DistinctTMS test to same-prefix-different-params, matching what you verified by hand, so it actually pins this regression now. Full suite green under -race.

AkramBitar:
Hi @sid200727,

Thank you very much.

Could you please take a look at the failed tests?

@atharrva01,

I will continue reviewing the code, and if there are no other issues, we will promote this PR.

I would really appreciate it if you could find some time to review this PR once more.

Thanks a million to both of you. I really appreciate all your help and support.

Akram

token/services/storage/db/sql/postgres/tokenlock.go
```diff
@@ -32,6 +32,15 @@ type TokenLockStore struct {
 	writeDB *sql.DB
 	ci      common3.CondInterpreter
 	lockID  int64
+
+	// cleanupLockID is derived from the table prefix, so it is unique per
```
AkramBitar: Only one minor one. 

The comment says the ID is "derived from the table prefix," but the code now derives it from tableNames.TokenLocks. That wording caused the original bug because two TMSs sharing a persistence block can have the same Prefix. The comment should be updated so it doesn’t mislead future readers.

AkramBitar (changes requested):

token/services/storage/db/sql/postgres/tokenlock_leadership_test.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	l3, acquired3, err := replicaA.AcquireCleanupLeadership(context.Background())
+	require.NoError(t, err)
+	require.True(t, acquired3, "leadership should be acquirable again after release")
+	require.NoError(t, l3.Close())
+}
+
+// TestTokenLockStore_AcquireCleanupLeadership_DistinctTMS verifies that two
+// TokenLockStores for *different* TMSes (different table prefixes) derive
```
atharrva01, line 102: leftover from the previous version, the test uses the same prefix now with different params, which is the point of it.

atharrva01 (approved):
this looks right now, nice one. i checked the ids rather than just reading it, two TMSes on the same prefix with different params:

```
A  schema=6005146497978177417   cleanup=357936614517608776
B  schema=-584983751925593000   cleanup=-2185387577765780972
```

all four distinct, and the test builds them the way `db/manager.go` actually does now, so it pins the regression properly. the schema lock going per table name is fine too, the schema is `CREATE TABLE IF NOT EXISTS` on differently named tables, so dropping the shared lock between TMSes can't race them against each other.

the round 2 test fixes all look good, the ordered `["cleanup", "closed"]` check, the `Stop()` calls, and the non blocking send to stop the cleaner getting stuck on the second tick.

one stale comment left inline (@AkramBitar already got the other one on tokenlock.go, i duplicated it by mistake and have removed mine). a couple of asserts that might be worth adding while you're in there:

- nothing pins the schema `lockID` being per TMS, `DistinctTMS` only checks `cleanupLockID`. a regression back to `createTableLockID("tokenlock")` would pass everything, and that's the line Angelo asked about.
- nothing checks the cleanup id differs from the schema id on the same store, though the comment says it does. i verified they do.

both are one liners on the existing test, entirely up to you.

on the failing checks @AkramBitar mentioned, i don't think any of them are yours:

- the 4 `benchmark` ones are the fork issue from #2103, `actions/checkout` refuses to check out fork PR code under `pull_request_target`, so they die before running anything
- `utest` is `TestGetRandomBytes_Success/single_byte` in `token/services/ttx`, which this PR doesn't touch. it asserts a single random byte isn't zero, so it fails about 1 run in 256 by construction. i'll add it to #2121
- both `itest` ones are docker registry, one of them an outright 502

token/services/storage/db/sql/postgres/tokenlock.go
sid200727: Fixed, updated the comment to say it's derived from `tableNames.TokenLocks` (the fully-qualified table name), not the prefix, and explained why the prefix-only wording was misleading (exactly the bug this PR fixes). Also added two assertions @atharrva01 suggested while I was in there. Thanks!

sid200727:
Fixed the comment wording, and added both suggested assertions, thanks @atharrva01. Full suite green under -race.

AkramBitar:
> Fixed the comment wording, and added both suggested assertions, thanks @atharrva01. Full suite green under -race.

@sid200727 

LGTM.
Thanks a lot, great work. 

@atharrva01,
Thanks a lot for your support on that PR. I appreciate it a lot.

Best Regards,
Akram

AkramBitar (approved):
LGTM

PR #1985, opened 2026-07-23
fix(token): instantiate validators with limits resolved from configuration

### LINKED ISSUE #1978: validator should be instantiated with the limits taken from the configuration
Currently both zkatdlog and fabtoken driver instantiante their validators using the default resource limits with driver.DefaultResourceLimits(). A solution would be to pass to the driver's constructor a ResourceLimitsProvider resolved via dig. In Dig, we register the provider based on configuration.

adecaro:
## Summary
- Both the fabtoken and zkatdlog drivers built their per-TMS validator with a hardcoded `driver.DefaultResourceLimits()`, silently ignoring any `token.validation.limits` the operator configured — while the separate validator-driver-service path already honored it.
- Register a single `driver.ResourceLimitsProvider` (config-backed) in dig and inject it into both drivers' `NewTokenService` and into `newValidatorDriverService`, so every validator-construction path resolves the same configured limits.
- Updated `docs/drivers/validation-resource-limits.md` to describe the unified flow.

Fixes #1978

## Test plan
- [x] `make checks`
- [x] `make lint-auto-fix`
- [x] `go test ./token/core/fabtoken/v1/driver/... ./token/core/zkatdlog/nogh/v1/driver/... ./token/sdk/dig/... ./token/core/... ./token/services/config/...`
- [x] `go test -race` on the same packages (includes `TestFabricWiring` dig dry-run)

### DIFF
--- a/docs/drivers/validation-resource-limits.md
+++ b/docs/drivers/validation-resource-limits.md
@@ -51,12 +51,25 @@ A `driver.StaticResourceLimits` provider (a trivial wrapper returning a fixed va
 tests, tools, and any caller that only needs the defaults (e.g.
 `cmd/token_validation_service`, the zkatdlog regression suite).
 
-The resolved `driver.ResourceLimits` flows: composition root → `core.NewValidatorDriverService(limits, ...)`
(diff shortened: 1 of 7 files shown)


KElkhiyaoui (approved):

AkramBitar (approved):
LGTM

PR #1987, opened 2026-07-23
fabricx: support ThresholdRule endorsement policies in nspolicy endorser selection

### LINKED ISSUE #1983: fabricx: nspolicy endorser selection does not support ThresholdRule endorsement policies
## Problem

`QueryServiceEndorserSelector.candidateMSPSets` in
[`token/services/network/fabricx/endorsement/nspolicy.go`](https://github.com/LFDT-Panurus/panurus/blob/main/token/services/network/fabricx/endorsement/nspolicy.go)
only understands MSP/identity-based endorsement policies (`NamespacePolicy.MspRule`,
a `common.SignaturePolicyEnvelope`). When a namespace's endorsement policy is
configured as a `NamespacePolicy.ThresholdRule` instead, `candidateMSPSetsFromPolicies`
returns a hard error:

```
namespace [%s] uses a threshold-rule endorsement policy, which is not identity/MSP
based and cannot be mapped to a subset of endorsers
```

(see `nspolicy.go` around line 111, `GetThresholdRule()` check).

This means `SelectEndorsers` — and therefore any FabricX token flow that relies on
selecting a satisfying subset of configured FSC endorsers for a namespace — fails
outright for namespaces protected by a threshold-signature rule rather than an
MSP-rule signature policy. There is currently no way to select endorsers for such a
namespace at all.

## Impact

Any FabricX deployment/network that configures a namespace's endorsement policy via
`ThresholdRule` cannot use `QueryServiceEndorserSelector` to select endorsers,
blocking token transaction endorsement for that namespace.

## Proposed enhancement

Extend `nspolicy.go` to also support `NamespacePolicy.ThresholdRule`:

- Determine how a `ThresholdRule` (scheme + public key) maps to a concrete
  set/subset of configured FSC endorser identities (e.g. via MSP/identity
  metadata associated with the threshold scheme's signers, if available).
- Extend `candidateMSPSetsFromPolicies` (or introduce a parallel code path) to
  produce candidate MSP-ID sets — or an equivalent endorser-selection strategy —
  for the `ThresholdRule` case instead of erroring out.
- Add unit test coverage mirroring the existing `mustThresholdRulePolicyItem`
  test helper in `nspolicy_test.go`, replacing the current "hard error" assertion
  with assertions on correct endorser-set selection.

## Current behavior (for reference)

Existing unit test `nspolicy_test.go` (`"threshold-rule namespace - hard error"`)
documents and locks in today's error-out behavior — it will need to be updated
once threshold-rule support lands.

adecaro:
## Summary
- `QueryServiceEndorserSelector.SelectEndorsers` previously hard-errored whenever a FabricX namespace's endorsement policy was a `ThresholdRule` (raw public key + signature scheme), since a `ThresholdRule` is a single-signer policy with no MSP principal to map onto a subset of endorsers.
- Added `endorserForThresholdRule`, which matches the `ThresholdRule`'s ECDSA public key directly against each configured endorser's own MSP/X.509 identity to find the one endorser that can sign for it. This is the objectively correct criterion, not just a workaround: FSC's pre-broadcast endorsement verification is already MSP/X.509-based regardless of the namespace's actual on-chain policy, so a signature can only pass both checks if the two keys are the same key.
- `BLS`/`EdDSA` `ThresholdRule` namespaces still get a clear, scheme-specific error, since FSC's signing path is X.509/ECDSA-based only.
- Updated `docs/services/network-fabricx.md`'s Namespace Policy section accordingly, and added a `FuzzEndorserForThresholdRuleNoPanic` fuzz test (wired into `.github/workflows/nightly-fuzz.yml`) covering the PEM/X.509 key-parsing path, since `ThresholdRule.PublicKey` is attacker-influenced data arriving over the wire via the query service RPC.

Fixes #1983

## Test plan
- [x] `go test ./token/services/network/fabricx/endorsement/...` — new `TestEndorserForThresholdRule` (7 subtests) and existing `TestCandidateMSPSetsFromPolicies` all pass
- [x] `go test ... -run='^$' -fuzz='^FuzzEndorserForThresholdRuleNoPanic$' -fuzztime=20s` — no panics
- [x] `make checks`
- [x] `make lint-auto-fix`
- [x] `make unit-tests` (full suite)


token/services/network/fabricx/endorsement/nspolicy.go
```diff
@@ -164,3 +193,72 @@ func mspIDsOf(principalSet []*msp.MSPPrincipal) ([]string, error) {
...
+	}
+
+	var matches []view.Identity
+	for _, id := range configured {
+		key, ok := ecdsaPublicKeyOf(id)
+		if !ok {
+			continue
+		}
+		if key.Equal(ruleKey) {
+			matches = append(matches, id)
```
AkramBitar, line 236: Minor comment, if single endorser accidentally written twice in the config. then this will fail in switch len(matches). Is that ok?

AkramBitar (approved):
LGTM, one minor issue

token/services/network/fabricx/endorsement/nspolicy.go
adecaro: Good catch, thanks — fixed in e4aa55f17. `endorserForThresholdRule` now dedupes matches by identity bytes before counting, so a duplicated entry in the `endorsers:` config for the *same* endorser collapses to a single match instead of tripping the "expected exactly one" error. The multi-match error is now reserved for genuinely distinct configured identities sharing a key (the real misconfiguration case). Also caught that the existing multi-match test was accidentally hitting this same bug — both fixtures had hardcoded `Mspid: "Org1MSP"`, making them byte-identical rather than two distinct endorsers — so I fixed that test too and added a dedicated duplicate-entry case.

AkramBitar (approved):
LGTM

PR #1988, opened 2026-07-24
feat(evm): endorsement flow (responder, initiator, registry)

atharrva01:
Wires the translator and signer from the last PR into an actual endorsement flow over FSC.

- A responder that endorses a request: authorize the caller, validate the token request against on-chain state (a small read-only ledger over getToken), translate to a StateDelta, sign the digest. It recomputes the digest itself and never signs one handed to it, so a bad initiator can't get it to sign something that doesn't match what it validated.
- An initiator that collects signatures and assembles a quorum. It recovers each signature to a distinct registered endorser before counting it, the same thing the contract checks, so it never builds a quorum the contract would reject.
- The registry (endorser address to FSC identity), the messages the two sides exchange (no digest in the request, on purpose), and a small dependency-free ABI codec for the reads. Both sides build the delta through one shared path, so everyone produces identical bytes.

For the gate, a test drives the real initiator and responders over in-memory sessions to assemble a 2-of-3 quorum, and a forge test verifies that exact quorum on the EndorsementVerifier. Same Go to Solidity loop as before, now over the whole flow.

The validation-record persistence, the per-TMS provider, and the client-side pp-version cross-check wait on the config and storage wiring in the next PR. Noted in the plan.


adecaro (approved):
LGTM. I have a few questions. I'll ask during our sync

Issue #1989, opened 2026-07-24
nwo: bootstrap a fabricx+EVM backend for token integration tests


arunabha003:
Problem

The EVM network driver under development (feature/evm-network-driver) has no way to be exercised inside NWO. NWO today can only bootstrap Fabric and FabricX backends — there is no machinery to stand up an EVM-speaking node, deploy the token contracts onto it, and wire FSC nodes to talk to it. Nothing in integration/ currently knows how to launch an external JSON-RPC node, deploy Solidity contracts, or emit an EVM network config block.

The token integration suites should be able to bootstrap a fabricx + EVM gateway backend by switching topology, so the existing fungible test bodies run against it exactly like dlog vs dlogx today. The platform should be buildable and testable in isolation, without waiting on the driver's end-to-end methods.

Solution

Add a standalone NWO platform + token backend for the fabricx+EVM gateway, mirroring integration/nwo/token/fabricx/, and make it testable in isolation.

1. Token-layer backend — new package integration/nwo/token/evm/ implementing the token Backend interface (integration/nwo/token/fabric/fabric.go): PrepareNamespace(tms) registers the token namespace; UpdatePublicParams(tms, ppRaw) seeds public parameters on-chain (EVM analog of the FabricX TMSDeploy path). Register it as a NetworkHandler under a new platform name in integration/nwo/token/factory.go, reusing the backend-agnostic crypto/PP generation.
2. FSC-level EVM platform/topology — boot the fabric-x-evm gateway node, reusing the FSC container-lifecycle pattern (nwo/fabricx/extensions/scv2: port allocation, container create/start, log stream, readiness wait, cleanup). Expose an evm.Topology implementing api.Topology + BackedTopology.
3. Contract deployment — deploy EndorsementVerifier + a per-TMS TokenState clone, seeding PP v0 + endorser set + threshold + graphHiding, and inject the resulting addresses back into node config via the existing TMS.BackendParams seam. (Deploy mechanism — forge vs a JSON-RPC/view path — to confirm; leaning forge.)
4. EVM config + identity generation — emit the token.tms.<id>.services.network.evm.* block per node (endpoint, chainID, contract addresses, finality/gas, submitter/endorser keystores), including the endorser Ethereum-address ↔ FSC view.Identity registry + threshold + allowlist the driver expects (design §6.1/§10). New secp256k1 crypto-material path alongside the x509/idemix generators.
[...]

PR #1992, opened 2026-07-26
feat(evm): NWO backend skeleton in the evm module

atharrva01:
The driver-owned seam for standing up an EVM backend in the token integration network, so the fungible suites run against an EVM backend by switching topology. Pushing it first (as discussed on #1989) so both EVM backends build against a fixed contract.

Moved into the evm module's own `nwo` folder per review: the module is developed as if it lived in a separate repo, so its integration surface ships with it and depends only on evm-module types.

What's here:

- a `Backend` interface: `Deploy` stands up a TMS's contracts (EndorsementVerifier + a per-TMS TokenState clone, seeding PP v0 + endorser set + threshold + graphHiding) and returns the addresses a node's config points at; `UpdatePublicParams` seeds the on-chain params.
- the supporting types (`DeploySpec`, `Deployment`, `TMSRef`), all evm-native.

No node bootstrap here. Besu (mine, Week 6) and the fabricx+EVM gateway (#1989) both implement this same Backend. The thin panurus adapter that plugs it into the token NWO framework lands with the concrete backend and the go.mod require+replace.

### DIFF
--- a/x/token/services/network/evm/nwo/backend.go
+++ b/x/token/services/network/evm/nwo/backend.go
@@ -1,0 +1,54 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package nwo
+
+import (
+	"github.com/LFDT-Panurus/panurus/x/token/services/network/evm/client"
+)
+
+// TMSRef identifies the token management system a backend operation targets. It is the evm module's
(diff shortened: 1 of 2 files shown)


adecaro:
Hi @atharrva01 , this must also go under `x/token/services/network/evm` in an `nwo` folder as well. 
Remember, the evm go module is developed as if it was living in a different repo 😄

atharrva01:
@adecaro , i have done the changess

PR #1996, opened 2026-07-26
fix(zkatdlog): enforce issuer membership in audit path, fix inverted upgrade precision check

### LINKED ISSUE #1993: zkatdlog: audit path admits unregistered issuers; upgrade eligibility check inverted
## Problem

An internal security review of `token/core/zkatdlog/nogh/v1` surfaced two independent,
narrowly-scoped correctness bugs:

**1. `audit/auditor.go`: `validateIssuer` never checked issuer membership**

`validateIssuer` (called from `IssueAuditValidate`) only checked that the issuer identity
matched the audit metadata attached to the action — it never checked that the issuer was
actually one of `PP.Issuers()`. The equivalent check already exists and is enforced for
redeems, but the issue-audit path silently admitted an issue action signed and audited by
an identity absent from the public parameters' issuer list.

**2. `crypto/upgrade/service.go`: inverted precision-eligibility check**

`NewService`'s loop building `UpgradeSupportedTokenFormatList` used
`precision > maxPrecision` where it should have used `precision <= maxPrecision`. This
inverts which token formats are considered eligible for upgrade — e.g. with
`maxPrecision=32`, only the 64-bit format was admitted into the supported-upgrade list
instead of the 16- and 32-bit ones that should actually be eligible.

## Why it matters

- (1) weakens the auditor's issuer-membership guarantee specifically on the audit path,
  even though the production validator path already enforces it elsewhere; an audit that
  should flag an unrecognized issuer does not.
- (2) means `UpgradeSupportedTokenFormatList` returns the wrong set of token formats,
  which can make eligible lower-precision tokens invisible to the upgrade flow while
  admitting formats that shouldn't be offered.

## What to do

- Fix `validateIssuer` to check membership in `PP.Issuers()` when the public parameters
  declare a non-empty issuer set, with an explicit open-policy bypass when the set is
  empty (mirroring the existing redeem-path check).
- Fix the inverted comparison in `crypto/upgrade/service.go`.
- Add regression tests pinning both corrected behaviors.

## Getting started

adecaro:
Fixes #1993

## Summary

Two independent, narrowly-scoped correctness fixes surfaced by an internal security review
of `token/core/zkatdlog/nogh/v1`:

- `audit/auditor.go`: `validateIssuer` only checked that the issuer identity matched the
  audit metadata, never that the issuer was actually one of `PP.Issuers()`. An issue action
  signed and audited by an identity absent from the public parameters' issuer list passed
  the audit unchanged, even though the equivalent check is already enforced for redeems.
  Fixed by checking membership when PP declares a non-empty issuer set, with an explicit
  open-policy bypass when it is empty.

- `crypto/upgrade/service.go`: the token-format eligibility loop for
  `UpgradeSupportedTokenFormatList` used `precision > maxPrecision` instead of
  `precision <= maxPrecision`, inverting which token precisions were considered eligible
  for upgrade (e.g. `maxPrecision=32` admitted only the 64-bit format instead of the 16-
  and 32-bit ones).

Each fix has a dedicated regression test pinning the corrected behavior. Also adds a few
standalone hardening tests (foreign-commitment rejection in the native range-proof
verifier, heterogeneous-token-type rejection in the bulletproof/CSP issue provers, and
owner-substitution rejection in the RWSet translator) confirming related attack scenarios
raised by the same review do not succeed against the current implementation.

## Test plan

- [x] `make checks`
- [x] `make lint-auto-fix`
- [x] `make unit-tests`
- [x] Targeted: `go test ./token/core/zkatdlog/nogh/v1/audit/... ./token/core/zkatdlog/nogh/v1/crypto/upgrade/... ./token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof/... ./token/core/zkatdlog/nogh/v1/issue/... ./token/services/network/common/rws/translator/...`

### DIFF
--- a/docs/upgradability.md
+++ b/docs/upgradability.md
@@ -25,7 +25,10 @@ The current driver determines compatibility based on several criteria:
 
 ### The "Burn and Re-issue" Mechanism
 
-When in-place upgrade is not possible (e.g., moving to a completely incompatible cryptographic curve or increasing precision beyond limits), Panurus implements an atomic "Burn and Re-issue" protocol. 
+When in-place upgrade is not possible (e.g., moving to a completely incompatible cryptographic curve or increasing precision beyond limits), Panurus implements an atomic "Burn and Re-issue" protocol.
+
+> [!NOTE]
+> **Eligibility is the complement of in-place support, by design.** For Fabtoken-to-DLog upgrades, a token's precision is either `<= maxPrecision` (in-place support, criterion 2 above — no issuer needed) or `> maxPrecision` (this path — issuer sign-off required because the value cannot be safely reinterpreted in a lower-precision format). These two ranges must never overlap: a driver that also accepted `<= maxPrecision` tokens into the Burn-and-Re-issue eligibility list would leave no token that ever needs this path, since every upgrade-eligible token would already be directly spendable. See the `Service` doc comment in `token/core/zkatdlog/nogh/v1/crypto/upgrade/service.go` for the implementation-level invariant.
 
 #### Step-by-Step Flow:
 1.  **Identification**: The owner identifies tokens that are no longer supported.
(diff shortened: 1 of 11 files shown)


AkramBitar (approved):
LGTM

PR #1997, opened 2026-07-26
perf(zkatdlog): reduce CPU time and allocations in CSP range-proof crypto

### LINKED ISSUE #1994: CSP range-proof crypto: reduce allocations and CPU time in small MSMs and Lagrange interpolation
## Problem

Profiling the Compressed Sigma Protocol (CSP) range-proof package
(`token/core/zkatdlog/nogh/v1/crypto/rp/csp`) found two avoidable sources of CPU time and
allocation overhead that are unrelated to the underlying cryptography:

**1. `MultiScalarMul` used unconditionally, even at n=1/n=2**

gnark-crypto's `MultiScalarMul` always spawns a goroutine fan-out sized for
`runtime.NumCPU()*2`, regardless of input size. CSP's folding rounds shrink every vector
down to length 2 and then 1 in their last two rounds — for every proof size, not just small
ones — and several range-proof call sites (the Schnorr proof-of-knowledge commitment)
always operate on exactly 2 generators/scalars. At these sizes, the fan-out overhead costs
more than it saves.

**2. `interpolateNative`'s outer loop reallocates scratch buffers every iteration**

`interpolateNative` evaluates a Lagrange interpolation at each of `n` points, calling
`NativeBatchInverse` once per outer iteration. `NativeBatchInverse` unconditionally
allocates 3 fresh slices per call, on top of the caller's own per-iteration scratch — 5
fresh same-size allocations per outer iteration, `n` times per call.

## Why it matters

CSP range-proof generation/verification is on the hot path for every private token
issue/transfer, so both the extra goroutine scheduling overhead and the repeated
allocations show up directly in proving/verification latency and GC pressure.

## What to do

- Add a small-MSM dispatch helper that special-cases n=1 (direct `Mul`) and n=2 (`Mul2`),
  falling through to `MultiScalarMul` for n>=3, and apply it at CSP's `MultiScalarMul` call
  sites.
- Add a zero-allocation variant of batch field inversion that writes into caller-supplied
  buffers, and rewrite `interpolateNative` to hoist its per-iteration scratch outside the
  outer loop using it.
- Back every change with a `benchstat`-verified before/after benchmark
  (`BenchmarkBFProver`); no public API, wire format, or verification behavior may change.

## Getting started

adecaro:
Fixes #1994

## Summary

Optimizes the Compressed Sigma Protocol (CSP) range-proof package
(`crypto/rp/csp`) with no change to its public API, wire format, or verification
semantics:

- `crypto/rp/csp/msm.go` (new): dispatches small, fixed-size multi-scalar
  multiplications to direct `Mul`/`Mul2` instead of gnark-crypto's
  `MultiScalarMul`. CSP's folding rounds shrink every vector down to length 2
  and then 1 in their last two rounds, and several call sites always operate
  on exactly 2 generators/scalars; `MultiScalarMul`'s goroutine fan-out (sized
  for `runtime.NumCPU()*2`) costs more than it saves at n<=2. Benchmarked on
  BLS12-381 (M1 Max): direct `Mul` is ~2.5x faster than `MultiScalarMul` at
  n=1, `Mul2` ~25-29% faster at n=2; from n=3 the two are at parity or
  `MultiScalarMul` wins, so `smallMSM` only special-cases n=1/n=2 and falls
  through to `MultiScalarMul` otherwise (strictly non-regressive). Applied at
  all 6 production `MultiScalarMul` call sites in `csp.go`/`rp.go`.

- `crypto/math/native.go`: adds `NativeBatchInverseInto`, a batch
  field-inversion helper that writes into caller-supplied buffers instead of
  allocating fresh ones on every call.

- `crypto/rp/csp/lagrange_native.go`: `interpolateNative`'s outer loop
  previously allocated 5 fresh same-size slices/scalars per iteration.
  Rewritten to hoist every per-x buffer and scalar scratch outside the outer
  loop via `NativeBatchInverseInto`, so each is allocated once per call
  instead of once per iteration.

- `crypto/rp/csp/transcript.go`: `Transcript.Absorb` is called many times per
  proving/verification round; reuses a scratch buffer across calls instead of
  growing a fresh slice on every `Absorb`.

Benchmarked on `BenchmarkBFProver` (n=32/n=64, `BLS12_381_BBS_GURVY`, count=8,
`benchstat`): combined effect of the MSM dispatch and batch-inversion changes
is roughly -6% ns/op and -35-40% allocs/op (geomean), each individually
significant at p<0.01. No public API, wire format, or verification behavior
changed.

## Test plan

- [x] `make checks`
- [x] `make lint-auto-fix`
- [x] `make unit-tests`
- [x] Targeted: `go test ./token/core/zkatdlog/nogh/v1/crypto/rp/csp/... ./token/core/zkatdlog/nogh/v1/crypto/math/... ./token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof/...`
- [x] `benchstat` before/after on `BenchmarkBFProver`, count=8

### DIFF
--- a/token/core/zkatdlog/nogh/v1/crypto/math/native.go
+++ b/token/core/zkatdlog/nogh/v1/crypto/math/native.go
@@ -91,6 +91,32 @@ func NativeBatchInverse[T any, E GnarkFr[T]](elems []E) []E {
 	return resultE
 }
 
+// NativeBatchInverseInto is the zero-allocation counterpart of NativeBatchInverse,
+// for hot loops that perform many same-size batch inversions: the caller hoists
+// prefix and out once and reuses them across calls instead of paying 3 fresh
+// allocations (prefix, result, resultE) per call.
+//
+// elems, prefix, and out must all have equal, non-zero length; prefix and out
+// must not alias elems or each other. out[i] must already be a valid pointer
+// into caller-owned storage (e.g. out[i] = E(&outBacking[i])).
+func NativeBatchInverseInto[T any, E GnarkFr[T]](elems []E, prefix []T, out []E) {
+	n := len(elems)
+
+	prefix[0] = *elems[0]
+	for i := 1; i < n; i++ {
+		E(&prefix[i]).Mul(E(&prefix[i-1]), elems[i])
+	}
+
+	var acc T
+	E(&acc).Inverse(E(&prefix[n-1]))
+
+	for i := n - 1; i > 0; i-- {
+		out[i].Mul(E(&acc), E(&prefix[i-1]))
+		E(&acc).Mul(E(&acc), elems[i])
+	}
+	*out[0] = acc
+}
+
 // DispatchCurve returns true for BLS and BN254 curves based on the curve ID.
 func DispatchCurve(curve *mathlib.Curve) (isBLS bool, isBN254 bool) {
 	switch curve.GroupOrder.CurveID() {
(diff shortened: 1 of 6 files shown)


HayimShaul (approved):
LGTM

Effi-S:
image
image

Effi-S:
image

PR #1998, opened 2026-07-26
test(zkatdlog): close regression-corpus coverage gaps for CSP scenarios

### LINKED ISSUE #1995: zkatdlog regression corpus: fixed-shape scenarios (upgrade witness, public metadata, multi-auditor, open issuer policy) not covered
## Problem

The ZKAT-DLog regression corpus (`token/core/zkatdlog/nogh/v1/regression`) is built by
driving the real production Issue/Transfer/Auditor service stack
(`testutils/env.go`) and serializing the resulting requests/proofs as fixed test vectors.
Several validator code paths are currently exercised only by hand-built unit tests that
construct structs directly, not by this production-stack corpus:

- proof-format upgrade-witness transfers
- public-metadata issue/transfer
- an unclaimed-metadata transfer (should be rejected)
- multi-auditor (1-of-N auditor policy) transfers
- an extra-signature transfer (should be rejected)
- issue/redeem under an open issuer policy (empty `PP.IssuerIDs`)

Because these paths never flow through `testutils/env.go`, a regression in the
production-stack serialization or validation of any of them would not be caught by the
regression suite, even though the corpus is specifically meant to catch exactly this class
of bug.

## Why it matters

The regression corpus is the primary defense against silent serialization/validation
drift across proof-system changes (e.g. IPA vs. CSP). Gaps in its scenario coverage mean
that drift in these specific paths could land undetected.

## What to do

- Extend `testutils/env.go` with fixture builders for each gap above, going through the
  same `buildTestCase` path used by existing fixtures.
- Extend `benchmark/setup.go` so the shared public parameters can carry a second auditor
  (for the multi-auditor fixture) and support an open-issuer-policy configuration.
- Wire the new fixtures into the corpus generator
  (`regression/testdata/zero/generator/main.go`) and regenerate `testdata/zero/**`.
- Extend `regression_test.go`'s fixture dispatch to validate the new fixtures, asserting
  the two negative fixtures fail with the expected error.
- Update `docs/drivers/benchmark/core/dlognogh/dlognogh_regression.md` to describe the new
  corpus subtrees and fixture keys.

## Getting started

adecaro:
Fixes #1995

## Summary

Extends the ZKAT-DLog regression corpus (`token/core/zkatdlog/nogh/v1/regression`) with
fixed-shape test vectors for validator code paths that were previously exercised only by
hand-built unit tests, not by the production Issue/Transfer/Auditor service stack that
`testutils/env.go` drives to build the corpus:

- `testutils/env.go`: adds fixture builders for a proof-format upgrade-witness transfer,
  public-metadata issue/transfer, an unclaimed-metadata transfer (negative fixture), a
  multi-auditor (1-of-N auditor policy) transfer, and an extra-signature transfer (negative
  fixture). Each goes through `buildTestCase`, the same path used by the existing
  size-sweep fixtures.

- `benchmark/setup.go`: adds a `SecondAuditorSigner` to `SetupConfiguration` so the shared
  public parameters carry two auditors, needed for the new multi-auditor fixture. Splits
  `NewSetupConfigurationsWithParams` into a private helper parameterized on
  `registerIssuer`, and adds `NewOpenIssuerPolicySetupConfigurations`, which skips issuer
  registration to produce public parameters under an open issuer policy (`PP.IssuerIDs`
  empty).

- `testutils/env_openpolicy.go` (new): `OpenPolicyEnv`/`NewOpenPolicyEnv` build issue/redeem
  fixtures against the open-issuer-policy setup above.

- `regression/testdata/zero/generator/main.go`: generates the 6 new fixed-shape fixtures
  per curve/bit-length configuration, and a new `generateOpenPolicyCorpus` step (IPA proof
  type only) writing `open_policy_issue_0`/`open_policy_redeem_0` fixtures under a new
  `testdata/zero/open-policy/` corpus subtree.

- `regression/regression_test.go`: switches the fixture-key dispatch to prefix-based
  matching to route the new keys to their validation functions, and asserts the two
  negative fixtures fail with the expected error substring rather than validating
  successfully.

- All `testdata/zero/**/{params.txt,pp.json,testdata.json}` are regenerated: existing
  fixtures' serialized bytes change because the shared public parameters now bake in a
  second auditor key; `testdata.json` additionally gains the 6 new fixture keys per
  configuration. `testdata/zero/open-policy/` is a new subtree (IPA proof type only, per
  curve/bit-length).

- `docs/drivers/benchmark/core/dlognogh/dlognogh_regression.md`: documents the new `csp/`
  and `open-policy/` corpus subtrees, the new fixed-shape fixture keys in a new table, and
  corrects stale path references and coverage counts.

No production validation logic changed; this closes test-coverage gaps only.

## Test plan

- [x] `make checks`
- [x] `make lint-auto-fix`
- [x] `make unit-tests`
- [x] Targeted: `go test ./token/core/zkatdlog/nogh/v1/regression/... ./token/core/zkatdlog/nogh/v1/testutils/... ./token/core/zkatdlog/nogh/v1/benchmark/...`


AkramBitar (approved):
LGTM

PR #1999, opened 2026-07-26
fix(lookup): isolate per-namespace failures in queryByID

### LINKED ISSUE #1990: lookup: one failing namespace aborts the whole key lookup batch in queryByID
## Problem

`DeliveryScanQueryByID.queryByID` in `token/services/network/fabric/lookup/deliveryqs.go` groups the
requested keys by namespace and then queries the token chaincode once per namespace. If any single
namespace fails, it returns from the whole function. There are three such spots in the loop:

- `json.Marshal` of the args fails
- the chaincode `Query` fails
- `json.Unmarshal` of the result fails

Because the function starts with `defer close(ch)`, returning early closes the channel, so every other
namespace in that batch is dropped too, including the ones that would have succeeded.

It also loses the fallback. On those paths `startDelivery` stays `false` and the keys are never removed
from `keySet`, so there is no block scan either. Listeners waiting on those keys just never fire.

## Why it matters

One misbehaving namespace takes down key lookup for every other namespace sharing the batch, and it
fails silently: the error is logged, the channel closes, and callers see an empty result rather than a
failure.

## Prior art

The same bug shape was fixed in the sibling finality path in #1426, in
`token/services/network/fabric/finality/deliveryqs.go`. That fix changed the early `return` inside the
loop to `continue`, and made the failed entry fall back to the block scan instead of vanishing. The
lookup copy never got the same treatment.

## What to do

- Make the three early returns skip only the failing namespace and carry on with the rest.
- Make a failed namespace fall back to the block scan rather than disappearing.
- Add unit tests. `finality/deliveryqs_test.go` (added in #1426) is a good model to copy from.

Heads up on testability: `q.Channel` is a concrete `*fabric.Channel` today, so this probably needs the
same small interface extraction #1426 did (it introduced `txLedger` and `blockScanner` interfaces so
the query could be faked in tests). That refactor is part of the task.

## Getting started

atharrva01:
Closes #1990.

`queryByID` queries the token chaincode once per namespace. If one of those queries failed, it
returned from the whole function, and since there is a `defer close(ch)` that dropped every
other namespace in the batch along with it. `startDelivery` also stayed false, so the block scan
never ran, and listeners on all those keys just never fired.

Now a failure only skips its own namespace. Its keys stay in `keysByNS` and `keySet` so the scan
picks them up, and the healthy namespaces still deliver. Same fix #1426 made on the finality path.

The type held a concrete `*fabric.Delivery` and `*fabric.Channel`, so it was not testable at all.
Pulled out three small interfaces plus a `ChannelStateQuerier` adapter that leaves the existing
chaincode call as it was. `deliveryllm.go` is the only construction site.

Four tests, and I checked they fail without the fix. One caveat worth knowing: the "other
namespaces still delivered" assertion only fails some of the time against the old code, because
map order decides whether the good namespace ran before the bad one returned. The "scan must
start" assertion fails every time, so that is the one really pinning the regression.

Not in here: the starting block is `max(FirstBlock, lastBlock-NumberPastBlocks)` with both
operands unsigned, so a low `lastBlock` underflows instead of clamping to `FirstBlock`. Same line
sits in `finality/deliveryqs.go`. Happy to open a separate issue for it.

### DIFF
--- a/docs/services/network-fabric.md
+++ b/docs/services/network-fabric.md
@@ -350,6 +350,13 @@ The setup listener watches for changes to a specific ledger key that stores publ
 
 The same delivery-based mechanism is reused by the [lookup service](../../token/services/network/fabric/lookup/deliveryllm.go) to detect transfer action metadata writes: it derives the transfer action metadata key prefix from [`KeyTranslator.TransferActionMetadataKeyPrefix`](../../token/services/network/common/rws/translator/rwset.go) and prefix-matches rwset writes against it, without needing to know each write's specific subkey ahead of time.
 
+When resolving a batch of keys, the lookup service groups them by namespace and issues one
+`QueryStates` chaincode call per namespace
+([`deliveryqs.go`](../../token/services/network/fabric/lookup/deliveryqs.go)). Each namespace is
+resolved independently: if its query cannot be built, does not reach the chaincode, or returns a
+response that cannot be decoded, only that namespace falls back to the block scan. The keys of the
+other namespaces in the same batch are still resolved and delivered.
+
 ## State Queries
 
 The Fabric implementation provides efficient state querying through the chaincode:
--- a/token/services/network/fabric/lookup/deliveryllm.go
+++ b/token/services/network/fabric/lookup/deliveryllm.go
@@ -129,7 +129,8 @@ func (p *deliveryBasedLLMProvider) NewManager(network, channel string) (Listener
 		},
 		&DeliveryScanQueryByID{
 			Delivery: ch.Delivery(),
-			Channel:  ch,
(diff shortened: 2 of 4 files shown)


plan.md
```diff
@@ -0,0 +1,55 @@
+# Fix #1990: one failing namespace aborts the whole key lookup batch
+
```
AkramBitar: Please remove this file

AkramBitar (changes requested):

plan.md
atharrva01: Removed

AkramBitar (approved):
LGTM
Please make sure that all the tests passes.
Thanks a lot

PR #2001, opened 2026-07-27
lint: enable rowserrcheck and fatcontext

### LINKED ISSUE #1991: lint: nine linters are configured in .golangci.yml but never enabled
## Problem

`.golangci.yml` carries `linters.settings` blocks for nine linters that are never enabled:

`revive`, `wrapcheck`, `iface`, `ireturn`, `lll`, `maintidx`, `gocognit`, `fatcontext`, `rowserrcheck`

None of them appear under `linters.enable`, and none are on by default in golangci-lint v2, whose
defaults are just `errcheck`, `govet`, `ineffassign`, `staticcheck` and `unused`. So none of that
configuration has any effect.

You can check it yourself:

```
golangci-lint linters -c .golangci.yml
```

and compare the enabled list against the `settings:` keys.

The `revive` block is the largest one. It sets `enable-all-rules: true` plus roughly fifteen rule
tweaks (`argument-limit`, `line-length-limit`, `function-result-limit`, a `unhandled-error` allowlist
and so on), and none of it runs today.

## Why it matters

The config reads as though these checks are enforced. Anyone looking at it will assume the rules apply,
and either we are missing lint coverage someone intended us to have, or we are carrying config that
misleads readers.

## What to do

Go linter by linter and make the config honest:

- If it was meant to run, add it to `linters.enable` and deal with the fallout, either fixing the hits
  or adding specific `nolint` directives. `nolintlint` is enabled with `require-specific: true`, so
  bare `//nolint` will not pass.
- If it was not, delete the dead settings block.

`revive` with `enable-all-rules: true` will produce a lot of findings across the repo, so it is
probably best left out of this PR and raised separately. Starting with the cheaper ones
(`rowserrcheck`, `fatcontext`, `ireturn`, `maintidx`) would make a good first change.

## Acceptance

- No `linters.settings` entry for a linter that is not enabled.
- `make lint` and `make checks` pass.

atharrva01:
First step of #1991, going linter by linter as you suggested.

I measured all nine first so the order is not guesswork. Issues and files touched, across every
module in `GO_MODULES`:

| linter | issues | files |
| --- | --- | --- |
| rowserrcheck | 9 | 3 |
| fatcontext | 2 | 2 |
| lll | 20 | 4 |
| maintidx | 8 | 8 |
| ireturn | 26 | 14 |
| iface | 52 | 37 |
| gocognit | 134 | 83 |
| wrapcheck | 2262 | 395 |
| revive | 3367 | 682 |

This PR enables the two cheapest, five files between them.

**rowserrcheck found a real bug.** `dedupedTokenRowsIterator.Next` never checked `rows.Err`, so
when iteration stopped because of an error rather than exhaustion it returned `(nil, nil)`. The
caller cannot tell that from a clean end of results, so the selector would quietly act on a
truncated token set. Every other query in `tokens.go` already checks `rows.Err`, this iterator
was the outlier. Added tests and confirmed they fail without the fix.

Six sites in `prepared_stmt_holder_test.go` now assert `rows.Err`. Three hits the linter cannot
see through, where rows are handed to the iterator or the call never produces rows, carry a
specific `//nolint` with the reason.

**fatcontext** flags the cleanup and recovery managers holding a context that `Stop` cancels.
That is a deliberate service lifecycle rather than a per-request context, so I suppressed both
with a reason instead of restructuring them. Happy to change that if you would rather the
managers took a context from the caller.

`golangci-lint run` is clean on all nine modules, `go test -race` passes on the touched
packages, and `make checks` is green.

### Next steps

`lll` looks free from the root module, which has zero violations, but the other modules have 20
across 4 files. 19 are long function signatures in integration support files and one is a string
literal that cannot be wrapped, so it is style churn with no correctness value. I left it for
its own PR.

`maintidx` only flags large table-driven tests. Satisfying it means splitting those tests up,
which does not seem worth it. Might be better to exclude `_test.go` from it or drop the setting,
let me know which you prefer.

The remaining six keep their settings blocks for now, since the plan is to enable them in later
steps rather than delete the config.


atharrva01:
@adecaro , @AkramBitar  , your thoughts on this...?

adecaro (approved):
Thanks for the effort @atharrva01 🙏

PR #2004, opened 2026-07-27
fix(ci): skip tidy-check in nightly FSC update workflow

### LINKED ISSUE #2003: Nightly FSC Update workflow: checks job fails at tidy-check after every dependency bump
## Problem

The `checks` job in `.github/workflows/nightly-fsc.yml` ("Nightly FSC Update")
fails on most nightly runs at the `make checks` step, always with the same
`tidy-check` output:

```
The following go.mod/go.sum files are not tidy:
cmd/artifactgen/go.mod
cmd/artifactgen/go.sum
...
go.mod
go.sum
...
make: *** [checks.mk:137: tidy-check] Error 1
```

Example failing run: https://github.com/LFDT-Panurus/panurus/actions/runs/30237531226

This has failed 4 of the last 6 nightly runs (2026-07-22, 2026-07-23, 2026-07-25,
2026-07-26, 2026-07-27), blocking the rest of that workflow's jobs (`utest`,
`itest`, `bench-check`) every time it fires, since they all `need: checks`.

## Root cause

`checks.mk`'s `tidy-check` target runs `go mod tidy` in every module and then
checks `git diff --name-only` for changes to `go.mod`/`go.sum`, failing if any
are found. That's a valid tidiness check only when the committed `go.mod`/`go.sum`
already match what `go mod tidy` produces — true for a normal PR.

`nightly-fsc.yml`'s "Bump FSC dependency" step
(`.github/actions/bump-fsc/action.yml`) deliberately rewrites every module's
`go.mod`/`go.sum` to the latest `fabric-smart-client` main commit via
`make update-dep` (which already ends with its own `make tidy`) *before*
`make checks` runs. So by the time `tidy-check` runs, the files are correctly
tidy for the new dependency version, but necessarily differ from git HEAD
(which still points at the old version) — `tidy-check` fails unconditionally,
regardless of whether the modules are actually untidy.

Confirmed locally: on an unmodified checkout, `go mod tidy -diff` exits 0 in
every module — i.e. the modules are genuinely tidy; only the git-diff-based
verification is wrong for this workflow.

`bump-fsc` is used only by `nightly-fsc.yml`, so no other workflow is affected.

## Fix

Add a `checks-no-tidy` target to `checks.mk` mirroring `checks` minus
`tidy-check`, and have `nightly-fsc.yml`'s `checks` job run
`make checks-no-tidy` instead of `make checks`. Tidiness for the bumped
dependency was already established by `make update-dep`'s own `make tidy` call;
re-verifying it against a stale git HEAD is what's broken. `tidy-check` and
`checks` remain unchanged for every other caller (e.g. `tests.yml`), where the
git-diff check is correct.

adecaro:
## Summary
- `tidy-check` flags any uncommitted `go.mod`/`go.sum` change as untidy by diffing against git HEAD. The nightly FSC bump workflow deliberately rewrites those files to a new `fabric-smart-client` version (and already runs `make tidy` itself via `update-dep`) before `make checks` runs, so the git-diff heuristic fails unconditionally on every nightly run even though the modules are genuinely tidy.
- Add a `checks-no-tidy` target to `checks.mk` (same check list as `checks`, minus `tidy-check`) and switch `nightly-fsc.yml`'s "Run checks" step to use it. `tidy-check` and the regular `checks` target are unchanged for normal PR/local usage.

Fixes #2003

## Test plan
- [x] `make checks-no-tidy` exits 0 and skips the tidy-check section entirely
- [x] `make checks` still exits 0 and prints "✓ All Go modules are tidy." (no regression for normal usage)
- [x] `make lint-auto-fix` — 0 issues in every module

### DIFF
--- a/.github/workflows/nightly-fsc.yml
+++ b/.github/workflows/nightly-fsc.yml
@@ -33,7 +33,7 @@ jobs:
         run: make install-tools
 
       - name: Run checks
-        run: make checks
+        run: make checks-no-tidy
 
       - name: Run linter
         run: make lint
--- a/checks.mk
+++ b/checks.mk
@@ -1,6 +1,12 @@
 .PHONY: checks
 checks: licensecheck gofmt goimports govet gofix misspell ineffassign staticcheck protos-lint buf-format tidy-check
 
+.PHONY: checks-no-tidy
+# same as 'checks' but without tidy-check; for use after a workflow step that
+# already rewrote go.mod/go.sum to a new dependency version and ran 'make tidy'
+# itself, where tidy-check's git-diff-against-HEAD heuristic would always fail
+checks-no-tidy: licensecheck gofmt goimports govet gofix misspell ineffassign staticcheck protos-lint buf-format
+
 .PHONY: licensecheck
 licensecheck:
 	@echo Running license check


AkramBitar (approved):
LGTM

PR #2009, opened 2026-07-28
point to fork FSC commit to test leak fix

SaidAltury-ibm:


adecaro:
it looks like we are good here, @SaidAltury-ibm 🙏

AkramBitar:
@SaidAltury-ibm thanks a lot. @Effi-S and I will run the benchmark and validate the fix. Will keep you updated.

AkramBitar (approved):

AkramBitar (commented):
Let's wait to the test and then close this PR and then promote with the latest FSC that will contain this FSC fix.

PR #2015, opened 2026-07-29
Benchmark Github Actions

### LINKED ISSUE #1973: Over Night benchmarks
### Problem

Github action that runs the benchmarks over night
Something similar to the fuzzers https://github.com/LFDT-Panurus/panurus/actions/workflows/nightly-fuzz.yml.

### Solution

- Jobs for some of the benchmarks (and uploads output to txt files)
- Second Job that creates a plot for these benchmarks
- Job that compares to previous runs  

### Alternatives

_No response_

### Additional Context

_No response_

Effi-S:
## Add a Token Validation benchmark GitHub Action for PR-vs-base comparison

Adds a CI workflow that automatically benchmarks the token validation service on every pull request and posts a comment showing how the PR's performance compares to its base branch. Also refactors the benchmark tooling to share a single parser between the plotting and comparison scripts.

What's included

New workflow — .github/workflows/token-validation-benchmark.yml
- Triggers on pull_request.
- benchmark job: runs a matrix of 2 refs (pr, base) × 2 crypto variants (ipa, csp) × 2 benchmarks (local, gRPC) = 8 isolated runs, each on its own runner so the fixed gRPC server port (8099) never collides.
  - local → BenchmarkLocalTokenValidation (in-process, sweeps -cpu).
  - grpc → BenchmarkAPIGRPC (full gRPC API path, sweeps client connections).
  - csp variant selects the CSP crypto testdata via -testRoot; ipa is the default.
  - Raw go test -bench output is uploaded as artifacts, tagged _pr_ / _base_.
- compare job: downloads all raw outputs, pairs each benchmark row across PR and base, and posts a Markdown table of the percentage change in latency (ns/op) and throughput (TPS) as a PR comment (pull-requests: write).

New — cmd/benchmarking/bench_parse.py
- Extracts the pure go test -bench parsing logic (simple_parser) into a standalone module with no plotting/UI dependencies, so it can be imported without pulling in plotly/streamlit. Single source of truth for turning a benchmark line into a row.

New — cmd/benchmarking/compare_benchmarks.py
- Takes a directory of base/PR benchmark outputs, pairs rows by their input identity (ignoring run-to-run measurement columns), and emits a Markdown comparison report suitable for a PR comment.

Refactor — cmd/benchmarking/plotly_plot_node.py
- Now imports simple_parser from bench_parse instead of carrying its own copy.

Docs — docs/drivers/benchmark/token_validation_service_benchmark.md
- Fixes the client run command (adds GOGC=10000 go run ./client/ -benchtime=30s -count=5 …) and adds plotly to the pip install line.

Local testing

The workflow header documents how to run it locally with act, including a sample PR event payload and the need to run one matrix variant at a time (act shares the host network, unlike real CI).

Notes

- Benchmark knobs (BENCHTIME=10s, COUNT=3, CPU=16, NUM_CONN=4) are kept modest to keep PR feedback responsive; raise COUNT for more statistically stable comparisons.
- GOFLAGS: -mod=mod lets go test update go.mod/go.sum in the ephemeral checkout.

### DIFF
--- a/.github/workflows/token-validation-benchmark.yml
+++ b/.github/workflows/token-validation-benchmark.yml
@@ -1,0 +1,216 @@
+name: Token Validation Benchmark
+
+# Runs the token-validation benchmarks on both the PR branch and its base
+# branch, then posts a comment on the PR with the percentage differences.
+#
+# Run locally with act (https://github.com/nektos/act):
+#   1. Create an event payload describing the PR to benchmark. base.sha/head.sha
+#      must be commits that exist locally; number is only used for the comment.
+#        cat > /tmp/pr-event.json <<'JSON'
+#        {
+#          "pull_request": {
+#            "number": 1,
+#            "base": { "sha": "<base-commit-sha>" },
+#            "head": { "sha": "<pr-commit-sha>" }
+#          }
+#        }
+#        JSON
+#   2. Run one variant at a time so the two gRPC jobs never share port 8099
+#      (act puts all matrix jobs on the host network; real CI isolates them):
+#        act pull_request \
+#          -W .github/workflows/token-validation-benchmark.yml \
+#          -e /tmp/pr-event.json \
+#          --matrix variant:ipa \
+#          --artifact-server-path /tmp/act-artifacts
+#   3. The final `gh pr comment` step needs a real token/PR; expect it to fail
+#      locally. Inspect the generated report in the compare job's logs instead,
+#      or add `--matrix ...` and run only the `benchmark` job with `-j benchmark`.
+#   If a run is interrupted, clear leftover containers before retrying:
+#        docker rm -f $(docker ps -aq --filter ancestor=catthehacker/ubuntu:act-latest)
+on:
+  pull_request:
+  workflow_dispatch:
+  
+# Needed so the compare job can post a comment on the PR.
+permissions:
+  contents: read
+  pull-requests: write
+
+env:
+  # -mod=mod lets `go test` update go.mod/go.sum in the ephemeral CI checkout.
+  GOFLAGS: -mod=mod
+  MODULE_DIR: cmd/token_validation_service
+  GO_MOD_FILE: cmd/token_validation_service/go.mod
+  OUTPUT_DIR: bench
+  # go test knobs. Kept modest so the PR feedback loop stays reasonable; raise
+  # COUNT for more statistically stable comparisons.
+  BENCHTIME: 10s
+  COUNT: 5
+  # -cpu is set at run time to the number of CPUs the runner actually has
+  # (nproc), rather than a fixed 16/32 that may exceed the runner and skew
+  # results. NUM_CONN is the gRPC client connection sweep, unrelated to -cpu.
+  NUM_CONN: 4
+  # -testRoot values selecting the crypto variant. Each variant now names its
+  # testdata explicitly rather than relying on a compiled-in default.
+  CSP_TEST_ROOT: -testRoot=../../token/core/zkatdlog/nogh/v1/regression/testdata/zero/csp/32-BLS12_381_BBS_GURVY
+  IPA_TEST_ROOT: -testRoot=../../token/core/zkatdlog/nogh/v1/regression/testdata/zero/32-BLS12_381_BBS_GURVY
+  # GOGC for the benchmark runs. A high value suppresses GC during the short
+  # benchtime windows so measurements aren't perturbed by collection pauses.
+  GOGC: 10000
+  # Percentage change treated as noise by the compare step. Deltas smaller than
+  # this (or whose base/PR sample ranges overlap) are reported as neutral (➖).
+  DELTA: 1
+
+jobs:
+  # Job 1: benchmark BOTH branches on the SAME runner, interleaved.
+  #
+  # The matrix is the cross product of:
+  #   * variant  - the crypto backend (ipa is the default, csp adds -testRoot)
+  #   * bench     - which benchmark to run and the flags unique to it
+  #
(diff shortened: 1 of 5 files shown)


AkramBitar:
@Effi-S thanks a lot. Good work.
1) Could you please remove the ns/op column or if you want to keep it then explain the deference's between it and between TPS column?
2) Regarding the CPUs number please get the number of CPUs that you have in CI/CD and run the benchmark with this number (maybe we need to check the option to have fix number of CPUs for this action)

.github/workflows/token-validation-benchmark.yml
```diff
@@ -0,0 +1,206 @@
...
+  BENCHTIME: 10s
+  COUNT: 5
+  # -cpu is set at run time to the number of CPUs the runner actually has
+  # (nproc), rather than a fixed 16/32 that may exceed the runner and skew
+  # results. NUM_CONN is the gRPC client connection sweep, unrelated to -cpu.
+  NUM_CONN: 4
+  # -testRoot value selecting the CSP crypto variant. The IPA variant is the
+  # default (no -testRoot), so it has no equivalent env var.
+  CSP_TEST_ROOT: -testRoot=../../token/core/zkatdlog/nogh/v1/regression/testdata/zero/csp/32-BLS12_381_BBS_GURVY
+
```
AkramBitar, line 63: Add IPA explicit root like CSP.

.github/workflows/token-validation-benchmark.yml
```diff
@@ -0,0 +1,206 @@
...
+          : > "$pr_out"
+
+          # Run one sample of base, then one of pr, and repeat COUNT times.
+          # Interleaving (base,pr,base,pr,...) cancels slow drift on the runner
+          # in addition to the host-to-host variance the same-runner design
+          # already removes. Each file ends up with COUNT benchmark lines, which
+          # is shape-identical to `go test -count=COUNT` for the parser.
+          run_one() {
+            # $1 = checkout dir, $2 = output file to append to
+            ( cd "$1/$MODULE_DIR" && GOGC=10000 go test \
```
AkramBitar: Maybe to add all the parameters to one section and use them as env variables.

cmd/benchmarking/bench_parse.py
```diff
@@ -0,0 +1,90 @@
...
+    if "TPS" in df.columns:
+        df = df.rename(columns={"TPS": "tps"})
+    for col in [c for c in df.columns if c.startswith("ns/op") and "(" in c]:
+        label = col.split("(")[1].rstrip(")")
+        df[f"{label} (ms)"] = df[col] / 1e6
+        df = df.drop(columns=[col])
+
+    if 'tps' in df.columns and 'workers' in df.columns:
+        # Little's Law (latency = concurrency / throughput)
+        df['avg (ms)'] = df['workers'] * 1000 / df['tps']
```
AkramBitar, line 89: Explain why 1000

cmd/benchmarking/compare_benchmarks.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def _emoji(metric: str, pct: float, base_samples: list, pr_samples: list) -> str:
+    """Pick an indicator for a change, given whether lower is better.
+
+    Returns the neutral marker when the change is within the ±1% noise band or
+    when the base/PR sample ranges overlap (i.e. the delta is not statistically
+    distinguishable from run-to-run jitter).
+    """
+    if abs(pct) < 1.0:  # treat sub-1% as noise
```
AkramBitar: Can we have this as environment variable defined on the action?

.github/workflows/token-validation-benchmark.yml
```diff
@@ -0,0 +1,206 @@
...
+  # COUNT for more statistically stable comparisons.
+  BENCHTIME: 10s
+  COUNT: 5
+  # -cpu is set at run time to the number of CPUs the runner actually has
+  # (nproc), rather than a fixed 16/32 that may exceed the runner and skew
+  # results. NUM_CONN is the gRPC client connection sweep, unrelated to -cpu.
+  NUM_CONN: 4
+  # -testRoot value selecting the CSP crypto variant. The IPA variant is the
+  # default (no -testRoot), so it has no equivalent env var.
+  CSP_TEST_ROOT: -testRoot=../../token/core/zkatdlog/nogh/v1/regression/testdata/zero/csp/32-BLS12_381_BBS_GURVY
```
AkramBitar, line 55: Could you please add all the environment variables that we need to this section. Explain that if any tune is needed then we can change it here.

AkramBitar (changes requested):

.github/workflows/token-validation-benchmark.yml
Effi-S: DOne

.github/workflows/token-validation-benchmark.yml
Effi-S: Done (GOGC var)

cmd/benchmarking/bench_parse.py
Effi-S: Added Explainer

cmd/benchmarking/compare_benchmarks.py
Effi-S: Done added to compare_bench.py `--delta` param + added to YAML `DELTA` param

.github/workflows/token-validation-benchmark.yml
Effi-S: Done

AkramBitar (approved):
LGTM
Thanks a lot

PR #2016, opened 2026-07-29
identity: back off instead of spinning when recipient provisioning fails (#1642)

### LINKED ISSUE #1642: Bounded and Capacity-Controlled Resource Pools [MED]
**Description**
Apply explicit upper bounds to every resource pool that wallet operations touch. This includes: capping the `RecipientDataCache` channel size and the background provisioning goroutine's throughput so that cache-miss storms cannot exhaust memory; setting maximum connection-pool sizes and query timeouts on the underlying wallet database so that concurrent iterator construction cannot hold locks indefinitely; and enforcing a hard limit on the total number of recipient identities that can be registered per wallet (or per enrollment ID) to prevent unbounded growth of the persistent registry and the `identitydb` store. Without these caps, rate limiting alone can be circumvented by a burst that fills pools before the limiter fires.

AkramBitar:
## Problem

`RecipientDataCache.provisionIdentities` refills a wallet's recipient cache in the background. On error it retried with no pause:

```go
id, err := c.backed(ctx)
if err != nil {
    continue
}
```

A backend that keeps failing spins this at **196,401 attempts in 250 ms** — one pinned core per anonymous owner wallet, until the backend recovers.

## Fix

Retry with exponential backoff: 50 ms, doubling to a 30 s ceiling, reset on success.

2 files, +74 lines. Nothing changes on the success path.

## Verification

The test asserts under 50 attempts in the same window and fails if the backoff is removed. `make checks`, `make lint-auto-fix` and `go test -race ./token/services/identity/...` all clean.

## Why this belongs to #1642

#1642 asks for upper bounds on the resources wallet operations consume, and names three: the cache channel size, **the provisioning goroutine's throughput**, and the number of identities per wallet.

This PR fixes the second one. The goroutine had no bound on how often it could retry, so a failing backend let it consume a whole core. Adding a delay is that bound.

## Why the database bounds are not part of this

#1642 also asks for connection-pool and query-timeout caps on the wallet database. Neither is this layer's to set:

- **Query timeouts.** Every `WalletStore` method already passes its `context.Context` to a context-aware `database/sql` call, so the caller's deadline already stops the query. Adding our own ceiling would override the caller rather than protect it.
- **Pool size.** The pool is opened and sized by FSC from `fsc.persistences.<name>.opts.maxOpenConns`. FSC memoises the `*sql.DB` per data source, so whichever store opens it first fixes the pool for everyone — Panurus cannot change it afterwards. The real defect is that FSC defaults `MaxOpenConns` to 0, which `database/sql` treats as unlimited. That is an FSC fix, not one we can make here.

Part of #1642


token/services/identity/role/registry.go
```diff
@@ -22,6 +22,52 @@ type WalletFactory interface {
...
+// nothing ever removes one. Because a counterparty can ask this node for recipient identities (see
+// ttx.RespondRequestRecipientIdentityView), that growth is remotely driven and otherwise unbounded,
+// in both the wallet registry and the identity store that holds each identity's audit and signer
+// data.
+//
+// The value is deliberately generous. An anonymous owner wallet consumes roughly one identity per
+// transaction it receives, so this is a backstop against runaway growth rather than an operational
+// quota; a wallet reaching it has either been attacked or long outlived the retention assumptions
+// of its deployment.
+const MaxIdentitiesPerWallet = 1 << 20
```
adecaro: what happens if the wallet belong to a merchant with millions of customers? This wallet will exhaust pretty quickly its quota then, no?
I mean, these should all be configuration customizable per wallet with an impact on the datamodel then.

token/services/identity/role/cache.go
```diff
@@ -30,23 +62,44 @@ type RecipientDataCache struct {
...
+	generations *semaphore.Weighted
 }
 
+// NewRecipientDataCache returns a cache of recipient data of the requested size, backed by backed.
+//
+// size is clamped to [0, MaxRecipientDataCacheSize]: a size of 0 disables pre-provisioning
+// altogether (every call generates on the spot), and a size above the maximum is reduced so that
+// misconfiguration cannot commit unbounded memory. Independently of size, at most
+// MaxConcurrentRecipientDataGenerations generations run concurrently.
 func NewRecipientDataCache(Logger logging.Logger, backed RecipientDataBackendFunc, size int, metrics *Metrics) *RecipientDataCache {
```
adecaro: There is still a fundamental issue here. Given the current code, we keep in memory all the wallets that have performed at least an operation for the entire lifetime of the node. For each of this wallet we keep a cache that will be start filling as soon as the wallet is loaded. In a scenario where there are many wallets operating at the same time consuming the cache, the system will also be busy refilling these caches. So, an attacher can still produce pressure to the system by pushing the cache to work in critical moment when the system needs to handle other parts of the code. Long story short, in which cases do we really need this cache? Can we get rid of it?

token/services/identity/role/registry.go
adecaro: Another point, an attacher might exhaust this number by asking recipient identities and never use them in actual transactions. So, the counting should take in consideration the actual used recipient identities and get rid of the recipinet identities that were requested and never used. So, if too much time has passed between the time of recipient generation and signature generation, the signature should not be produced.

AkramBitar: Agreed. Removing the cap.  Refiling as a design issue.

token/services/identity/role/cache.go
AkramBitar: Based on preliminary analysis. There's no case here that justifies it. So let's drop it. Generate on demand under a single shared semaphore, and I'll track it in #1642.

AkramBitar: Agreed — dropped that cap.

AkramBitar:
@adecaro please let me know if you have more comments on that, thanks a lot.

Effi-S (approved):

SaidAltury-ibm (commented):
@AkramBitar This looks superseded by #2137 (merged into main on 2026-08-10), which fixes the same busy-spin defect in provisionIdentities, I'd suggest closing this one

AkramBitar:
Dup off https://github.com/LFDT-Panurus/panurus/pull/2137

PR #2083, opened 2026-07-30
Add a default token selection rate limiter

### LINKED ISSUE #1986: Add a default token rate limiter
Currently we have an infrastructure to let the application inject a rate limiter.
Add a default rate limiter that takes effect when no rate limiter is given by the application

HayimShaul:
Add a default rate limiter that an application can attach if it does not want to supply its own. It is provided as a default — or as an example — for clients that would like to use one instead of writing their own.

It is not attached: nothing is rate limited unless `token.selector.rateLimitEnabled` is set or a limiter is passed with `WithLimiter`, so current behaviour is unchanged. The default limiter is a per-wallet token bucket (100 selections/s, burst 200) charged once per `Select` call, and a throttled selection fails fast with `token.SelectorRateLimited` before any token is locked.

Addresses #1986

needs https://github.com/LFDT-Panurus/panurus/pull/2151 to be merged before CI passes


token/services/selector/ratelimit/limiter.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	b.tokens--
+
+	return true
+}
+
+// Stop terminates the eviction goroutine. Buckets are left in place, so a limiter that is
+// still consulted after Stop keeps enforcing the limit; it simply stops reclaiming the
+// memory of idle wallets.
+func (l *TokenBucketRateLimiter) Stop() {
```
AkramBitar: **Blocking: leaks after a public-params update.**

`Update` (`token/provider.go:135`) calls `Shutdown()` to reap the *evicted TMS* — but this limiter is process-wide, and `Stop` is `sync.Once` with no restart path. So one routine public-params update kills eviction for the rest of the process: buckets then grow unbounded, losing the memory bound the docs promise.

Simplest fix: evict lazily inside `take` and drop the goroutine entirely. Or make the limiter per-TMS, which also fixes the keying issue below.

token/services/selector/ratelimit/limiter.go
```diff
@@ -0,0 +1,238 @@
...
+// reporting whether a token was available.
+func (l *TokenBucketRateLimiter) take(identity string) bool {
+	l.mu.Lock()
+	defer l.mu.Unlock()
+
+	now := l.now()
+	b, ok := l.buckets[identity]
+	if !ok {
+		// A wallet not seen recently starts with a full bucket.
+		b = &bucket{tokens: l.burst, last: now}
```
AkramBitar, line 216: Buckets are keyed on wallet ID alone, but the limiter spans all TMSs. Two TMSs with a wallet named `issuer` share one bucket: each gets half the rate, and one network can throttle the other.

`Allow` refuses this exact trade-off for empty identities ("unrelated callers throttle each other"), so it reads as inconsistent. Suggest keying on `TMSID + walletID`.

token/services/selector/ratelimit/selector.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+// Select consults the limiter for the selecting wallet and, if the request is allowed,
+// delegates to the wrapped selector. On denial it returns the limiter's error (which
+// wraps token.SelectorRateLimited) without touching the wrapped selector, so a throttled
+// request queries no storage and locks no tokens.
+func (s *selector) Select(ctx context.Context, ownerFilter token.OwnerFilter, q string, tokenType token2.Type) ([]*token2.ID, token2.Quantity, error) {
+	// A nil owner filter carries no wallet id to key on; let the wrapped selector
+	// report the error it already reports for this case.
+	if ownerFilter != nil {
```
AkramBitar, line 73: `ownerFilter == nil` here, plus `identity == ""` in `Allow`, means a selection with an empty wallet ID skips the limiter entirely. Is that reachable? If so it is an unmetered path around a security control, and rejecting it would be safer than exempting it.

token/services/selector/sherdlock/service.go
```diff
@@ -59,13 +95,19 @@ func (s *SelectorService) SelectorManager(tms *token.ManagementService) (token.S
...
+// the rate limiter's own goroutine.
 func (s *SelectorService) Shutdown() {
 	s.mu.Lock()
 	managers := s.managers
 	s.managers = nil
+	limiter := s.limiter
 	s.mu.Unlock()
 
+	if limiter != nil {
+		limiter.Stop()
```
AkramBitar, line 124: This is what triggers the leak above: `Shutdown` means "a TMS was evicted", but `limiter` is process-wide. Same in `simple/service.go`. Correct as written if the limiter becomes per-TMS.

integration/token/fungible/support.go
```diff
@@ -338,20 +338,26 @@ func CheckCoOwnedBalance(network *integration.Infrastructure, ref *token3.NodeRe
 	CheckCoOwnedBalanceForTMSID(network, ref, wallet, typ, expected, nil)
 }
 
+// CheckCoOwnedBalanceForTMSID asserts that ref reports expected as the balance of the
```
AkramBitar, line 382: Good fix, but unrelated to rate limiting — could it go in its own PR? Keeps this one reviewable as a security change, and the flake fix bisectable on its own.

token/services/selector/config/driver.go
```diff
@@ -105,3 +118,26 @@ func (c *Config) GetFetcherCacheMaxQueries() int {
...
+
+// GetRateLimit returns the number of selections per second per wallet the built-in limiter
+// allows: the configured RateLimit, the built-in default when it is unset, or zero when a
+// negative RateLimit disables rate limiting.
+func (c *Config) GetRateLimit() int {
+	if c.RateLimit < 0 {
+		return rateLimitDisabled
+	}
+	if c.RateLimit == 0 {
+		return defaultRateLimit
```
AkramBitar, line 140: - `rateLimit: -1` to mean "off" is easy to misread in YAML; an explicit `rateLimitDisabled: true` would be clearer. Minor.
- Pre-existing: `config.New` returns `nil, err`, yet both `NewService`s log *"using defaults"* and continue with the nil `cfg` — `Default(cfg)` is now the first nil deref. Returning `&Config{}` alongside the error fixes both.

token/services/selector/ratelimit/selector.go
HayimShaul: Both bypass paths are benign:

**ownerFilter == nil**
only reachable via a programming mistake, and the inner selector either errors or panics immediately anyway. Nothing is silently passed through.

**ownerFilter.ID() == ""**
 intentional by design. Putting an empty ID in a shared bucket would let unrelated anonymous callers throttle each other, which is a worse outcome than skipping the meter. The inner selectors already reject or crash on an empty ID independently. An adversary who can supply a crafted OwnerFilter already has in-process code execution.
No change is needed.

integration/token/fungible/support.go
HayimShaul: moved to https://github.com/LFDT-Panurus/panurus/pull/2151

token/services/selector/sherdlock/service.go
HayimShaul: This is an observation about a potential future hazard, not a current bug. If the design were ever changed so that SelectorService became per-TMS (one instance per TMS instead of one per process), then Shutdown() would be called on TMS teardown, which would be correct — the limiter would be local to that TMS and safe to stop. The code at lines 107–110 anticipates that model and would be correct under it too.

Bottom line: no bug here, no change needed. The limiter.Stop() in both Shutdown() methods is safe because Shutdown() is only ever called at full-process teardown.

AkramBitar (approved):

PR #2084, opened 2026-07-30
feat(identity): observe and escalate signature throttling

### LINKED ISSUE #1643: Observability, Alerting, and Automated Throttle Escalation [LOW]
**Description**
Instrument every Signer and Verifier service operation with metrics (call rate, latency, error rate, cache hit/miss ratio) and expose them via the existing `token/core/common/metrics` infrastructure. Define alert thresholds that trigger automated responses -- such as temporarily blocking a caller or reducing its quota -- when anomalous call volumes are detected. Comprehensive audit logging of caller identity, operation type, and outcome provides the forensic trail needed to detect abuse early, attribute it to a specific principal, and tune throttling policies over time.

HayimShaul:
Signature operations had no observability, and an identity flooding the node with invalid-signature calls looked no different from a healthy one.

This adds a leaf observer package, `sigobserve`, that the signature surface reports to, plus three sinks:

- **metrics** — operation counts by `op`/`role`/`outcome`, operation and `GetSigner` duration histograms, signer-cache lookups, throttle escalations, and a gauge of currently throttled principals
- **audit log** — one greppable record per operation, naming the principal by identity hash only, so identity material never reaches a log file
- **throttle policy** — a per-principal token bucket driving an escalating `normal → soft → blocked` state machine that de-escalates once the principal goes quiet

`identity.Provider`, `common.Deserializer` and the signers/verifiers they hand out are instrumented. The throttle gate is consulted **only** at `token.SignatureService`: driver validators are instrumented but never gated, because per-node call history would make transaction validation non-deterministic across nodes. `AreMe`/`IsMe` are ungated too — refusing them would return a wrong answer rather than an error.

Configured per TMS under `token.tms.<name>.identity.throttle` and defaulting to **monitor** mode, so a deployment gets the metrics and the audit trail without any identity being blocked until enforcement is switched on deliberately. Denials surface as `token.SignatureThrottled`.

### Hot-path cost

`BenchmarkGetSignerAndSign`, warm signer cache, `GetSigner` + `Sign` per iteration:

| | ns/op | B/op | allocs/op |
|---|---|---|---|
| observer off | 1401 | 522 | 9 |
| observer on (metrics + audit + policy) | 2167 | 863 | 15 |

The audit logger asks the logger whether the level is enabled before rendering a record, which keeps the string build off every routine operation at a production log level.

### Docs

`docs/security/signature_observability.md` (metrics reference, audit format, escalation policy, rollout), linked from `docs/README.md`, with the configuration block in `docs/configuration.md`.

Fixes #1643


token/services/identity/throttle/throttle.go
```diff
@@ -0,0 +1,504 @@
...
+// Callers must hold e.mu.
+func (e *Escalator) escalate(ctx context.Context, principalID string, p *principal, reason string) {
+	p.lastViolation = e.now()
+
+	switch p.level {
+	case LevelNormal:
+		e.transition(ctx, principalID, p, LevelSoft, reason)
+	case LevelSoft, LevelBlocked:
+		e.transition(ctx, principalID, p, LevelBlocked, reason)
+	}
```
AkramBitar, line 448: **HIGH — `escalate` ignores the minimum soft period, so a principal reaches `blocked` two requests after its first over-quota request.**

This switch consults `p.level` only; nothing here reads `p.levelUntil` (the end of `SoftDuration`). Combined with two things upstream:

- `decide` calls `escalate(ReasonRate)` on **every** request that finds an empty bucket (line 227), not just the first one;
- `transition(LevelSoft)` clamps the bucket balance to 0 via the quota override.

…the request immediately after the `normal → soft` transition also fails `Take`, re-enters `escalate` with `p.level == LevelSoft`, and takes the `LevelSoft, LevelBlocked` branch.

Verified empirically with `Rate: 10, Burst: 10, ModeEnforce`:

```
request 11 -> soft     ("principal exceeded its quota and is now at level soft")
request 12 -> blocked  ("principal exceeded its quota and is now at level blocked")
request 13 -> "principal is blocked until ..."
```

There is no window in which the principal actually experiences the reduced quota. This contradicts the package doc ("`soft` … for at least `SoftDuration`") and `docs/security/signature_observability.md:137`.

Consequence: the mechanism is documented as graduated, but in practice one momentary spike past `burst` costs the principal a full `blockDuration` (default 1m) of hard refusal rather than 5 minutes of reduced quota. See finding 2 for why that is worse than it sounds on the auditing path.

Suggested shape — refuse to escalate again while the current level is still serving its minimum term:

```go
func (e *Escalator) escalate(ctx context.Context, principalID string, p *principal, reason string) {
	now := e.now()
	// A principal still serving its minimum time at this level has already been
	// penalised for this episode; re-arm the violation clock but do not push it up.
	if p.level != LevelNormal && now.Before(p.levelUntil) {
		p.lastViolation = now
		return
	}
	p.lastViolation = now
	...
}
```

(That also makes the `LevelBlocked` re-arm in the current `case` meaningful rather than reachable on the second request.)

token/services/identity/throttle/throttle.go
```diff
@@ -0,0 +1,504 @@
...
+	e.mu.Unlock()
+
+	if e.buckets.Take(principalID) {
+		return false, ""
+	}
+
+	e.mu.Lock()
+	e.escalate(ctx, principalID, p, ReasonRate)
+	level := p.level
+	e.mu.Unlock()
```
AkramBitar, line 236: **HIGH (same defect as the `escalate` comment, second route into it) — `e.mu` is dropped around `buckets.Take`, so concurrent over-quota requests each escalate a level.**

The lock is released at line 220, `Take` runs unlocked, then the lock is re-acquired at 226 to mutate `p`. Two goroutines that both fail `Take` for the same principal will therefore both call `escalate`, and the second one observes `p.level == LevelSoft` set by the first — producing the same `normal → blocked` jump from a **single** bucket-exhaustion event, with no second round-trip needed.

At the default `rate: 200 / burst: 400` on a concurrent call site this is the likely path in practice, not the sequential one.

Fixing `escalate` to honour `levelUntil` closes this too, since the second caller would find `levelUntil` in the future. If you'd rather fix it here, re-check the bucket under the lock (or have `Take` return the level-relevant decision) instead of trusting the pre-unlock read.

token/sig.go
```diff
@@ -22,54 +30,127 @@ type Verifier = driver.Verifier
...
+		opt(s)
+	}
+
+	return s
 }
 
 // AuditorVerifier returns a signature verifier for the given auditor identity
 func (s *SignatureService) AuditorVerifier(ctx context.Context, id Identity) (Verifier, error) {
+	if err := s.allow(ctx, sigobserve.OpAuditorVerifier, sigobserve.RoleAuditor, id); err != nil {
+		return nil, err
```
AkramBitar: **HIGH — the default quota is a hard per-TMS throughput ceiling on these call sites, because they all key on one fixed identity.**

`allow` keys the throttle principal on the identity passed in. But the two gated call sites that run on the hot path use a *constant* identity per node:

- `AuditorVerifier` — called once per transaction from `token/services/ttx/auditor.go:302`, always with an auditor ID from the public parameters. That set is fixed and tiny (usually one).
- `GetSigner` — called once per transaction from `token/services/ttx/endorse.go:129` with the node's own long-term signing identity.

So all of a node's traffic is charged to a single throttle principal, and `DefaultRate = 200` becomes a global cap on transactions per second for that node — not a per-counterparty abuse limit.

What happens past the cap, on the auditing path:

1. The auditor's principal exhausts its bucket.
2. Via the `escalate` defect above, it reaches `blocked` within two requests.
3. `AuditorVerifier` returns `SignatureThrottled`.
4. `auditor.go:302` treats that as "cannot get a verifier" and `continue`s — it does not distinguish throttling from a malformed identity.
5. The loop exhausts every auditor and returns `failed verifying auditor signature`.

So an auditor sustaining >200 tx/s doesn't get slowed down; auditing fails outright for `blockDuration`, and the error surfaced to the operator names the wrong cause.

`docs/security/signature_observability.md` describes the defaults as "generous … not a throughput cap on healthy traffic". For these call sites they are precisely a throughput cap on healthy traffic. Options, roughly in order of preference:

- Don't gate operations whose principal is the node's own identity or a public-parameters auditor — those are not attacker-controlled, so there is nothing to defend against. Observe them, but leave them ungated.
- Or key those call sites on something attacker-influenced (the requesting session/peer) rather than the verified identity.
- At minimum, propagate `SignatureThrottled` distinctly at `auditor.go:302` so a throttle is retried/reported as a throttle rather than silently becoming a signature-verification failure.

token/services/identity/throttle/throttle.go
```diff
@@ -0,0 +1,504 @@
...
+// evictIdle performs one eviction sweep.
+func (e *Escalator) evictIdle() {
+	e.mu.Lock()
+	defer e.mu.Unlock()
+
+	cutoff := e.now().Add(-e.cfg.IdleTTL)
+	for id, p := range e.principals {
+		if p.level == LevelNormal && p.lastSeen.Before(cutoff) {
+			delete(e.principals, id)
+		}
```
AkramBitar, line 591: **MEDIUM — escalated principals are retained forever, so the DoS defence is itself an attacker-controlled unbounded allocation.**

The `p.level == LevelNormal` guard means any principal that has ever escalated is never deleted, regardless of `IdleTTL`. Its companion in `ratelimit/bucket.go:252` skips `overridden` buckets — which is exactly the flag `transition(LevelSoft)` sets — so *both* halves of the state survive indefinitely.

A principal that escalates once and never calls again therefore keeps its `principal` struct (the slot ring + map key) and its bucket for the process lifetime, and `e.counts` / the `identity_throttled_principals` gauge never return to zero — so the gauge reads as a persistent incident rather than a current one.

The cost to an attacker is low: minting fresh identities and spending either `burst`+1 requests or ~`MinSamples` (50) failing operations per identity permanently pins an entry. There is no cap on `len(e.principals)`. Since the identities are unverified at this point (that's the point of gating verification), nothing bounds the key space.

Worth doing both: evict escalated principals whose `levelUntil` has passed and which have been idle for `IdleTTL` (the block has expired, so there is no state left worth keeping), and add a hard cap on `len(e.principals)` with an eviction policy for when it's hit, so the worst case is bounded regardless.

token/services/ratelimit/bucket.go
```diff
@@ -0,0 +1,256 @@
...
+// being throttled.
+func (s *BucketSet) evictIdle() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	cutoff := s.now().Add(-s.idleTTL)
+	for key, b := range s.buckets {
+		if !b.overridden && b.last.Before(cutoff) {
+			delete(s.buckets, key)
+		}
```
AkramBitar, line 267: Companion to the `Escalator.evictIdle` comment: `!b.overridden` is the second half of the retain-forever behaviour. The doc comment's reasoning — "their bucket is the only record that they are still being throttled" — holds only while they *are* still being throttled. Once the escalation has expired, the override is stale and the entry is pure retained memory.

If the escalator drops the principal, this bucket becomes unreachable-but-retained, so the two eviction policies need to agree. Simplest coupling: have the escalator clear the override when it de-escalates or evicts, which lets the normal `idleTTL` path here reclaim the bucket without a special case.

token/core/fabtoken/v1/driver/ws.go
```diff
@@ -145,4 +154,13 @@ func (d *WalletServiceFactory) NewWalletService(tmsConfig driver.Configuration,
 		true,
 		&disabled.Provider{},
 	)
+	if err != nil {
+		return nil, err
+	}
+	// This factory builds a standalone wallet service with no client-facing signature service to
+	// gate, so the policy's background eviction has nothing to serve. Instrumentation keeps
+	// working: stopping the stack only releases its goroutine.
+	sigStack.Stop()
```
AkramBitar, line 163: **MEDIUM — `Stop()` stops eviction but leaves the escalator wired as an observer, so its state grows with nothing to reclaim it.**

By this point `newWalletService` has already run `identityProvider.SetObserver(sigStack.Observer())` (line 66) and `deserializer.SetObserver(sigStack.Observer())` (line 114), and those references outlive this function. `Stop()` only ends the background loops.

So after this line: `Escalator.Observe` still runs (`cfg.Enabled()` is true by default) and still inserts into `e.principals`, `BucketSet` still creates buckets — but *both* eviction loops are dead. Nothing is ever reclaimed for the lifetime of this wallet service.

The comment "Instrumentation keeps working: stopping the stack only releases its goroutine" is the part I'd push back on — it reads as though the goroutine were the only thing the stack owned, when it is in fact the only thing that bounded it.

If this factory genuinely has no gate to serve, the cheaper and safer construction is to not build an escalator at all here: pass an observer that is metrics+audit only (the `reporting` chain in `sigpolicy/stack.go:73` is exactly that), so there is no unbounded map to keep. If you want to keep `Stop()`, it should also detach the observers, or at least keep the eviction loop alive.

token/core/zkatdlog/nogh/v1/driver/ws.go
```diff
@@ -43,28 +48,69 @@ func (d *BaseWalletServiceFactory) NewWalletService(
...
+		fscIdentity,
+		networkDefaultIdentity,
+		publicParams,
+		ignoreRemote,
+		metricsProvider,
+	)
+	if err != nil {
+		return nil, err
+	}
+	sigStack.Stop()
```
AkramBitar, line 66: **MEDIUM — same `Stop()`-leaves-observer-wired issue as the fabtoken factory, and worse here because of the key space.**

`NewWalletService` has already installed `sigStack.Observer()` on the identity provider (line 110) and the deserializer (line 182); this `Stop()` kills only the eviction loops, so `principals`/`buckets` grow monotonically for the service's lifetime.

The zkatdlog-specific aggravation: principals here are keyed on identity hashes, and zkatdlog owner wallets are *anonymous* — a fresh pseudonym per recipient. So the key space is not merely attacker-influenced, it is unbounded by design under normal operation.

Worth noting this also makes the throttling ineffective on the non-stopped path: each single-use nym gets a fresh full bucket, so it can never exceed a per-principal rate, while still costing a map entry retained for `IdleTTL` (10 min). At moderate throughput that's hundreds of thousands of live entries buying no rate-limiting at all. Consider keying anonymous-wallet operations on the long-term identity behind the nym (if reachable at this layer) or excluding them from the escalator entirely.

token/services/identity/sigobserve/audit.go
```diff
@@ -0,0 +1,172 @@
...
+	}
+}
+
+// levelFor maps an event to the level its record is written at. Everything an operator would
+// investigate - errors, rejected signatures, throttled calls - is warn, a level change is info,
+// and routine successes are debug.
+func levelFor(e Event) zapcore.Level {
+	switch e.Outcome {
+	case OutcomeError, OutcomeInvalid, OutcomeThrottled:
+		return zapcore.WarnLevel
```
AkramBitar, line 95: **MEDIUM — `OutcomeError` at warn fires on a documented happy path, and `OutcomeThrottled` at warn is unbounded.**

Two separate problems with mapping these three outcomes to `WarnLevel`:

**1. `OutcomeError` is routine control flow at one call site.** `token/services/ttx/collectendorsements.go:301` deliberately uses a *failing* `GetSigner` as the "is this party local?" probe:

```go
if signer, err := ...GetSigner(ctx, signerIdentity); err == nil {
    // local: sign here
} else {
    logger.DebugfContext(ctx, "failed to find a signer for party [%s]: [%s]", ...)
}
```

Note that call site logs its own failure at **debug**, precisely because it is expected. With `AuditLogger` on the main observer chain, every transaction involving a remote signer now also emits `sig-audit op=get_signer … outcome=error` at **warn**. That is one warn line per remote party per transaction on a fully healthy network.

Secondary effect: those same expected failures feed the escalator's `errors` window for the *counterparty's* principal, so a legitimate busy counterparty accumulates error-ratio samples for something it did not do and may escalate on `errorRateThreshold` alone.

**2. `OutcomeThrottled` at warn is one line per denied request.** `SignatureService.allow` emits an event per denial, so a principal that is already `blocked` still drives a warn log line for every request it makes. Blocking a flooder therefore converts a signature-verification flood into a log flood — the amplification the gate was meant to prevent.

Suggestions: keep `OutcomeInvalid` at warn (a rejected signature is genuinely notable), drop `OutcomeError` to debug or introduce a distinct "expected error" outcome for the probe call site, and rate-limit or sample throttled records — logging the level *transition* (already info via `OpEscalation`) is the operator-actionable event, not each subsequent denial.

token/services/identity/sigobserve/decorator.go
```diff
@@ -0,0 +1,100 @@
...
+// keep working through the wrapper.
+//
+// ctx is the context the events are reported with: Sign takes no context of its own, so the
+// resolution-time context is the only one available.
+func InstrumentSigner(ctx context.Context, signer driver.Signer, o Observer, principal string, role Role) driver.Signer {
+	if signer == nil || o == nil || o == Nop {
+		return signer
+	}
+
+	base := instrumentedSigner{signer: signer, observer: o, principal: principal, role: role, ctx: ctx}
```
AkramBitar: **MEDIUM — capturing `ctx` in the wrapper pins a request-scoped context for the whole life of the signer.**

The doc comment above is right that `Sign` has no context of its own, but storing the resolution-time context and reusing it on every later `Sign` has two consequences beyond "the events have a context":

1. **Misattributed traces.** `logging.Logger.*fContext` derives trace/span fields from the context, so every future audit record for this signer is correlated to whatever span was open at *resolution* time. For long-lived signers that span is long dead, and the records attach to it rather than to the transaction actually being signed.
2. **Retention.** The context and everything reachable from its values stay alive as long as the signer does.

See the `provider.go:363` comment for why that lifetime is "process lifetime" in the common case.

If the goal is just to carry trace correlation, consider extracting the span context at wrap time into a plain value (or storing `context.WithoutCancel` of a stripped context) rather than retaining the full request context — or reporting these events against `context.Background()` and attaching the identity/role as fields, which is what the record actually needs.

token/services/identity/provider.go
```diff
@@ -264,30 +324,43 @@ func (p *Provider) areMe(ctx context.Context, identities ...driver.Identity) []s
...
 	start := time.Now()
 	signer, _, path, err := p.getSignerAndCache(ctx, identity, idHash, true)
 	p.metrics.GetSignerDuration.With("path", path).Observe(time.Since(start).Seconds())
 	if err == nil {
 		p.metrics.SignerResolutions.With("outcome", path).Add(1)
 	}
+	t.DoneResolution(ctx, path, err)
 
-	return signer, err
+	return sigobserve.InstrumentSigner(ctx, signer, p.observer, idHash, sigobserve.RoleUnknown), err
```
AkramBitar: Companion to the `decorator.go:30` comment — this is where the captured context becomes process-lifetime.

Signers are resolved during wallet construction (`token/services/identity/role/wfactory.go:115` and `:133`), and those wallets live for the process lifetime. The `ctx` handed to `InstrumentSigner` here is therefore the wallet-*creation* context, and every signature that wallet ever produces is reported under it.

So for the wallets that do most of the signing, the trace correlation on the audit records is not merely stale, it is constant — every record for the life of the node points at the startup span. Worth confirming that's not what the observability story intends.

token/services/identity/sigpolicy/stack.go
```diff
@@ -0,0 +1,138 @@
...
+	var auditObserver sigobserve.Observer
+	if logger != nil {
+		auditObserver = sigobserve.NewAuditLogger(logger)
+	}
+
+	reporting := sigobserve.Multi(metricsObserver, auditObserver)
+	escalator := throttle.New(cfg, throttle.WithObserver(reporting), throttle.WithLevelGauge(gauge))
+
+	s := &Stack{
+		observer:  sigobserve.Multi(metricsObserver, auditObserver, escalator),
```
AkramBitar: **LOW — `mode: off` doesn't restore the zero-cost path, because the escalator is unconditionally in the observer chain.**

`s.gate` is correctly guarded by `cfg.Enabled()` three lines down, but `observer` is not: `escalator` is always a non-`Nop` observer here. Since `InstrumentSigner`/`InstrumentVerifier` skip wrapping only when `o == nil || o == Nop`, the result with `mode: off` and no metrics/logger is that every signer and verifier is still wrapped, every `Sign`/`Verify` still allocates a `Timer` and walks the `Multi` chain, and `Escalator.Observe` is still called on each one just to return early.

That makes "off" mean "gate disabled" rather than "feature disabled", which is probably not what an operator turning it off expects — and it's the one configuration where you'd want to be able to say the feature costs nothing.

```go
observers := []sigobserve.Observer{metricsObserver, auditObserver}
if cfg.Enabled() {
    observers = append(observers, escalator)
}
```

With that, `Multi` collapses to `Nop` when metrics and logging are also absent, and the decorators drop out entirely. Worth a benchmark assertion in `provider_observe_bench_test.go` pinning the disabled path at zero allocations, so this doesn't regress silently.

token/core/fabtoken/v1/driver/driver.go
```diff
@@ -191,6 +191,9 @@ func (d *Driver) NewTokenService(tmsID driver.TMSID, publicParams []byte) (drive
 	if err != nil {
 		return nil, errors.WithMessagef(err, "failed to create token service")
 	}
+	// The stack gates this service's client-facing signature service and is released when the
+	// service is done with.
+	service.SetSignatureInstrumentation(sigStack)
```
AkramBitar, line 202: **LOW — `sigStack` goroutines leak on every `NewTokenService` failure after the wallet service is built.**

The comment is accurate for the success path: ownership transfers to the service at line 196 and is released with it. But `newWalletService` returns a *live* stack (escalator evict loop + `BucketSet` evict loop), and every error return between that call and this line drops it without `Stop()` — `driver.go:161` (`failed to initialize token service`), `:165` (`failed resolving validator resource limits`), and the `err != nil` on line 191 itself.

`TMSProvider.GetTokenManagerService` does not cache failures, so a TMS that fails init repeatedly — bad public params, unreachable storage — leaks two goroutines per attempt for as long as callers keep retrying.

Same pattern in `token/core/zkatdlog/nogh/v1/driver/driver.go` at `:172` and `:213`. A `defer` that stops the stack unless ownership was successfully transferred is the least invasive fix:

```go
transferred := false
defer func() {
    if !transferred {
        sigStack.Stop()
    }
}()
...
service.SetSignatureInstrumentation(sigStack)
transferred = true
```

token/services/identity/throttle/config.go
```diff
@@ -0,0 +1,204 @@
...
+	// Burst is the bucket capacity. Zero selects DefaultBurst; values below Rate are raised
+	// to Rate.
+	Burst float64 `yaml:"burst,omitempty"`
+	// Window is the evaluation period for the ratio thresholds. Zero selects DefaultWindow.
+	Window time.Duration `yaml:"window,omitempty"`
+	// MinSamples is the minimum number of observations in a window before a ratio can
+	// escalate a principal. Zero selects DefaultMinSamples.
+	MinSamples int `yaml:"minSamples,omitempty"`
+	// ErrorRateThreshold is the failing-operation fraction that escalates. Zero selects
+	// DefaultErrorRateThreshold; a value of 1 or more effectively disables this trigger.
```
AkramBitar: **LOW — the documented way to disable this trigger doesn't work.**

`Observe` compares with `>=`:

```go
float64(errCount)/float64(total) >= e.cfg.ErrorRateThreshold
```

so at exactly `1` a principal with a 100% error rate over `MinSamples` observations still escalates. Only a value strictly greater than 1 disables the trigger. An operator following the comment to switch it off gets the most aggressive possible setting instead — which, combined with the `escalate` defect, means a straight jump to `blocked`.

Same wording in `docs/configuration.md:924` and `docs/security/signature_observability.md:191`. `InvalidSignatureRateThreshold` uses the same comparison, so it's worth checking its docs too.

Either fix the comparison to `>` for the sentinel, or fix the docs:

```suggestion
	// DefaultErrorRateThreshold; a value greater than 1 disables this trigger.
```

AkramBitar (commented):
## Code review — signature observability & throttling

Reviewed at `c199202`. Overall the layering is clean (observer chain → escalator → gate) and the test coverage on the new packages is good. The findings below are mostly about the **escalation state machine** and the **lifetime/keying of per-principal state**, which is where this feature can turn into an availability problem rather than a defence against one.

Two findings I'd consider blocking:

| # | Severity | Where | What |
|---|---|---|---|
| 1 | **HIGH** | `throttle.go:378` | `escalate` ignores `levelUntil`, so a principal goes `normal → blocked` in two consecutive requests |
| 2 | **HIGH** | `sig.go:89` | Gated call sites key on a single fixed identity, making `rate: 200` a hard per-TMS throughput ceiling |
| 3 | MEDIUM | `throttle.go:500`, `bucket.go:252` | Escalated principals are never evicted → attacker-controlled unbounded memory growth |
| 4 | MEDIUM | `ws.go` (both drivers) | `sigStack.Stop()` kills eviction but leaves the escalator wired as an observer |
| 5 | MEDIUM | `audit.go:94` | Warn-level audit lines for routine control flow, and one per throttled request |
| 6 | MEDIUM | `decorator.go:30`, `provider.go:363` | Wrapped signer pins the resolution-time `ctx` for its whole lifetime |
| 7 | LOW | `stack.go:77` | `mode: off` does not restore the zero-cost path |
| 8 | LOW | `driver.go:192` (both drivers) | `sigStack` goroutines leak on every `NewTokenService` failure |
| 9 | LOW | `config.go:99` | Documented way to disable the error-rate trigger is off by one comparison |

Findings 1 and 2 compound: with the default config, a single momentary spike past `burst` on the auditing path blocks the auditor for `blockDuration`, and `AuditView` then fails every transaction with `failed verifying auditor signature`. Finding 1 is reproducible — details inline.

Details are in the inline comments.

AkramBitar (dismissed):

AkramBitar:
@adecaro 

I would like to ask your help to review this PR. 

Regards,
Akram

AkramBitar:
@HayimShaul 

Could you please fix the lint issues?

Regards,
Akram

token/services/identity/throttle/throttle.go
```diff
@@ -0,0 +1,524 @@
...
+	// quiet-period counter restarts) without pushing it to blocked. This preserves the
+	// graduated response: normal → soft (reduced quota) → blocked, where "soft" lasts at
+	// least SoftDuration before the next escalation can fire.
+	//
+	// A blocked principal is intentionally excluded from this guard: a fresh violation
+	// while blocked must re-arm the block deadline (there is no higher level, and extending
+	// the block is the correct response).
+	if p.level == LevelSoft && now.Before(p.levelUntil) {
+		p.lastViolation = now
+		return
```
adecaro, line 438: **Blocking: `make lint` fails here (`nlreturn`), which is why CI's `checks` job is red at this commit and `utest`/`itest`/`bench-check` never ran against `8d6dcc33e`.**

```
token/services/identity/throttle/throttle.go:392:3: return with no blank line before (nlreturn)
```

This is the exact line that fixes the escalation bug found in the earlier review round (verified locally: reverting it makes `TestEscalatorSoftDurationIsHonouredBeforeBlocking` fail again with the original symptom), so it's worth landing cleanly rather than letting the lint failure block the rest of the pipeline from ever running against it.

**Suggestion**:
```go
	if p.level == LevelSoft && now.Before(p.levelUntil) {
		p.lastViolation = now

		return
	}
```

adecaro (commented):
## Review: does this fully address #1643?

@HayimShaul

**Requesting changes: CI's `checks` job is currently failing on a lint error at the exact line that fixes the escalation bug this review's earlier round found, so `utest`/`itest`/`bench-check` haven't run against this commit yet.**

Akram already did a thorough pass on an earlier commit (12 inline threads, changes requested). I checked what's changed since then rather than re-reviewing from scratch.

```
go build ./...                                                    -> clean
go vet (throttle, ratelimit, sigobserve, sigpolicy, common, token) -> clean
go test -count=1 ./token/services/identity/... ./token/services/ratelimit/... ./token/core/common/...  -> ok, all packages
go test -race -run TestEscalator ./token/services/identity/throttle/...  -> ok, no data race
go test -coverprofile ./token/services/identity/throttle/...      -> 97.0% of statements
gofmt -l <touched files>                                           -> clean
```

I reverted the new `escalate` soft-duration guard (throttle.go:390-393) and re-ran
`TestEscalatorSoftDurationIsHonouredBeforeBlocking`: it fails exactly as expected ("request 12
must stay at soft" gets `blocked` instead), so the test pins the fix rather than just decorating
the diff. Restored, rebuilt clean.

CI's `checks` is red at this commit ([run](https://github.com/LFDT-Panurus/panurus/actions/runs/32626932356/job/97163575236)) on an `nlreturn` finding, and because `checks` gates the rest of the
pipeline here, `utest`/`itest`/`bench-check`/`publish-coverage` never ran against `8d6dcc33e` —
the green marks on the PR predate this commit's fix.

### What it fixes

| # | Finding | Verified |
|---|---|---|
| 1643.1 | Instrument Signer/Verifier operations with metrics | ✅ `sigobserve` + `identity.Metrics` cover op/role/outcome counts, latency, cache hit/miss, escalation counters — read in full |
| 1643.2 | Automated defensive response (quota reduction / temporary block) | ✅ `throttle.Escalator`'s `normal → soft → blocked` state machine, now with the soft-duration guard fixed and test-pinned |
| 1643.3 | Audit logging of caller identity, operation, outcome | ✅ `sigobserve.AuditLogger`, identity hashed rather than logged raw |
| 1643.4 | Configurable thresholds, not hardcoded | ✅ `throttle.Config` + `docs/configuration.md`, docs now match the `>1 disables` code behavior |

All four are substantively covered. The three HIGH defects from the earlier review round — the
escalate logic reaching `blocked` after one violation, the same defect via the unlocked
`buckets.Take` race, and `AuditorVerifier`/`GetSigner` turning the default quota into a per-node
throughput ceiling — are fixed and, for the first, test-pinned. Full detail in the write-up.

### Blocking 1 — `make lint` fails on the line that fixes the escalation bug

`throttle.go:392` trips `nlreturn` ("return with no blank line before"), which is why CI's
`checks` job is red at this commit and `utest`/`itest` never ran. It's a one-line fix, but it's
blocking merge right now, and I want to flag it rather than let it slip past on the strength of
the already-green `Analyze`/`build`/`benchmark` jobs.

### Non-blocking

- **Still open, not re-raised:** three of Akram's earlier MEDIUM findings remain unresolved in this
  commit and are worth another look before merge — unbounded retention of escalated principals
  that go silent (`throttle.go` `evictIdle` only reclaims `LevelNormal`), `Stop()` leaving the
  escalator wired as an observer with no eviction running in both `ws.go` factories, and
  `OutcomeError`/`OutcomeThrottled` logged at warn on what are, at two call sites, expected/routine
  outcomes.
- **Resolved since the earlier round:** the `sigpolicy/stack.go` mode-off cost (now backed by a
  test asserting the observer collapses to `Nop`), the goroutine leak on `NewTokenService` init
  failure in both fabtoken and zkatdlog `driver.go` (the suggested `transferred` defer-guard is now
  in both), and the `config.go` docs/code mismatch on the disabling threshold.

### Recommendation

Fix the `nlreturn` lint error so `checks` (and therefore `utest`/`itest`) actually run against this
commit, and decide whether the three still-open MEDIUM items from the earlier round need addressing
before merge or can be tracked as follow-up. Everything else checked out clean.

AkramBitar:
@HayimShaul 

Could you please fix lint issues? 

Regards,
Akram

adecaro (commented):
## Review: does this fully address #1643?

@HayimShaul

**Triage of the outstanding review round against `e6ead413` (current head): the two HIGH-severity concurrency/escalation bugs and the unbounded-retention issue from @AkramBitar's round are fixed and covered by new race-tested unit tests; the throughput-ceiling HIGH and five MEDIUM/LOW items are still open.**

Checked out `1643_observe_and_escalate_signer_verifier_throttling` at `e6ead413`, diffed it against `8d6dcc33e` (my last review round) to see exactly what changed, and ran:

```
go test ./token/services/identity/throttle/... -race -run 'TestEscalator' -v
  → 26/26 PASS, including TestEscalatorSoftDurationIsHonouredBeforeBlocking and
    TestEscalatorConcurrentQuotaExhaustionEscalatesOnce (new tests targeting the two HIGH findings)
```

CI (`checks`, `lint`, `utest`, all `itest` shards, `bench-check`, CodeQL, DCO) is green at this SHA — the `nlreturn` lint failure I flagged last round is resolved.

I'm not re-deriving @AkramBitar's findings from scratch here — replying in each of their threads with a fixed/still-open verdict instead of restating them.

### Fixed since the last round

| Finding | Verified |
|---|---|
| HIGH — `escalate` ignores `SoftDuration` | `escalate` now guards `p.level == LevelSoft && now.Before(p.levelUntil)` and absorbs the violation instead of escalating |
| HIGH — lock dropped around `Take` lets concurrent requests double-escalate | `decide` re-acquires `e.mu` and re-looks-up the principal via `principalFor` before calling `escalate`, so the second concurrent caller observes the updated level |
| MEDIUM — escalated principals retained forever | New `Config.MaxPrincipals` cap (default 10,000) with oldest-`LevelNormal` eviction, plus `evictIdle` now reclaims entries whose `levelUntil` has expired |
| MEDIUM — `bucket.go`'s stale-override companion | Covered by the same fix: `evictPrincipal` calls `e.buckets.ClearRate(id)` on both eviction paths, which clears `overridden` |

### Still open

- HIGH — default quota is a hard throughput ceiling on `AuditorVerifier`/`GetSigner` hot paths (`token/sig.go`)
- MEDIUM — `Stop()` leaves the escalator wired as an observer after the eviction loops die (both `ws.go` factories)
- MEDIUM — `OutcomeError`/`OutcomeThrottled` at warn level (routine-control-flow log flood) (`sigobserve/audit.go`)
- MEDIUM — signer wrapper captures resolution-time `ctx` for process lifetime (`sigobserve/decorator.go`, `provider.go:363`)
- LOW — `mode: off` doesn't restore the zero-cost path (`sigpolicy/stack.go`)
- LOW — `sigStack` goroutine leak on `NewTokenService` failure (`fabtoken/v1/driver/driver.go`, and the same pattern in the zkatdlog driver)
- LOW — `>=` vs `>` on the threshold-disable comparison (`throttle/config.go`)

### Recommendation

The two HIGH concurrency/escalation fixes and the retention fix are solid, tested, and safe to land as-is. I'd still want the throughput-ceiling HIGH resolved before merge — as documented today, a healthy auditor sustaining normal traffic gets treated as an attacker. The five MEDIUM/LOW items are reasonable to track as fast-follow if the team wants to unblock the fixed work first.

token/services/identity/throttle/throttle.go
adecaro: **Fixed** — verified against `e6ead413`. `escalate` now guards `p.level == LevelSoft && now.Before(p.levelUntil)` and absorbs the violation instead of escalating. `TestEscalatorSoftDurationIsHonouredBeforeBlocking` passes.

token/services/identity/throttle/throttle.go
adecaro: **Fixed** — verified against `e6ead413`. `decide` re-acquires `e.mu` and re-looks-up the principal via `principalFor` before calling `escalate`, so the second concurrent caller observes the already-updated level. `TestEscalatorConcurrentQuotaExhaustionEscalatesOnce` passes under `-race`.

token/sig.go
adecaro: **Still open** as of `e6ead413` — no changes to this file's gating logic or to `token/services/ttx/auditor.go` since the last review round. The `sig.go` diff in this range is unrelated (`AreMe`/`IsMe` gaining error returns for cache-tombstone/context-cancellation handling).

token/services/identity/throttle/throttle.go
adecaro: **Fixed** — verified against `e6ead413`. New `Config.MaxPrincipals` (default 10,000) evicts the oldest `LevelNormal` entry when the map is full, and `evictIdle` now also reclaims principals whose `levelUntil` has expired. `TestEscalatorPrincipalCapEvictsOldestNormal` and `TestEscalatorEvictIdleReleasesExpiredThrottledPrincipals` pass.

token/services/ratelimit/bucket.go
adecaro: **Fixed** by the companion change — `evictPrincipal` now calls `e.buckets.ClearRate(id)` on both eviction paths, which flips `overridden = false`, so the bucket-side pin is released whenever the escalator drops the principal.

token/core/fabtoken/v1/driver/ws.go
adecaro: **Still open** as of `e6ead413` — no diff to this file since the last review round.

token/core/zkatdlog/nogh/v1/driver/ws.go
adecaro: **Still open** as of `e6ead413` — no diff to this file since the last review round.

token/services/identity/sigobserve/audit.go
adecaro: **Still open** as of `e6ead413` — the `audit.go` diff in this range is purely `_, _ = b.WriteString(...)` errcheck wrapping, no change to the outcome→level mapping.

token/services/identity/sigobserve/decorator.go
adecaro: **Still open** as of `e6ead413` — no diff to this file since the last review round.

token/services/identity/provider.go
adecaro: **Still open** as of `e6ead413` — companion to the decorator.go comment; no change to this call site's context handling.

token/services/identity/sigpolicy/stack.go
adecaro: **Still open** as of `e6ead413` — the only change to this file is a `//nolint:ireturn` annotation, unrelated to the observer-filtering fix suggested here.

token/core/fabtoken/v1/driver/driver.go
adecaro: **Still open** as of `e6ead413` — the driver.go diff in this range is an unrelated authorization-multiplexer refactor; no `defer`/ownership-transfer fix around `sigStack.Stop()`.

token/services/identity/throttle/config.go
adecaro: **Still open** as of `e6ead413` — `ErrorRateThreshold`/`InvalidSignatureRateThreshold` comparisons in throttle.go are still `>=`; docs still say "greater than 1 disables".

Effi-S (approved):

AkramBitar (commented):
Review of the escalating throttle policy. Nice feature overall — the observability surface is thorough and the Godoc explaining *why* each op is or isn't gated is genuinely useful. `go build ./...` is clean and every touched package's tests pass, `-race` included on the four new packages.

One finding I'd treat as a blocker, two that affect whether the control keeps working over time, and six minor ones.

**Blocker — the rate limiter adds an unbounded attacker-driven allocation of its own.** `Escalator.principals` is carefully capped, but `BucketSet.buckets` is not, and its key is a remote-supplied identity. See the comment on `ratelimit/bucket.go`.

**Reliability of the control itself.** Findings 2 and 3 combine: a principal can escalate and then never de-escalate *or* be evicted, and once `MaxPrincipals` slots are held above `normal`, ratio-based detection — the main point of the feature — silently switches off for every newly seen identity. Pinning a slot only needs `burst+1` requests to reach `soft`, not a sustained blocked-level campaign.

The remaining six are small: one comment with wrong arithmetic, one unreachable branch, two doc inconsistencies, a missing config validation, and an observer left wired after `Stop()`.

Findings 1 and 2 were confirmed empirically with throw-away probes, not just by reading — details in the individual comments.

I'd hold merge on finding 1, and ideally 2 and 3 too since they determine whether this works as a security control a month after deployment.

token/services/ratelimit/bucket.go
```diff
@@ -0,0 +1,269 @@
...
+
+// bucketFor returns key's bucket, creating it full when the key is new and refilling it for
+// the time elapsed since its last update. Callers must hold s.mu.
+func (s *BucketSet) bucketFor(key string) *bucket {
+	now := s.now()
+	b, ok := s.buckets[key]
+	if !ok {
+		// A key not seen recently starts with a full bucket at the set's defaults.
+		b = &bucket{tokens: s.burst, last: now, rate: s.rate, burst: s.burst}
+		s.buckets[key] = b
```
AkramBitar, line 246: **high — the per-key bucket map is unbounded, which defeats the `MaxPrincipals` cap it works alongside.**

This insert is unconditional. `Escalator.principals` is deliberately capped (`MaxPrincipals`, default 10 000, with the whole detached-principal mechanism to enforce it), but `BucketSet.buckets` has no cap at all — the only reclamation is the 1-minute sweep against `idleTTL` (default 10m). And `decide` calls `Take(principalID)` for *every* principal, including detached ones; the detached-principal comments explicitly rely on the bucket being persistent.

The key is attacker-supplied. In `token/services/ttx/recipients.go:945` and `:1176` the node calls `SigService().OwnerVerifier(ctx, resp.RecipientData.Identity)` on an identity chosen by the remote peer, so every request bearing a fresh random identity allocates a bucket (~200 B with the 64-char hash key plus map overhead) that is retained for at least 10 minutes. At 1 000 distinct identities/s that is ~600 000 live buckets, >100 MB, with no upper bound.

This is reachable by default: `DefaultMode = ModeMonitor` makes `Config.Enabled()` true, so the gate is installed and `Take` runs on every metered op even though nothing is ever denied.

Confirmed empirically — with `MaxPrincipals: 10` and 5 000 distinct principals the result is `principals=10 buckets=5000`.

A defence against resource exhaustion shouldn't itself be an unbounded attacker-driven allocation. Either give `BucketSet` a capacity bound (it already exposes `Len()`), or only create a bucket when the principal is `attached`.

token/services/identity/throttle/throttle.go
```diff
@@ -0,0 +1,620 @@
...
+	return true, "principal exceeded its quota and is now at level " + string(level)
+}
+
+// Observe records the outcome of an operation and escalates the principal when the observed
+// failure ratios cross their thresholds.
+//
+// Every observed operation contributes to the window's sample count, so the ratios are fractions
+// of what the principal actually did. Denied operations are the exception: they never ran, and
+// counting them would let a throttled principal dilute the very ratio that throttled it.
+func (e *Escalator) Observe(ctx context.Context, ev sigobserve.Event) {
```
AkramBitar, line 286: **medium — a ratio-escalated principal can never de-escalate, so `identity_throttled_principals` latches permanently.**

`maybeDeescalate` is only reachable from `decide`, i.e. only on the four *gated* ops (`OwnerVerifier`, `IssuerVerifier`, `RegisterSigner`, `GetAuditInfo`). But escalation via `errorRateThreshold` / `invalidSignatureRateThreshold` is driven entirely by `Observe`, which is fed by the *ungated* ops (`sign`, `verify`, `get_signer`, `is_me`, `bind`, …) — and `Observe` never de-escalates.

So a principal whose traffic is all ungated escalates on a transient error burst and then stays there forever. The node's own long-term signing identity is exactly this case; it's ungated by design (`token/sig.go:53`, and documented at `docs/security/signature_observability.md:62-64`). `evictIdle` can't rescue it either: it requires `lastSeen.Before(cutoff)`, and `Observe` refreshes `lastSeen` on every call.

Verified: after a 100-event invalid-signature burst the principal is `soft`; after a further 200 simulated minutes of nothing but clean `OpSign`/`OutcomeOK` it is *still* `soft`, `Throttled() == (1, 0)`, and `evictIdle` does not reclaim it.

This contradicts `docs/security/signature_observability.md:161` ("A principal that goes `deescalateAfter` without a violation is restored one level at a time") and makes that doc's own suggested alert — `identity_throttled_principals{level="blocked"} > 0` — fire forever once tripped. It also pins the bucket override, which `BucketSet.evictIdle` skips by design, feeding the `bucket.go` finding.

Calling `maybeDeescalate` from `Observe` as well (or driving de-escalation from the window roll) would fix it.

token/services/identity/throttle/throttle.go
```diff
@@ -0,0 +1,620 @@
...
+
+	cap := e.cfg.MaxPrincipals
+	if cap > 0 && len(e.principals) >= cap {
+		e.evictOldestNormal()
+	}
+	// If the map is still at capacity (all entries are above LevelNormal), return the new
+	// principal without inserting it. The caller proceeds as if the principal is at normal
+	// with a full bucket — a safe conservative choice compared with evicting a throttled
+	// entry and silently restoring its quota.
+	if cap > 0 && len(e.principals) >= cap {
```
AkramBitar, line 412: **medium — saturating the map with cheap `soft` entries silently disables ratio escalation for every new principal.**

`evictOldestNormal` only considers `LevelNormal` entries. Once all `MaxPrincipals` slots hold entries above normal, this returns `attached=false`, and then both `decide` and `Observe` refuse to record samples or escalate at all — so the invalid-signature detector, which is the main point of this feature, is off for every newly seen identity.

The cost of pinning a slot is low: `burst+1` requests is enough to reach `soft`. It does not require driving a principal to `blocked`. And per the `Observe` finding above, a `soft` entry with ongoing ungated traffic is never released, so saturation is permanent rather than transient.

The rationale in this doc comment and in `DefaultMaxPrincipals` covers only the *blocked* case — "avoids silently restoring a blocked principal's quota" — which is sound. Extending non-evictability to `soft` is what makes the bypass cheap. Allowing `soft` entries to be evicted under pressure (keeping `blocked` pinned) would preserve the stated safety property without the bypass.

token/services/identity/throttle/config.go
```diff
@@ -0,0 +1,223 @@
...
+	// DefaultIdleTTL is how long per-principal state is kept after the last operation.
+	DefaultIdleTTL = 10 * time.Minute
+	// DefaultMaxPrincipals is the upper bound on the number of per-principal entries the
+	// Escalator will hold simultaneously. When this limit is reached and a new, unseen
+	// principal arrives, the idle LevelNormal entry with the oldest lastSeen timestamp is
+	// evicted to make room. If no LevelNormal entry is available the new principal is
+	// admitted without creating state, so it is treated as unthrottled — a conservative
+	// choice that avoids silently restoring a blocked principal's quota.
+	// At the default rate (200/s) and burst (400) an attacker needs 401 requests to pin
+	// one entry; 10 000 entries therefore cost ~50 000 requests and roughly 28 minutes of
```
AkramBitar: **low — the cost model is off by ~80x.**

"an attacker needs 401 requests to pin one entry; 10 000 entries therefore cost ~50 000 requests" — 10 000 × 401 is ~4 010 000 requests, not ~50 000.

The "roughly 28 minutes" figure in the next line implies ~2 400 req/s, which is consistent with 4 M and not with 50 k, so the minutes look right and only the request count is wrong. Worth fixing since this comment is the stated justification for the default.

token/services/ttx/auditor.go
```diff
@@ -301,6 +301,13 @@ func (a *AuditingViewInitiator) verifyAuditorSignature(context view.Context, sig
 	for _, auditorID := range a.tx.TokenService().PublicParametersManager().PublicParameters().Auditors() {
 		v, err := a.tx.TokenService().SigService().AuditorVerifier(context.Context(), auditorID)
 		if err != nil {
+			if errors.Is(err, token.SignatureThrottled) {
```
AkramBitar, line 304: **low — unreachable branch that changes the loop's semantics for a condition that can't occur.**

`AuditorVerifier` is explicitly ungated (`token/sig.go:39-48`), and this PR's own docs say so at `docs/security/signature_observability.md:54-61`: the deserializer is instrumented but never gated. Nothing below this call returns `SignatureThrottled`, so `errors.Is(err, token.SignatureThrottled)` is always false and this branch is dead.

More importantly, if it ever *did* fire it would be wrong: the early `return` aborts the whole auditor loop on the first auditor, where the pre-existing behaviour is to log at debug (line 311) and try the remaining auditors from the public parameters. Either drop the branch, or `continue` rather than `return` if you want it as future-proofing.

docs/security/signature_observability.md
```diff
@@ -0,0 +1,248 @@
...
+state *is* the record that they are throttled.
+
+## Handling a denial
+
+A denied operation returns an error wrapping the `token.SignatureThrottled` sentinel.
+Callers should treat it as "back off", distinct from "unknown identity" or "invalid
+signature":
+
+```go
+signer, err := signatureService.GetSigner(ctx, id)
```
AkramBitar: **low — this example contradicts the same document.**

The snippet illustrates the `token.SignatureThrottled` sentinel with `GetSigner`, but line 54 and lines 62-64 of this file state that `GetSigner` is **not gated** (and the Godoc at `token/sig.go:53` agrees). So this call can never return the sentinel, and a reader who copies the example gets dead code.

`OwnerVerifier` is one of the four actually-gated ops and would illustrate it correctly.

token/services/identity/throttle/throttle.go
```diff
@@ -0,0 +1,620 @@
...
+
+	if oldestID != "" {
+		e.evictPrincipal(oldestID, e.principals[oldestID])
+	}
+}
+
+// advanceWindow rolls the ring forward to cover the current time, clearing the slots that
+// have aged out. Callers must hold e.mu.
+func (e *Escalator) advanceWindow(p *principal) {
+	slotDuration := e.cfg.Window / windowSlots
```
AkramBitar, line 460: **low — a `Window` below `windowSlots` nanoseconds freezes the sliding window.**

For `Window < 6ns` this division yields 0, the guard on the next line returns immediately, and slots are therefore never cleared — `totals` accumulates for the entire process lifetime, so every principal eventually crosses `errorRateThreshold` on a single error and escalates.

`Config.Defaults()` only rejects `Window <= 0`. Given that this config deliberately rejects values it can't honour elsewhere (negative rate, burst below rate), enforcing a minimum of `windowSlots` — or validating `Window/windowSlots > 0` — belongs there rather than letting the window silently stop rolling.

docs/configuration.md
```diff
@@ -1089,4 +1089,74 @@ Default values:
...
\ No newline at end of file
+- `owner` identifies the replica holding each lease and **must be non-empty and unique per replica**: every lease query (acquire, release, renew, and the pre-write `AssertLocksHeld` check) is scoped by it. If several replicas shared one owner value, those predicates would match each other's rows and mutual exclusion would be lost cluster-wide, so an empty or blank resolved owner is rejected at startup rather than tolerated. A typical cause is a templated node configuration with `fsc.id` left unset. No owner is synthesized as a fallback, because an owner that changed on each restart would leave a replica unable to renew or release the leases it still holds. The `memory` backend has no owner and is unaffected.
+
+---
+
+### Optional: token.tms.<name>.identity.throttle
```
AkramBitar, line 1096: **low — `maxPrincipals` is undocumented.**

It's a supported YAML field (`token/services/identity/throttle/config.go:133`) and arguably has the most operationally surprising behaviour in the whole policy — at the cap, newly seen principals are silently untracked and ratio detection stops for them — yet it appears in neither the example block nor the parameter table here, and it's absent from `docs/security/signature_observability.md` too (grep finds no mention of it anywhere under `docs/`).

It needs an entry in this section's example plus a row in the parameter table at `docs/security/signature_observability.md:205`, ideally calling out the saturation behaviour so operators know to size it.

token/core/fabtoken/v1/driver/ws.go
```diff
@@ -145,4 +165,13 @@ func (d *WalletServiceFactory) NewWalletService(tmsConfig driver.Configuration,
...
```
AkramBitar, line 177: **low — the stack is stopped but the escalator stays wired as an observer.**

The comment above says instrumentation keeps working and only the goroutine is released, which is true, but it undersells the consequence: the identity provider and deserializer keep `sigStack.Observer()`, and in the default `monitor` mode that chain includes the escalator. So `Escalator.Observe` carries on populating `e.principals` while `evictLoop` and the bucket sweeper are both dead — the map grows monotonically to `MaxPrincipals` and never shrinks for the lifetime of the wallet service.

It's bounded, so not a leak in the serious sense, but it's retained state with no consumer: nothing ever reads those escalation levels on this path. Passing `Observer()` minus the escalator, or building the stack with `ModeOff` here, avoids it.

Same pattern at `token/core/zkatdlog/nogh/v1/driver/ws.go:66`.

AkramBitar (commented):
@Effi-S 

Please see my comments.

Regards,
Akram

token/services/ratelimit/bucket.go
Effi-S: `ratelimit.BucketSe`t now takes a `maxKeys` cap and evicts the least-recently-used unmetered bucket to make room. Overridden buckets (throttled principals' reduced quotas) are never evicted.
if the set is entirely overridden, a new key gets a transient, unstored full bucket, mirroring the escalator's existing "conservatively unthrottled once saturated" choice, and keeping the key space bounded. The Escalator passes `cfg.MaxPrincipals`, so both maps share one count bound.
A non-positive value disables both together.

token/services/identity/throttle/config.go
Effi-S: Fixed

token/services/ttx/auditor.go
Effi-S: continue rather than return (for future-proofing)

docs/security/signature_observability.md
Effi-S: Updated example

docs/configuration.md
Effi-S: +- **maxPrincipals**: Maximum number of per-principal entries (and token buckets) held at once, bounding the policy's memory footprint under a flood of distinct attacker-supplied identities. When the cap is reached and a new principal arrives, the oldest idle `normal` entry is evicted (or the oldest `
soft` entry when no `normal` one exists); `blocked` entries are never evicted. A negative value disables the cap (not recommended in production)

token/services/identity/throttle/throttle.go
Effi-S: validating Window/windowSlots > 0

token/services/identity/throttle/throttle.go
Effi-S: Fixed

token/core/fabtoken/v1/driver/ws.go
Effi-S: Adding an `atomic.Bool` 
Cost: one struct lingers inert in memory (no growth).
In the future we can implement a more robust solution

Effi-S:
@AkramBitar, All comments have been addressed. 
Please re-review

AkramBitar (commented):
## Code review: signature throttling observability + escalation

Reviewed commit `b89646f1e`. Build, `go vet` and `go test -race` pass on all touched packages.

The feature is well-tested and the layering (observe → policy → throttle) is clean, but I don't think it should merge as-is: **the escalation gate is wired into the wrong place**, and it doesn't cover the flood it was built for.

**Two findings I'd treat as blocking, and they share one root cause** — the drivers hand the validator the *same* deserializer that carries the escalation observer:

1. Ledger-supplied token owners become throttle principals, so anyone can escalate — and eventually `block` — an identity they don't control (remote DoS against a third party, and it contradicts the `throttle` package's own doc).
2. That same wiring puts a single per-TMS mutex plus a SHA-256 on every `Verify` during validation, serializing concurrent validation on the signature path.

Decoupling validation from escalation (an un-observed deserializer for validators) addresses both.

**Separately, the gate misses its primary case:** fresh pseudonyms get a fresh 400-token bucket every round, so the ttx recipient flood the PR describes is never throttled — while still allocating per-pseudonym state on the defender.

**Follow-up-able:** the `"<empty>"` guard that never fires, `SetRate` silently no-op'ing at `maxKeys`, the latched gauge after `Stop`, the dead auditor branch, and warn-log amplification.

Details inline — 9 comments, ordered roughly by severity.

---
*🤖 Review generated with [Claude Code](https://claude.com/claude-code)*

token/core/zkatdlog/nogh/v1/driver/driver.go
```diff
@@ -146,6 +146,12 @@ func (d *Driver) NewTokenService(tmsID driver.TMSID, publicParams []byte) (drive
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to initiliaze wallet service for [%s:%s]", tmsID.Network, tmsID.Namespace)
 	}
+	transferred := false
+	defer func() {
+		if !transferred {
+			sigStack.Stop()
+		}
+	}()
 	deserializer := ws.Deserializer
```
AkramBitar, line 155: **HIGH — validator-fed escalation lets an attacker block a third-party identity.**

This is the same `*common.Deserializer` instance that receives `SetObserver(sigStack.Observer())`, and it is then handed to the validator further down this function. The chain contains the `Escalator` that also gates `token.SignatureService`, so validation *drives* the gate's state.

In `validator_transfer.go:64-73` the principal is `tok.Owner` — the input token's owner, read from ledger state. That is a **victim the attacker merely names**, not the attacker. Each bogus transfer action contributes 2 samples for that victim (`owner_verifier`=ok, `verify`=invalid), so roughly 25 invalid transfer actions per minute referencing a victim's publicly-visible unspent tokens crosses `MinSamples=50` with an invalid ratio of 0.5, well over `InvalidSignatureRateThreshold=0.2`. The victim escalates to `soft`, and after `SoftDuration` to `blocked`.

In `enforce` mode the node then refuses that identity's own `OwnerVerifier`, `IssuerVerifier`, `GetAuditInfo` and `RegisterSigner` calls with `SignatureThrottled` — a remote, cheap DoS against an arbitrary identity.

This also contradicts the package doc at `throttle/throttle.go:23-26` ("it must not be applied inside driver validators... denying them based on local per-node call history"). The gate isn't *invoked* during validation, but its state is *driven* by validation, which makes gate decisions per-node and attacker-steerable.

Suggested direction: give the validator its own un-observed deserializer, so escalation is fed only by this node's client-facing signature paths. That also resolves the lock-contention finding on `deserializer.go`.

token/core/fabtoken/v1/driver/driver.go
```diff
@@ -144,6 +144,12 @@ func (d *Driver) NewTokenService(tmsID driver.TMSID, publicParams []byte) (drive
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to initiliaze wallet service for [%s:%s]", tmsID.Network, tmsID.Namespace)
 	}
+	transferred := false
+	defer func() {
+		if !transferred {
+			sigStack.Stop()
+		}
+	}()
 	deserializer := ws.Deserializer
```
AkramBitar, line 153: **HIGH — same validator-fed escalation issue as the zkatdlog driver** (see the detailed comment on `token/core/zkatdlog/nogh/v1/driver/driver.go`).

This observed deserializer is passed to `validator.NewValidator(...)` at line 165, so ledger-supplied token owners become throttle principals here too.

token/sig.go
```diff
@@ -146,3 +231,20 @@ func (s *SignatureService) GetAuditInfo(ctx context.Context, ids ...Identity) ([
...
 }
+
+// allow consults the gate and reports a denial as a throttled event. It returns nil when no
+// gate is installed, so an unconfigured service behaves exactly as before.
+func (s *SignatureService) allow(ctx context.Context, op sigobserve.Op, role sigobserve.Role, id Identity) error {
+	if s.gate == nil {
+		return nil
+	}
+
+	principal := id.UniqueID()
```
AkramBitar, line 279: **MEDIUM — the gate is bypassed by pseudonym rotation on the very path it protects.**

`allow` keys the bucket on `id.UniqueID()`. On the ttx recipient path the identity is a fresh pseudonym chosen by the *requester*: `ttx/recipients.go:1176` reaches `OwnerVerifier(ctx, recipientData.Identity)`, and `recipients.go:640` / `:751` reach `RegisterSigner`.

A flooder that sends a new pseudonym per round therefore always gets a brand-new full bucket (`burst`=400) and is never throttled or escalated — while each attempt still allocates a `BucketSet` entry and an `Escalator.principals` entry, so the memory cost lands on the defender.

The PR's stated motivation ("an identity flooding the node with invalid-signature calls") is exactly this path, so as written the feature does not cover its primary case. Keying on something more stable than the pseudonym (enrollment ID, or the sender's session/peer identity) would be needed.

token/services/identity/throttle/throttle.go
```diff
@@ -0,0 +1,692 @@
...
+// thresholds can be tuned against production traffic before they bite.
+//
+// An empty principal is never throttled: without attribution, a shared bucket would let
+// unrelated callers throttle each other.
+func (e *Escalator) Allow(ctx context.Context, principalID string, op sigobserve.Op) error {
+	// A stopped Escalator is inert (see Stop): it neither denies nor mutates state, so a gate
+	// that is still consulted after teardown lets operations through rather than escalating a
+	// principal whose state could never be reclaimed. decide is reached only from here, so this
+	// guard covers it too.
+	if e.inert.Load() || !e.cfg.Enabled() || principalID == "" {
```
AkramBitar: **MEDIUM — the "empty principal is never throttled" guard never fires.**

`Allow` and `Observe` both guard on `principalID == ""`, but every caller derives the principal from `view.Identity.UniqueID()`, which returns the literal string `"<empty>"` for a zero-length identity (fabric-smart-client `platform/common/services/identity/identity.go:24-27`).

So all empty/nil identities share one bucket and one principal entry — precisely the "shared bucket would let unrelated callers throttle each other" outcome this guard documents. It is reachable from `SignatureService.RegisterSigner`, where `allow` runs *before* `checkSignerIdentity` rejects the empty identity (`token/sig.go:133-137`), and from `OwnerVerifier`/`GetAuditInfo` with a wire-supplied `RecipientData.Identity`. 401 requests with an empty identity push `"<empty>"` to `soft`.

Either compare against the `"<empty>"` sentinel as well, or have callers pass the raw identity length alongside the ID.

token/services/ratelimit/bucket.go
```diff
@@ -0,0 +1,315 @@
...
+//
+// The current balance is clamped to the new capacity, so lowering a quota cannot hand the
+// key more tokens than the new bucket holds; raising it never grants the difference
+// retroactively either, the bucket simply refills faster from where it is. A key with an
+// override is kept until ClearRate is called, so idle eviction cannot restore the default
+// quota behind the caller's back.
+//
+// A non-positive rate is ignored: an unmetered exception for a single key would be a
+// footgun, and the caller that wants one can stop consulting the set for that key.
+func (s *BucketSet) SetRate(key string, rate, burst float64) {
```
AkramBitar, line 158: **MEDIUM — `SetRate` can silently drop a quota override.**

`SetRate` mutates whatever `bucketFor` returns, but `bucketFor` (line 231, the `return b` at 244) hands back a **transient, unstored** bucket when the set is at `maxKeys` and `evictOneForRoom` finds nothing evictable — i.e. when every bucket is overridden. The `rate`, `burst` and `overridden` writes then land on a value that is immediately garbage.

Concrete scenario: `maxKeys == MaxPrincipals == 10000` and all 10000 buckets carry overrides (10000 principals at soft/blocked). A principal escalated via the ungated `Observe` path (error/invalid ratio, no prior `Take`, so no bucket yet) gets `transition` → `LevelSoft`; the per-level count and the `identity_throttled_principals` gauge are bumped and an escalation event is emitted — **but its quota is untouched**. Enforcement silently does nothing while reporting that it is in force.

`ClearRate` on such a key is likewise a no-op. `SetRate`/`ClearRate` should either report that they could not store the override, or force room for an override write.

token/core/common/deserializer.go
```diff
@@ -35,22 +40,47 @@ func NewDeserializer(
...
 // GetAuditorVerifier returns the verifier associated to the passed auditor identity.
 func (d *Deserializer) GetAuditorVerifier(ctx context.Context, id driver.Identity) (driver.Verifier, error) {
-	return d.auditorDeserializer.DeserializeVerifier(ctx, id)
+	return d.getVerifier(ctx, d.auditorDeserializer, id, sigobserve.OpAuditorVerifier, sigobserve.RoleAuditor)
+}
+
+// getVerifier resolves a verifier through vd, reports the resolution as op, and wraps the result
+// so that the verifications it performs are reported too.
+func (d *Deserializer) getVerifier(ctx context.Context, vd driver.VerifierDeserializer, id driver.Identity, op sigobserve.Op, role sigobserve.Role) (driver.Verifier, error) {
+	principal := id.UniqueID()
```
AkramBitar, line 78: **MEDIUM — unconditional per-resolution hashing, plus a single global mutex on the transaction-validation hot path.**

`getVerifier` computes `id.UniqueID()` (SHA-256 + base64 of the full identity) and two `time.Now()` calls for **every** verifier resolution, regardless of whether an observer is actually installed. Worth gating on `d.observer != sigobserve.Nop`.

The bigger cost is contention. On a live TMS the observer *is* installed on the validator's deserializer (see the comment on `zkatdlog/.../driver.go`), so every verifier resolution and every `Verify` during validation also takes `Escalator.mu` — one mutex per TMS guarding `principals` and `counts`, held across the window roll, `maintainLevel` and the ratio checks. Concurrent validation/endorsement of transactions is therefore serialized on a single lock on the signature path.

The PR's benchmark (1401 → 2167 ns/op) is single-goroutine and does not measure this; a parallel benchmark would.

Also note `validator_transfer.go:62` already computes `driver.Identity(tok.Owner).UniqueID()`, so the same hash is now computed twice per input.

token/services/ttx/auditor.go
```diff
@@ -301,6 +301,15 @@ func (a *AuditingViewInitiator) verifyAuditorSignature(context view.Context, sig
...
```
AkramBitar, line 304: **LOW — dead branch.**

`SignatureService.AuditorVerifier` is explicitly and deliberately *not* gated (`token/sig.go:91-97`), so it can never return an error satisfying `errors.Is(err, token.SignatureThrottled)`.

Even if it could, this branch and the pre-existing one below both `continue`, so the only observable effect would be a different debug message. As written it reads as though a real recovery path exists where none does.

token/services/identity/throttle/throttle.go
```diff
@@ -0,0 +1,692 @@
...
+// disabled Escalator. It is idempotent.
+//
+// Inertness is required, not merely convenient. An Escalator may outlive its Stop while still
+// wired into a live observer chain — a wallet-only service stops the stack the moment it is
+// built yet keeps the deserializer and identity-provider instrumentation reporting to it (see
+// zkatdlog/fabtoken driver ws.go). Its eviction goroutines are gone by then, so any state a
+// post-Stop Observe or Allow accumulated (principal windows, bucket overrides, per-level counts
+// and the gauge) could never be reclaimed and would grow without bound. Refusing to mutate
+// after Stop closes that leak at its source rather than relying on every caller to unwire first.
+func (e *Escalator) Stop() {
```
AkramBitar, line 386: **LOW — `Stop` leaves the throttled-principals gauge latched.**

`Stop` sets `inert`, closes `stopped` and stops the buckets, but never re-reports the gauge. After `Service.Done()` → `Stack.Stop()` with, say, 3 principals at `blocked`, `identity_throttled_principals{level="blocked"}` keeps reporting 3 for the rest of the process lifetime: `evictPrincipal` (the only other writer) can no longer run, and `Allow`/`Observe` are inert.

An alert on `identity_throttled_principals > 0` would fire permanently after a TMS teardown. `Stop` should `SetThrottledPrincipals(soft, 0)` and `SetThrottledPrincipals(blocked, 0)`.

token/services/identity/sigobserve/audit.go
AkramBitar, line 95: **LOW — unbounded warn-level log amplification under the attack being detected.**

`levelFor` maps every `OutcomeInvalid`, `OutcomeError` and `OutcomeThrottled` to `WarnLevel`, which is enabled in production, so the level probe does not suppress it: one full record is rendered and written per rejected signature.

Combined with the validator-fed escalation issue, an attacker pushing invalid signatures through validation produces one warn line per attempt with no rate limiting of its own — the observability feature becomes a log-volume amplifier for exactly the flood it is meant to surface. Worth a per-principal suppression or sampling on the warn path.

token/core/zkatdlog/nogh/v1/driver/driver.go
Effi-S: `zkatdlog` validator gets a fresh un-observed `NewDeserializer(pp)`  instead of `ws.Deserializer`

token/core/fabtoken/v1/driver/driver.go
Effi-S: fabtoken validator gets NewDeserializer() instead of `ws.Deserializer`

token/sig.go
Effi-S: PrincipalKeyResolver seam + WithPrincipalKeyResolver; residual limitation documented

token/services/identity/throttle/throttle.go
Effi-S: unattributed() helper catches both "" and "<empty>" in the Allow/Observe guards

token/services/ratelimit/bucket.go
Effi-S: bucketForOverride always stores the override (bounded overflow when all buckets overridden)

token/core/common/deserializer.go
Effi-S: getVerifier Nop short-circuit skips hash + time.Now(); removes validator lock contention

token/services/ttx/auditor.go
Effi-S: dead SignatureThrottled branch removed

token/services/identity/throttle/throttle.go
Effi-S: Stop zeroes the throttled-principals gauge for both levels

token/services/identity/sigobserve/audit.go
Effi-S: goroutine-free per-principal warnSampler caps warn-log amplification

PR #2085, opened 2026-07-31
fix(recovery): derive the recovery lock id per TMS

atharrva01:
Follow-up to #1982, where you asked for the same fix on the other locks.

Recovery managers are created per TMS but all took the same advisory lock id from config, so
several TMSes on one postgres competed for a single lock and every one but the winner skipped
its sweep. Pending transactions on those TMSes just never got recovered.

The id now derives from the requests table name, which already carries network, channel and
namespace. Worth flagging that the table prefix on its own is not enough here: TMSes normally
share a persistence config and are told apart only by those params, so a prefix-derived id
still collides. `TestRecoveryLockID_DistinctPerTMSOnSharedPrefix` pins exactly that, and I
checked it fails against the prefix version.

`AcquireRecoveryLeadership` no longer takes a lockID, the factory is bound to its id when the
store is built. `AdvisoryLockID` is dropped from the config since there is nothing left to keep
unique by hand. Extra yaml keys are ignored on load so existing configs won't break, the setting
just stops doing anything. Happy to keep it as an override instead if you'd rather.

One thing I left alone: `AuditTransactionStore` passes a nil recovery factory, so its
`AcquireRecoveryLeadership` always grants locally and audit recovery has no cross-replica
coordination at all. That is pre-existing and looked like its own issue rather than widening
this one.

### Breaking change

This changes exported interfaces, so out-of-tree implementors will not compile until they drop the
`lockID` parameter: `driver.TransactionStore`, `driver.AuditTransactionStore`, `driver.TokenStore`,
`recovery.Storage`, `cleanup.Storage`. `recovery.Config` and `cleanup.Config` also lose their
`AdvisoryLockID` field. Existing yaml keeps working, a stale `advisoryLockID:` is simply ignored,
which `TestLoadConfig_IgnoresStaleAdvisoryLockID` covers.

Keystore cleanup is fixed here too, same shape, since it had the identical bug and is started per
TMS right beside recovery.

No separate tracking issue for this one, it's the direct follow-up requested in #1982's comments
rather than a fresh report, so there isn't a standalone bug report to link.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -253,14 +253,6 @@ token:
               # Relationship: Should be greater than the expected network latency + processing time for transaction status queries.
               leaseDuration: 30s
               
-              # advisoryLockID is the PostgreSQL advisory lock identifier used for recovery leader election.
-              # This ensures only one replica performs recovery sweeps at a time in multi-instance deployments.
-              # Default: 8389190333894887286 (hex: 0x74746b7265636f76, ASCII: "ttkrecov")
-              # The default value is derived from the ASCII encoding of "ttkrecov" (Token Transaction Recovery).
-              # Only change this if you need to run multiple independent recovery managers on the same database.
-              # Note: PostgreSQL advisory locks use 64-bit integers. This value must be unique across your application.
-              advisoryLockID: 8389190333894887286
-              
               # instanceID identifies this replica as the owner of recovery claims.
               # If empty, a process-local identifier is generated automatically at startup using a UUID.
               # Set this explicitly in containerized environments to maintain consistent identity across restarts.
@@ -314,14 +306,6 @@ token:
             # Decrease to reduce resource consumption on constrained systems.
             # Performance impact: More workers increase CPU utilization during cleanup sweeps.
             workerCount: 1
-            
-            # advisoryLockID is the PostgreSQL advisory lock identifier used for cleanup leader election.
-            # This ensures only one replica performs cleanup sweeps at a time in multi-instance deployments.
-            # Default: 8389190333894887277 (hex: 0x74746b636c65616e, ASCII: "ttkclean")
-            # The default value is derived from the ASCII encoding of "ttkclean" (Token Transaction Keystore Cleanup).
-            # Only change this if you need to run multiple independent cleanup managers on the same database.
-            # Note: PostgreSQL advisory locks use 64-bit integers. This value must be unique across your application.
-            advisoryLockID: 8389190333894887277
             
             # instanceID identifies this replica in logs and monitoring.
             # If empty, a unique identifier is generated automatically at startup.
@@ -657,7 +641,6 @@ token:
               batchSize: 100
               workerCount: 4
               leaseDuration: 30s
-              advisoryLockID: 8389190333894887286
               instanceID:
               notFoundGracePeriod: 30m
 ```
@@ -670,7 +653,6 @@ Default values:
 - batchSize: 100
 - workerCount: 4
 - leaseDuration: 30s
-- advisoryLockID: 8389190333894887286 (`0x74746b7265636f76`)
 - instanceID: empty, auto-generated when the recovery manager starts
 - notFoundGracePeriod: 30m (set to 0 to disable promotion to Orphan)
 
@@ -679,7 +661,7 @@ Default values:
 - **Recovery is enabled by default** to ensure automatic recovery of lost finality listeners.
 - **Only pending transactions older than `ttl` are considered for recovery** to avoid interfering with active transaction processing.
 - **The manager validates** that `ttl`, `scanInterval`, `batchSize`, `workerCount`, and `leaseDuration` are all greater than zero.
-- **`advisoryLockID`** is used to acquire PostgreSQL advisory-lock leadership so that only one replica performs a recovery sweep at a time. The default value (8389190333894887286 or 0x74746b7265636f76) represents the ASCII string "ttkrecov" (Token Transaction Recovery) encoded as a 64-bit integer.
+- **Recovery leadership** uses a PostgreSQL advisory lock so that only one replica performs a recovery sweep at a time, on the owner transaction path. The lock identifier is no longer configurable: it is derived from the TMS's own `requests` table name, which already carries the network, channel and namespace. That makes it unique per TMS automatically, so several TMSes sharing one PostgreSQL database no longer compete for a single lock, and there is nothing left for an operator to keep unique by hand. The audit transaction store does not yet participate in this coordination — its recovery leader factory is nil, so every replica is granted leadership and the audit sweep runs redundantly on each of them; see [#2143](https://github.com/LFDT-Panurus/panurus/issues/2143).
 - **`instanceID`** is used as the lease owner identifier for claimed transactions; if omitted, the manager generates a unique UUID automatically at startup.
 - **`notFoundGracePeriod`** controls how long a transaction whose status query keeps returning `NotFound` is left in `Pending` before being promoted to the terminal `Orphan` status. This protects the recovery sweep from being permanently blocked by transactions whose audit log was persisted but whose broadcast never reached the ledger. The promoted row is marked `Orphan` rather than `Deleted` so operators can distinguish broadcast failures from ledger-rejected transactions. Raise the default if your network has long catch-up windows after committer/orderer restarts; set it to `0` to disable the promotion entirely.
 
@@ -784,7 +766,6 @@ token:
             scanInterval: 1h
             batchSize: 100
             workerCount: 1
-            advisoryLockID: 8389190333894887277
             instanceID:
 ```
 
@@ -795,7 +776,6 @@ Default values:
 - scanInterval: 1h
 - batchSize: 100
 - workerCount: 1
-- advisoryLockID: 8389190333894887277 (`0x74746b636c65616e`)
 - instanceID: empty, auto-generated when the cleanup manager starts
 
 **Parameter Relationships and Tuning:**
@@ -803,7 +783,7 @@ Default values:
 - **Cleanup is disabled by default** and must be explicitly enabled. This is a conservative default to prevent unexpected key deletion in existing deployments.
 - **Only deleted tokens older than `ttl` are considered for cleanup** to ensure tokens are truly finalized before key deletion.
 - **The manager validates** that `ttl`, `scanInterval`, `batchSize`, and `workerCount` are all greater than zero.
-- **`advisoryLockID`** is used to acquire PostgreSQL advisory-lock leadership so that only one replica performs a cleanup sweep at a time. The default value (8389190333894887277 or 0x74746b636c65616e) represents the ASCII string "ttkclean" (Token Transaction Keystore Cleanup) encoded as a 64-bit integer.
+- **Cleanup leadership** uses a PostgreSQL advisory lock so only one replica sweeps at a time. The identifier is not configurable: it is derived from the TMS's own token SKI cleanup table name, which already carries network, channel and namespace, so it is unique per TMS automatically.
 - **`instanceID`** is used to identify this replica in logs and monitoring; if omitted, the manager generates a unique identifier automatically at startup.
 
 **Tuning Recommendations:**
@@ -825,7 +805,6 @@ Default values:
 
 4. **For Multi-Instance Deployments:**
    - **PostgreSQL Required**: Multi-instance deployments require PostgreSQL for distributed coordination via advisory locks
-   - Keep default `advisoryLockID` unless running multiple independent cleanup systems
    - Consider setting explicit `instanceID` values for easier debugging and monitoring
 
 5. **For Single-Node Deployments:**
@@ -910,7 +889,6 @@ multiply `acquireDeadline` there.
 
 4. **For Multi-Instance Deployments:**
    - **PostgreSQL Required**: Multi-instance deployments require PostgreSQL for distributed coordination via advisory locks
-   - Keep default `advisoryLockID` unless running multiple independent recovery systems
    - Consider setting explicit `instanceID` values for easier debugging and monitoring
    - Ensure all instances share the same PostgreSQL database for proper coordination
 
--- a/docs/services/storage/keystore_cleanup.md
+++ b/docs/services/storage/keystore_cleanup.md
@@ -105,7 +105,7 @@ The Storage interface abstracts database operations needed for cleanup:
 ```go
 type Storage interface {
     // AcquireCleanupLeadership obtains distributed lock for leader election
-    AcquireCleanupLeadership(ctx context.Context, lockID int64) (Leadership, bool, error)
+    AcquireCleanupLeadership(ctx context.Context) (Leadership, bool, error)
     
     // GetDeletedTokensPendingSKICleanup queries deleted, owned tokens older than TTL that haven't been cleaned.
     // Tokens for which this node is only an auditor or issuer are excluded, since this node
@@ -168,7 +168,6 @@ services:
       scanInterval: 1h           # How often to scan
       batchSize: 100            # Max tokens per scan
       workerCount: 1            # Parallel workers (default: 1)
-      advisoryLockID: 0x74746b636c65616e  # Lock ID for leader election ("ttkclean" in hex)
       instanceID: "cleanup-1"   # Instance identifier (auto-generated if empty)
 ```
 
@@ -179,8 +178,12 @@ services:
 - **scanInterval**: How frequently the manager scans for eligible tokens
 - **batchSize**: Maximum number of tokens processed in a single sweep
 - **workerCount**: Number of parallel workers processing tokens within a sweep (default: 1)
-- **advisoryLockID**: PostgreSQL advisory lock ID for distributed coordination
 - **instanceID**: Identifies the cleanup instance; auto-generated as `cleanup-<pointer>` if not provided
+
+Leader election uses a PostgreSQL advisory lock whose identifier is not configurable. It is derived
+from the TMS's own token SKI cleanup table name, which already carries network, channel and
+namespace, so it is unique per TMS automatically and several TMSes sharing one database no longer
+compete for a single lock.
 
 **Configuration Loading:**
 
@@ -203,7 +206,6 @@ config := cleanup.Config{
     ScanInterval:    1 * time.Hour,
     BatchSize:       100,
     WorkerCount:     4,
-    AdvisoryLockID:  0x74746b636c65616e,
     InstanceID:      "cleanup-instance-1",
 }
 
@@ -390,7 +392,6 @@ The `token_ski_cleanups` table is automatically created by the schema initializa
 - **Scan Interval**: 1 hour (less aggressive than recovery's 5 seconds)
 - **Batch Size**: 100 tokens per sweep
 - **Worker Count**: 1 parallel worker (conservative default)
-- **Advisory Lock ID**: `0x74746b636c65616e` ("ttkclean" in hex)
 - **Instance ID**: Auto-generated as `cleanup-<pointer>` if not provided
 
 ### Tuning Recommendations
@@ -412,7 +413,6 @@ The `token_ski_cleanups` table is automatically created by the schema initializa
 
 4. **For Multi-Instance Deployments:**
    - **PostgreSQL Required**: Multi-instance deployments require PostgreSQL for distributed coordination
-   - Keep default `advisoryLockID` unless running multiple independent cleanup systems
    - Consider setting explicit `instanceID` values for easier debugging and monitoring
 
 ### Performance Considerations
--- a/token/services/storage/db/sql/postgres/advisorylock.go
+++ b/token/services/storage/db/sql/postgres/advisorylock.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/LFDT-Panurus/panurus/token/services/logging"
 	tokensdriver "github.com/LFDT-Panurus/panurus/token/services/storage/db/driver"
+	common5 "github.com/LFDT-Panurus/panurus/token/services/storage/db/sql/common"
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/common/utils"
 )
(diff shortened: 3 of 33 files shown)


atharrva01:
hi @AkramBitar , @adecaro  , would love your thoughts on this...

token/services/storage/db/sql/postgres/advisorylock.go
```diff
@@ -115,6 +115,20 @@ func NewAdvisoryLockFactory() func(context.Context, *sql.DB, int64) (tokensdrive
 	}
 }
 
+// NewAdvisoryLockFactoryForID returns a recovery leader factory bound to lockID. Binding the id
+// at construction keeps it out of the call path, so no caller has to know how a lock id is made
+// unique and none can accidentally pass one that is shared with another TMS.
+func NewAdvisoryLockFactoryForID(lockID int64) func(context.Context, *sql.DB) (tokensdriver.RecoveryLeadership, bool, error) {
```
AkramBitar, line 118: Now that this exists, the old `NewAdvisoryLockFactory` just above (line 107) has no production callers left — `transactions.go` uses this one, and the only remaining use is `advisorylock_test.go:170`. Can we delete it? Keeping it exported keeps the caller-supplied lock ID around, which is the footgun this function's comment says we're removing.

token/services/storage/db/sql/postgres/recovery_lockid_test.go
```diff
@@ -0,0 +1,70 @@
...
+func tmsTables(t *testing.T, prefix, network, channel, namespace string) sqlcommon.TableNames {
+	t.Helper()
+	tables, err := sqlcommon.GetTableNames(prefix, network, channel, namespace)
+	require.NoError(t, err)
+
+	return tables
+}
+
+// recoveryLockID mirrors the derivation in NewTransactionStoreWithNotifier.
+func recoveryLockID(tables sqlcommon.TableNames) int64 {
```
AkramBitar: This re-implements the derivation instead of calling it, so all four tests still pass if the string in `transactions.go:77` changes. Could we export a small `recoveryLockID(tables)` and have both use it?

AkramBitar:
@atharrva01 

Great work, thanks a lot!

Two things beyond the inline comments:

- Cleanup still has the same bug: `cleanup/config.go:39` uses one shared `defaultLockID`, `cleanup/manager.go:208` still passes it, and the cleanup manager is per-TMS and started right next to recovery (`network.go:483`). Since this PR is framed as the follow-up for "the other locks", could you either fix it here or open an issue and note it in the description?

- Worth adding to the description: this breaks the exported `TransactionStore`, `AuditTransactionStore` and `recovery.Storage` interfaces plus `recovery.Config`, so downstream implementors won't compile. And a small test that `LoadConfig` still succeeds with a stale `advisoryLockID:` key would back up the "existing configs won't break" claim.

Agree on leaving audit recovery alone — worth its own issue though.

atharrva01:

```
```

atharrva01:
thanks, all four addressed.

**cleanup**: fixed here rather than in a follow-up, since you're right that the PR is framed as the one for "the other locks". `keystoreCleanupLockID` derives from `tables.TokenSKICleanups`, `AcquireCleanupLeadership` drops its lockID param, and `AdvisoryLockID` is gone from `cleanup.Config` too. It sits on `TokenStore` rather than `TokenLockStore`, so it doesn't collide with #1982.

**`NewAdvisoryLockFactory`**: deleted, and `NewCleanupLeaderFactory` with it for the same reason. You were right that the only caller left was the test.

**the mirrored derivation**: fair hit, that's exactly the weakness I flagged on #1982 and I wrote the same thing here. The tests now call the production `recoveryLockID(tables)`, so changing the string breaks them.

**stale config key**: added `TestLoadConfig_IgnoresStaleAdvisoryLockID` with a testdata config that still carries `advisoryLockID`, asserting the load succeeds and the rest of the block still applies.

On the breaking change, added to the description: this changes `driver.TransactionStore`, `driver.AuditTransactionStore`, `driver.TokenStore`, `recovery.Storage`, `cleanup.Storage`, `recovery.Config` and `cleanup.Config`, so anyone implementing those outside the repo will need to drop the parameter.

Audit recovery I've left alone as agreed, will open an issue for it.

AkramBitar:
@atharrva01 thanks a million. 

@adecaro I would like to ask your review on that one as well. Could you please have a look once you have time?

Thanks a lot,
Akram

atharrva01:
opened #2143 for the audit recovery one, as mentioned above.

turned out to be a bit more than just the nil factory: `AuditTransactionStore` also doesn't override `ClaimPendingTransactions`, so it inherits the permissive common one rather than the atomic `UPDATE ... RETURNING` that `postgres.TransactionStore` uses. so on that path there's no leader election and no claim persistence, and every replica sweeps the same pending audit transactions each tick.

atharrva01:
@adecaro , a gentle ping on this , whenever you have time , thanks

atharrva01:
Pushed a fix, this branch was not compiling.

Renaming `NewCleanupLeaderFactory` to `NewCleanupLeaderFactoryForID` covered the keystore cleanup path but missed `postgres/tokenlock.go`, which is the token lock cleanup from #1982 and calls the same factory. That file is not otherwise touched here so it slipped through. `cgo-check`, `checks` and `Analyze` were all failing on it, and `utest` and `itest` were being skipped because they need `checks`, so none of the tests on this PR had actually run.

`tokenlock.go` already derived its id per TMS, it just kept it in a field and passed it on every call. It now binds it at construction like the other two, through a named `tokenLockCleanupLockID` so the three derivations sit together in `advisorylock.go`.

Two assertions in `tokenlock_leadership_test.go` referenced the `cleanupLockID` field and now call the derivation directly. I dropped the one checking that two replicas of the same TMS derive the same id: once the id is bound at construction that only restated the constructor, and the concurrent acquire right below it already proves the two contend over one lock.

`cgo-check`, `Analyze` and `build` are green again, `checks` is still running.

@adecaro whenever you have time, this one is ready. I have the audit side ready as a follow-up on top of it: `AuditTransactionStore` still passes a nil factory, which is the pre-existing gap I mentioned above and filed as #2143. It needs the per-store id from this PR to avoid contending with owner recovery, so I am holding it until this lands.

adecaro (commented):
**Requesting changes — one blocking CI failure; the fix itself is correct and well tested.** Verified locally (not just from the diff): the lock-id derivation is right, every affected interface/mock/test was updated consistently, and `recovery_lockid_test.go` genuinely pins the bug — reverting `recoveryLockID` back to deriving from `tables.Prefix` instead of `tables.Requests` makes `TestRecoveryLockID_DistinctPerTMSOnSharedPrefix` fail (`Should not be: 5406896795709311132`, both simulated TMSes collide on the same lock id).

### Verification performed locally

| Check | Result |
|---|---|
| `go build ./token/...` | clean |
| `go vet ./token/services/storage/...` | clean |
| `go test -count=1` recovery/cleanup/postgres packages | pass (postgres suite against a real Testcontainers Postgres) |
| `go test -race -count=1` (same packages) | pass |
| Coverage | recovery 81.9%, cleanup 63.3%, postgres 80.2% |
| Revert-and-fail check | new test fails without the fix (see above) |
| `gofmt -l -s` | flags `token/services/storage/services/recovery/manager_test.go` |
| `golangci-lint run` (after local gofmt fix) | 0 issues |
| Mock regeneration (`go generate`) | no diff |
| Base drift | true merge-base is `0a45e0ea0` (60 commits behind `main`); diff there matches the PR's own `+282/-194`/31-files exactly |

### What it fixes

No issue is linked via the Development sidebar (`issueSource` isn't `closingIssuesReferences`), so there's no acceptance-criteria checklist to check off here — the description's own narrative (per-TMS advisory lock collision on shared persistence config, for both recovery and keystore/token-lock cleanup) is what I verified above.

### Blocking

**1. `gofmt` failure in `manager_test.go` fails CI's `checks` job outright**, which in turn means `utest`/`itest`/`lint`/`bench-check` never ran on this SHA at all (they're gated on `checks` succeeding). See inline comment below for the exact line. `gofmt -l -s -w` on that file and pushing should be enough — `golangci-lint` came back clean once I applied it locally.

### Non-blocking

- No linked issue and no labels on the PR (assignee/milestone/project are set). Per `docs/development/general.md` this should either link an issue or the description should note explicitly that none exists.

### Recommendation

Once the `gofmt` fix is pushed and `checks`/`utest`/`itest` actually run and pass, this looks ready to merge — the derivation, tests, and doc updates are solid.

token/services/storage/services/recovery/manager_test.go
```diff
@@ -338,10 +332,9 @@ func TestManager_StopDuringFanOutDoesNotDeadlock(t *testing.T) {
 		TTL:            10 * time.Millisecond,
 		ScanInterval:   10 * time.Millisecond,
 		BatchSize:      100,
-		WorkerCount:    1,
-		LeaseDuration:  time.Second,
-		AdvisoryLockID: 1,
-		InstanceID:     "test-instance",
+		WorkerCount:   1,
+		LeaseDuration: time.Second,
+		InstanceID:    "test-instance",
```
adecaro, line 337: **Blocking: `gofmt` misalignment fails CI's `checks` job.**

```go
config := recovery2.Config{
	Enabled:        true,
	TTL:            10 * time.Millisecond,
	ScanInterval:   10 * time.Millisecond,
	BatchSize:      100,
	WorkerCount:   1,
	LeaseDuration: time.Second,
	InstanceID:    "test-instance",
}
```

`WorkerCount:   1,` is indented one column short relative to the fields above it — left over from removing `AdvisoryLockID:` without re-running `gofmt`. Confirmed locally: `gofmt -l -s` flags exactly this file, matching the `checks` job failure in CI (https://github.com/LFDT-Panurus/panurus/actions/runs/31479453154/job/97584520232). Because `utest`/`itest`/`lint`/`bench-check` are all gated on `checks`, none of them have actually run on this SHA.

**Suggestion**: `gofmt -l -s -w token/services/storage/services/recovery/manager_test.go` and push.

atharrva01: Good catch, fixed in 432516bb6. Confirmed `gofmt -l -s` is clean across the repo now, and `go build`, `go vet`, `golangci-lint`, and `go test -race` on recovery/cleanup/postgres (real Testcontainers) all pass locally.

atharrva01:
@adecaro 
Fixed in 432516bb6 (`gofmt -l -s -w` on `manager_test.go`), `checks` should go green now.

Also added a line to the description noting there's no separate tracking issue, this is the direct follow-up requested in #1982's comments rather than a fresh report, so nothing to link.

Left labels alone, my account doesn't have push/triage access on this repo so setting them silently no-ops (confirmed via the collaborators permission endpoint). Assignee and milestone are already set though, so that part must have been done by a maintainer.

atharrva01:
Turned out the gofmt fix alone wasn't enough, checks still failed after that push: go vet flagged an unknown AdvisoryLockID field at manager_test.go:420.

Root cause was the stale base, not another formatting slip. The branch was 60+ commits behind main, and a commit that landed on main after this branch was created ("fix: no such transaction ID") added a test using AdvisoryLockID in a recovery.Config literal. GitHub actually tests the PR's merge with main, not just this branch's own diff, so that field reference (which exists on main but not in this PR's own changes) broke go vet on the merged state even though it never showed up as a git conflict.

Rebased onto current upstream/main and dropped that stale field. Pushed as 7e26c8bf6 (force-push, history rewritten by the rebase). PR's own diff is unchanged in shape, still 31 files, now +286/-198 vs the previous +282/-194, the delta is just this fix plus the earlier gofmt one.

Verified clean at this SHA: `go build ./...`, `go vet ./...` (matches make checks' govet target exactly, `-all` across all modules), `gofmt -l -s`, `golangci-lint run` on recovery/cleanup, `go test -race` on recovery/cleanup, and the postgres suite against real Testcontainers Postgres (ran twice to be sure, no flakes).

atharrva01:
Sorry, that last push didn't actually include the fix, my mistake: I edited the file locally and verified it, but never committed it before pushing, so 7e26c8bf6 only carried the rebase and the earlier gofmt fix, not the AdvisoryLockID removal. Same failure, same line, for that reason.

Actually committed and pushed now as f2746eb2b. This time I checked `git status` was clean and reran `go vet -all` across every module in GO_MODULES (same loop `make checks`' govet target uses) before pushing, not just the recovery package, to make sure nothing like this slips through again.

AkramBitar:
@atharrva01 

Could you please rebase and squash the commits to one? 

Thanks,
Akram

docs/services/storage/keystore_cleanup.md
```diff
@@ -412,7 +414,6 @@ The `token_ski_cleanups` table is automatically created by the schema initializa
 
 4. **For Multi-Instance Deployments:**
```
AkramBitar, line 414: **Blocking — two stale references left in this file.** Anchoring here because neither spot is inside a diff hunk; the actual lines are 108 and 395. This file was otherwise updated by the PR, so these look like misses rather than intent.

**Line 108** still has the old signature:

```go
AcquireCleanupLeadership(ctx context.Context, lockID int64) (Leadership, bool, error)
```

`cleanup/manager.go:34` is now `AcquireCleanupLeadership(ctx context.Context) (Leadership, bool, error)`.

**Line 395**, under "Default Values":

> - **Advisory Lock ID**: `0x74746b636c65616e` ("ttkclean" in hex)

The constant is deleted and the id is now `createTableLockID(tables.TokenSKICleanups + "_cleanup")` — neither a default nor configurable, so the whole bullet should go rather than be updated. A grep for `advisoryLockID|AdvisoryLockID|ttkrecov|ttkclean` shows this is the last stale reference outside the intentional ones in the stale-config test.

docs/configuration.md
```diff
@@ -679,7 +661,7 @@ Default values:
 - **Recovery is enabled by default** to ensure automatic recovery of lost finality listeners.
 - **Only pending transactions older than `ttl` are considered for recovery** to avoid interfering with active transaction processing.
 - **The manager validates** that `ttl`, `scanInterval`, `batchSize`, `workerCount`, and `leaseDuration` are all greater than zero.
-- **`advisoryLockID`** is used to acquire PostgreSQL advisory-lock leadership so that only one replica performs a recovery sweep at a time. The default value (8389190333894887286 or 0x74746b7265636f76) represents the ASCII string "ttkrecov" (Token Transaction Recovery) encoded as a 64-bit integer.
+- **Recovery leadership** uses a PostgreSQL advisory lock so that only one replica performs a recovery sweep at a time. The lock identifier is no longer configurable: it is derived from the TMS's own `requests` table name, which already carries the network, channel and namespace. That makes it unique per TMS automatically, so several TMSes sharing one PostgreSQL database no longer compete for a single lock, and there is nothing left for an operator to keep unique by hand.
```
AkramBitar: This new bullet says "only one replica performs a recovery sweep at a time", stated unconditionally — but it is not true for the audit path. `NewAuditTransactionStore` passes a nil recovery factory, so `AcquireRecoveryLeadership` hits the `noopRecoveryLeadership{}, true, nil` branch at `common/transactions.go:385`, and the audit store also inherits the permissive `ClaimPendingTransactions` with no claim or lease. On three replicas all three sweep the same audit rows.

I agree with leaving the fix itself out of scope. The ask is just to not widen the claim: either scope this sentence to the owner/transaction path, or wire `recoveryLockID(tableNames)` into `NewAuditTransactionStore` too, which is a one-liner given the factory shape you have now.

token/services/storage/db/sql/postgres/tokenlock.go
```diff
@@ -83,7 +81,7 @@ func NewTokenLockStore(dbs *common2.RWDB, tableNames common5.TableNames) (*Token
 // duration (see NewAdvisoryLock), so writeDB's connection pool needs at
 // least one spare connection beyond normal write traffic. See #1798.
 func (db *TokenLockStore) AcquireCleanupLeadership(ctx context.Context) (driver.CleanupLeadership, bool, error) {
-	return db.cleanupLeaderFactory(ctx, db.writeDB, db.cleanupLockID)
+	return db.cleanupLeaderFactory(ctx, db.writeDB)
```
AkramBitar, line 90: The note just above this ("`writeDB`'s connection pool needs at least one spare connection beyond normal write traffic") was accurate when the recovery and keystore-cleanup ids were global constants — one lock holder node-wide each. Now that both are per-TMS, a node running N TMSes on one shared pool can pin up to 3N connections in lock holders, and each holder needs a further connection to do its actual work.

Worth updating this comment to state the new floor in terms of TMS count, since it is the one place in the tree that documents the requirement, and an operator sizing `maxOpenConns` off it would now undershoot.

token/services/storage/services/recovery/config.go
```diff
@@ -90,9 +87,6 @@ func LoadConfig(cfg *config.Configuration) (Config, error) {
 	if config.LeaseDuration > 0 {
 		result.LeaseDuration = config.LeaseDuration
 	}
-	if config.AdvisoryLockID != 0 {
-		result.AdvisoryLockID = config.AdvisoryLockID
-	}
 	if config.InstanceID != "" {
```
AkramBitar, line 93: Nit, and it applies to `cleanup/config.go` too: the key is now read and discarded with no diagnostic. The doc comment this PR deletes explicitly sanctioned `advisoryLockID` for "multiple independent recovery managers on the same database", so an operator who set a custom value silently loses that isolation on upgrade — the yaml still loads, the setting just stops meaning anything.

A warn would make the upgrade visible, and line 97 of this file already uses exactly this shape for `notFoundGracePeriod`:

```go
if cfg.IsSet(ConfigKeyRecovery + ".advisoryLockID") {
    logger.Warnf("advisoryLockID is no longer used; the recovery lock id is derived per TMS")
}
```

Separately, `cleanup` has no equivalent of `TestLoadConfig_IgnoresStaleAdvisoryLockID` — the back-compat guarantee is only pinned on the recovery side.

AkramBitar (commented):
## Review

The core fix is correct and the reasoning in the description is the important part: deriving the id from `tables.Requests` rather than the prefix is what actually fixes this, because TMSes share one persistence configuration and are distinguished only by network/channel/namespace. `TestRecoveryLockID_DistinctPerTMSOnSharedPrefix` pins exactly that, and `TestRecoveryLockID_DistinctFromSchemaLock` covers the collision with the schema-creation lock — worth having, since both ids come out of the same `createTableLockID` space.

Binding the id at construction instead of passing it per call is the right shape: it makes a shared id unrepresentable rather than leaving it as a caller's responsibility. I checked that all four new helpers are wired to real call sites (`transactions.go:77`, `tokens.go:76`, `tokenlock.go:72`), the old unbound factories are gone, and no dead code is left behind. The config back-compat is real, and `testdata/stalecfg/core.yaml` exercises the actual `UnmarshalKey` path rather than the defaults early-return — easy thing to get wrong in a test like that.

**Two blocking items, both doc staleness in `docs/services/storage/keystore_cleanup.md`** (inline below). Everything else is a follow-up, not a merge blocker.

### Non-blocking

**Pinned connections now scale with TMS count.** FSC keys its pool by data source alone (`key(o) = o.DataSource`) and returns one `*sql.DB` as both `ReadDB` and `WriteDB`, so every TMS on a node shares one pool. `NewAdvisoryLock` pins a dedicated `db.Conn(ctx)` for the winner, held across the whole sweep (`recovery/manager.go:225-231` defers `Close` around `recoverTransactions`). Before this change the global ids meant at most one recovery lock and one keystore-cleanup lock existed node-wide; now it is one of each per TMS, on top of the already-per-TMS token-lock cleanup. Each winner then needs a second connection from the same pool for `ClaimPendingTransactions`. With `maxOpenConns: 10` and 4 TMSes, winners can block waiting on connections held by other winners that are themselves blocked, and the manager context has no timeout. This is a consequence of the fix rather than a reason to hold it — see the `tokenlock.go` comment below.

**The audit recovery path has no leadership at all.** Confirmed: `NewAuditTransactionStore` → `NewOwnerTransactionStore` → `newTransactionStore(..., nil, nil)`, and `AcquireRecoveryLeadership` returns `noopRecoveryLeadership{}, true, nil` on a nil factory (`common/transactions.go:385`). Agreed this is pre-existing and its own issue — the only thing this PR needs to do about it is not claim otherwise (see the `docs/configuration.md` comment).

### Metadata

No labels are set. Given the exported-interface changes to `driver.TransactionStore`, `driver.AuditTransactionStore`, `driver.TokenStore`, `recovery.Storage` and `cleanup.Storage`, this wants a breaking-change label so it shows up in the release notes. Assignee, milestone and project are already set.

_Verified against `e1fe58e` on a clean checkout: `go build ./...` and `go vet ./token/...` clean, unit tests green for the recovery, cleanup and postgres lock-id packages, `golangci-lint` clean on the changed packages._

AkramBitar (approved):
Approving. Re-checked at `acdd0519`: the branch was rebased onto `4c6e5e9`, but the PR's own content is unchanged from what I reviewed at `e1fe58e` — same 31 files, +286/-199, no edits to the code itself.

The fix is correct. Deriving the id from `tables.Requests` rather than the table prefix is the part that actually resolves the bug, since TMSes share one persistence configuration and differ only by network/channel/namespace, and binding it at construction means no caller can reintroduce a shared id. Build, vet, lint and the recovery/cleanup/postgres unit tests are all clean.

Two things still to do before this merges, neither of them code:

1. `docs/services/storage/keystore_cleanup.md:108` still documents the old `AcquireCleanupLeadership(ctx context.Context, lockID int64)` signature.
2. `docs/services/storage/keystore_cleanup.md:395` still lists `**Advisory Lock ID**: 0x74746b636c65616e` under "Default Values", where the whole bullet should go — the id is no longer a default or configurable.

My approval covers these being fixed as-described; no need for another round on my account.

The three non-blocking points from my earlier review stand as follow-ups rather than changes here: the connection-pool floor now scales with TMS count (worth updating the note above `tokenlock.go`'s `AcquireCleanupLeadership`), the audit recovery path still has no leadership at all, and the removed `advisoryLockID` key is dropped without a warning. Happy for those to be separate issues.

One correction to my earlier comment: I suggested a breaking-change label, but no such label exists in this repo — 67 labels and none for API breaks. The PR does still need labels set per the contributing rules; how the interface break gets surfaced in release notes is a separate question for the maintainers.

AkramBitar (changes requested):
@atharrva01 

The PR looks good for me now, please have a look at my above comments before we will merger it.

Thanks a lot,
Akram

docs/services/storage/keystore_cleanup.md
atharrva01: Fixed both in b0556169a: line 108's signature now matches `AcquireCleanupLeadership(ctx context.Context)`, and the Advisory Lock ID bullet is gone from the Default Values list.

docs/configuration.md
atharrva01: Scoped the sentence to the owner path and added a note pointing at #2143 for the audit gap, in b0556169a. Left the audit fix itself out of this PR since #2143 is already tracking the decision on whether it should coordinate at all.

token/services/storage/db/sql/postgres/tokenlock.go
atharrva01: Updated in b0556169a to describe the new floor in terms of TMS count (up to 3N pinned connections across the recovery, keystore-cleanup and token-lock locks) instead of the old flat one-spare-connection note.

token/services/storage/services/recovery/config.go
atharrva01: Added the Warnf in b0556169a for both recovery and cleanup configs, same shape as the notFoundGracePeriod check. Also added cleanup/config_test.go with a TestLoadConfig_IgnoresStaleAdvisoryLockID mirroring the recovery one, since cleanup didn't have a config test at all before this.

atharrva01:
Pushed b0556169a addressing your latest round: the two blocking doc fixes in keystore_cleanup.md, plus the three follow-ups (scoped the configuration.md claim and linked #2143, updated the tokenlock.go pool comment for the new per-TMS lock count, and added the stale-advisoryLockID warning + a cleanup config test). Replied inline on each thread.

AkramBitar (approved):

AkramBitar:
@atharrva01 

Thanks a lot for all the effort that you put on that PR. 
One more thing before merging. Could you please squash all the commits to one? 

Regards,
Akram

atharrva01:
Squashed to one commit as a8fc45474, force-pushed. Thanks for all the review time on this one.

atharrva01:
The failing race check is unrelated to this PR. TestBatchingStatusDB_CoalescesConcurrentGetStatus in token/services/ttx/dep/wrapper/dbs_test.go is a pre-existing timing-based test (from #1896, not touched here) that coalesces concurrent calls within a 300us window. Under -race the extra instrumentation overhead can push a call past that window, so it occasionally sees 2 calls instead of 1. This PR doesn't touch that package at all.

unit-tests-regression shows cancelled only because it's in the same matrix and got killed by fail-fast when unit-tests-race failed, not a separate issue.

Recent main runs are all green, so this looks like a flake rather than a real regression. Worth a re-run.

PR #2088, opened 2026-07-31
fix(token): keep request-carried audit info when the local lookup is empty

### LINKED ISSUE #2087: AuditRecord overwrites request-carried owner audit info with an empty local lookup — intended or bug?
### Context

An auditor validating a transfer needs the audit info of each input owner to resolve enrollment IDs. The request already carries that material: `extractTransferInputs` fills `Input.OwnerAuditInfo` from `metadata.Inputs[j].Senders[k].AuditInfo` (`token/request.go`, L945 on `main`).

### What happens

`Request.AuditRecord` then overwrites it unconditionally (`token/request.go`, L1467-L1473 on `main`):

```go
ownerAuditInfo, err := r.TokenService.tms.WalletService().GetAuditInfo(ctx, toks[i].Owner)
if err != nil {
    return nil, errors.Wrapf(err, "failed getting audit info for owner [%s]", toks[i].Owner)
}
in.OwnerAuditInfo = ownerAuditInfo
```

`Provider.GetAuditInfo` is documented as returning "the audit information associated to the passed identity, nil otherwise" — it only knows identities registered on the local node. An auditor is not the owner's node and has no reason to have registered that identity, so the lookup returns nil and the audit info the request carried is replaced with nothing.

### Why it is invisible for x509 owners

`in.EnrollmentID` is resolved earlier, in `extractTransferInputs`, via `GetEIDAndRH(ctx, sender.Identity, sender.AuditInfo)` while the metadata audit info is still intact, and it is stored in a separate field. Overwriting `OwnerAuditInfo` afterwards does not disturb it, and an x509 identity can in any case be re-read from the certificate it embeds.

Owners whose audit info cannot be recomputed from the identity have neither fallback. For a boolpolicy owner, `AuditInfo.EnrollmentID()` is hard-coded to `""` (`token/services/identity/boolpolicy/identity.go`, L100) — deliberately, since members may span enrollments — so a consumer that resolves the effective enrollment ID from the per-member audit info gets an empty field instead. An idemix member cannot be re-read from the identity at all: a nym carries no transparent material.

### Question

Is overwriting intended — i.e. is the locally registered audit info meant to be authoritative over what the request carries — or should the local result only be preferred when it is non-empty?

Keeping the request-carried value when the local lookup finds nothing also seems more correct across policy versions: the request carries the material that matched the owner *at the time the transaction was created*, whereas a later local lookup reflects the current registration.

We have a patch for the second reading (prefer local when non-empty, otherwise keep what the request carries; fail closed when neither exists) with a red/green test, and are happy to open the PR — it is behavior-preserving for every owner whose identity is registered locally.

### Repro sketch

1. Transfer from a wallet whose owner audit info cannot be derived from the identity (boolpolicy with an idemix member, or a plain idemix owner).
2. Audit it on a node that has not registered that owner identity.
3. `AuditRecord` returns the input with an empty `OwnerAuditInfo`, although the request carried it.

EvanYan1024:
Refs #2087.

`Request.AuditRecord` replaced the owner audit info of every input with the result of a local wallet-service lookup:

```go
ownerAuditInfo, err := r.TokenService.tms.WalletService().GetAuditInfo(ctx, toks[i].Owner)
...
in.OwnerAuditInfo = ownerAuditInfo
```

`Provider.GetAuditInfo` returns nil for identities that were never registered on this node, and an auditor has no reason to have registered the owner of a transfer it is auditing. So the audit info the request carried — filled a few frames earlier in `extractTransferInputs` from `metadata.Inputs[j].Senders[k].AuditInfo` — was discarded and replaced with nothing.

x509 owners never surfaced this. Their enrollment ID is resolved while the metadata audit info is still intact and kept in a separate field, and the identity embeds a certificate the value can be re-read from. Owners whose audit info cannot be recomputed from the identity have neither fallback: `boolpolicy.AuditInfo.EnrollmentID()` returns `""` by design, and an idemix nym carries no transparent material.

This keeps the local result when it is non-empty and falls back to what the request carries otherwise. For every owner registered on the auditing node the behavior is unchanged; when neither source has anything, consumers still see an empty field and can fail closed as before.

Keeping the request-carried value also holds up better across policy versions: the request carries the material that matched the owner when the transaction was created, whereas the local lookup reflects the current registration.

`TestRequest_AuditRecord_OwnerAuditInfo` covers both directions and fails on the pre-patch code (the metadata audit info comes back nil).

Verified with `gofmt`, `go build ./token/...`, `go vet`, `go test -race ./token/`, and golangci-lint v2.12.2 with the repo configuration.

### DIFF
--- a/token/request.go
+++ b/token/request.go
@@ -1470,7 +1470,12 @@ func (r *Request) AuditRecord(ctx context.Context) (*AuditRecord, error) {
 		if err != nil {
 			return nil, errors.Wrapf(err, "failed getting audit info for owner [%s]", toks[i].Owner)
 		}
-		in.OwnerAuditInfo = ownerAuditInfo
+		// The wallet service only knows the identities registered on this node
+		// and returns nil for the others, so keep the audit info the request
+		// metadata already carries when the local lookup finds nothing.
+		if len(ownerAuditInfo) != 0 {
+			in.OwnerAuditInfo = ownerAuditInfo
+		}
 	}
 
 	return &AuditRecord{
--- a/token/request_test.go
+++ b/token/request_test.go
@@ -1105,3 +1105,89 @@ func TestRequest_PublicParamsHash(t *testing.T) {
 	hash := r.PublicParamsHash()
 	assert.Equal(t, expectedHash, hash)
 }
+
+// TestRequest_AuditRecord_OwnerAuditInfo checks how AuditRecord fills the audit
+// info of an input: the local wallet service wins when it knows the owner, and
+// the audit info carried by the request metadata survives when it does not.
+func TestRequest_AuditRecord_OwnerAuditInfo(t *testing.T) {
+	ctx := t.Context()
+	owner := Identity("owner1")
+	metadataAuditInfo := []byte("audit-info-from-metadata")
+
+	newRequest := func(localAuditInfo []byte) *Request {
+		// A transfer with one input and no output: AuditRecord only reads back
+		// the inputs, and no output keeps the action deobfuscation out of play.
+		transferAction := &driver2.TransferAction{}
+		transferAction.NumInputsReturns(1)
+		transferAction.NumOutputsReturns(0)
+
+		transferService := &driver2.TransferService{}
+		transferService.DeserializeTransferActionReturns(transferAction, nil)
+
+		walletService := &driver2.WalletService{}
+		walletService.GetAuditInfoReturns(localAuditInfo, nil)
+
+		pp := &driver2.PublicParameters{}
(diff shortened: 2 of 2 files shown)


AkramBitar (approved):
LGTM

AkramBitar:
@EvanYan1024 thanks a lot for submitting this fix.

Please ignore the Token Validation Benchmark failures—we're currently fixing them. The issue affects Git forks.

The fix looks good to me.

However, I noticed that one test is failing (dlog-fabric-t7). I'll rerun it and hope it's just a flaky test that passes on the next run.

Thanks a lot,
Akram

PR #2089, opened 2026-07-31
fix(token): record one issue receiver per recipient of a composite owner

### LINKED ISSUE #2081: Issuing directly to a composite (boolpolicy) identity fails AuditCheck with "invalid recipient" — intended limitation or bug?
### Context

We issue tokens (zkatdlog nogh v1 driver, v0.14.1 — the relevant code is unchanged on current `main`) directly to a boolpolicy identity created with `boolpolicy.WrapPolicyIdentity("$0", member)`. The member is an idemix nym identity in our case, but the behavior below is independent of the member type.

### What happens

`Request.AuditCheck` → `extractIssueOutputs` fails deterministically with:

```
failed verifying inputs and outputs: invalid recipient [0,0] [<member>:<policy>]
```

The comparison in `token/request.go` (L696 on `main`) puts two different representations side by side:

- `recipients` comes from `TokensService().Deobfuscate(...)`, which enumerates the output owner through the identity deserializer. For a boolpolicy owner, `boolpolicy.TypedIdentityDeserializer.Recipients` (`token/services/identity/boolpolicy/deserializer.go`, L142) unwraps the policy identity into its **member identities**.
- `issueMeta.Outputs[j].RecipientAt(k).Identity` holds the receiver identity **as passed to `Request.Issue`** — the wrapped policy identity itself.

For any composite identity these can never be equal (a member identity != the wrapped policy identity), so auditing an issue action whose recipient is a policy identity always fails. Multi-member policies additionally diverge in cardinality (N unwrapped members vs 1 metadata recipient).

Transfers to the same policy identity pass `AuditCheck` in our test environments under sustained use, so the asymmetry appears specific to the issue path.

### Question

Is issuing directly to a composite identity intended to be unsupported (i.e., funds are expected to enter a policy-owned position only via transfer), or is this a bug in the issue-path comparison — e.g., should it compare against the original owner bytes, or unwrap the metadata identity with the same `Recipients` logic applied on the action side?

If it is the latter, we are happy to work on a PR — a pointer to the preferred fix direction would help (adjusting the comparison in `extractIssueOutputs` vs. recording unwrapped recipients in the issue metadata).

### Repro sketch

1. Build a policy identity `id` via `boolpolicy.WrapPolicyIdentity("$0", member)` and register it with `RegisterRecipientIdentity` (audit info wrapped via `boolpolicy.WrapAuditInfo`).
2. `Request.Issue(ctx, issuerWallet, id, typ, q)`.
3. Run `AuditCheck` on the auditing side: it fails with `invalid recipient [0,0]`.

EvanYan1024:
Refs #2081.

Issue metadata recorded the output owner as a single receiver:

```go
Receivers: []*driver.AuditableIdentity{
    {Identity: owner, AuditInfo: auditInfo},
},
```

while the transfer path expands the owner through `Deserializer.Recipients` and records one receiver per member. The auditing side rebuilds the recipients from the ledger output — `Deobfuscate` unwraps a composite owner into its members — and matches them positionally against that list, so an issue to a policy or multisig identity compared its first member against the whole composite identity and failed with `invalid recipient [0,0]`. Multi-member owners additionally diverged in cardinality.

#2081 asked which of two directions was preferred: adjusting the comparison in `extractIssueOutputs`, or recording unwrapped recipients in the issue metadata. This takes the second, which makes the issue path agree with the transfer path rather than adding a second convention — the expansion is lifted out of `transfer.go` into `common.AuditableRecipients` and used on both issue paths.

A plain identity expands to itself, so the common case is unaffected: same single receiver, same audit info. Both drivers had the same shape and both are fixed.

`TestAuditableRecipients` covers the plain and composite expansions and both error paths.

Verified with `gofmt`, `go build ./token/...`, `go vet`, `go test -race` over `token/core/common`, `token/core/zkatdlog/...` and `token/core/fabtoken/...`, and golangci-lint v2.12.2 with the repo configuration.

End to end, with this and #2088 applied, issuing directly to a boolpolicy identity whose member is an idemix nym now passes auditing in our test environments — the flow previously had to route funds in through a transfer.


AkramBitar (approved):
LGTM

AkramBitar:
@EvanYan1024 thanks a lot for submitting this fix. 

Please ignore the Token Validation Benchmark failures—we're currently fixing them. The issue affects Git forks.

The fix looks good to me.

However, I noticed that one test is failing (unit-tests-race). I'll rerun it and hope it's just a flaky test that passes on the next run.

Thanks a lot,
Akram

PR #2097, opened 2026-08-02
Added Error handling and < to <=

### LINKED ISSUE #2057: network-driver: finality MapTxData block-metadata off-by-one guard + unguarded index can crash the process on malformed block metadata
### Summary

`EndorserTxInfoMapper.MapTxData` has two distinct, compounding bugs guarding access to a block's transaction-filter metadata: an off-by-one length guard that lets a metadata slice missing the transactions-filter entry through, followed by a second index into that entry's byte slice with no bounds check at all. Either gap alone panics the finality/delivery listener on malformed block metadata.

### Where

`token/services/network/fabric/finality/deliveryflm.go:193-196`:

```go
if len(block.Metadata) < int(common.BlockMetadataIndex_TRANSACTIONS_FILTER) {   // TRANSACTIONS_FILTER == 2, guard uses strict `<`
    return nil, errors.Errorf("block metadata lacks transaction filter")
}
code, message := committer.MapValidationCode(int32(
    committer.ValidationFlags(block.Metadata[common.BlockMetadataIndex_TRANSACTIONS_FILTER])[txNum]  // deliveryflm.go:195-196
))
```

1. **Outer guard is off by one.** The check uses strict `<` against the raw index value (2), so it only rejects `len(block.Metadata) < 2`. A `Metadata` slice of length exactly 2 — containing SIGNATURES and LAST_CONFIG but missing TRANSACTIONS_FILTER at index 2 — passes the guard, and the very next line's `block.Metadata[2]` access panics with an out-of-range index. The guard should use `<=` (or compare against `3`).
2. **Inner index has no bounds check at all.** Even when the outer guard is satisfied and `block.Metadata[2]` is a valid access, the code never checks `len(block.Metadata[2])` against `txNum` before indexing into it via `ValidationFlags(...)[txNum]`. A transaction-filter byte slice shorter than `txNum + 1` panics on this second, independent index.

### Impact

Either a block metadata slice of length exactly 2 (missing the transactions-filter entry entirely) or a present-but-truncated transactions-filter entry (shorter than the current transaction's index) — both plausible outcomes of malformed, truncated, or adversarially crafted block metadata, e.g. from a byzantine/misbehaving orderer or peer, or a deserialization bug upstream — panics `MapTxData` on the delivery/finality listener's block-processing path. Depending on how that path is invoked, an unrecovered panic here can take down the finality listener goroutine and potentially the whole process if nothing recovers it upstream (see the related unrecovered-goroutine-panic issues filed for the `lookup` and `fabricx factory` packages). This is the most severe finding of this hardening pass: it is triggered by data structurally similar to normal block metadata, making it a plausible target for a malformed-block-based denial-of-service.

### Reproduction

Reproduced locally with two independent unit tests (not yet committed), one per bug:

```go
// Bug 1: outer guard off-by-one — Metadata length exactly 2, TRANSACTIONS_FILTER missing
block := &common.BlockMetadata{Metadata: [][]byte{{}, {}}}
require.Panics(t, func() { _, _ = mapper.MapTxData(ctx, tx, block, 0, 0) })

// Bug 2: inner index unchecked — TRANSACTIONS_FILTER present but shorter than txNum
block := &common.BlockMetadata{Metadata: [][]byte{{}, {}, {0x01}}}
require.Panics(t, func() { _, _ = mapper.MapTxData(ctx, tx, block, 0, 5) })  // txNum=5, filter has 1 byte
```

Happy to include these regression tests in the fix PR.

### Suggested fix

- Change the outer guard to `<=` (or compare `len(block.Metadata)` against `int(common.BlockMetadataIndex_TRANSACTIONS_FILTER) + 1`) so a metadata slice missing the transactions-filter entry is rejected.
- Add a bounds check on `len(block.Metadata[TRANSACTIONS_FILTER])` against `txNum` before indexing, returning an error instead of panicking when the filter entry is too short.

### Severity

CRITICAL — most severe finding of this hardening pass; reachable via malformed block metadata plausibly producible by a byzantine/misbehaving orderer or peer, with a realistic path to an unrecovered process crash.

Effi-S:
Applied suggested fix in: 
https://github.com/LFDT-Panurus/panurus/issues/2057

### DIFF
--- a/token/services/network/fabric/finality/deliveryflm.go
+++ b/token/services/network/fabric/finality/deliveryflm.go
@@ -190,10 +190,14 @@ func (m *EndorserTxInfoMapper) MapTxData(ctx context.Context, tx []byte, block *
 		return nil, errors.Wrapf(err, "failed extracting rwset")
 	}
 
-	if len(block.Metadata) < int(common.BlockMetadataIndex_TRANSACTIONS_FILTER) {
+	if len(block.Metadata) <= int(common.BlockMetadataIndex_TRANSACTIONS_FILTER) {
 		return nil, errors.Errorf("block metadata lacks transaction filter")
 	}
-	code, message := committer.MapValidationCode(int32(committer.ValidationFlags(block.Metadata[common.BlockMetadataIndex_TRANSACTIONS_FILTER])[txNum]))
+	txFilter := committer.ValidationFlags(block.Metadata[common.BlockMetadataIndex_TRANSACTIONS_FILTER])
+	if txNum >= uint64(len(txFilter)) {
+		return nil, errors.Errorf("transaction filter too short: index [%d] out of range [%d]", txNum, len(txFilter))
+	}
+	code, message := committer.MapValidationCode(int32(txFilter[txNum]))
 
 	return m.mapTxInfo(rwSet, chdr.TxId, code, message)
 }
--- a/token/services/network/fabric/finality/deliveryflm_test.go
+++ b/token/services/network/fabric/finality/deliveryflm_test.go
@@ -1,0 +1,157 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package finality_test
+
+import (
+	"context"
+	"testing"
+
+	"github.com/LFDT-Panurus/panurus/token/services/network/common/rws/keys"
+	"github.com/LFDT-Panurus/panurus/token/services/network/fabric/finality"
+	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/proto"
+	"github.com/hyperledger/fabric-protos-go-apiv2/common"
+	"github.com/hyperledger/fabric-protos-go-apiv2/ledger/rwset"
+	"github.com/hyperledger/fabric-protos-go-apiv2/ledger/rwset/kvrwset"
+	"github.com/hyperledger/fabric-protos-go-apiv2/msp"
+	"github.com/hyperledger/fabric-protos-go-apiv2/peer"
+	"github.com/stretchr/testify/require"
+)
+
+// buildEndorserTx assembles a minimal but well-formed ENDORSER_TRANSACTION
+// envelope that passes UnmarshalTx and rwset.Read, so that MapTxData proceeds
+// to the transaction-filter checks that this test exercises.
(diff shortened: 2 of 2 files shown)


Effi-S:
Seems like low hanging fruit. Applied suggested fix.

token/services/network/fabric/finality/deliveryflm.go
```diff
@@ -190,10 +190,14 @@ func (m *EndorserTxInfoMapper) MapTxData(ctx context.Context, tx []byte, block *
 		return nil, errors.Wrapf(err, "failed extracting rwset")
 	}
 
-	if len(block.Metadata) < int(common.BlockMetadataIndex_TRANSACTIONS_FILTER) {
+	if len(block.Metadata) <= int(common.BlockMetadataIndex_TRANSACTIONS_FILTER) {
 		return nil, errors.Errorf("block metadata lacks transaction filter")
 	}
-	code, message := committer.MapValidationCode(int32(committer.ValidationFlags(block.Metadata[common.BlockMetadataIndex_TRANSACTIONS_FILTER])[txNum]))
+	txFilter := committer.ValidationFlags(block.Metadata[common.BlockMetadataIndex_TRANSACTIONS_FILTER])
```
AkramBitar, line 196: Need to check if we have tests that covers these changes and especially the new added error.
We need good and bad path tests.

We need also to make sure that we have tests for the bugs reported in the issue #2097

// Bug 1: outer guard off-by-one — Metadata length exactly 2, TRANSACTIONS_FILTER missing
block := &common.BlockMetadata{Metadata: [][]byte{{}, {}}}
require.Panics(t, func() { _, _ = mapper.MapTxData(ctx, tx, block, 0, 0) })

// Bug 2: inner index unchecked — TRANSACTIONS_FILTER present but shorter than txNum
block := &common.BlockMetadata{Metadata: [][]byte{{}, {}, {0x01}}}
require.Panics(t, func() { _, _ = mapper.MapTxData(ctx, tx, block, 0, 5) })  // txNum=5, filter has 1 byte

AkramBitar (changes requested):

token/services/network/fabric/finality/deliveryflm.go
Effi-S: Added

AkramBitar (approved):
LGTM

PR #2098, opened 2026-08-02
lint: enable lll

atharrva01:
Next step of #1991, after #2001.

`lll` had a 240 char limit configured but was never enabled, so nothing enforced it. 32 lines
were over. Most are long function signatures and call expressions and are just wrapped. Four are
long string literals, split with `+`, and I checked the values are unchanged (two of them are
expected-error assertions in the asn1 tests, so a silent change there would have been a real
bug).

Touches 16 files across the root, integration and tokengen modules.

### A correction on the numbers

The table I put on #2001 said `lll` was 20 issues over 4 files. That was wrong, it is 32 over 12.

I had measured by running the remaining linters together in one `--enable-only` pass. That
under-reports: running `lll` on its own surfaced 16 issues in 8 files that the combined run never
showed. So the figures I gave for `ireturn`, `iface`, `gocognit`, `wrapcheck` and `revive` are
unreliable too, and I would not plan the remaining steps around them until each one has been
measured on its own. Happy to redo that and post a corrected table if useful.

Still open from #2001, no rush but they block the next steps:

- `maintidx` only fires on large table-driven test functions, so satisfying it means splitting
  those tests up. Excluding `_test.go` or dropping the setting seems better, your call.
- `revive` runs with `enable-all-rules: true`, which is far too much for one PR and probably
  wants a narrowed rule list first.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -17,6 +17,7 @@ linters:
     - importas # Enforces consistent import aliases. [auto-fix]
     - intrange # Intrange is a linter to find places where for loops could make use of an integer range. [auto-fix]
     - loggercheck # Checks key value pairs for common logger libraries (kitlog,klog,logr,slog,zap).
+    - lll # Reports long lines. [fast]
     - mirror # Reports wrong mirror patterns of bytes/strings usage. [auto-fix]
     - misspell # Finds commonly misspelled English words. [fast, auto-fix]
     - nakedret # Checks that functions with naked returns are not longer than a maximum size (can be zero). [fast, auto-fix]
--- a/cmd/tokengen/main_test.go
+++ b/cmd/tokengen/main_test.go
@@ -306,7 +306,10 @@ func TestGenFailure(t *testing.T) {
 				"--idemix", "./testdata/idemix",
 				"--issuers", "Error: failed to generate public parameters: failed to get issuer identity [aOrg1MSP]: invalid input [aOrg1MSP]",
 			},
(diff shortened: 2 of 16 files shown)


atharrva01:
hi @AkramBitar , @adecaro  , would love your thoughts on this...

AkramBitar:
> hi @AkramBitar , @adecaro , would love your thoughts on this...

@atharrva01 thanks a lot for submitting this PR:
Looks good from my side. 

Do we still need to have https://github.com/LFDT-Panurus/panurus/issues/1991 opened after this PR?

atharrva01:
thanks! and no, i'd keep it open. there are 6 left: `gocognit`, `iface`, `ireturn`, `maintidx`, `revive`, `wrapcheck`. this PR only does `lll`.

two of them are waiting on a call from you or @adecaro before i pick them up:

- `maintidx` only fires on big table-driven test functions, so satisfying it means splitting those tests up. excluding `_test.go` or just dropping the setting feels saner to me, but that's your call.
- `revive` runs with `enable-all-rules: true`, which is thousands of findings across most of the repo. probably wants a narrowed rule list agreed first rather than one enormous PR.

`ireturn` and `iface` are the next unblocked ones, happy to take those whenever.

also worth repeating from the PR description since it affects planning: the per-linter counts i posted on #1991 were measured by running the linters together, and that under-reports. `lll` turned out to be 32 issues over 12 files, not the 20 over 4 i quoted. so i wouldn't trust the other numbers until i redo them one at a time.

AkramBitar (approved):
LGTM

AkramBitar:
@atharrva01 thanks a lot for submitting this one. 
We can ignore the failure Token Validation Benchmark. We have an issue that we are fixing to run it on git forks.

atharrva01:

```
```

Issue #2040, opened 2026-07-30
auditdb/locker: misc medium/low findings (empty-lock semantics, key-namespace collision, retry storms)


adecaro:
### Summary

A batch of medium/low-severity findings from a systematic bug-hunting review of `token/services/storage/auditdb` and its `locker/**` subpackages, not severe enough to warrant individual issues but worth tracking and fixing. (Two more severe findings from the same review — Postgres locker self-theft and empty-`Owner` cluster-wide lock disablement — are tracked separately as #2033 and #2035.)

### 1. Inconsistent empty-enrollment-IDs semantics between memory and postgres lockers (MEDIUM)

- `auditdb/locker/memory/memory.go:87-89` — `AssertLocksHeld` always returns `nil` regardless of whether any locks were ever acquired.
- `auditdb/locker/postgres/postgres.go:107-111` — `AcquireLocks` returns early with **no session recorded** when the deduplicated enrollment-ID set is empty; `AssertLocksHeld` (`postgres.go:259-261`) then treats "no session" as `ErrLockNotHeld` unconditionally.

A token request whose inputs+outputs yield zero enrollment IDs succeeds at `AcquireLocks` but then fails `Append` ("locks lost before write") **only under the Postgres backend** — a backend-dependent correctness divergence for an edge case that is rare but real and verified in code.

**Suggested fix:** align behavior — either both backends should treat zero-locks as always-satisfied (mirroring memory), or both should require an explicit no-op session marker so `AssertLocksHeld` succeeds consistently.

### 2. Shared `sync.Map` namespace risks a panic on key collision (MEDIUM)

`auditdb/locker/memory/memory.go` — `m.locks sync.Map` stores both `eID -> *semaphore.Weighted` (line 46) and `anchor -> []string` (line 59) in the same map, with no prefix/namespace separation. If an anchor string ever collides with an enrollment-ID string, `AcquireLocks`'s type assertion `sem.(*semaphore.Weighted)` (line 47) or `ReleaseLocks`'s equivalent (line 79) panics. Both namespaces are unconstrained strings; not currently exercised by tests.

**Suggested fix:** use two separate maps, or prefix keys by namespace (e.g. `"eid:" + eID` vs `"anchor:" + anchor`) before storing.

### 3. Doubled, uncoordinated retry loops can cause ~10 minutes of blocking under contention (LOW/MEDIUM)

[...]

Issue #2033, opened 2026-07-30
auditdb: postgres locker's acquire upsert lets a node steal its own lock across unrelated anchors


adecaro:
### Summary

The Postgres-backed audit locker (`auditdb.Locker`) can let a single replica silently steal its own lock across two unrelated transactions that happen to share an enrollment ID, because the acquire upsert's conflict-resolution clause only checks `owner`, not `owner AND anchor`.

### Where

`token/services/storage/auditdb/locker/postgres/postgres.go:194-222` (`buildAcquireQuery`):

```go
// WHERE clause (paraphrased):
// cond.InPast(expires_at) OR cond.Cmp(owner, "=", excluded.owner)
```

`owner` is `p.cfg.Owner`, a value fixed once per `Locker` instance (the replica/FSC-node ID) — it does **not** vary per anchor (transaction) or per call.

### Reachability / failure scenario

1. A single auditor node runs two concurrent `Audit()` calls for two different transactions (`anchor1`, `anchor2`) whose input/output sets share at least one enrollment ID.
2. Call 1 acquires the lock row for that enrollment ID with `anchor = anchor1`.
3. Call 2's upsert hits the same enrollment-ID row. Its `WHERE` clause matches because `owner = excluded.owner` is true (same node, same configured owner) — regardless of whether the row's `anchor` differs. The row is overwritten with `anchor = anchor2`, and `AcquireLocks` returns `nil` (success) for **both** calls.
4. Both transactions now proceed believing they exclusively hold the enrollment-ID lock. All work done before the eventual DB write (limit checks in `AuditView`, holdings computation) runs on a false assumption of exclusivity.
5. The only backstop is `AssertLocksHeld`, called just before the DB write in `store.go:222` — so one of the two transactions eventually fails, but only at the write step, after doing all its prior (now-wasted, and briefly unsafe) work under a false lock guarantee.

This is a routine concurrency pattern (same auditor node processing multiple in-flight token transactions), not an edge case.

By contrast, the in-memory locker implementation (`auditdb/locker/memory/memory.go`) correctly serializes via per-enrollment-ID semaphores and does not have this issue — this is specific to the Postgres backend.

### Suggested fix

The upsert `WHERE` clause needs to require **both** owner and anchor match for "this is my own re-acquisition, refresh the lease":

```
[...]

PR #2099, opened 2026-08-02
fix(auditdb): reject an empty postgres locker owner at construction

### LINKED ISSUE #2035: auditdb: empty locker Owner default can silently disable cluster-wide mutual exclusion
### Summary

If the Postgres audit locker's configured `Owner` is left empty (e.g. `fsc.id` unset in a cloned/templated node config, and `cfg.Owner` not explicitly set in YAML), every replica in the cluster silently defaults to the same empty-string owner. Combined with the self-lock-theft bug in the acquire query's `WHERE` clause (see the companion issue on `owner = excluded.owner` matching), this turns the distributed lock into a cluster-wide no-op — not just a same-node issue.

### Where

- `token/services/storage/auditdb/locker/postgres/config.go:45-47` — `Owner` falls back to `replicaID.ID()` with no validation that the result is non-empty, and no panic/error if both `cfg.Owner` and `replicaID.ID()` are empty.
- `token/services/storage/auditdb/locker/postgres/postgres.go:53-57` — constructs the locker with whatever `Owner` value it was given, again with no non-empty check.

### Failure scenario

A misconfigured or templated deployment where `fsc.id` is accidentally left unset (or empty) on multiple nodes results in every node's audit locker using `Owner = ""`. Any acquire-query logic that treats `owner = excluded.owner` as "this is my own lock, refresh it" (see companion critical issue) then matches across *every* node, not just within one node — silently disabling mutual exclusion across the entire cluster with no error, warning, or log line indicating misconfiguration.

### Suggested fix

Validate at locker construction time that the resolved `Owner` is non-empty, and fail fast (return a clear configuration error) rather than silently proceeding with an empty/shared owner identity. This should be fixed independently of, but is compounded by, the acquire-query self-theft bug.

### Severity

HIGH — silent, cluster-wide loss of mutual exclusion under a plausible misconfiguration, with no diagnostic signal; discovered via a systematic bug-hunting review of `token/services/storage/auditdb` and related locker packages.

HayimShaul:
Fixes #2035

## What

The Postgres audit locker resolved its lease `owner` with no validation. `Config.withDefaults` fell back to the replica ID without checking the result was non-empty, and `New` accepted a nil `ReplicaIDProvider` or one reporting `""`. In production that provider is the FSC config provider, whose `ID()` is a bare `fsc.id` lookup — so a templated or cloned node config with `fsc.id` unset left every replica with `owner = ""`.

The owner scopes **every** lease query — the acquire upsert, `releaseAnchor`, `AssertLocksHeld` and `renewLeases` all compare against it. A shared empty owner therefore makes those predicates match rows belonging to *other* replicas: a replica can release or renew leases it does not hold, and `AssertLocksHeld` can be satisfied by a foreign row. Mutual exclusion is lost cluster-wide with no error, warning, or log line.

## Change

- Validate the resolved owner in `postgres.New`, **before** the lease table is created, returning an error wrapping the new `errs.ErrLockerOwnerRequired` and naming both remedies (`...auditor.locker.postgres.owner` or `fsc.id`).
- Blank-after-`TrimSpace` counts as empty — a whitespace-only `fsc.id` is just as broken and just as silent.
- Log the resolved owner once at construction, so a healthy replica leaves a positive trace of the identity it locks under.
- Docs: `docs/services/auditor.md` gains a *Replica owner identity* section (resolution order, failure mode, rationale); `docs/configuration.md` records the requirement and the exact error string.

Deliberately **not** synthesizing a fallback owner (UUID/hostname): an owner that changed on each restart would leave a restarted replica unable to renew or release the leases it still holds, trading a loud misconfiguration for a quiet correctness bug.

## ⚠️ Behaviour change

A node running `backend: postgres` with an empty resolved owner starts today and will now fail to build its audit store. That is the intended fix — such a node is already broken, it just fails silently — but it is worth calling out for operators. The `memory` backend (the default) has no owner concept and is unaffected.

## Out of scope

- #2033 — the `owner = excluded.owner` self-theft in the acquire `WHERE` clause. That predicate is untouched here; the new owner-scoping test is written so it stays valid either way.
- #2040 — empty-EID semantics, memory-locker key namespacing, doubled retry loops.

## Tests

New, all passing:
- `postgres/config_test.go` — table-driven `withDefaults` (defaulting, precedence, negative durations) and `validate` (empty / blank / tab-newline / inner-spaces).
- `postgres/postgres_test.go` — `TestLocker_OwnerRequired` (empty owner, nil provider, blank owner, blank replica ID), `TestLocker_OwnerFromReplicaID`, and `TestLocker_OwnerScopingAcrossReplicas`, which pins the property the empty-owner default destroyed: owner-2 cannot release owner-1's lease. The validation cases need no live Postgres, since validation precedes `createSchema`.
- `locker/factory_test.go` — new file; confirms the owner requirement stays Postgres-only (memory / `""` / `DefaultConfig` still build with an empty replica ID) and pins the existing backend-selection errors.

`make checks`, `make lint`, and `make lint-auto-fix` all clean. `go test ./token/services/storage/auditdb/...` passes, including the Docker-backed Postgres suite (9/9).

### DIFF
--- a/token/services/storage/auditdb/locker/postgres/config.go
+++ b/token/services/storage/auditdb/locker/postgres/config.go
@@ -6,7 +6,13 @@ SPDX-License-Identifier: Apache-2.0
 
 package postgres
 
-import "time"
+import (
+	"strings"
+	"time"
+
+	"github.com/LFDT-Panurus/panurus/token/services/storage/auditdb/locker/errs"
+	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
+)
 
 const (
 	defaultTTL             = 30 * time.Second
@@ -25,7 +31,10 @@ type Config struct {
 	AcquireDeadline time.Duration `yaml:"acquireDeadline"`
 	// Heartbeat is the interval at which held leases are renewed (~TTL/3).
 	Heartbeat time.Duration `yaml:"heartbeat"`
-	// Owner identifies this replica. Defaults to the FSC node ID when empty.
+	// Owner identifies this replica and is required: it scopes every lease
+	// query, so two replicas sharing one owner value share their leases.
+	// Defaults to the FSC node ID when empty; locker construction fails when
+	// both this field and the node ID are empty or blank.
 	Owner string `yaml:"owner"`
 }
 
@@ -42,9 +51,30 @@ func (c Config) withDefaults(owner string) Config {
 	if c.Heartbeat <= 0 {
 		c.Heartbeat = defaultHeartbeat
 	}
+	c.Owner = strings.TrimSpace(c.Owner)
 	if c.Owner == "" {
-		c.Owner = owner
+		c.Owner = strings.TrimSpace(owner)
 	}
 
 	return c
+}
+
+// validate reports whether the defaulted configuration is usable. Only Owner is
+// checked: it is the identity every lease query is scoped by (the acquire
+// upsert, releaseAnchor, AssertLocksHeld and renewLeases all compare against
+// it), so a blank value shared by every replica would make those predicates
+// match cluster-wide and silently turn the distributed lock into a no-op.
+// Failing here is deliberate — synthesizing an owner instead would make lease
+// ownership unstable across restarts, so a restarted replica could no longer
+// renew or release the leases it still holds.
+func (c Config) validate() error {
+	if strings.TrimSpace(c.Owner) == "" {
+		return errors.WithMessage(
+			errs.ErrLockerOwnerRequired,
+			"resolved owner is empty: set token.tms.<name>.auditor.locker.postgres.owner "+
+				"to a value unique per replica, or set fsc.id so the replica ID is non-empty and unique per node",
+		)
+	}
+
+	return nil
 }
--- a/token/services/storage/auditdb/locker/postgres/postgres.go
+++ b/token/services/storage/auditdb/locker/postgres/postgres.go
@@ -46,6 +46,13 @@ type lockSession struct {
 
 // New creates a Postgres-backed distributed Locker.
 // The table is created if it does not exist. db must be a *sql.DB connected to Postgres.
+//
+// cfg.Owner defaults to the identifier reported by replicaID (the FSC node ID in
+// production). It returns an error wrapping errs.ErrLockerOwnerRequired when the
+// resolved owner is empty or blank — including when replicaID is nil or reports
+// an empty ID — because an owner shared by several replicas disables mutual
+// exclusion across all of them. That check runs before the table is created, so
+// a misconfigured node fails without touching the database.
 func New(db *sql.DB, table string, cfg Config, replicaID id.ReplicaIDProvider) (*Locker, error) {
 	if db == nil {
 		return nil, errors.New("postgres locker requires a non-nil *sql.DB")
@@ -55,6 +62,9 @@ func New(db *sql.DB, table string, cfg Config, replicaID id.ReplicaIDProvider) (
 		owner = replicaID.ID()
 	}
 	cfg = cfg.withDefaults(owner)
+	if err := cfg.validate(); err != nil {
+		return nil, err
+	}
 	l := &Locker{
 		db:       db,
 		table:    table,
@@ -65,6 +75,7 @@ func New(db *sql.DB, table string, cfg Config, replicaID id.ReplicaIDProvider) (
 	if err := l.createSchema(); err != nil {
 		return nil, err
 	}
+	logger.Infof("postgres auditor locker for table [%s] locking as owner [%s]", table, cfg.Owner)
 
 	return l, nil
 }
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -850,7 +850,9 @@ Default values:
 - postgres.acquireBackoff: 100ms
 - postgres.acquireDeadline: 1m
 - postgres.heartbeat: 10s
-- postgres.owner: empty, defaults to the FSC node ID (`config.Provider.ID()`)
+- postgres.owner: empty, defaults to the FSC node ID (`config.Provider.ID()`). Required
+  when `backend: postgres` — if both this value and `fsc.id` are empty or blank, the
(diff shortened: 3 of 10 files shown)


AkramBitar:
Worth naming explicitly: this validates **emptiness**, not **uniqueness**.

`validate()` is `strings.TrimSpace(c.Owner) == ""`. But an empty owner is just the accidental case of a broader property: two replicas both set to `owner: auditor-1` lose mutual exclusion in exactly the same way, and nothing detects it. The docs, the `Config.Owner` godoc and the error message all say "non-empty **and unique per replica**" — only the first half is enforced.

Not asking for uniqueness here: detecting a collision needs coordination (a registry table keyed on `owner` with a per-process session token, so a second live claimant fails at startup). That's real design work and the fail-fast fix stands on its own. Happy to file the follow-up.

One thing that *is* in scope: the error message's remedies aren't symmetric. The first says "a value unique per replica"; the second says "or set `fsc.id` so the replica ID is **non-empty**". Following the second half, an operator could set the same `fsc.id` everywhere and land right back here. Suggest "non-empty and unique per node".

token/services/storage/auditdb/locker/postgres/config.go
```diff
@@ -48,3 +57,23 @@ func (c Config) withDefaults(owner string) Config {
...
+// validate reports whether the defaulted configuration is usable. Only Owner is
+// checked: it is the identity every lease query is scoped by (the acquire
+// upsert, releaseAnchor, AssertLocksHeld and renewLeases all compare against
+// it), so a blank value shared by every replica would make those predicates
+// match cluster-wide and silently turn the distributed lock into a no-op.
+// Failing here is deliberate — synthesizing an owner instead would make lease
+// ownership unstable across restarts, so a restarted replica could no longer
+// renew or release the leases it still holds.
+func (c Config) validate() error {
+	if strings.TrimSpace(c.Owner) == "" {
```
AkramBitar, line 71: **Code and docs disagree here.** `validate()` treats blank-after-`TrimSpace` as empty, but `withDefaults` tests `c.Owner == ""` exactly — so `owner: "   "` is *kept*, never replaced by the replica ID, and then rejected here even when `fsc.id` is perfectly valid.

`docs/services/auditor.md` says the opposite:

> 1. `…postgres.owner`, when set. 2. Otherwise the FSC node ID… If **both** are empty or blank, the locker fails to start.

That reads as: blank owner falls through to `fsc.id`, failure only when both are blank.

One change fixes this *and* the normalization gap — trim before the empty check in `withDefaults`:

```go
c.Owner = strings.TrimSpace(c.Owner)
if c.Owner == "" {
    c.Owner = strings.TrimSpace(owner)
}
```

Blank owner then falls back to `fsc.id` as documented, blank `fsc.id` still fails here, and the stored owner is normalized. If you prefer the current strictness, the docs need to say so: "a blank `owner` is not treated as unset — it fails rather than falling back."

(Your test name `"blank config owner is kept, not replaced by the replica id"` shows the behaviour is deliberate, which is why I think the docs are the thing that's wrong — but either way they should agree.)

AkramBitar (commented):
Reviewed the code path: the fix is sound. `validate()` runs after `withDefaults` and before `createSchema()`, `errors.WithMessage` preserves the sentinel so `ErrorIs` works, and `store.go:69` wraps the error with the TMS ID so context isn't lost. I also grepped — nothing in-tree configures `backend: postgres` outside this package's tests, so the behaviour change breaks no existing test or integration setup.

Two things I want to call out as genuinely good, since they're easy to miss: `unconnectedDB` turns "validation precedes any DB access" into an enforced invariant rather than a comment — if the ordering regressed, the test fails with a connection error instead of `ErrLockerOwnerRequired`. And `TestLocker_OwnerScopingAcrossReplicas` pins the actual security property while staying valid whichever way #2033 lands.

Three inline notes; only the first needs resolving.

token/services/storage/auditdb/locker/postgres/config.go
```diff
@@ -25,7 +31,10 @@ type Config struct {
 	AcquireDeadline time.Duration `yaml:"acquireDeadline"`
 	// Heartbeat is the interval at which held leases are renewed (~TTL/3).
 	Heartbeat time.Duration `yaml:"heartbeat"`
-	// Owner identifies this replica. Defaults to the FSC node ID when empty.
+	// Owner identifies this replica and is required: it scopes every lease
+	// query, so two replicas sharing one owner value share their leases.
```
AkramBitar, line 35: Related to the above: the owner is *validated* with `TrimSpace` but *stored* untrimmed. `owner: "replica-1 "` (trailing space from templating or YAML) passes and becomes the lease identity verbatim.

Harmless for mutual exclusion — distinct owners stay distinct — but if that replica later renders without the space, it can no longer renew or release the leases it still holds. That's the same restart-instability failure mode this PR cites to justify not synthesizing an owner, arriving by a different route. Trim-on-resolve closes it.

token/services/storage/auditdb/locker/postgres/postgres.go
```diff
@@ -55,6 +62,10 @@ func New(db *sql.DB, table string, cfg Config, replicaID id.ReplicaIDProvider) (
 		owner = replicaID.ID()
 	}
 	cfg = cfg.withDefaults(owner)
+	if err := cfg.validate(); err != nil {
+		return nil, err
+	}
+	logger.Infof("postgres auditor locker for table [%s] locking as owner [%s]", table, cfg.Owner)
```
AkramBitar: Nit: this fires before `createSchema()`, so it announces `locking as owner [x]` for a locker that may then fail to build its table. Moving it below the `createSchema()` error check makes the line mean what it says — a locker that is actually ready to lock.

AkramBitar (changes requested):

AkramBitar:
Hi @HayimShaul 

Thanks a lot for your efforts on that PR.

One thing the error message still says "set fsc.id so the replica ID is non-empty" — should be "non-empty and unique per node",

AkramBitar (approved):
LGTM

Added one minor fix.

AkramBitar (approved):

PR #2100, opened 2026-08-02
fix(fabtoken): validate every input of an HTLC transfer

### LINKED ISSUE #2025: fabtoken: TransferHTLCValidate only validates InputTokens[0] for multi-input HTLC transfers
### Summary

`TransferHTLCValidate` iterates over `ctx.InputTokens` with index `i`, but every check inside the HTLC branch is hardcoded to index `0` instead of `i`. For an HTLC-owned transfer with more than one input, only `InputTokens[0]` is ever actually validated; `InputTokens[1..n]` are accepted regardless of their type, quantity, or owner script. fabtoken is also missing the single-input restriction that the equivalent zkatdlog validator enforces for HTLC-owned transfers.

This is a genuine invalid→valid validity flip: a structurally-invalid multi-input HTLC transfer is accepted as valid.

### Where

`token/core/fabtoken/v1/validator/validator_transfer.go`, `TransferHTLCValidate`:

```go
for i, in := range ctx.InputTokens {
    owner, err := identity.UnmarshalTypedIdentity(in.GetOwner())
    ...
    if owner.Type == htlc.ScriptType {
        if len(ctx.TransferAction.GetOutputs()) != 1 {
            return errors.New("invalid transfer action: an htlc script only transfers the ownership of a token")
        }

        output := ctx.TransferAction.GetOutputs()[0].(*actions.Output)
        tok := output
        if ctx.InputTokens[0].Type != tok.Type {        // line 172 — hardcoded [0], not [i]
            return errors.New(...)
        }
        if ctx.InputTokens[0].Quantity != tok.Quantity { // line 175 — hardcoded [0], not [i]
            return errors.New(...)
        }
        ...
        script, op, err := htlc2.VerifyOwner(ctx.InputTokens[0].GetOwner(), tok.Owner, now) // line 183 — hardcoded [0]
        ...
        sigma := ctx.Signatures[i]
        ...
    }
}
```

(`token/core/fabtoken/v1/validator/validator_transfer.go:172,175,183`)

Compare with the equivalent zkatdlog check (`token/core/zkatdlog/nogh/v1/validator/validator_transfer.go:207-208`), which additionally *restricts* HTLC-owned transfers to exactly one input:

```go
if len(ctx.InputTokens) != 1 || len(ctx.TransferAction.GetOutputs()) != 1 {
    return ErrInvalidHTLCAction
}
```

fabtoken has no analogous restriction — it only checks `len(ctx.TransferAction.GetOutputs()) != 1`, never the input count.

### Impact

A transfer with 2+ HTLC-owned inputs, one output, where:
- `InputTokens[0]` is a valid-looking match for the single output (e.g. an expired/Reclaim-eligible HTLC script owned by the sender), and
- `InputTokens[1..n]` are arbitrary — wrong type, wrong quantity, a completely different/invalid HTLC script

...passes `TransferHTLCValidate` with `nil` error, because indices `1..n` are never checked against anything.

### Reproduction

Reproduced locally with a unit test (not yet committed) that constructs exactly this scenario — 2 HTLC-owned inputs, input[0] matching an expired/Reclaim-branch script against the sole output, input[1] with a completely different type/quantity/script — and asserts `TransferHTLCValidate` returns `nil`. Happy to include this regression test in the fix PR.

### Suggested fix

- Replace every hardcoded `ctx.InputTokens[0]` inside the HTLC branch with `in` / `ctx.InputTokens[i]`.
- Add the missing single-input restriction for HTLC-owned transfers, mirroring zkatdlog's `len(ctx.InputTokens) != 1` guard — or, if multi-input HTLC transfers are intentionally meant to be supported, validate each HTLC-owned input against its own corresponding output rather than assuming output index 0 for all of them.

### Severity

HIGH — validity-flip bug in on-chain transaction validation.

AkramBitar:
Fixes #2025

`TransferHTLCValidate` looped with index `i` but checked `ctx.InputTokens[0]`, so in a
multi-input HTLC transfer inputs `1..n` were never validated — a structurally invalid
action was accepted with a `nil` error.

## Changes

- Use the iterated input (`in`) for the type, quantity, and `VerifyOwner` checks.
- Restore the single-input pin for HTLC-owned transfers (1 input, 1 output), which
  fabtoken had before it was dropped in `a6b159e321` and which zkatdlog still has.
- Guard against a nil input token and a missing signature (both previously panicked).

Multi-input HTLC remains unsupported, as in zkatdlog: `Claim`/`Reclaim` each spend a
single token, and the action carries no input→output mapping to validate pairs against.

## Tests

- 5 subtests in `TestTransferHTLCValidate`, including the issue's exact reproduction.
  Verified by reverting the fix: all 5 fail, the 10 pre-existing ones pass.
- `FuzzTransferHTLCValidateNoPanic`, wired into `nightly-fuzz.yml`.
- HTLC validation rules documented in `docs/services/interop.md`.

### DIFF
--- a/token/core/fabtoken/v1/validator/validator_transfer.go
+++ b/token/core/fabtoken/v1/validator/validator_transfer.go
@@ -127,8 +127,13 @@ func TransferBalanceValidate(c context.Context, ctx *Context) error {
 			return errors.Errorf("input type %s does not match type %s", input.Type, typ)
 		}
 	}
-	for _, output := range ctx.TransferAction.GetOutputs() {
-		out := output.(*actions.Output)
+	for i, output := range ctx.TransferAction.GetOutputs() {
+		// A nil *actions.Output boxed in the driver.Output interface satisfies the type
+		// assertion, so the nil pointer must be rejected explicitly to avoid a panic.
+		out, ok := output.(*actions.Output)
+		if !ok || out == nil {
+			return errors.Errorf("invalid output at index [%d]", i)
+		}
 		q, err := token.ToQuantity(out.Quantity, ctx.PP.QuantityPrecision)
 		if err != nil {
 			return errors.Wrapf(err, "failed parsing quantity [%s]", out.Quantity)
@@ -151,11 +156,15 @@ func TransferBalanceValidate(c context.Context, ctx *Context) error {
 }
 
 // TransferHTLCValidate checks the validity of the HTLC scripts, if any.
-// A nil input token or a signature missing at the index of an HTLC-owned input
-// yields a validation error rather than a panic, regardless of the order in which
+// An HTLC-owned input may only be spent by a 1-to-1 transfer of ownership: exactly one
+// input and exactly one output. Every check below is performed against the input being
+// iterated (`in`), never against a fixed index, so no input can escape validation.
+// A nil input token, a nil output or a signature missing at the index of an HTLC-owned
+// input yields a validation error rather than a panic, regardless of the order in which
 // the validation steps of the pipeline are executed.
 func TransferHTLCValidate(c context.Context, ctx *Context) error {
 	now := time.Now()
+	outputs := ctx.TransferAction.GetOutputs()
 
 	for i, in := range ctx.InputTokens {
 		// guard: a nil token in the input slice must return an error, not panic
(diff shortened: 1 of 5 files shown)


token/core/fabtoken/v1/validator/validator_transfer.go
```diff
@@ -150,42 +150,59 @@ func TransferBalanceValidate(c context.Context, ctx *Context) error {
...
+			// without it, the outputs of a multi-input action could not be matched
+			// one-to-one against their inputs.
+			if len(ctx.InputTokens) != 1 || len(ctx.TransferAction.GetOutputs()) != 1 {
 				return errors.New("invalid transfer action: an htlc script only transfers the ownership of a token")
 			}
 
 			// check type and quantity
 			output := ctx.TransferAction.GetOutputs()[0].(*actions.Output)
 			tok := output
-			if ctx.InputTokens[0].Type != tok.Type {
```
Effi-S, line 179: Best to make sure `len(ctx.InputTokens) >= 1` before we try accessing `len(ctx.InputTokens[0]`

token/core/fabtoken/v1/validator/validator_transfer.go
```diff
@@ -150,42 +150,59 @@ func TransferBalanceValidate(c context.Context, ctx *Context) error {
...
+			// An htlc script must be a 1-to-1 transfer of ownership. The input count is
+			// restricted as well, mirroring zkatdlog's ErrInvalidHTLCAction check:
+			// without it, the outputs of a multi-input action could not be matched
+			// one-to-one against their inputs.
+			if len(ctx.InputTokens) != 1 || len(ctx.TransferAction.GetOutputs()) != 1 {
 				return errors.New("invalid transfer action: an htlc script only transfers the ownership of a token")
 			}
 
 			// check type and quantity
 			output := ctx.TransferAction.GetOutputs()[0].(*actions.Output)
```
Effi-S: Maybe add a type assertion as well on `ctx.TransferAction.GetOutputs()`
```go 
output, ok := ctx.TransferAction.GetOutputs()[0].(*actions.Output)
if !ok {
    return errors.New("invalid transfer action: output has unexpected type")
}
```

token/core/fabtoken/v1/validator/validator_transfer.go
```diff
@@ -150,42 +150,59 @@ func TransferBalanceValidate(c context.Context, ctx *Context) error {
...
+			return errors.Errorf("nil input token at index [%d]", i)
+		}
 		owner, err := identity.UnmarshalTypedIdentity(in.GetOwner())
 		if err != nil {
 			return errors.Wrap(err, "failed to unmarshal owner of input token")
 		}
 		// is it owned by an htlc script?
 		if owner.Type == htlc.ScriptType {
-			// Then, the first output must be compatible with this input.
-			if len(ctx.TransferAction.GetOutputs()) != 1 {
```
Effi-S, line 172: `ctx.TransferAction.GetOutputs()` is called a few times 
Perhaps  store it in a variable

Effi-S (changes requested):

Effi-S (approved):

PR #2101, opened 2026-08-02
fix(identity): make conf_id collision-free across identity configurations

### LINKED ISSUE #2070: identity/membership: conf_id cache key is an unescaped concatenation, so a collision routes to the wrong KeyManager
### Summary

The `conf_id` used to route signer resolution is built by unescaped string concatenation of three
fields (`ID + "@" + Type + "@" + URL`). Two distinct configurations can concatenate to the same
string (e.g. `ID="a@b", Type="c", URL="d"` vs. `ID="a", Type="b@c", URL="d"`), producing the same
`conf_id` and routing to the wrong `KeyManager` via `DeserializeSignerNoProbe`, whose explicit
contract is that callers must already know the signer belongs to that key manager — i.e. this skips
the verification that would otherwise catch the mismatch.

### Where

`token/services/identity/membership/lm.go:910-912` (`configKey`):

```go
return config.ID + "@" + config.Type + "@" + config.URL
```

`token/driver/wallet.go:226-228` (`IdentityConfiguration.UniqueID`) hashes the same unescaped
concatenation with sha256. This becomes both the DB `conf_id` column and the key into
`SignerRouter.byConfID` (`lm.go:824`).

`SignerRouter.Resolve` (`token/services/identity/signer_router.go:103-104`) uses the routed
`KeyManager` to call `DeserializeSignerNoProbe` — whose documented contract
(`idemix/km.go:421-426`, `driver/deserialization.go:52-55`) is that the caller already knows the
signer belongs to that specific key manager, precisely because the probe (which would normally catch
a mismatch) is skipped for performance.

### Impact

If two configurations collide on `configKey`, `Resolve` returns a signer built by the wrong
`KeyManager`. Because the probe is intentionally skipped for this path, the mismatch isn't caught
here — it surfaces later as invalid signatures produced by that signer, which is a confusing failure
mode to diagnose (looks like a signing bug, not a routing bug).

`config.ID` derives from on-disk directory names (`lm.go:717`), so triggering this requires control
over local configuration/identity directory naming — this is a local-misconfiguration risk, not a
remote attack surface.

### Reproduction

Not yet committed. Construct two `IdentityConfiguration` values whose `ID`/`Type`/`URL` fields
concatenate to the same string via the unescaped `"@"`-joined scheme (e.g. `{ID: "a@b", Type: "c"}`
vs `{ID: "a", Type: "b@c"}`, same `URL`), register both, and assert `SignerRouter.byConfID` ends up
with only one entry (the collision) instead of two, or that `configKey` produces the same string for
both.

### Suggested fix

Use an unambiguous encoding for the composite key — e.g. length-prefix each field, or hash each field
independently before concatenating (`sha256(ID) + sha256(Type) + sha256(URL)`), or use a delimiter
that's rejected/escaped in the individual fields. Apply consistently to both `configKey` in
`membership/lm.go` and `IdentityConfiguration.UniqueID` in `token/driver/wallet.go`, since they must
stay in agreement.

### Severity

LOW–MEDIUM — requires local configuration control to trigger, and the resulting failure (bad
signatures from a misrouted signer) is contained rather than a crash, but is confusing to diagnose.

AkramBitar:
Fixes #2070

`conf_id` was `sha256(ID + "@" + Type + "@" + URL)` with no escaping, so two different
configurations could produce the same value. `SignerRouter` then held one entry for the two
and resolved signers through the wrong `KeyManager` — undetected, because that path uses
`DeserializeSignerNoProbe`, whose contract is that the caller already knows the signer belongs
to that key manager. It surfaces later as invalid signatures, which reads as a signing bug
rather than a routing bug.

## Why escaping is needed

The three fields are joined with `@`, but nothing stopped a field from *containing* `@`. The
join is then ambiguous: the same output can be cut into fields in more than one way.

**Before — two distinct configurations, one `conf_id`:**

| `ID` | `Type` | `URL` | joined key | `conf_id` |
|:--|:--|:--|:--|:--|
| `a@b` | `c` | `d` | `a@b@c@d` | `XYHVaqPaIYcsLbO60TNDgrx375wRXfrYJwD3kZ+WDB4=` |
| `a` | `b@c` | `d` | `a@b@c@d` | `XYHVaqPaIYcsLbO60TNDgrx375wRXfrYJwD3kZ+WDB4=` |

Same string, same hash. `SignerRouter.Register` is called once per configuration with
`config.UniqueID()`, so the second registration silently overwrites the first: an identity
belonging to the first configuration is handed to the second configuration's `KeyManager`, and
the probe that would have caught it is skipped by design.

**After — `@` inside a field is escaped as `\@`, so the join is unambiguous:**

| `ID` | `Type` | `URL` | joined key | `conf_id` |
|:--|:--|:--|:--|:--|
| `a@b` | `c` | `d` | `a\@b@c@d` | `KCiegkvTuMbTtKG/yT+nDAKzxi8zX/tHKclYu6k9Upg=` |
| `a` | `b@c` | `d` | `a@b\@c@d` | `Bn6wZxoPdWfJloDM0k0sCv1JmfWgFOWnlLl0ux0X5Oo=` |

`\` is escaped too (`\\`), so a field that already contains a backslash cannot forge an escape
sequence and re-introduce the ambiguity.

### The escaping is byte-wise, not rune-wise

`escapeConfigKeyField` walks its input one **byte** at a time. Both delimiters are ASCII, so
decoding UTF-8 buys nothing — and costs correctness: `for _, r := range field` yields
`utf8.RuneError` for every byte that is not valid UTF-8, substituting one replacement character
for the original byte. `{ID: "\xff@"}` and `{ID: "\xfe@"}` would then encode identically,
re-introducing exactly the collision this change removes. `URL` is a filesystem path and paths
are not required to be valid UTF-8, so every field is treated as an opaque byte string.

### How this is reachable in practice

`config.ID` comes from on-disk identity directory names (`membership/lm.go`), and `@` in a
directory name is ordinary — MSP folders are routinely named after enrollment IDs:

| `ID` | `Type` | `URL` | joined key (before) |
|:--|:--|:--|:--|
| `alice@org1` | `idemix` | `/msp` | `alice@org1@idemix@/msp` |
| `alice` | `org1@idemix` | `/msp` | `alice@org1@idemix@/msp` |

Triggering it needs control over local configuration naming, so this is a
local-misconfiguration risk rather than a remote attack surface — matching the issue's severity.

## Why escaping, and not one of the other two suggested encodings

`conf_id` is persisted: it is `UNIQUE` on `identity_configurations` and the target of the
`wallets.conf_id` foreign key. Length-prefixing or per-field hashing change the value for
*every* configuration, so every existing deployment would need a migration — and there is no
schema-migration framework in `token/services/storage`. Escaping leaves a field untouched when
it contains no `@` or `\`, so the key is byte-identical for the overwhelming majority of stored
configurations:

| `ID` | `Type` | `URL` | key | `conf_id` (unchanged) |
|:--|:--|:--|:--|:--|
| `alice` | `idemix` | `/msp/alice` | `alice@idemix@/msp/alice` | `045gSx98I6vKzphB3QGahdL5JzbCtdeiF3KKH94VIqE=` |

## `conf_id` is stored state, so it is read back — not recomputed

Escaping still changes the value for any configuration that *does* contain `@` or `\` in one of
the three fields — a set wider than the configurations that could actually collide: a lone
`{ID: "alice@org1", Type: "idemix", URL: "/msp"}` gets a new `conf_id` even though it never had
an ambiguous partner. Since `ID` comes from directory entries and `URL` is a filesystem path,
that is not exotic.

For those configurations, recomputing the identifier is a **hard failure**, not a cosmetic
change. `commitLocalIdentity` looks a configuration up by `(id, type, url)` and never rewrites a
stored `conf_id`, so the recomputed value matches no `identity_configurations` row and
`WalletStore.StoreIdentity` fails the foreign key (`FOREIGN KEY constraint failed`, reproduced on
sqlite with `ID: "alice@org1"`). The node can then no longer mint a pseudonym or serve
`RegisterRecipient` — it cannot receive tokens.

So `conf_id` is treated as what it is, persisted state:

- `IdentityStoreService.GetConfigurationID(id, typ, url)` returns the stored `conf_id`, or `""`
  when the configuration is not stored yet. SQL reads the column; kvs derives it from the stored
  record (it keeps no separate column).
- `LocalMembership.confIDFor` binds identities under the stored value verbatim, and computes
  `UniqueID()` only for a configuration being inserted — where it is exactly what the following
  `AddConfiguration` writes. It also reports whether the configuration was already stored, which
  replaces the previous `ConfigurationExists` call in `commitLocalIdentity`.

Configurations stored before the change keep their original `conf_id` indefinitely; new ones get
the unambiguous encoding. No migration is needed, and nodes on either release agree on the
identifier for the same configuration — which an in-place migration could not guarantee during a
rolling upgrade.

**kvs caveat:** the kvs store has no `conf_id` column, so `GetConfigurationID` re-derives it
with the current encoding. There is no foreign key there for a changed identifier to violate, so
nothing hard-fails, but a configuration stored before the change is reported with a new
`conf_id` while `WalletStore.GetConfID` still returns the original one for its existing
identities. `SignerRouter.byConfID` misses for those identities and resolution falls back to the
probing deserializer: correct, just without the fast path. Documented at the source and in
`docs/services/identity.md`.

## kvs row-key collisions

Reading the value back needs a guard on the kvs store, which keys a configuration row by
`base64(id||url)` with **no separator**: a lookup for `{ID: "bob/msp", URL: "/alice"}` returns
the record stored for `{ID: "bob", URL: "/msp/alice"}`. Two changes:

- `GetConfigurationID` verifies the record it read back is the one asked for, and reports "not
  stored" otherwise. Handing back the colliding record's `conf_id` would bind identities under it
  and take over its `SignerRouter` entry — the wrong-`KeyManager` route this PR exists to prevent.
- `AddConfiguration` refuses to store a configuration whose row key is occupied by a *different*
  one. It has to: reporting "not stored" above makes `commitLocalIdentity` go on to insert, and a
  `Put` on the shared key would replace the stored record, dropping its `Config`/`Raw` and
  leaving that configuration unable to reload from the store. The collision is now reported to
  the caller instead of resolving in favour of whoever wrote last.

## Changes

- `token/driver/wallet.go`: `IdentityConfiguration.CompositeKey()` escapes `@` and `\` inside
  each field before joining, byte-wise; `UniqueID()` hashes it.
- `token/services/identity/membership/lm.go`: `configKey` delegates to `CompositeKey()`, so the
  in-memory `localIdentitiesByConfig` index and the persisted `conf_id` cannot drift apart; new
  `confIDFor`; `commitLocalIdentity` uses its `stored` flag.
- `token/services/identity/driver/storage.go`: `GetConfigurationID` added to
  `IdentityStoreService` (mocks regenerated).
- `token/services/storage/db/sql/common/identity.go`: SQL implementation (selects `conf_id`).
- `token/services/storage/db/kvs/identitydb.go`: kvs implementation, the row-key mismatch guard,
  and the `AddConfiguration` overwrite refusal.
- `docs/services/identity.md`: conf_id-collision-free section, byte-wise rationale, stored-state
  rationale, kvs caveat.

## Tests

- `TestIdentityConfigurationCompositeKey_Collision` — reproduction of the original collision.
- `TestIdentityConfigurationCompositeKey_Injective` — 18 adversarial tuples (separators, escapes,
  empty fields, multi-byte runes, invalid UTF-8) pairwise distinct in both `CompositeKey` and
  `conf_id`.
- `TestIdentityConfigurationCompositeKey_InvalidUTF8Collision` and
  `TestEscapeConfigKeyField_PreservesBytes` — the byte-wise property.
- `TestIdentityConfigurationUniqueID_BackwardCompatible` — pins the pre-fix hashes for fields
  containing neither delimiter.
- `TestSignerRouter_CollidingConfigurationsRouteIndependently` — two colliding configurations
  route to their own `KeyManager`, in both directions, asserting the other is never touched.
- `lm_confid_test.go` — binds under the stored `conf_id`, computes one for a new configuration,
  and keys `SignerRouter` by the stored value.
- `dbtest.TGetConfigurationID` (all SQL backends + kvs) — round-trip, misses, and the row-key
  collision: whichever way a store resolves the insert, the configuration already stored must
  come back intact.
- `TestIdentityStoreGetConfigurationID` — the kvs row-key collision and the refused overwrite.
- `FuzzIdentityConfigurationUniqueIDInjective` (`conf_id` equality ⟺ same tuple) and
  `FuzzConfigKeyFieldRoundTrip` (escape/unescape round-trip against an independently written
  decoder), both wired into `nightly-fuzz.yml`.
- Verified by reverting the fix: each test fails.


atharrva01 (commented):
thanks for picking this up, the escaping approach looks right to me. i brute forced `CompositeKey` over `""`/`a`/`@`/`\` up to three chars a field and couldn't find a collision for valid UTF-8.

i think there may be one case left, though tell me if i'm reading it wrong. `escapeConfigKeyField` walks the field with `for _, r := range field`, and that yields `U+FFFD` for each invalid UTF-8 byte, so `WriteRune` writes the replacement character and the original byte is lost. when i tried it:

```
ID "\xff@"  ->  conf_id 8unv8vxDrEvEG++mEn8wETnFf9faD3xnifTkrelHOY0=
ID "\xfe@"  ->  conf_id 8unv8vxDrEvEG++mEn8wETnFf9faD3xnifTkrelHOY0=
```

both land on the same conf_id. it only bites when the field also contains `@` or `\`, so the early return does keep it fairly narrow.

if you think it's worth closing, a byte loop would do it, since both delimiters are ASCII:

```go
for i := 0; i < len(field); i++ {
	if field[i] == configKeySeparator || field[i] == configKeyEscape {
		b.WriteByte(configKeyEscape)
	}
	b.WriteByte(field[i])
}
```

honestly not something likely to come up in practice, `URL` is `ci.Path` so a filesystem path, so entirely your call on whether it's worth the churn.

one other thing you might find useful, not a problem as such: the fuzz target compares two independently fuzzed tuples, so the mutator has to land two different inputs that happen to collide. i ran it for 60s and 470k execs and it didn't surface the above. a single input round trip found it in 0.32s, if that shape appeals:

```go
f.Fuzz(func(t *testing.T, field string) {
	require.Equal(t, field, unescapeConfigKeyField(escapeConfigKeyField(field)))
})
```

one minor note inline. thanks again for the fix.

token/driver/wallet.go
```diff
@@ -219,12 +220,64 @@ type IdentityConfiguration struct {
...
+
 // UniqueID returns a stable identifier for this identity configuration, derived
 // deterministically from its (ID, Type, URL) tuple. It is used to link records that
 // originate from this configuration (e.g. wallet entries) back to it, without requiring
 // a database round-trip to compute.
+//
+// It hashes CompositeKey, so distinct configurations always get distinct UniqueIDs. Because
+// CompositeKey is byte-identical to the older unescaped encoding whenever no field contains a
+// separator or an escape character, the value is unchanged for such configurations and the
+// persisted conf_id column stays valid across an upgrade.
```
atharrva01: minor, and quite possibly you've already weighed this up. it holds when no field has a separator or escape in it, but for the ones that do `UniqueID` changes, and those rows already exist: `lm.go:787` persists it as `ConfigurationID` through `BindIdentity`. after an upgrade `Lookup` reads the old value back from storage while `Register` used the new one, so `byConfID` misses and it drops to the probe fallback.

as far as i can tell nothing breaks, it's just slower for those identities, and arguably fine since they were ambiguous anyway. might be worth a sentence noting it, but happy to leave it if you'd already considered it.

AkramBitar:
> thanks for picking this up, the escaping approach looks right to me. i brute forced `CompositeKey` over `""`/`a`/`@`/`\` up to three chars a field and couldn't find a collision for valid UTF-8.
> 
> i think there may be one case left, though tell me if i'm reading it wrong. `escapeConfigKeyField` walks the field with `for _, r := range field`, and that yields `U+FFFD` for each invalid UTF-8 byte, so `WriteRune` writes the replacement character and the original byte is lost. when i tried it:
> 
> ```
> ID "\xff@"  ->  conf_id 8unv8vxDrEvEG++mEn8wETnFf9faD3xnifTkrelHOY0=
> ID "\xfe@"  ->  conf_id 8unv8vxDrEvEG++mEn8wETnFf9faD3xnifTkrelHOY0=
> ```
> 
> both land on the same conf_id. it only bites when the field also contains `@` or `\`, so the early return does keep it fairly narrow.
> 
> if you think it's worth closing, a byte loop would do it, since both delimiters are ASCII:
> 
> ```go
> for i := 0; i < len(field); i++ {
> 	if field[i] == configKeySeparator || field[i] == configKeyEscape {
> 		b.WriteByte(configKeyEscape)
> 	}
> 	b.WriteByte(field[i])
> }
> ```
> 
> honestly not something likely to come up in practice, `URL` is `ci.Path` so a filesystem path, so entirely your call on whether it's worth the churn.
> 
> one other thing you might find useful, not a problem as such: the fuzz target compares two independently fuzzed tuples, so the mutator has to land two different inputs that happen to collide. i ran it for 60s and 470k execs and it didn't surface the above. a single input round trip found it in 0.32s, if that shape appeals:
> 
> ```go
> f.Fuzz(func(t *testing.T, field string) {
> 	require.Equal(t, field, unescapeConfigKeyField(escapeConfigKeyField(field)))
> })
> ```
> 
> one minor note inline. thanks again for the fix.

 Good catch,  range decodes UTF-8 and loses the byte. Will fix with the byte loop, and switch to your round-trip fuzz target. Thanks!

token/driver/wallet.go
AkramBitar: ▎You're right — added a paragraph on UniqueID and in docs/services/identity.md.

AkramBitar:
@atharrva01 

Thanks a lot for the great review that you give me. 

Do you have other comments? 

Regards,
Akram

atharrva01:
hi @AkramBitar 

nothing further from me, it all looks good.

i re-checked the invalid utf-8 case on the new code and it's gone, `"\xff@"` and `"\xfe@"` now land on different conf_ids. also ran the round-trip target for 45s, 1.1m execs, clean, where the old encoder fell over in 0.32s.

one thing i thought was neat: writing `unescapeConfigKeyField` independently of the encoder so the fuzz acts as a real oracle instead of mirroring it. that catches drift on either side, which is stronger than what i'd suggested.

and the `UniqueID` paragraph traces the upgrade path a lot more thoroughly than my note did. thanks for taking the comments on board.

AkramBitar:
Hi @atharrva01 thanks a lot for the comment.

Fuzz numbers are reassuring, and agreed the independent decoder is the better shape.

on the UniqueID paragraph though: I have to walk it back. "nothing breaks" holds on the read path only — `StoreIdentity` hits the `wallets.conf_id` FK and fails hard (reproduced on sqlite with `alice@org1`), so an upgraded node with `@` in ID/type/URL can't register recipients. Fixing the code and the docs rather than documenting it.

atharrva01:
@AkramBitar , my inline was wrong on that. i traced the read side, `GetConfID` handing `Resolve` the stored value, `byConfID` missing, falling back to probing, and stopped there having concluded "correct, just slower". i never followed the write side, and `wallets.conf_id` has `FOREIGN KEY(conf_id) REFERENCES <identity_configurations>(conf_id)` (`wallet.go:171`), so a fresh `StoreIdentity` on an upgraded node points at a conf_id that isn't in the referenced table yet. that's a hard failure, not a slow path.

so "nothing breaks" was me checking one direction and generalising from it. sorry for sending you down that route in the first place.

agreed fixing it beats documenting it. if it helps, the affected set is exactly the configurations whose `ID`, `Type` or `URL` contains `@` or `\`, since everything else still encodes byte-identically and keeps its old conf_id.

AkramBitar:
"Hi @atharrva01, I’ve addressed the feedback and added the fixes. Could you please take another look when you have a chance?"

atharrva01 (commented):
this looks right to me, and the shape is nicer than i expected. reading the stored value rather than recomputing is the bit that actually closes it, and folding the old `ConfigurationExists` call into the same lookup is a bonus.

i checked the sql path end to end: `confIDFor` returns the persisted conf_id untouched, so `StoreIdentity` now writes a wallets row pointing at a conf_id that exists in `identity_configurations` and the FK holds. the tests cover both branches and the router key, and `token/services/identity/...` plus `token/driver/...` are green under `-race`.

one small thing on the kvs side, inline. and thanks for chasing the FK down, my original note only looked at the read path and i'd have left you with a much worse bug.

token/services/storage/db/kvs/identitydb.go
```diff
@@ -88,6 +88,36 @@ func (s *IdentityStore) GetConfiguration(ctx context.Context, id, typ, url strin
 	return &res, nil
 }
 
+// GetConfigurationID returns the conf_id of the stored configuration with the given id, type,
+// and url, or the empty string if that configuration is not stored yet.
+//
+// Unlike the SQL backend, this store serialises the whole IdentityConfiguration under a
+// composite key and keeps no separate conf_id, so the identifier is derived from the stored
+// value rather than read back. There is no foreign key here for a stale identifier to violate,
```
atharrva01, line 117: the FK part is right, nothing hard-fails here. but i think kvs still ends up on the degraded path, so the comment reads a bit more reassuring than it is.

this derives via `c.UniqueID()`, so it returns the new encoding. meanwhile the kvs `WalletStore` does persist confID (`walletdb.go:49`) and `GetConfID` reads the old one back, and `SetConfIDResolver(walletDB)` makes that store the resolver. so after an upgrade `Register` is keyed on the new value while `GetConfID` hands back the old one, `byConfID` misses, and signer deserialization falls back to probing. same degradation as before, just without the crash.

not suggesting you plumb a stored conf_id through kvs if you'd rather not, it's a fair call that it isn't worth the bookkeeping. maybe just a clause saying the lookup still misses there, so nobody reads this as kvs being unaffected.

AkramBitar: Agreed, and the clause you are asking for is in the committed version of this comment — you anchored on the last line of the reassuring sentence, and the next two lines say exactly this: "a configuration stored under an earlier encoding is reported with the current one, so SignerRouter lookups for its identities miss and fall back to probing". So no code change needed for that part.

You were right that the doc was the gap, though: `docs/services/identity.md` asserted "No migration is needed, and nodes running either release agree on the identifier for the same configuration" without qualification, which is only true of the SQL backend. The kvs caveat is now spelled out there too, naming `WalletStore.GetConfID` as the resolver that still returns the old value.

Chasing this also turned up a real bug in the guard itself, now fixed in e41b369: reporting "not stored" is correct, but `commitLocalIdentity` then goes on to insert, and kvs `AddConfiguration` did a plain `Put` on the shared row key — so the colliding insert replaced `{ID: "bob", URL: "/msp/alice"}`, dropping its Config/Raw and leaving it unable to reload from the store. Before conf_id was read back, `ConfigurationExists` returned true for that lookup and the insert was skipped, so this was a regression introduced by the read-back. `AddConfiguration` now refuses to write when the row key holds a different (ID, Type, URL), and the shared dbtest case asserts the backend-agnostic property: accepted or refused, the configuration already stored must come back intact.

token/driver/wallet.go
```diff
@@ -219,12 +220,80 @@ type IdentityConfiguration struct {
...
+// for every byte that is not valid UTF-8, which would write a replacement character in place of
+// the original byte and make distinct fields encode identically (issue #2070 review). Copying
+// bytes through verbatim keeps the encoding injective for arbitrary field contents, valid UTF-8
+// or not.
+//
+// A field containing neither delimiter is returned unchanged. That is deliberate: it keeps the
+// composite key — and therefore the conf_id derived from it — exactly byte-identical to what an
+// earlier release computed, which is the common case.
+//
+// Every other configuration gets a different key than before. That set is wider than the set of
```
Effi-S: nit: 4 paragraph comment for function is excessive. Can we shorten to 1 / 1.5 ?

Effi-S (approved):

AkramBitar:
@atharrva01 

I would like to thanks a lor for your review. 

Any other comments before I will merge?

Regards,
Akram

atharrva01:
No @AkramBitar , this looks good to be , you can merge!

SaidAltury-ibm (approved):
LGTM

PR #2102, opened 2026-08-02
fix(identity): reject trailing-byte and non-canonical DER in identity decode paths (#2071)

### LINKED ISSUE #2071: identity: trailing-byte and non-canonical DER accepted in every identity decode path
### Summary

Every identity-decoding path in this tree discards the "remaining bytes" return from `asn1.Unmarshal`
without checking it's empty, and the low-level SEQUENCE-length check in `marshal/marshal.go` doesn't
verify the parsed position reaches the end of the buffer. This means distinct byte strings (one
canonical, one with trailing garbage or non-minimal encoding) can decode to the same logical
`TypedIdentity` while still hashing to *different* `UniqueID()` values, since `UniqueID()` hashes the
raw bytes, not the decoded value. This is filed as a robustness issue scoped narrowly to identity
decoding — not signatures — because all identities in this system are produced by our own canonical
`asn1.Marshal`, so no known live path actually substitutes a re-encoded identity for the original
today.

### Where

- `token/services/identity/multisig/identity.go:35` — discards `asn1.Unmarshal`'s `rest` return.
- `token/services/identity/multisig/sig.go:29` — same.
- `token/services/identity/boolpolicy/identity.go:78` — same.
- `token/services/identity/boolpolicy/sig.go:33` — same.
- `token/services/identity/marshal/marshal.go:55` — reads the outer SEQUENCE's length and discards
  it with no `pos != len(b)` check after parsing the declared fields.

### Impact

Because `UniqueID()` (used as cache keys throughout the identity/wallet layers — `role/registry.go`
fast-path cache, `provider.go` signer cache, etc.) hashes the *raw bytes* rather than a canonicalized
form of the decoded value, two byte strings that decode to the same logical identity via a lenient
parse (one with trailing bytes appended, say) are treated as different cache keys even though they'd
behave identically once decoded. This is a latent robustness gap rather than a demonstrated exploit:
no code path in this repo currently constructs or substitutes a non-canonical re-encoding of an
existing identity in place of the original.

**Explicitly out of scope for any fix here:**
- Signature paths (`x509/crypto/ecdsa.go:47`, `idemixnym/nym/signer.go:45`) parse signatures produced
  by external signers and HSMs, whose DER encoders are routinely non-minimal in ways that are still
  valid DER for signature purposes; tightening these would risk rejecting live counterparty traffic
  and is a much larger discussion (borders on a consensus-rule change).
- `marshal/marshal.go:177` `parseInt32`'s 5-byte allowance — a leading `0x00` byte to keep a
  high-bit-set value positive is *correct*, canonical DER; this is not a laxness bug.

This needs maintainer sign-off before any implementation, since tightening decode strictness for
identities that flow into consensus-relevant equality/hash checks touches behavior that other nodes
may depend on if any non-canonical identity bytes are already in flight anywhere in a live network.

### Reproduction

Not yet committed. Construct a valid `MultiIdentity`/policy identity, ASN.1-marshal it, append trailing
garbage bytes, and confirm `Deserialize` still succeeds (ignoring the trailing bytes) while
`UniqueID()` on the two byte strings (with and without the trailing garbage) differs.

### Suggested fix

At each of the four call sites, check that `rest` is empty after `asn1.Unmarshal` and return an error
otherwise. In `marshal/marshal.go`, after parsing all declared fields, verify the read position
equals the buffer length. Scope any change to identity decode paths only, per the exclusions above.

### Severity

LOW (latent — no demonstrated live exploitation path, but worth tracking as a hardening item).

AkramBitar:
`Identity.UniqueID()` hashes the **raw** identity bytes, not the decoded value, and it is the cache key across the identity and wallet layers. So two byte strings that decode to the same identity but differ as bytes give that one identity two cache slots. Every identity decode path allowed exactly that.

## The problem

Alice and Bob's 2-of-2 multisig identity:

```
canonical : 300e 300c 0405 616c696365 0403 626f62
variant   : 3011 300c 0405 616c696365 0403 626f62 02012a
                 ^^ 14 -> 17                      ^^^^^^ undeclared junk
```

Both decode to `MultiIdentity{alice, bob}`; both were accepted; their `UniqueID()`s differ.

Transfer a token to `variant` and it **verifies** — verification works on the decoded identities — but never resolves to Alice's or Bob's wallet, because the lookup works on `UniqueID()`. A valid, signed, spendable token neither owner can see.

Four spellings of one identity did this:

| spelling | caught by |
|---|---|
| garbage appended **after** the value | `rest != 0` |
| garbage smuggled **inside** the SEQUENCE, outer length grown to cover it | re-encode + compare |
| `T61String`/`IA5String`/`GeneralString` where `UTF8String` was declared | re-encode + compare |
| non-minimal lengths (`0x81 0x06`) and integers (`02 02 00 05`) | `readLen` / `parseInt32` |

The second is the subtle one: `rest` only reports bytes *after* the top-level TLV, and `encoding/asn1` silently discards SEQUENCE elements the destination struct has no field for. The outer TLV consumed the whole input, so `rest` is empty and a check on `rest` alone sees a clean parse.

## The fix

* **`marshal.UnmarshalCanonicalDER`** (`MultiIdentity`, `PolicyIdentity`, `MultiSignature`, `PolicySignature`) checks `rest`, then re-encodes the decoded value and requires it to reproduce the input byte-for-byte (`ErrNonCanonical`). One check for the first three rows, and for any further `encoding/asn1` leniency. Cost: one extra marshal of a small struct per decode.
* **`marshal.DecodeIdentity`** (the `TypedIdentity` envelope, and the hot path — hand-walks TLVs, no re-marshal) pins the outer SEQUENCE length to `len(b)`, requires the OCTET STRING to end exactly at `len(b)`, and rejects non-minimal lengths (`ErrNonMinimalLen`) and non-minimal INTEGER contents (`ErrNonMinimalInt`).

## No existing identity can be rejected

The envelope is single-sourced: all seven construction sites funnel through `TypedIdentity.Bytes()` → `marshal.EncodeIdentity`, and the four inner envelopes only through their own `asn1.Marshal`. Every one of those encoders is canonical — `appendTLV` emits minimal lengths, `encodeInt32` strips exactly the bytes `parseInt32` now rejects, and `asn1.Marshal` is canonical by construction, including the legacy string-typed spellings written by older versions of this SDK. There is no non-Go producer of these bytes.

So nothing previously written — on a ledger or in storage — becomes undecodable, and old and new nodes cannot disagree during a rolling upgrade. Pinned by `TestUnmarshalCanonicalDER_AcceptsWhatStdlibMarshalEmits`, `TestDecodeIdentity_AcceptsEveryLengthFormOurEncoderEmits`, `TestDecodeIdentity_AcceptsEveryIntegerOurEncoderEmits` and `TestDecodeIdentity_StdlibEncodingsRemainDecodable`. **If you know of a producer outside this repo, say so** — the two non-minimal checks are the only ones that could reject a legitimately-encoded foreign identity, and can be dropped or gated separately.

Key material is unaffected: x509 certs and idemix credentials live inside the `OCTET STRING` and never reach these decoders.

## Scope

Identity **envelopes** only, per the issue. Deliberately not covered, and now stated in `docs/services/identity.md`:

* Signature parsing (`x509/crypto/ecdsa.go`, `idemixnym/nym/signer.go`) stays lenient — those bytes come from external signers and HSMs. The `MultiSignature`/`PolicySignature` *envelopes* are strict since we always produce them; the signatures they carry are not.
* **The legacy type fold is still a `UniqueID()` split, and it is the same class of problem.** `INTEGER 2`, `UTF8String "x509"` and `PrintableString "x509"` all decode to one x509 identity under three different `UniqueID()`s, so a token paid to the second or third spelling of a victim's identity verifies but does not reach their wallet. This PR narrows that set from unbounded to exactly three; closing it to one cannot be done in the decoder, because older versions of this SDK wrote those spellings and they may exist in persisted data. It needs a rule at the validator boundary — require the `INTEGER` spelling for new transactions, keep decoding the others for reads — tracked separately.
* The payload inside the `OCTET STRING` is **protobuf** for x509/idemix identities, not DER, and remains malleable.

`UnmarshalCanonicalDER`'s round-trip means "`b` is what `asn1.Marshal` would emit", which is narrower than "valid DER": a field tagged `optional`/`omitempty`, or a `time.Time`, would false-reject legal encodings. None of the four types has one, and `TestUnmarshalCanonicalDER_FourCallSitesHaveNoOptionalFields` reflects over them so a future field trips there rather than in production.

## Tests

Every vector is tested at each affected site, asserting the bypass really did decode to the same value with an empty `rest` and a different `UniqueID()` — not just that it now errors. The five fuzz targets also assert canonicality; note this is a contract-level regression guard at the four `asn1` sites (`UnmarshalCanonicalDER` enforces it the same way `Bytes()` computes it), and a genuinely independent check only in `FuzzDecodeIdentityNoPanic`, where `DecodeIdentity` and `EncodeIdentity` are separate implementations. boolpolicy's two targets are new and wired into the nightly-fuzz matrix.

Two `TestDecodeErrors` fixtures had an outer length disagreeing with their own buffer, so the new check fired before the inner failure they were named for; their lengths are corrected rather than their expectations.

Squashed to a single commit, rebased onto current `main`.

Fixes #2071

### DIFF
--- a/token/services/identity/marshal/marshal.go
+++ b/token/services/identity/marshal/marshal.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package marshal
 
 import (
+	"bytes"
 	"encoding/asn1"
 	"errors"
 	"math"
@@ -32,7 +33,105 @@ var (
 	ErrUnexpectedTag = errors.New("asn1: unexpected tag")
 	ErrIntOverflow   = errors.New("asn1: integer overflows int32")
 	ErrInvalidLen    = errors.New("asn1: invalid length encoding")
+	ErrTrailingBytes = errors.New("asn1: trailing bytes after value")
+	ErrNonMinimalLen = errors.New("asn1: non-minimal length encoding")
+	ErrNonMinimalInt = errors.New("asn1: non-minimal integer encoding")
+	// ErrNonCanonical means the input is not the encoding asn1.Marshal produces
+	// for the destination type. For the four identity envelopes that is the same
+	// thing as "not canonical DER" — a property their field types guarantee and
+	// TestUnmarshalCanonicalDER_FourCallSitesHaveNoOptionalFields pins. For a type
+	// with an optional/omitempty/time.Time field the two would diverge, which is
+	// why UnmarshalCanonicalDER documents those as unsupported destinations.
+	ErrNonCanonical = errors.New("asn1: not the canonical encoding for this type")
+
+	// ErrInvalidDestination reports a caller programming error — a nil decode
+	// destination — as opposed to the Err* values above, which all report
+	// something wrong with the bytes being decoded. The *T parameter makes the
+	// other misuses (non-pointer, untyped nil) compile errors instead.
+	ErrInvalidDestination = errors.New("asn1: destination pointer is nil")
 )
+
+// UnmarshalCanonicalDER decodes DER-encoded data into val with encoding/asn1 and
+// accepts b only if b is the one canonical encoding of the value it decoded to.
+//
+// The destination is *T rather than any so that the compiler rejects the two
+// misuses that used to be runtime errors — a non-pointer, and an untyped nil.
+// A typed nil pointer is the only one left, and returns ErrInvalidDestination
+// without inspecting b. Every other error value this returns describes the
+// bytes, not the destination.
+//
+// This matters because Identity.UniqueID() hashes the *raw* identity bytes
+// rather than a canonicalised form of the decoded value, and it is the cache
+// key throughout the identity and wallet layers (role/registry.go's fast path,
+// provider.go's signer cache, …). Any two byte strings that decode to the same
+// logical identity but hash differently give that one identity two cache slots:
+// a token paid to the non-canonical spelling still verifies (the verifier works
+// on the decoded value) but never resolves to its owner's wallet (the lookup
+// works on UniqueID()). Use this instead of calling asn1.Unmarshal directly
+// whenever the bytes being decoded came off the wire.
+//
+// encoding/asn1.Unmarshal is lenient in two ways that both have to be closed:
+//
+//   - it reports bytes left over after the top-level value through its first
+//     return value and is happy to ignore them;
+//   - it silently discards SEQUENCE elements the destination struct has no
+//     field for. Those never appear in rest — the top-level TLV did consume the
+//     whole input — so checking rest alone leaves the SEQUENCE body itself as a
+//     free-form channel for garbage. It is similarly happy to accept
+//     T61String/IA5String/GeneralString where the struct asks for a UTF8String.
+//
+// The rest check catches the first. Re-encoding the decoded value and requiring
+// it to reproduce b byte-for-byte catches both, and generalises to any future
+// encoding/asn1 leniency, at the cost of one extra marshal of a small struct per
+// decode. The hot TypedIdentity envelope path does not come through here: it
+// uses DecodeIdentity, which walks the TLVs itself and pins every length.
+//
+// The property this establishes is precisely "b is something asn1.Marshal would
+// have emitted for this type". That is the right definition here because every
+// producer in this tree goes through asn1.Marshal, but it is narrower than "b is
+// valid DER", so it constrains what val may be:
+//
+//   - Do not use it on a type with an `asn1:"optional"` or `asn1:"omitempty"`
+//     field. asn1.Marshal omits such a field when it holds the zero value, while
+//     asn1.Unmarshal accepts it explicitly present — a legal encoding that the
+//     round-trip then rejects. (A field with `default:` is different: DER does
+//     require the default to be absent, so rejecting an explicit one is correct.)
+//   - Do not use it on a type with a time.Time field. UTCTime and
+//     GeneralizedTime both decode, but asn1.Marshal picks one by year, so the
+//     other is rejected.
+//
+// None of the four identity envelopes hit either case: they carry only int32,
+// string and [][]byte fields, with no optional, omitempty or default tags. Check
+// before adding a fifth caller.
+func UnmarshalCanonicalDER[T any](b []byte, val *T) error {
+	// asn1.Unmarshal would reject this too, but only as an untyped structure
+	// error that a caller cannot tell apart from "these bytes are malformed" —
+	// the one distinction this package's error values exist to make.
+	if val == nil {
+		return ErrInvalidDestination
+	}
+
+	rest, err := asn1.Unmarshal(b, val)
+	if err != nil {
+		return err
+	}
+	if len(rest) != 0 {
+		return ErrTrailingBytes
+	}
+
+	// asn1.Marshal takes the value, not a pointer to it. Typing val as *T makes
+	// this a plain dereference: no reflection, and no pointer-to-pointer case to
+	// reason about, because *T is exactly one level.
+	reencoded, err := asn1.Marshal(*val)
+	if err != nil {
+		return err
+	}
+	if !bytes.Equal(b, reencoded) {
+		return ErrNonCanonical
+	}
+
+	return nil
+}
 
 // Result holds the decoded payload. IsInt distinguishes the two variants.
 // Data is a zero-copy sub-slice of the input — do not modify input while using it.
@@ -52,9 +151,19 @@ func DecodeIdentity(b []byte) (Result, error) {
 	if len(b) == 0 || b[0] != tagSequence {
 		return r, ErrUnexpectedTag
 	}
-	_, pos, err := readLen(b, 1) // skip SEQUENCE length; we trust inner bounds checks
+	seqLen, pos, err := readLen(b, 1)
 	if err != nil {
 		return r, err
+	}
+	// The outer SEQUENCE must declare exactly the bytes it was given: no
+	// fewer (trailing garbage after the value) and no more (a truncated
+	// buffer). Without this, b and b+garbage decode identically while
+	// hashing to different Identity.UniqueID()s — see UnmarshalCanonicalDER.
+	if pos+seqLen > len(b) {
+		return r, ErrTruncated
+	}
+	if pos+seqLen < len(b) {
+		return r, ErrTrailingBytes
 	}
 
 	// Dispatch on first element's tag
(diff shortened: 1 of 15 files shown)


token/services/identity/marshal/marshal.go
```diff
@@ -32,8 +34,87 @@ var (
...
+//     round-trip then rejects. (A field with `default:` is different: DER does
+//     require the default to be absent, so rejecting an explicit one is correct.)
+//   - Do not use it on a type with a time.Time field. UTCTime and
+//     GeneralizedTime both decode, but asn1.Marshal picks one by year, so the
+//     other is rejected.
+//
+// None of the four identity envelopes hit either case: they carry only int32,
+// string and [][]byte fields, with no optional, omitempty or default tags. Check
+// before adding a fifth caller.
+func UnmarshalStrict(b []byte, val any) error {
```
Effi-S: Very interesting solution, 
The name here is a bit ambiguous, 
Perhaps `UnmarshalCanonicalDER` ?

Effi-S: Since val accepts `any` we probably need to add guards like: 
```go
 if rv.Kind() != reflect.Pointer || rv.IsNil() {
    return ErrInvalidDestination
}
```
Instead of implicitly relying on encoding/asn1 to validate all of the possible weird cases

token/services/identity/marshal/marshal.go
```diff
@@ -32,8 +34,87 @@ var (
...
+	reencoded, err := asn1.Marshal(v.Interface())
+	if err != nil {
+		return err
+	}
+	if !bytes.Equal(b, reencoded) {
+		return ErrNonCanonical
+	}
+
+	return nil
+}
```
Effi-S, line 134: The important thing I'd change is to make the helper's API narrower and its errors more semantically accurate, rather than trying to make it a generic "strict DER decoder." The round-trip equality itself is the strongest part of the implementation.

AkramBitar:
@Effi-S 

Thanks a lot for the comments I fixed all of them
Could you please have additional review? 

Thanks a lot,
Akram

adecaro (commented):
## Review: does this fully address #2071?

@AkramBitar

**Yes — the fix is correct and well-tested; one non-blocking note about the PR description being stale relative to the pushed code.**

Checked out `fix-2071-identity-canonical-der` at `38e645a` and verified:

```
go build ./...                                            clean
go vet ./token/services/identity/...                      clean
go test -count=1 -race ./token/services/identity/...      pass (marshal, multisig, boolpolicy)
go test -fuzz=Fuzz... -fuzztime=20s (5 targets)            clean, no panics
fuzz-matrix.sh                                             all 5 targets wired into nightly-fuzz.yml
golangci-lint run ./token/services/identity/...            0 issues (3 pre-existing revive hits elsewhere, unrelated)
git merge --no-commit --no-ff main (2 commits behind)      clean, build+test pass, merge aborted
```

I also hand-decoded the Alice/Bob smuggled-junk example from the description byte-for-byte against `DecodeIdentity` (confirms both spellings decode to the same identities with different `UniqueID()`s), and grepped every `asn1.Unmarshal`/`TypedIdentity{...}` construction site in the tree for anything the diff might have missed — nothing left over.

Reverting `MultiIdentity.Deserialize`/`MultiSignature.FromBytes`/`PolicyIdentity.Deserialize`/`PolicySignature.FromBytes` to their pre-fix bodies fails all 9 issue-#2071 reproduction tests cleanly. `marshal.go`'s own new guards (`readLen`'s non-minimal-length check, `parseInt32`'s non-minimal-integer check) can't be tested by a full-file revert — the new test files reference symbols that don't exist pre-fix — so I inverted just those two checks in place and confirmed the 8 non-minimal-encoding tests in `minimal_length_test.go` fail cleanly, then restored the file and re-verified `go build`/`go test` clean.

`checks`, `utest`, `itest`, `bench-check`, `cgo-check` are all green, and the Token Validation Benchmark comment shows no regression (all deltas within ±1% noise).

@Effi-S's three threads on `marshal.go` all look resolved to me: the rename to `UnmarshalCanonicalDER` addresses the naming ambiguity exactly as suggested, the generic `*T` signature closes the non-pointer/untyped-nil cases your `reflect` guard was aimed at (a typed-nil pointer is the one case left, and that now has its own named error and test), and the API-narrowing point is addressed via semantically distinct errors plus `TestUnmarshalCanonicalDER_FourCallSitesHaveNoOptionalFields` rather than a type-level change — worth a quick nod from you that that's an acceptable resolution to the third point specifically, since your reply to Akram was general ("fixed all of them") rather than per-thread.

### What it fixes

| # | Finding | Verified |
|---|---|---|
| 2071.1 | Trailing bytes after a canonical identity/signature envelope were silently accepted | ✅ `TestTrailingBytesRejectedOnIdentityDeserialize`/`...OnSignatureFromBytes` pass; confirmed the bypass at `readLen`/`asn1.Unmarshal`'s `rest` |
| 2071.2 | Garbage smuggled *inside* the SEQUENCE (undeclared extra field) bypassed the `rest` check entirely | ✅ hand-decoded the description's own byte example; `TestExtraElementInsideSequenceRejectedOn...` pass |
| 2071.3 | `T61String`/`IA5String`/`GeneralString` accepted where `UTF8String` was declared | ✅ `TestAlternateStringTagRejectedOnIdentityDeserialize` (boolpolicy only — the only envelope with a UTF8String field) |
| 2071.4 | Non-minimal DER lengths and INTEGER encodings accepted | ✅ `TestDecodeIdentity_RejectsNonMinimalLengthEncoding`/`...IntegerEncoding`, and the targeted revert check above |

All four rows are solid — no gaps found in any of them.

### Non-blocking

- **Stale description:** the description names `marshal.UnmarshalStrict` and two `TestUnmarshalStrict_*` tests three times; the shipped, renamed symbols are `marshal.UnmarshalCanonicalDER` / `TestUnmarshalCanonicalDER_AcceptsWhatStdlibMarshalEmits` / `TestUnmarshalCanonicalDER_FourCallSitesHaveNoOptionalFields`. Worth a quick edit to the description before this is someone's only reference later.
- **Hygiene:** title doesn't carry `#2071` — not a blocker, `Fixes #2071` in the body already links it via the Development sidebar.
- **Out of scope, correctly so:** the legacy `INTEGER`/`UTF8String`/`PrintableString` type-fold `UniqueID()` split you flagged in the description is real and distinct from this fix (can't be closed in the decoder without breaking reads of pre-existing persisted data) — good call tracking it separately rather than trying to squeeze it into this PR.

### Recommendation

Update the three stale `UnmarshalStrict`/`TestUnmarshalStrict_*` references in the description to their shipped names; everything else is good to merge as-is.

AkramBitar:
@adecaro 

Thanks a lot for reviewing — fixed: all 5 stale UnmarshalStrict references in the description are now UnmarshalCanonicalDER, and the title carries (#2071). No code changed, head still 38e645a.


Please let me know if any changes required. 

Regards,
Akram

Effi-S (approved):
Nice👍

Issue #2017, opened 2026-07-30
Token selection is amount-blind, and lock acquisition is serial and pinned to the vault database


adecaro:
## Summary

Panurus ships two token selectors — `sherdlock` (DB-backed locking, the default) and `simple` (in-memory). Both decide *where* candidate tokens are enumerated from; neither decides *how* they are chosen. Selection is a randomized greedy first-fit: candidates are shuffled (`collections.NewPermutatedIterator`), locked as they are encountered, and accumulated until the running sum reaches the target, at which point the selector returns immediately (`token/services/selector/sherdlock/selector.go:154-216`). The token amount never enters the decision.

Separately, the locking layer has no batch primitive and its table is structurally pinned to the database carrying vault/commit load.

## Problem 1 — amount-blind selection

Because amount is ignored:

- **The number of inputs a transaction ends up with is uncontrolled.**
- **The wallet fragments over time.** Every payment leaves change and dust is never preferentially swept, so the candidate set grows and skews small.
- **In `sherdlock`, every extra candidate is one more lock row**, i.e. one more DB write per selection.

Selection strategy is not a pluggable dimension today: there is no interface for it and no configuration key. `token/services/selector/config/driver.go` exposes `driver`, `retryInterval`, `numRetries`, `leaseExpiry`, `leaseCleanupTickPeriod` and the fetcher cache knobs — nothing for *how* tokens are chosen.

### Measured impact (ZK driver)

`BenchmarkTransferProofGeneration`, Apple M1 Max, BN254, `-benchtime=10x`:

| bits | inputs x outputs | ns/op | note |
|---|---|---|---|
| 64 | **1 x 1** | **381,804** | range proof skipped entirely |
| 64 | 1 x 2 | 19,931,125 | **52x worse than 1x1** |
| 64 | 2 x 1 | 10,383,725 | |
| 64 | 3 x 1 | 10,463,150 | +0.8% over 2x1 |
| 64 | 3 x 3 | 30,317,346 | |
| 32 | **1 x 1** | **480,204** | |
| 32 | 1 x 2 | 12,954,988 | 27x worse than 1x1 |
| 32 | 2 x 1 | 6,824,138 | |
| 32 | 3 x 1 | 6,993,154 | +2.5% over 2x1 |
| 32 | 2 x 2 | 13,270,267 | |

Three observations:

[...]

PR #2106, opened 2026-08-03
2023_document_actual_randomized_greedy_selection_algorithm

### LINKED ISSUE #2023: docs/services/selector.md documents a smallest-first selection strategy that does not exist in the code
## Summary

`docs/services/selector.md` describes a selection strategy the implementation does not have. Four places make the claim:

- **line 10**: "**Selection Strategy Implementation**: Providing different algorithms (e.g., First-In-First-Out, smallest-first) to optimize for transaction size, cost, or privacy."
- **line 42**: "**Apply Selection Strategy**: Selector picks optimal tokens (e.g., smallest-first to minimize transaction size)"
- **line 51**: "The service supports various strategies for picking tokens... A common strategy is to pick the smallest number of tokens that cover the requested amount to minimize the transaction size and the associated verification overhead on the ledger."
- **line 56**: "**Apply Selection Strategy**: Algorithm picks optimal tokens based on configured strategy (e.g., smallest-first)"

The architecture diagram (lines 28-35) also shows a distinct `Strategy[Apply Selection Strategy]` node between the query and lock stages.

## What the code actually does

There is no selection strategy abstraction, no configurable strategy, and no amount-aware ordering. `Selector.selectInternal` (`token/services/selector/sherdlock/selector.go:154-216`) walks a **shuffled** iterator, locks each candidate as it is encountered, accumulates the running sum, and returns as soon as the target is covered. Token amount plays no part in choosing candidates — only in deciding when to stop. The `simple` selector (`token/services/selector/simple/selector.go:49`) follows the same shape.

Nothing selects "the smallest number of tokens", and there is no configuration key for a strategy. The documented `driver: sherdlock` key (line 92) — commented as "Selection strategy (default: sherdlock)" — chooses the *locking backend*, not a selection algorithm.

## Why it matters

The page reads as a description of current behaviour rather than of intent, so a reader plans against amount-aware, size-minimising selection that they will not get. The `driver` key comment additionally conflates the locker with the selection algorithm, which is the specific distinction that matters when reasoning about selector performance.

## Suggested direction

Describe the actual algorithm — randomized greedy first-fit over a shuffled candidate set — and note explicitly that randomization is deliberate, since it is what spreads concurrent selectors across different candidates and keeps lock contention down. Correct the `driver` key comment to say it selects the locking backend. If smallest-first and FIFO are retained on the page, mark them clearly as possible future strategies rather than supported options.

Related: #2017.

HayimShaul:
`docs/services/selector.md` described a configurable, amount-aware selection strategy
(smallest-first, First-In-First-Out) that the implementation does not have. This replaces
those claims with the algorithm the code actually runs.

- The selection section now describes the **randomized greedy first-fit**: candidates are
  enumerated in randomized order, locked as they are encountered, and the selector returns
  as soon as the running sum covers the request. Token amounts only decide when the loop
  stops, never which candidate is picked.
- States that the randomization is deliberate — it spreads concurrent selectors across
  different candidates and keeps lock contention down — and locates it in the sherdlock
  fetcher, noting that the `simple` driver walks the database order instead.
- The mermaid diagram's `Apply Selection Strategy` node becomes `Take Next Candidate`, with
  the lock-failure loop-back edge that the code actually has.
- The `driver` key comment now says it selects the selector implementation and its locking
  backend, not a selection algorithm; a new `Driver` subsection spells out the two values.
- smallest-first and FIFO are kept only under an explicit "Strategies that are not
  implemented" note pointing at #2017.
- `docs/services.md` carried the same overclaim in its Selector Service summary and is
  corrected to match.

Documentation only — no code changes. `make checks` and `make lint` are clean.

Fixes #2023

### DIFF
--- a/docs/services/selector.md
+++ b/docs/services/selector.md
@@ -1,13 +1,13 @@
 # Selector Service
 
-The **Selector Service** (`token/services/selector`) implements strategic token selection algorithms to ensure that Panurus can efficiently and correctly select the best set of unspent tokens (UTXOs) for any given transaction.
+The **Selector Service** (`token/services/selector`) picks the unspent tokens (UTXOs) that fund a transaction and holds them under a temporary lock while the transaction is assembled, so that concurrent transactions of the same wallet do not try to spend the same tokens.
 
 ## Core Responsibilities
 
 The Selector Service is responsible for:
 *   **UTXO Selection**: Finding a set of spendable tokens that cover the total quantity required for a transfer operation.
 *   **Double-Spending Mitigation**: Temporarily locking selected tokens during the transaction assembly phase to prevent multiple concurrent transactions from attempting to spend the same tokens.
-*   **Selection Strategy Implementation**: Providing different algorithms (e.g., First-In-First-Out, smallest-first) to optimize for transaction size, cost, or privacy.
+*   **Candidate Enumeration**: Walking the wallet's candidate tokens in randomized order, locking each one as it is encountered, and stopping as soon as the accumulated amount covers the request. Token amounts do not order or rank the candidates.
 
 ## Interaction with TTX and Storage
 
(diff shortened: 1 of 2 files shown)


AkramBitar (commented):
Looks good overall. One blocker remains: the error referenced on line 76 is incorrect. Everything else is optional.

Separately, I noticed a pre-existing docs mismatch: docs/configuration.md:454 says leaseCleanupTickPeriod: 90s, but the code default is 1m. That should be a separate one-line fix.

docs/services/selector.md
```diff
@@ -25,43 +25,85 @@ graph LR
...
+
+*   **The number and size of the inputs is not minimized.** A request that a single large
+    token could have covered may well be funded by several small ones.
+*   **The result is not deterministic.** The same request against the same wallet can select
+    a different set of tokens, and a different number of inputs, on each run.
+
+**The randomization is deliberate.** It is what spreads concurrent selectors of the same
+wallet across different candidates: walking a fixed order would make every selector contend
+for the same first tokens, driving up lock failures and, with them, the retry and backoff
+path that ends in `token.SelectorSufficientButLockedFunds`.
```
AkramBitar: **Wrong error — please fix before merge.**

The backoff path does not end here. `SelectorSufficientButLockedFunds` is the *inner* loop's give-up value after `maxImmediateRetries` (`selector.go:185`), and `StubbornSelector` swallows it at `:63`, ultimately returning `SelectorInsufficientFunds` wrapped with "aborted too many times and no other process unlocked or added tokens" (`:97`). It only reaches a caller under `NoBackoff`.

Suggest: "…the immediate-retry path that gives up with `token.SelectorSufficientButLockedFunds`, and beyond it the backoff path that ends in `token.SelectorInsufficientFunds`."

docs/services/selector.md
```diff
@@ -25,43 +25,85 @@ graph LR
...
+more exposed to colliding on the same leading candidates.
 
 **How it works in the flow (see "Selection Logic" subgraph in diagram):**
 1. **TTX Request**: TTX Service requests token selection for a transfer operation
 2. **Query Spendable Tokens**: Selector queries via Fetcher (Cache Hit → fast path, Cache Miss → Token Store - TokenDB)
-3. **Apply Selection Strategy**: Algorithm picks optimal tokens based on configured strategy (e.g., smallest-first)
-4. **Acquire Temporary Lock**: Selected tokens are locked in TokenLocks table to prevent double-spending
+3. **Take Next Candidate**: Selector takes the next token from the randomized candidate set
+4. **Acquire Temporary Lock**: The candidate is locked to prevent double-spending (in the `TokenLocks` table under the `sherdlock` driver, in memory under `simple`); on success its amount is added to the running sum, on failure the loop moves to the next candidate
+5. **Return or Retry**: The selector returns as soon as the sum covers the request; if the candidate set is exhausted while other processes hold locks, it refreshes the token cache and retries a bounded number of times before failing
```
AkramBitar: Two points on this step:

1. **"a bounded number of times" collapses two retry layers.** The hardcoded `maxImmediateRetries = 5` refetch loop keeps its locks (`selector.go:177-194`); the configurable `numRetries`/`retryInterval` backoff loop releases them. A reader told `numRetries` is "Retry attempts for token selection" will assume it governs the refresh loop — it doesn't. Since this section is now the authority on what's configurable, worth naming both bounds.
2. **"refreshes the token cache" is sherdlock-only.** `simple` has no fetcher cache; it re-queries the query service. Reads as covering both drivers because step 4 just did.

docs/services/selector.md
```diff
@@ -129,6 +171,17 @@ token:
...
+### Driver
+
+`driver` selects the selector implementation and, with it, the locking backend:
+
+- **sherdlock** (default): locks in the `TokenLocks` table of the Storage Service, with
+  leases governed by `leaseExpiry` and `leaseCleanupTickPeriod`.
+- **simple**: keeps its locks in memory (see [In-Memory Locker Internals](#in-memory-locker-internals)).
+
+It does **not** select a selection algorithm: both drivers use the randomized greedy
+first-fit described above, and `simple` uses it without the shuffle.
```
AkramBitar: "randomized greedy first-fit … without the shuffle" contradicts itself, and `simple` diverges further than the shuffle: it unlocks everything between retries (`simple/selector.go:180-181`) where sherdlock deliberately holds locks across immediate retries, tracks `potentialSumWithLocked`, and runs a `GetTokens` concurrency check that yields a fourth error, `SelectorSufficientFundsButConcurrencyIssue` (`:195`).

Suggest: "both drivers walk candidates greedily and stop on first cover, but `simple` does so in database order and with different retry semantics."

docs/services/selector.md
```diff
@@ -25,43 +25,85 @@ graph LR
...
-The service supports various strategies for picking tokens (see "Strategy" box in diagram above). A common strategy is to pick the smallest number of tokens that cover the requested amount to minimize the transaction size and the associated verification overhead on the ledger.
+### Token Selection Algorithm
+
+Selection is a **randomized greedy first-fit**. It is not configurable, and it is not
+amount-aware. `Selector.selectInternal` (`token/services/selector/sherdlock/selector.go`)
+does the following:
+
+1. the candidate tokens of the wallet and token type are enumerated in randomized order,
+2. each candidate is locked as it is encountered — a candidate that is already locked by
+   another process is skipped,
```
AkramBitar: Nit: now that the loop is enumerated step by step, this reads as exhaustive — but a `TryLock` failure wrapping `token.SelectorRateLimited` is a hard abort, not a skip (`selector.go:196-199`, same in `simple/selector.go:125`). One clause would cover it.

AkramBitar (changes requested):

AkramBitar (approved):

PR #2107, opened 2026-08-03
Revert "Workfow splitting for forks"

Effi-S:
Reverts LFDT-Panurus/panurus#2103


Effi-S:
Hi @AkramBitar , 
I'm reverting this change because it didn't fix the issue. 
And the  commenting  Github action will not cause a failure of the CI if it fails (but the benchmarks will run and cause failure if they fail. 
This will mean that fork PRs will not have a benchmark comment

AkramBitar (approved):

PR #2108, opened 2026-08-03
2032_guard_fabtoken_htlc_validate_missing_signatures

### LINKED ISSUE #2032: fabtoken: TransferHTLCValidate panics on out-of-range ctx.Signatures index
### Summary

`TransferHTLCValidate` indexes into `ctx.Signatures` using the loop index `i` with no bounds check. If `ctx.Signatures` is nil or shorter than `i+1` for an HTLC-owned input, this panics with an index-out-of-range error instead of returning a validation error.

### Where

`token/core/fabtoken/v1/validator/validator_transfer.go:189`:
```go
for i, in := range ctx.InputTokens {
    ...
    if owner.Type == htlc.ScriptType {
        ...
        sigma := ctx.Signatures[i]   // no bounds check
        metadataKey, err := htlc2.MetadataClaimKeyCheck(ctx.TransferAction, script, op, sigma)
        ...
    }
}
```

Compare with zkatdlog's equivalent, which guards this explicitly (`token/core/zkatdlog/nogh/v1/validator/validator_transfer.go:224`):
```go
if i >= len(ctx.Signatures) {
    return errors.New(...)
}
```

fabtoken has no equivalent guard.

### Reachability

In the default pipeline, `ctx.Signatures` is populated by `TransferSignatureValidate` (one entry per input, plus an optional trailing issuer signature) before `TransferHTLCValidate` runs, so `len(ctx.Signatures) >= len(ctx.InputTokens)` normally holds. Like finding (e), this is a defense-in-depth gap that only becomes directly reachable with a reordered/custom pipeline, or if `TransferSignatureValidate`'s signature-collection logic changes to append fewer entries than inputs in some future edge case — at which point a HTLC-owned transfer would panic instead of returning a controlled validation error. Every panic here is caught by `tcc.go`'s top-level `recover()`, but the underlying gap should be closed the same way zkatdlog already closed it.

### Reproduction

Reproduced locally with a unit test (not yet committed) that constructs an HTLC-owned input at index `i` with `ctx.Signatures` shorter than `i+1`, and asserts `TransferHTLCValidate` panics. Happy to include this regression test in the fix PR.

### Suggested fix

Add a bounds check (`if i >= len(ctx.Signatures) { return errors.New(...) }`) before indexing, mirroring zkatdlog's guard.

### Severity

MEDIUM-HIGH — panic surface directly analogous to a gap zkatdlog's validator has already been hardened against.

HayimShaul:
`TransferHTLCValidate` in the fabtoken transfer validator indexed `ctx.Signatures` with the
`ctx.InputTokens` loop index without a bounds check, so a nil or short `ctx.Signatures` made an
HTLC-owned input panic with index-out-of-range instead of returning a validation error. The
equivalent zkatdlog validator already guards this case.

This PR adds the same bounds check, plus a nil guard for entries of `ctx.InputTokens` whose owner
was dereferenced unconditionally in the same loop, and regression tests for both. Neither state is
reachable through the default pipeline — `TransferSignatureValidate` populates one signature per
input before this step runs — so this is a defense-in-depth fix for reordered or custom validation
pipelines. The validator extension guide now notes that a validation function must not assume other
pipeline steps have already run.

Fixes #2032

### DIFF
--- a/docs/drivers/extending_validator.md
+++ b/docs/drivers/extending_validator.md
@@ -69,6 +69,11 @@ func MyCustomTransferValidation(ctx validator.Context, tr *transfer.Action) erro
 }
 ```
 
+A validation function must not assume that any other step of the pipeline has already run:
+fields that earlier steps populate (`validator.Context.InputTokens`, `Context.Signatures`, ...)
+may be empty or shorter than expected, so bound-check them and return an error instead of
+indexing blindly.
+
 ### 2. Create a custom Validator Driver
 
 Implement the `driver.ValidatorDriver` interface by wrapping the standard one.
--- a/token/core/fabtoken/v1/validator/validator_test.go
+++ b/token/core/fabtoken/v1/validator/validator_test.go
@@ -997,6 +997,93 @@ func TestTransferHTLCValidate(t *testing.T) {
 		require.Error(t, err)
 		assert.Contains(t, err.Error(), "expiration date has already passed")
 	})
+
+	t.Run("MissingSignature_NoPanic", func(t *testing.T) {
+		sender, _ := identity.WrapWithType(x509.IdentityType, []byte("sender"))
+		htlcOwner := newExpiredHTLCOwner(t, sender)
+
+		ta := &actions.TransferAction{
+			Outputs: []*actions.Output{
+				{Owner: sender, Type: "ABC", Quantity: "100"},
+			},
+		}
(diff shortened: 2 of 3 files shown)


AkramBitar (approved):
LGTM

AkramBitar (approved):
LGTM

PR #1749, opened 2026-05-27
docs: add Token SDK metrics reference and identify coverage gaps

SuyashAlphaC:
Closes #1745 

Adds `docs/metrics.md`, a complete reference of the 50 metrics the SDK emits (driver services, ttx lifecycle, finality, envelope sessions, auditor, selection, certification, identity caches, Fabric-X queue), each with type, labels, and description. Linked from the monitoring guide.
  
  Also adds a coverage-gap section ranking the uninstrumented layers — storage, the auditor lock manager, the standard Fabric network path, validation, a transaction-failure counter, and wallet resolution — with concrete suggested metrics for follow-up.

PR #2111, opened 2026-08-03
docs: document exported metric names and coverage gaps

### LINKED ISSUE #1745: Document Token SDK Monitoring Metrics and Identify Missing Metrics
1) List and document all monitoring information (metrics) collected by the Token SDK.
2) Review the list and identify any missing monitoring metrics that should be added.

HayimShaul:
Fixes #1745

## Problem

Panurus exports 54 metrics and documented none of them under the name Prometheus actually serves.

Nothing in the SDK spells those names out. The fully-qualified name is assembled at registration time
from the Go package of the *caller*, and the TMS-scoped provider in `token/core/common/metrics` adds a
stack frame — so metrics created through it are exported under *that* package, not their own. Docs
written from the bare `Name` field therefore list names no query matches. That is why #1749 was closed.

## What this PR does

**1. Documents the exported names** — `docs/development/metrics.md`: how a name is derived, one table
per subsystem (exported name, type, labels, meaning, source), PromQL examples, and a ranked
`## Coverage gaps` section on what a node still cannot report.

**2. Guards the page** — `token/services/metricsdoc` instantiates all 16 metrics constructors the way
production wires them, reads the names back out of a Prometheus registry, and checks them against
`testdata/metrics.golden` and against the page in both directions. It also pins the wiring the names
depend on: the two token drivers are asserted to be the only places a TMS-scoped provider is built, so
dropping that wrapper fails the build instead of silently renaming 21 metrics.

**3. Adds a Grafana dashboard** — `docs/monitoring/grafana/token-sdk.json`, revived from #1749 with the
correct names, plus a row for the four `identity_*` metrics that had no panel. All 54 metrics covered,
19 panels over 9 rows. Its queries are guarded too: every metric must be registered, every name must
carry its package prefix, every label filtered or grouped on must be declared, every `$variable` must
exist. These are the failures Grafana renders as "No data" rather than as an error.

**4. Corrects `docs/drivers/metrics.md`**, which listed declared names as if they were queryable, and
makes `docs/development/monitoring.md` the entry point.

`NewMetrics`/`NewClientMetrics` are exported in the auditor, certification and Fabric-X queue packages
so the guard can build their instrumentation without constructing the surrounding service.

## Verification

- All 54 names read out of a live Prometheus registry, not transcribed from source.
- Each guard check proven to fire by mutating the source: renamed metric, dropped TMS wrapper, stale
  wiring entry, truncated name in the page, absent label, undeclared variable.
- Run against #1749's original dashboard, the query guard reports 53 findings — that PR's defect would
  have failed the build.
- `make checks` and `make lint` exit 0; package green under `-count=3 -race -shuffle=on`.

## Known limitation

The dashboard guard checks **names, not rendering**. PromQL is not parsed (it would need a new
dependency, the exprs contain Grafana variables that are not valid PromQL, and it is the one failure
mode Grafana surfaces itself), and the file has not been imported into a live Grafana against a live
Prometheus — so panel presentation is unverified.

## Follow-up

The metric-hygiene items in the gap analysis (package attribution, stuttering names, missing
`network`/`channel`/`namespace` labels) all change exported names, so they belong in a single
deliberate follow-up with a release note rather than here.

### DIFF
--- a/token/services/metricsdoc/reference_test.go
+++ b/token/services/metricsdoc/reference_test.go
@@ -1,0 +1,841 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package metricsdoc
+
+import (
+	"io/fs"
+	"os"
+	"path/filepath"
+	"regexp"
+	"slices"
+	"strings"
+	"sync"
+	"testing"
+
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/metrics"
+	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/metrics/disabled"
+	promprovider "github.com/hyperledger-labs/fabric-smart-client/platform/view/services/metrics/prometheus"
+	prom "github.com/prometheus/client_golang/prometheus"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/LFDT-Panurus/panurus/token"
+	commonmetrics "github.com/LFDT-Panurus/panurus/token/core/common/metrics"
+	"github.com/LFDT-Panurus/panurus/token/services/auditor"
+	"github.com/LFDT-Panurus/panurus/token/services/certifier/interactive"
+	"github.com/LFDT-Panurus/panurus/token/services/identity"
+	idemixcache "github.com/LFDT-Panurus/panurus/token/services/identity/idemix/cache"
+	"github.com/LFDT-Panurus/panurus/token/services/identity/role"
+	_ "github.com/LFDT-Panurus/panurus/token/services/logging" // registers the "panurus" metric namespace replacer
+	fabricxqueue "github.com/LFDT-Panurus/panurus/token/services/network/fabricx/finality/queue"
+	"github.com/LFDT-Panurus/panurus/token/services/selector/sherdlock"
+	"github.com/LFDT-Panurus/panurus/token/services/ttx"
+	ttxfinality "github.com/LFDT-Panurus/panurus/token/services/ttx/finality"
+	jsession "github.com/LFDT-Panurus/panurus/token/services/utils/json/session"
+)
+
+const (
+	goldenFile = "testdata/metrics.golden"
+	// referenceDoc is the documentation page the golden file mirrors. Every name
+	// in the golden file must appear there, and vice versa.
+	referenceDoc = "../../../docs/development/metrics.md"
+	// repoRoot locates the repository root relative to this package, for the
+	// checks that read production wiring out of the source tree.
+	repoRoot = "../../.."
+	// updateEnv regenerates the golden file instead of comparing against it.
+	updateEnv = "UPDATE_GOLDEN"
+	// gapsHeading starts the part of the reference documentation that discusses
+	// metrics the SDK does *not* export yet. Names proposed there are deliberately
+	// not registered, so the coverage check stops at this heading.
+	gapsHeading = "\n## Coverage gaps\n"
+)
+
+// tmsID supplies the network/channel/namespace values that the TMS-scoped
+// provider binds to every metric it hands out. The values themselves never
+// reach an exported metric name, only its labels.
+var tmsID = token.TMSID{Network: "testnet", Channel: "testchannel", Namespace: "testns"}
+
+// scope records how production reaches a group of metrics: directly with the
+// provider from the dependency-injection container, or through the TMS-scoped
+// wrapper returned by commonmetrics.NewTMSProvider. The distinction is not
+// cosmetic - the wrapper adds a stack frame, and the exported name is derived
+// from the package of the caller that Prometheus sees, so wrapped metrics are
+// exported under the wrapper's package rather than their own.
+//
+// The column is not taken on trust: TestTMSScopedProviderWiringIsIntact pins the
+// complete set of files that may build a TMS-scoped provider, so a group is
+// tmsScoped exactly when its provider originates in one of them.
+type scope int
+
+const (
+	// direct means production passes the container's provider straight in.
+	direct scope = iota
+	// tmsScoped means production passes commonmetrics.NewTMSProvider(tmsID, provider).
+	tmsScoped
+)
+
+// wiringSite is a production call site of a group's constructor. Both fields are
+// checked by TestWiringSitesArePresent, so a stale entry fails the test instead
+// of sending a reader to the wrong file.
+type wiringSite struct {
+	// file is repository-relative.
+	file string
+	// call is the constructor invocation as it appears in that file. Call sites
+	// of the same constructor do not always spell it the same way: some reach the
+	// unexported constructor from inside its own package.
+	call string
+}
+
+// group is one set of metrics created by a single constructor, mirroring one
+// wiring site in the SDK.
+type group struct {
+	// subsystem is the human-readable label used in the documentation.
+	subsystem string
+	// source is the file that declares the metric options.
+	source string
+	// scope is the provider flavour production uses for this constructor.
+	scope scope
+	// wiring lists the production call sites of the constructor. It records where
+	// the metrics are built; which provider flavour reaches them is established by
+	// TestTMSScopedProviderWiringIsIntact, not by this field.
+	wiring []wiringSite
+	// build invokes the constructor exactly as production does.
+	build func(metrics.Provider)
+}
+
+// tokenDrivers are the two files that turn the container's metrics provider into
+// a TMS-scoped one. Every tmsScoped group is reached from here.
+var tokenDrivers = []string{
+	"token/core/fabtoken/v1/driver/driver.go",
+	"token/core/zkatdlog/nogh/v1/driver/driver.go",
+}
+
+// groups enumerates every metrics constructor in the SDK. A new constructor must
+// be added here, otherwise its metrics are absent from the golden file and from
+// the reference documentation.
+var groups = []group{
+	{
+		subsystem: "Driver: issue service",
+		source:    "token/core/common/metrics/issue.go",
+		scope:     tmsScoped,
+		wiring: []wiringSite{
+			{file: "token/core/fabtoken/v1/driver/driver.go", call: "metrics.NewIssueService("},
+			{file: "token/core/zkatdlog/nogh/v1/driver/driver.go", call: "metrics.NewIssueService("},
+		},
+		build: func(p metrics.Provider) { commonmetrics.NewIssueService(nil, p) },
+	},
+	{
+		subsystem: "Driver: transfer service",
+		source:    "token/core/common/metrics/transfer.go",
+		scope:     tmsScoped,
+		wiring: []wiringSite{
+			{file: "token/core/fabtoken/v1/driver/driver.go", call: "metrics.NewTransferService("},
+			{file: "token/core/zkatdlog/nogh/v1/driver/driver.go", call: "metrics.NewTransferService("},
+		},
+		build: func(p metrics.Provider) { commonmetrics.NewTransferService(nil, p) },
+	},
+	{
+		subsystem: "Driver: auditor service",
+		source:    "token/core/common/metrics/auditor.go",
+		scope:     tmsScoped,
+		wiring: []wiringSite{
+			{file: "token/core/fabtoken/v1/driver/driver.go", call: "metrics.NewAuditorService("},
+			{file: "token/core/zkatdlog/nogh/v1/driver/driver.go", call: "metrics.NewAuditorService("},
+		},
+		build: func(p metrics.Provider) { commonmetrics.NewAuditorService(nil, p) },
+	},
+	{
+		subsystem: "Driver: tokens service",
+		source:    "token/core/common/metrics/tokens.go",
+		scope:     tmsScoped,
+		wiring: []wiringSite{
+			{file: "token/core/fabtoken/v1/driver/driver.go", call: "metrics.NewTokensService("},
+			{file: "token/core/zkatdlog/nogh/v1/driver/driver.go", call: "metrics.NewTokensService("},
+		},
+		build: func(p metrics.Provider) { commonmetrics.NewTokensService(nil, p) },
+	},
+	{
+		subsystem: "Driver: tokens upgrade service",
+		source:    "token/core/common/metrics/upgrade.go",
+		scope:     tmsScoped,
+		wiring: []wiringSite{
+			{file: "token/core/fabtoken/v1/driver/driver.go", call: "metrics.NewTokensUpgradeService("},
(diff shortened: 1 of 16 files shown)


token/services/metricsdoc/reference_test.go
```diff
@@ -0,0 +1,525 @@
...
+
+// group is one set of metrics created by a single constructor, mirroring one
+// wiring site in the SDK.
+type group struct {
+	// subsystem is the human-readable label used in the documentation.
+	subsystem string
+	// source is the file that declares the metric options.
+	source string
+	// scope is the provider flavour production uses for this constructor.
+	scope scope
```
AkramBitar, line 101: **medium — the guard can't catch the failure it exists to catch.**

`scope` is hand-written metadata that nothing verifies, and the test builds its provider from it rather than from production. That means the one change which actually invalidates the documentation is invisible here.

Concretely: change `token/core/fabtoken/v1/driver/driver.go:134` (and the zkatdlog twin at `:135`) to pass `d.metricsProvider` instead of `metrics.NewTMSProvider(tmsConfig.ID(), d.metricsProvider)`. In a real node, all 21 driver + identity metrics then export under their own package prefixes instead of `panurus_core_common_metrics_*` — but the golden file, `docs/development/metrics.md` and `docs/drivers/metrics.md` all keep the old names, and **all three tests still pass**. That is exactly the "documentation lists names no query matches" failure this PR exists to prevent.

Worth asserting the wiring itself — either by constructing the real driver/DI path, or at minimum a comment-anchored grep on `driver.go` for `NewTMSProvider`.

token/services/metricsdoc/reference_test.go
```diff
@@ -0,0 +1,525 @@
...
+// expected to pass only the part of the page that documents existing metrics; see
+// gapsHeading.
+func extractMetricNames(doc string) map[string]struct{} {
+	out := make(map[string]struct{})
+	for _, match := range metricNamePattern.FindAllStringSubmatch(doc, -1) {
+		name := match[2]
+		if strings.HasSuffix(name, "_") {
+			continue
+		}
+		for _, suffix := range []string{"_bucket", "_sum", "_count"} {
```
AkramBitar: **medium — this trim corrupts real metric names and will fail on a correct page.**

The suffix strip is unconditional, so it also rewrites metric families whose exported name genuinely ends in one of those suffixes. A table row for `panurus_services_ttx_retry_count` lands in `documented` as `panurus_services_ttx_retry`, so the forward assertion at line 274 (`assert.Contains(documented, "panurus_services_ttx_retry_count")`) fails and reports a correctly documented metric as undocumented. Reproduced standalone with this logic in isolation.

No metric hits this today, so the guard is green — but the first future metric ending in `_count`, `_sum` or `_bucket` breaks it. Suggest only trimming when the trimmed base is itself a registered histogram family.

token/services/metricsdoc/reference_test.go
```diff
@@ -0,0 +1,525 @@
...
+// collect instantiates every group twice: once against a recording provider that
+// captures the declared options, and once against a real Prometheus provider
+// that yields the exported names. Both passes run the same constructors in the
+// same order, so the two results line up index by index.
+func collect(t *testing.T) []entry {
+	t.Helper()
+
+	declared := declaredOptions(t)
+	exported := exportedNames(t)
+	require.Len(t, exported, len(declared))
```
AkramBitar: **low/medium — index pairing breaks silently when two groups share a metric name.**

`declared[i]` is paired to `exported[i]` positionally, but `exportedNames` shares a single `promprovider.Provider` across all groups, and FSC's provider calls `prom.Register` only on the first request per `kind|fqName` (see `getOrCreate`/`register` in `platform/view/services/metrics/prometheus/provider.go`). So a duplicate fully-qualified name adds a `declared` entry with no matching `exported` one and shifts every subsequent pair.

I hit this by adding a second group for sherdlock's other production wiring site (`fetcher.go:68`, mirroring the server/client split already used for `interactive`): the failure is `"[...54 names...]" should have 58 item(s), but has 54` — a wall of text with no hint that provider-level dedup is the cause. Pairing by `(kind, fqName)` with an explicit duplicate check, or a fresh `Provider` per group, would make this diagnosable.

token/services/metricsdoc/reference_test.go
```diff
@@ -0,0 +1,525 @@
...
+
+	return out
+}
+
+// isPrefixOfRegistered reports whether a name mentioned in the documentation is a
+// proper prefix of a metric the SDK registers. Such a mention is prose about the
+// naming scheme - a namespace such as "panurus_services" - rather than a claim
+// that a metric by that name exists. A misspelled or stale metric name is not a
+// prefix of anything and is therefore still reported.
+func isPrefixOfRegistered(name string, registered map[string]struct{}) bool {
```
AkramBitar: **low — the prefix escape hatch lets truncated names through.**

This makes the reverse check ("the page must not mention names that are no longer exported") accept any documented name that is a proper prefix of a registered one. `panurus_services_ttx_accepted` (missing `_transactions`) or `panurus_core_common_metrics_cache` both pass.

The doc comment above says a misspelled or stale name "is not a prefix of anything and is therefore still reported" — that holds for misspellings but not for truncation, which is the most likely copy-paste error in a hand-maintained 54-row table and precisely the class of error this page exists to eliminate. An explicit allowlist of the handful of prose namespaces (`panurus_services`, …) would close the hole without losing the intent.

token/services/metricsdoc/reference_test.go
```diff
@@ -0,0 +1,525 @@
...
+// wiring site in the SDK.
+type group struct {
+	// subsystem is the human-readable label used in the documentation.
+	subsystem string
+	// source is the file that declares the metric options.
+	source string
+	// scope is the provider flavour production uses for this constructor.
+	scope scope
+	// wiring points at the production call site, so a reader can check the scope.
+	wiring string
```
AkramBitar: **low — `wiring` is never read, and points at the wrong file for 5 of the 16 groups.**

Nothing in this file consumes `group.wiring`, and where it is set it often names the constructor's own file rather than "the production call site" the comment promises:

- `"TTX: finality listener"` → says `token/services/ttx/finality/{listener,recovery}.go`; actual sites are `token/services/ttx/db.go:70`, `token/services/auditor/auditor.go:239`, `token/services/network/fabric/network.go:513`.
- `"Certification service (server)"` / `"Certification client"` → say `interactive/{service,client}.go`; both providers originate at `token/services/certifier/interactive/driver.go:195` and `:170`.
- `"Fabric-X finality queue"` → says `queue/queue.go`; the provider comes from `token/services/network/fabricx/driver.go:78`.

Since `scope` correctness is the linchpin of the whole guard (see the comment on the `scope` field), a reviewer who follows `wiring` to check it gets sent to the wrong file.

token/services/metricsdoc/reference_test.go
```diff
@@ -0,0 +1,525 @@
...
+	t        *testing.T
+	previous prom.Registerer
+	names    []string
+}
+
+func newCapturingRegisterer(t *testing.T) *capturingRegisterer {
+	t.Helper()
+
+	c := &capturingRegisterer{Registerer: prom.NewRegistry(), t: t, previous: prom.DefaultRegisterer}
+	prom.DefaultRegisterer = c
```
AkramBitar, line 660: **low — unsynchronized swap of a package-level global.**

`prom.DefaultRegisterer` is assigned with no synchronization, and FSC's provider documents the opposite assumption ("It assumes the default registerer is not swapped while the provider is in use"). This is safe today only because the three tests run strictly sequentially — adding `t.Parallel()` to any of them, or importing a package that registers a collector from a background goroutine, gives a data race on the global plus non-deterministic name capture.

Relatedly, `capturingRegisterer` overrides only `Register`; `MustRegister` is promoted from the embedded `prom.Registerer` and would bypass name capture entirely. If FSC ever switches to `promauto`, that surfaces as the same opaque `require.Len` mismatch described in the comment on line 323.

AkramBitar (commented):
Reviewed the docs against the code: the content is accurate. Every metric name, label value and `scope` claim matches production wiring today (traced the `NewTMSProvider` wrapper in `token/core/{fabtoken,zkatdlog}/v1/driver/driver.go` through `ws.go` into `identity/role/idemix`, and the DI-container providers for ttx / auditor / sherdlock / certifier / fabricx). All 15 files that declare metric opts are covered by the group list, the 54 table rows match the 54 golden lines, the `monitoring.md` URL change fixes a real 404, and `go vet` / `gofmt` / `goimports` / `misspell` / `golangci-lint` plus `go test -count=3 -race -shuffle=on` are clean.

The comments below are all about the new `reference_test.go` guard rather than the documentation. Two of them I'd suggest addressing before merge:

- The guard cannot detect the one change that would actually invalidate the whole page (the `scope` field is unverified metadata).
- The unconditional histogram-suffix trim will fail on a *correct* page as soon as someone adds a metric whose name ends in `_count`, `_sum` or `_bucket`.

The remaining four are polish — confusing failure output, a gap in the reverse check, a dead field, and a global that is only safe while the tests stay sequential.

AkramBitar (dismissed):

AkramBitar:
**Follow-up: bring back `token-sdk.json`, with the names this PR pins.**

#1749 shipped a Grafana dashboard next to its metrics doc — https://github.com/LFDT-Panurus/panurus/pull/1749/files ([permalink](https://github.com/LFDT-Panurus/panurus/blob/b3dfae3157481f079695b41d07a07e1e6d6ce7b5/docs/monitoring/grafana/token-sdk.json)) — 17 panels over 9 rows, with `network`/`channel`/`namespace`/`method` template variables. The panel layout and label filters are sound; the one defect is the one I flagged there: https://github.com/LFDT-Panurus/panurus/pull/1749#discussion_r3431242480 — every `expr` used the bare option name from the Go source, so no query matched anything in a real Prometheus.

That PR closed unmerged, and `docs/monitoring/` is in neither this PR nor `main`, so the dashboard went with it. Worth reviving, because this PR's `testdata/metrics.golden` is exactly the input needed to fix it — the substitution is mechanical:

| `expr` in #1749's dashboard | Correct name |
|---|---|
| `issue_service_operations_total` | `panurus_core_common_metrics_issue_service_operations_total` |
| `endorsed_transactions` | `panurus_services_ttx_endorsed_transactions` |
| `finality_listener_confirmed_total` | `panurus_services_ttx_finality_finality_listener_confirmed_total` |
| `ttx_envelope_sent_total` | `panurus_services_utils_json_session_ttx_envelope_sent_total` |
| `auditor_audit_duration_seconds` | `panurus_services_auditor_auditor_audit_duration_seconds` |
| `unspent_tokens_invocations` | `panurus_services_selector_sherdlock_unspent_tokens_invocations` |
| `certified_tokens` | `panurus_services_certifier_interactive_certified_tokens` |
| `cache_level` | `panurus_core_common_metrics_cache_level` |
| `finality_queue_pending_events` | `panurus_services_network_fabricx_finality_queue_finality_queue_pending_events` |

Two things beyond the rename:

- the four `panurus_core_common_metrics_identity_*` signer-resolution metrics postdate #1749 and have no panel yet;
- the `metricsdoc` package added here could guard the dashboard too — extract every metric name from the panel `expr` fields and assert each one is registered. Without that, the queries can drift out of sync again exactly the way #1749's did, and the failure mode is a silent "No data" rather than an error.

**Do we want the dashboard in this PR, or as a follow-up?** I have no strong preference. Adding it here keeps the dashboard and the names it depends on in one changeset, which is nice for review — but this PR is already sizeable and purely documentation, so a separate PR is equally reasonable. A follow-up would need to be based on this branch rather than `main`, since the `metricsdoc` guard package doesn't exist on `main` yet. Happy either way — let me know which you prefer and I can open an issue to track it.

token/services/metricsdoc/reference_test.go
AkramBitar: Fixed in 83d60d76. The scope column is now proved by exhaustion rather than trusted: `TestTMSScopedProviderWiringIsIntact` greps the repository for `*metrics.NewTMSProvider(` call sites, asserts the set is exactly the two token drivers, and asserts each still contains the wrapper expression verbatim.

Verified against the exact mutation described above — passing `d.metricsProvider` in the fabtoken driver now fails with an actionable message, where previously all three tests stayed green. A new wrapping site anywhere else also fails, which a driver-constructing fixture would have missed.

token/services/metricsdoc/reference_test.go
AkramBitar: Fixed in 83d60d76. `foldPromQLSuffix` leaves a name alone when it is itself registered, and otherwise trims a single suffix only if the base is a registered histogram.

Proven end-to-end by renaming a counter to `accepted_count`, regenerating the golden file and updating the page: the old unconditional trim rejects that correct page, the new fold accepts it. `TestFoldPromQLSuffix` now covers the case permanently, since no metric exercises it today.

token/services/metricsdoc/reference_test.go
AkramBitar: Fixed in 83d60d76. Each group is instantiated against its own `promprovider.Provider` and its own registry, so provider-level dedup can no longer shift the pairing, and the length check now names the group and lists both sides.

Your reproduction — a second group for sherdlock's `fetcher.go` wiring — now passes cleanly, which is the right outcome: two wiring sites legitimately share a collector. Identical definitions are folded into one row; a genuine clash is reported by name.

token/services/metricsdoc/reference_test.go
AkramBitar: Fixed in 83d60d76. `isPrefixOfRegistered` is gone, replaced by an explicit `prosePrefixes` allowlist — `panurus_services` is the only entry the page actually needs.

Confirmed both directions: an extra truncated mention of `panurus_services_ttx_accepted` passes under the old prefix rule and fails under the allowlist.

token/services/metricsdoc/reference_test.go
AkramBitar: Fixed in 83d60d76. `wiring` is now `[]wiringSite{file, call}` and is read by `TestWiringSitesArePresent`, which requires each file to exist and to contain the call verbatim. All five stale entries corrected (TTX lifecycle, TTX finality, certification server/client, Fabric-X queue).

The field's documented purpose is also narrowed: it records where the metrics are built, while the scope claim is established by the exhaustive `NewTMSProvider` check instead.

token/services/metricsdoc/reference_test.go
AkramBitar: Fixed in 83d60d76. A package-level mutex is held for the lifetime of each capture, and `MustRegister` is overridden so a future switch to `promauto` cannot bypass name capture.

Verified by adding `t.Parallel()` to every test in the package: race-free with the mutex, and reproducibly racy without it.

AkramBitar:
@HayimShaul will be out of the office for the next two weeks. I'll be covering for him during that time, so I won't be able to continue reviewing this PR. Someone else will need to take over the review.

Effi-S (approved):
Tests are Backed by a real Prometheus registry, so a rename fails the build

PR #2103, opened 2026-08-03
Workfow splitting for forks

Effi-S:
Minor change for fork permission issue

PR #2114, opened 2026-08-03
lint: enable ireturn

atharrva01:
Next step of #1991, after #2098.

All 64 `ireturn` hits are returns of third-party interfaces, and in none of them is the signature
ours to change. Either the method implements that interface, so the shape is fixed:

- `(p *keyset[I, V]) GoToOffset(...) (driver.Pagination, error)` implements `driver.Pagination`
- `(k *ecdsaPrivateKey) PublicKey() (bccsp.Key, error)` implements `bccsp.Key`
- `(p *provider) NewGauge(...) metrics.Gauge` implements `metrics.Provider`

or it is the framework's own constructor convention, like the `view.View` constructors.

So this allows those 15 interfaces instead of touching code. No source files change, and `ireturn`
still catches the case it is actually for, returning an interface where one of our own concrete
types would do.

### Corrected measurements

I said on #2098 that the per-linter counts were unreliable because I measured them in one combined
run. Redone one linter at a time, root module:

| linter | issues | files | what I said before |
| --- | --- | --- | --- |
| maintidx | 17 | 15 | 10 / 8 |
| ireturn | 64 | 26 | 34 / 17 |
| iface | 88 | 58 | 64 / 46 |
| gocognit | 137 | 83 | 137 / 83 |
| wrapcheck | 2307 | 397 | 2301 / 397 |
| revive | 3452 | 683 | 2663 / 532 |

So four of the six were under-reported, `revive` by about 30%.

On `maintidx`, worth knowing before deciding: 12 of its 17 hits are in `_test.go` and 4 more are in
test helpers (`testutils`, `dbtest`). Only one is production code, `Prove` in
`crypto/rp/csp/rp.go`, which is in the middle of the CSP perf work. So excluding tests would leave
almost nothing, which makes me think the setting is not earning its keep, but that is your call.

`iface` at 58 files is the next unblocked one after this.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,6 +16,7 @@ linters:
     - govet # Vet examines Go source code and reports suspicious constructs. It is roughly the same as 'go vet' and uses its passes. [auto-fix]
     - importas # Enforces consistent import aliases. [auto-fix]
     - intrange # Intrange is a linter to find places where for loops could make use of an integer range. [auto-fix]
+    - ireturn # Accept Interfaces, Return Concrete Types.
     - loggercheck # Checks key value pairs for common logger libraries (kitlog,klog,logr,slog,zap).
     - lll # Reports long lines. [fast]
     - mirror # Reports wrong mirror patterns of bytes/strings usage. [auto-fix]
@@ -90,6 +91,31 @@ linters:
         - github.com/prometheus/client_golang/prometheus.Counter
         - github.com/prometheus/client_golang/prometheus.Gauge
         - github.com/prometheus/client_golang/prometheus.Histogram
+        # Third-party interfaces we are obliged to return: either the signature is fixed by an
+        # interface we implement (Pagination, bccsp.Key, KeyStore) or it is the framework's own
+        # constructor convention (view.View, view.Session). Returning a concrete type instead is
+        # not an option in these cases, so allow them rather than carry a nolint at every site.
+        - github.com/hyperledger-labs/fabric-smart-client/platform/common/driver.Pagination
+        - github.com/hyperledger-labs/fabric-smart-client/platform/view/view.View
+        - github.com/hyperledger-labs/fabric-smart-client/platform/view/view.Session
+        - github.com/hyperledger/fabric-lib-go/bccsp.Key
+        - github.com/hyperledger/fabric-lib-go/bccsp.BCCSP
+        - github.com/hyperledger/fabric-lib-go/bccsp.HashOpts
+        - github.com/IBM/idemix/bccsp/types.BCCSP
+        - github.com/IBM/idemix/bccsp/types.KeyStore
+        - go.opentelemetry.io/otel/trace.Span
+        - github.com/hyperledger/fabric-lib-go/common/metrics.Gauge
+        - github.com/hyperledger/fabric-lib-go/common/metrics.Histogram
+        - github.com/hyperledger-labs/fabric-smart-client/platform/common/driver.ConfigService
+        - github.com/hyperledger-labs/fabric-smart-client/integration/nwo/api.ViewClient
+        - github.com/hyperledger-labs/fabric-smart-client/integration/nwo/api.Platform
+        - github.com/hyperledger-labs/fabric-smart-client/integration/nwo/api.Context
+        - github.com/hyperledger/fabric-lib-go/common/metrics.Gauge
+        - github.com/hyperledger/fabric-lib-go/common/metrics.Histogram
+        - github.com/hyperledger-labs/fabric-smart-client/platform/common/driver.ConfigService
+        - github.com/hyperledger-labs/fabric-smart-client/integration/nwo/api.ViewClient
+        - github.com/hyperledger-labs/fabric-smart-client/integration/nwo/api.Platform
+        - github.com/hyperledger-labs/fabric-smart-client/integration/nwo/api.Context
     lll:
       # Max line length, lines longer will be reported.
       line-length: 240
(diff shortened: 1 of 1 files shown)


AkramBitar:
@atharrva01 is that ready for review?

atharrva01:
yes @AkramBitar

atharrva01:

```
```

atharrva01:
@AkramBitar your thoughts on this?

AkramBitar:
@atharrva01 

Two things:

**1.** Six entries are duplicated verbatim — `metrics.Gauge`, `metrics.Histogram`, `driver.ConfigService`, `api.ViewClient`, `api.Platform`, `api.Context`. That's 21 lines for the 15 interfaces; looks like a mis-resolved rebase against #2098.

**2.** The pre-existing `- T` on line 89 disables `ireturn` for all first-party code — allow entries are [unanchored regexes](https://github.com/butuzov/ireturn/blob/master/analyzer/internal/config/config.go), so `T` matches anything containing a capital `T`, i.e. all of `github.com/LFDT-Panurus/...`. Deleting it surfaces **50 hits in the root module**, and `generic` on line 87 already covers what it was for. Predates this PR, but it means the case `ireturn` exists for stays excluded — suggest dropping it here and handling the 50 as a follow-up.

Thanks a lot for you support.
Akram

atharrva01:
thanks,

**1.** fixed, the six were listed twice. not the rebase though, i ran the "add the remaining interfaces" step twice while putting this together and didn't spot it. list is 25 entries now.

**2.** you're right, and it's worse than the number suggests. `- T` is an unanchored regex so it matches anything containing a capital T, which includes every `github.com/LFDT-Panurus/...` interface. with it in place `ireturn` reports 0 hits repo-wide, so this PR as it stands only really documents the third-party allows.

one correction on the count: 50 is golangci-lint's default `--max-issues-per-linter` cap. running it uncapped gives **330 hits across 123 files**:

```
golangci-lint run --enable-only=ireturn                              -> 50   (capped)
golangci-lint run --enable-only=ireturn --max-issues-per-linter=0    -> 330
```

and agreed that `generic` on line 87 already covers what `T` was presumably for.

that does make it awkward to drop here though: removing `- T` makes `make lint` fail on all 330 straight away, so it can't land on its own without either fixing them in the same PR or carrying nolints. given Angelo asked for steps that minimise files touched, i'd rather keep `- T` in this one and do the removal properly as its own change, with the 330 either fixed or explicitly allowed.

happy to open an issue for it and pick it up after `iface`, unless you'd rather it block this PR.

AkramBitar:
Thanks a lot @atharrva01,

I agree on point 2. Let’s open a follow-up issue for removing T properly. I think it’s better handled separately with the full set of changes needed rather than expanding this PR’s scope.

Let’s keep this PR focused and revisit the ireturn cleanup after iface.

Regards,
Akram

AkramBitar (approved):
LGTM.

By the way I noticed that [Tests / utest (unit-tests-race) (pull_request)](https://github.com/LFDT-Panurus/panurus/actions/runs/31074760407/job/92531713286?pr=2114) still failing 

I am wondering if you rebase with main. I already merged your fix for this issue. 
Could you please have a look? 

Thanks a lot,
Akram

atharrva01:
thanks! opened #2155 for the `T` removal, with the 330 figure and the note about the default cap so whoever picks it up doesn't measure it as 50. will take it after `iface`.

rebased onto main, so this is on top of everything including the flaky test fixes.

on the failing `utest` though, i don't think the rebase will clear it, it's a different problem to the one you merged. that run failed on:

```
FAIL  github.com/LFDT-Panurus/panurus/token/services/selector/simple/inmemory   600.288s
panic: test timed out after 10m0s
  running tests:
    TestScannerDoesNotDeleteReclaimed (9m50s)
```

different package to the three in #2150 (`utils`, `sherdlock`, `ttx`), and a different failure mode: it hangs rather than asserting something wrong. the dump has the test blocked on a channel receive for 9 minutes while `(*locker).scan` sits in a select at `locker.go:429`, with `scan goroutine did not stop within timeout` logged just before.

i tried to reproduce it: passes on clean main, and 40 runs under `-race` locally were all green in 13s total, so it's rare rather than reliable. this PR is `.golangci.yml` only, no go code, so it can't be causing it either.

feels like a fourth one for the flaky pile, though a hang is worse than a bad assertion since it burns the full 10m timeout. happy to open an issue and dig into it, that locker is one i've worked in before. let me know if you'd rather just re-run the job first and see if it comes back.

AkramBitar:
@atharrva01,

Great. Thanks a lot.

I will more than happy if you can open  a new issue and dig into the unit test flaky test that you described above. 


Thanks a lot,
Akram

atharrva01:
opened #2156 for it.

dug in a bit further than the last comment: the dump has the scanner sitting in the idle branch of its collector loop (`locker.go:429` is the `time.After`/`ctx.Done()` select), so it isn't stuck, it's sleeping. the test is the one blocked, and its only unguarded receive is `<-entered`.

`entered` is closed by whoever first observes tx-A as `Deleted`, gated behind a `sync.Once`. so the hang is consistent with that hook never firing for the scanner, leaving the test waiting on it forever. the `scan goroutine did not stop within timeout` line just before the panic fits too, that's `t.Cleanup` calling `Stop()` once the test is already wedged.

still couldn't reproduce it, clean on main and 40 runs under -race. so i've written it up as what the dump supports rather than a confirmed root cause, and i'd want to pin the interleaving before touching it. the property it tests is a real one, it guards the re-validation in the scanner's delete phase, so it's worth keeping rather than loosening.

PR #2116, opened 2026-08-03
fix(storage): allow digits in SQL table names and return errors instead of panicking

### LINKED ISSUE #2034: db: escapeForTableName panics on channel/network names containing digits
### Summary

`escapeForTableName` panics on any channel/network name that contains a digit, because its character allow-list never includes digits. This is reachable from normal store construction with no `recover()` anywhere in the call chain, so a channel name like `channel1` crashes the node.

### Where

`token/services/storage/db/sql/common/tablename.go:126-134` (`escapeForTableName`): the replacer/allow-list covers letters and a fixed set of punctuation but omits `0-9`. Any character outside the allow-list fails the final `validName` regex check, which panics rather than returning an error.

### Call chain (verified end-to-end)

`db.NewStoreServiceManager` / `auditdb/store.go:58-62` and friends pass `tmsID.Channel` / network params into `GetTableNamesWithOverrides(...)` → `tableNameFormatter.Format` → `escapeForTableName`. None of `token/services/storage/db/**` wraps this call chain in a `recover()`.

### Failure scenario

Any deployment/test topology using a channel name containing a digit (extremely common — e.g. `channel1`, `testchannel1`, `mychannel01`) panics at store-construction time instead of getting a clear configuration error.

### Test coverage gap

`tablename_test.go` has `TestEscapeTableNameError`, but it only exercises `!`/`#` as invalid characters — it never tests digits, and it never goes through the real call chain (`db.manager` → driver constructors), so the reachable panic path is untested.

### Suggested fix

Add `0-9` to the allow-list in `escapeForTableName` (most natural fix, since digits are unambiguous, safe SQL-identifier characters), or explicitly validate/return an error rather than panicking when an unsupported character is encountered. Add a regression test that exercises a digit-containing channel name through the real `NewStoreServiceManager`/driver-construction path, not just the escaping function in isolation.

### Severity

CRITICAL — reachable process crash from ordinary, common configuration values; discovered via a systematic bug-hunting review of `token/services/storage/db/sql/common`.

AkramBitar:
Fixes #2034

`escapeForTableName` omitted digits from its character allow-list **and** panicked on anything outside it, while `buildTableNames` formatted through `MustFormat` — so a channel name like `channel1` crashed the node at store-construction time, with no `recover()` anywhere in the chain.

- digits are now accepted, validated against the identifier actually emitted (prefix included);
- unsupported characters return a configuration error instead of panicking.

Regression tests cover the real driver-construction and `NewStoreServiceManager` paths, not just the escaping helper, plus a `FuzzGetTableNamesNoPanic` target wired into the nightly fuzz matrix.

### DIFF
--- a/token/services/storage/db/sql/common/init.go
+++ b/token/services/storage/db/sql/common/init.go
@@ -8,10 +8,19 @@ package common
 
 import (
 	"github.com/LFDT-Panurus/panurus/token/services/logging"
+	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/view/services/storage/driver/common"
 )
 
-const defaultPrefix = "fsc"
+const (
+	defaultPrefix = "fsc"
+	// sharedNameProbe is the short code used to validate the part of a table name
+	// that every short code shares: the prefix and the params. It is a real
+	// canonical short code rather than a synthetic one, so a failure reports a name
+	// the operator actually recognises, and it is never read from the overrides, so
+	// the check does not depend on them.
+	sharedNameProbe = "movements"
+)
 
 var (
 	logger     = logging.MustGetLogger()
@@ -92,7 +101,7 @@ func GetTableNamesWithOverrides(prefix string, overrides TableNamesConfig, param
 		return TableNames{}, err
 	}
 
-	return buildTableNames(prefix, params, overrides, nc.MustFormat)
+	return buildTableNames(prefix, params, overrides, nc.Format)
 }
 
 // GetTableNamesWithOverridesSkipPrefix is like GetTableNamesWithOverrides but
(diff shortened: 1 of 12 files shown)


token/services/storage/db/sql/common/init.go
```diff
@@ -103,48 +104,66 @@ func GetTableNamesWithOverridesSkipPrefix(prefix string, overrides TableNamesCon
...
+		Wallets:                name("wallets"),
+		IdentityConfigurations: name("id_cfgs"),
+		IdentityInfo:           name("id_info"),
+		Signers:                name("id_signers"),
+		KeyStore:               name("key_store"),
+		EIDLeases:              name("eid_leases"),
+		TokenSKICleanups:       name("tkn_ski_cleanups"),
+	}
+	if firstErr != nil {
+		return TableNames{}, firstErr
```
Effi-S: @AkramBitar , 

We are returning the first error but  letting all formats run . 

I see the comment "reporting all of them would just repeat the same message 17 times" 

I don't see a reason to assume that every time 

Maybe we can store them is a set (map with empty `struct{}`) 

```go
errSet := make(map[string]struct{})

...

if err != nil {
			msg := errors.Wrapf(err, "failed to build table name for short code [%s]", code).Error()
			errSet[msg] = struct{}{}
		}
... 

if len(errSet) > 0 {
		var errMsgs []string
		for msg := range errSet {
			errMsgs = append(errMsgs, msg)
		}
		return TableNames{}, errors.New(strings.Join(errMsgs, "; "))
	}
```

Effi-S (changes requested):

AkramBitar:
@Effi-S could you please have a second review and let me if you still have issues?

AkramBitar:
@Effi-S the collect-all-errors feedback is addressed: `buildTableNames` now runs every `name()` call, appending failures to `errs`, and joins them all with `errors.Join` before returning. The shared-part early-exit (`sharedNameProbe`) surfaces a bad prefix or params once instead of 17 times. CI is green. Ready for your re-review.

Effi-S (approved):

SaidAltury-ibm (approved):
LGTM

PR #2118, opened 2026-08-03
fix(recovery): guard claim fan-out send on context cancellation to unblock Stop

### LINKED ISSUE #2038: services/recovery: unguarded fan-out send can deadlock Stop() on shutdown race
### Summary

The recovery service manager's claim fan-out loop can deadlock `Stop()`. Unlike the analogous cleanup manager, the recovery fan-out send has no escape hatch on context cancellation, so it can block forever on shutdown, which in turn prevents `wg.Done()` from ever firing and hangs `Stop()`'s `wg.Wait()` — deadlocking all subsequent `Start()`/`Stop()` calls on the manager.

### Where

`token/services/storage/services/recovery/manager.go:255-260` — the fan-out loop does an unconditional `work <- claim` send with no `select` on `ctx.Done()`.

Compare with the cleanup manager's equivalent fan-out, `token/services/storage/services/cleanup/manager.go:256-264`, which correctly does:
```go
select {
case work <- claim:
case <-m.ctx.Done():
    return ...
}
```

### Failure scenario

1. `recoveryLoop`'s goroutine is mid-fan-out, with claims still queued to send to `work`.
2. `Stop()` cancels `m.ctx`.
3. All `WorkerCount` worker goroutines observe `ctx.Done()` in their `select` (manager.go:294) and exit before draining the remaining items from `work`.
4. The fan-out loop's next `work <- claim` send now has no receiver and blocks forever.
5. Because this send is inside the goroutine that `wg.Done()` is deferred on (line 145), that deferred call never runs.
6. `Stop()`'s `m.wg.Wait()` (line 136) — called while holding `m.mu` — hangs indefinitely, and every subsequent `Start()`/`Stop()` call on the manager deadlocks trying to acquire `m.mu`.

This is a real shutdown-path race, not merely theoretical: it requires `Stop()` to race a batch drain, which is exactly the condition shutdown logic needs to handle correctly.

### Suggested fix

Mirror the cleanup manager's guarded send in the recovery fan-out loop:
```go
select {
case work <- claim:
case <-m.ctx.Done():
    return ... // or break out of the fan-out loop cleanly
}
```

### Severity

HIGH — a real deadlock in the shutdown path that can wedge the manager and any component waiting on `Stop()`; discovered via a systematic bug-hunting review of `token/services/storage/services/recovery` and `services/cleanup`.

AkramBitar:
Fixes #2038

The recovery manager fanned claims out to its worker pool with an unguarded `work <- claim` send. On shutdown the workers return from their own `select` on `ctx.Done()` without draining `work`, so that send blocked forever: `close(work)` was never reached, the recovery loop's deferred `wg.Done()` never ran, and `Stop()`'s `wg.Wait()` hung while holding `m.mu` — wedging every later `Start()`/`Stop()`.

## Changes

- Extract the fan-out into `Manager.fanOut` and guard the send on `ctx.Done()`, mirroring `services/cleanup/manager.go`.
- Keep `close(work)` / `workerWG.Wait()` on the cancellation path so no worker goroutine leaks, and surface the cancellation as `errors.Join(fanOutErr, firstErr)` instead of swallowing it.
- Undispatched claims stay `Pending`, so the next sweep re-claims them once their lease expires — no work is lost.
- `docs/services/storage/recovery.md`: new "Shutdown Behaviour" section documenting the above.

## Test

`TestManager_StopDuringFanOutDoesNotDeadlock` reproduces the issue's numbered failure scenario step by step: 64 claims to a single worker parked inside `Recover` (fan-out blocked on the unbuffered send), `Stop()` from a goroutine, worker released only after cancellation so both its select arms are ready, then a guard on `Stop()` returning and a final `Start()`/`Stop()` pair proving the manager is not wedged.

Verified both directions: it fails against the unguarded send (`Stop() deadlocked while the sweep was fanning out claims`) and passes with the fix under `-count=3 -race`. `gofmt`, `go vet`, and `golangci-lint` are clean.

### DIFF
--- a/token/services/storage/services/recovery/manager.go
+++ b/token/services/storage/services/recovery/manager.go
@@ -162,7 +162,7 @@ func (m *Manager) recoveryLoop() {
 	}
 
 	if err := m.runSweep(m.ctx); err != nil {
-		m.logger.Warnf("initial transaction recovery sweep failed: %v", err)
+		m.logSweepError("initial transaction recovery sweep", err)
 	}
 
 	for {
@@ -173,10 +173,26 @@ func (m *Manager) recoveryLoop() {
 			return
 		case <-ticker.C:
 			if err := m.runSweep(m.ctx); err != nil {
-				m.logger.Warnf("transaction recovery sweep failed: %v", err)
+				m.logSweepError("transaction recovery sweep", err)
 			}
 		}
 	}
+}
+
+// logSweepError reports a sweep that returned an error.
+//
+// A sweep aborted because Stop cancelled the manager context is an ordinary
+// shutdown rather than a failure, so it is logged at debug level: the fan-out
+// surfaces the cancellation as an error to keep the undispatched claims visible
+// to the caller, and warning about it would flag a healthy node shutdown.
+func (m *Manager) logSweepError(what string, err error) {
+	if errors.Is(err, context.Canceled) {
+		m.logger.Debugf("%s stopped: %v", what, err)
+
+		return
+	}
+
+	m.logger.Warnf("%s failed: %v", what, err)
 }
 
 func (m *Manager) validateConfig() error {
@@ -248,16 +264,7 @@ func (m *Manager) recoverTransactions(ctx context.Context) error {
 		go m.worker(ctx, &workerWG, work, errCh)
 	}
 
-	// ClaimPendingTransactions reads directly from the requests table where
-	// tx_id is the primary key, so each claim is already unique. Fan out
-	// straight to the workers; nil entries are defensive but should never
-	// occur in practice.
-	for _, claim := range records {
-		if claim == nil {
-			continue
-		}
-		work <- claim
-	}
+	dispatched, fanOutErr := m.fanOut(ctx, records, work)
 	close(work)
 
 	workerWG.Wait()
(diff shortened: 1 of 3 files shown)


token/services/storage/services/recovery/manager.go
```diff
@@ -283,9 +274,43 @@ func (m *Manager) recoverTransactions(ctx context.Context) error {
 		m.logger.Debugf("completed recovery sweep: claimed=%d, all succeeded", len(records))
```
atharrva01, line 295: I think the sweep summary may misreport a cancelled fan-out now, though I could be misreading the counting here.

`failures` only counts entries that arrive on `errCh`, and undispatched claims never reach a worker, so as far as I can tell they contribute nothing. In the 64-claim setup your test builds, if the one dispatched claim succeeds then `failures` is 0 and this takes the `else` branch, logging `claimed=64, all succeeded` when most were never touched. The `failures > 0` branch above looks like it has the same thing from the other side, since `len(records)-failures` would count undispatched claims as successes.

If that reading is right it would be new, because previously the fan-out either completed or hung. Returning the dispatched count from `fanOut` alongside the error would cover both branches. It is only logging though, so entirely your call whether it is worth doing here or at all.

token/services/storage/services/recovery/manager.go
```diff
@@ -283,9 +274,43 @@ func (m *Manager) recoverTransactions(ctx context.Context) error {
...
+// is the primary key, so each claim is already unique. Fan out straight to the
+// workers; nil entries are defensive but should never occur in practice.
+//
+// The send is guarded on ctx.Done() — mirroring the cleanup manager — because
+// the workers return from their own select as soon as the context is cancelled,
+// without draining what is left in work. An unguarded send would then block
+// forever with no receiver, so close(work) and the deferred wg.Done() of the
+// recovery loop would never run and Stop's wg.Wait() would hang while holding
+// m.mu, wedging every later Start/Stop call.
+func (m *Manager) fanOut(ctx context.Context, records []*ttxdb.RecoveryClaim, work chan<- *ttxdb.RecoveryClaim) error {
```
atharrva01: Small note on the doc comment, and I may be missing the intent.

Reading `cleanup/manager.go`, `cleanupTokens` returns on `<-m.ctx.Done()` before it reaches `close(work)` and `workerWG.Wait()` (lines 259-265), so it skips both. It also selects on `m.ctx` in the fan-out while passing `ctx` down to the workers. Same value today so the workers still exit, but they would diverge if a caller ever passed a different ctx.

So this version looks stricter than the one it points at. Might be worth not sending readers to cleanup as the model, or picking that side up separately since it looks like the same class as #2038. Happy to open an issue for it if that would help.

token/services/storage/services/recovery/manager.go
```diff
@@ -283,9 +274,43 @@ func (m *Manager) recoverTransactions(ctx context.Context) error {
...
+		if claim == nil {
+			continue
+		}
+		select {
+		case work <- claim:
+			// dispatched
+		case <-ctx.Done():
+			m.logger.Debugf("recovery fan-out cancelled: %d of %d claim(s) not dispatched", len(records)-i, len(records))
+
+			return errors.Wrapf(ctx.Err(), "recovery fan-out cancelled")
```
atharrva01: Very minor, and possibly deliberate: this makes an ordinary shutdown log at warn level. `runSweep` passes the error up to `recoveryLoop`, which hits `m.logger.Warnf("transaction recovery sweep failed: %v", err)`, so stopping a node mid-sweep would print `transaction recovery sweep failed: recovery fan-out cancelled: context canceled` on a path where nothing actually went wrong.

Certainly not worth reshaping the return for. The loop could skip the warn when the error is `ctx.Err()`, if you think it is worth the noise reduction.

atharrva01 (commented):
hi @AkramBitar , took a careful read since I have been working in this file recently. The deadlock traces through exactly as you describe: `worker` returns from its own select on `ctx.Done()` without draining `work`, `work` is unbuffered and `fanOut` is the only sender, so the send parks with no receiver and `close(work)` is never reached. Guarding the send looks right to me, and keeping `close(work)` and `workerWG.Wait()` on the cancellation path seems like the important detail.

Ran the test at `-count=3 -race` and it passes. I liked that it drives the failure through the real Start/Stop path rather than reaching for the unexported fan-out.

Three observations below, none of them blocking, and some possibly things you already weighed up.

Effi-S (approved):
LGTM

AkramBitar:
@atharrva01 

Thanks a lot for the review. I fixed all the comments. Hope now it is ok. 

Regards,
Akram

PR #2119, opened 2026-08-03
fix(network-fabric): propagate ledger info failures instead of rescanning from genesis

### LINKED ISSUE #2058: network-driver: finality ScanBlock silently rescans from genesis on any GetLedgerInfo error
### Summary

`Delivery.ScanBlock` swallows any error from `GetLedgerInfo` (logging it, but not returning it) and silently falls back to scanning from block 0 (genesis), rather than propagating the error or preserving the last known-good position.

### Where

`token/services/network/fabric/finality/delivery.go:22-34`:

```go
func (d *Delivery) ScanBlock(background context.Context, callback fabric.BlockCallback) error {
    startingBlock := uint64(0)
    if d.Ledger != nil {
        info, err := d.GetLedgerInfo()
        if err == nil {
            startingBlock = info.Height
        } else {
            d.Logger.ErrorfContext(background, "failed to get ledger info: %s", err)   // logged, not returned
        }
    }
    return d.ScanBlockFrom(background, startingBlock, callback)
}
```

### Impact

A transient failure to fetch ledger info (rather than a genuine "this is a fresh chain" condition) causes a full rescan of the entire chain from block 0 instead of resuming near the current height or failing loudly. On a chain with significant history, this silently turns a transient RPC hiccup into a very expensive full-chain rescan every time it recurs, with no error surfaced to the caller and no way to distinguish "genuinely fresh chain" from "ledger info temporarily unavailable" from the returned error alone (there is none).

### Reproduction

Reproduced locally with a unit test (not yet committed), built via unexported-field injection (mirroring FSC's own `TestNewDeliveryFLM` test) around a fake `driver.Ledger` whose `GetLedgerInfo` returns a fixed transient error, and a fake `driver.Delivery` recording the block number it was asked to scan from:

```go
ledgerDriver := &fakeLedgerDriver{err: errors.New("peer connection reset")}
deliveryDriver := &recordingDeliveryDriver{}
d := &finality.Delivery{Delivery: newTestDelivery(deliveryDriver), Ledger: newTestLedger(ledgerDriver), ...}

err := d.ScanBlock(context.Background(), callback)
require.NoError(t, err)   // error is swallowed, not propagated
require.Equal(t, uint64(0), deliveryDriver.scanBlockFromCalledWith,
    "ScanBlock falls back to block 0 on any GetLedgerInfo error, forcing a full rescan from genesis")
```

### Suggested fix

Propagate the `GetLedgerInfo` error from `ScanBlock` instead of swallowing it after a log line, so callers can distinguish a transient failure from a genuinely fresh chain and decide whether to retry rather than triggering an implicit full rescan from genesis.

### Severity

MEDIUM — resource/correctness risk (unbounded rescan cost, delayed/duplicate finality notifications) triggered by an ordinary transient RPC failure.

AkramBitar:
Fixes #2058

## The issue

`ScanBlock` asks the peer for the current ledger height to know where to resume the block scan. If that call failed, the error was only logged and the height stayed 0, so the scan silently restarted from genesis — rescanning the whole chain and replaying finality notifications for every past transaction, with no error for the caller.

## The fix

Retry the height lookup a few times, and if it still fails return the error instead of starting from block 0. A passing RPC hiccup is absorbed by the retry; a real outage is reported.

## Before

```
ScanBlock(callback):
    start = 0
    if ledger exists:
        info, err = GetLedgerInfo()
        if err == nil: start = info.Height
        else:          log(err)          # error dropped, start stays 0
    return ScanBlockFrom(start, callback)
```

One failed RPC → scan restarts from genesis, caller sees no error.

## After

```
ScanBlock(callback):
    start = 0
    if ledger exists:
        start, err = LedgerHeight()
        if err != nil: return err        # no scan at all
    return ScanBlockFrom(start, callback)

LedgerHeight():
    delay = LedgerInfoRetryDelay        # default 500ms
    for attempt = 1 .. LedgerInfoAttempts:   # default 7
        info, err = GetLedgerInfo()
        if err == nil and info != nil:
            return info.Height           # done
        log(err)
        if attempt == last: break
        wait delay or return if context cancelled
        delay = delay * 2
    return error "refusing to rescan from genesis"
```

## What an attempt does

One attempt is one `GetLedgerInfo` RPC to the peer:

1. If it returns a height, use it and start the scan there. Height is the next block to be delivered, so nothing is replayed.
2. If it returns an error — or `(nil, nil)`, which used to nil-deref — remember the error and log it as `attempt N/M`.
3. On the last attempt, stop looping. Otherwise wait `delay`, then double it. A cancelled context ends the wait immediately and returns.

Defaults: 7 attempts, 0.5s + 1s + 2s + 4s + 8s + 16s of waiting, ~31.5s worst case. Both bounds are configurable — see below.

## If every attempt fails

- The error is returned, wrapping the last ledger error so `errors.Is` still reaches it: `failed to get ledger info after 7 attempt(s), refusing to rescan from genesis`.
- `ScanBlockFrom` is never called — no scan, so no genesis rescan and no duplicate finality notifications.
- The caller, FSC's `ListenerManager.start()`, logs the error, so block-delivery finality does not start for that manager. Finality is degraded, not lost: timed-out listeners are queried directly against the ledger (`QueryByID` → `fetchTxs`), and the recovery service re-registers listeners for pending transactions.
- Not a regression: if the peer cannot answer `GetLedgerInfo` across ~31.5s of retries, the `ScanBlockFrom` stream to that same peer would have failed too. The old code just failed later, after starting a full-chain rescan.

Block 0 is still used in one case only: no ledger configured at all.

## Notes

- Retry, not just propagate: `ListenerManager.start()` calls `ScanBlock` once and only logs the error, so returning on the first hiccup would leave block-based finality dead until the process restarts. That is also why the budget has to outlast a peer restart rather than a dropped packet.
- `*fabric.Delivery` / `*fabric.Ledger` replaced by two small interfaces so the code is unit-testable. Field names unchanged, wiring untouched.

## Tests

`delivery_test.go`, 11 cases: height resumption, error propagated with no scan started, transient failure retried, attempt budget honoured, nil info, cancelled context, nil ledger → block 0, scan error propagated, plus the cancellation/nil-info sentinel overlap, last-failure-only classification, and the default attempt budget.

`config/config_test.go`, 4 cases (the package had none): defaults, configured values, non-positive rejection, and the budget appearing in `String()`.

Against the old code 4 fail and one panics. All pass with the fix under `-count=2 -race`; `gofmt`, `go vet`, `golangci-lint`, `go build ./...` clean.

## Configurable retry budget

The 7-attempt default is a judgement call, not a measurement: it is the smallest count on the existing doubling schedule that clears ~30s, chosen as a rough proxy for how long a peer restart takes. How long a peer actually takes to come back is a deployment property, so both bounds are exposed alongside the five existing `token.finality.delivery.*` settings rather than baked in:

```yaml
token:
  finality:
    delivery:
      ledgerInfoAttempts: 7        # attempts at reading the starting height
      ledgerInfoRetryDelay: 500ms  # first retry pause; doubles each attempt
```

Non-positive values fall back to the defaults: zero attempts would refuse every scan, and a non-positive delay would busy loop.

`lookup` is wired too, not just `finality` — it constructs the same `Delivery`, so leaving it out would have silently kept the old budget for lookup listeners. The exported provider constructors (`NewDeliveryBasedFLMProvider`, `NewDeliveryBasedLLMProvider`) take the values through a variadic option, so their existing signatures stay source-compatible for downstream callers.

## Error classification

Failures are classifiable with `errors.Is`, so a caller never has to match on message text:

| Sentinel | Meaning |
| --- | --- |
| `ErrLedgerHeightUnavailable` | the starting block could not be resolved, so no scan started |
| `ErrNoLedgerInfo` | the **last** attempt saw the ledger return neither info nor an error — a driver contract violation |
| `context.Canceled` / `context.DeadlineExceeded` | a wait between attempts was cut short, or the scan itself was cancelled |

`ErrLedgerHeightUnavailable` accompanies *every* failure to resolve the starting block, including a cancelled one, which makes it the single test for "no scan started" — and makes "an error without it came from the scan" true.

The two refining sentinels are deliberately documented as *not* discriminators on their own:

- `ErrNoLedgerInfo` reports only the last observed failure, so a driver that violates the contract intermittently may surface an ordinary ledger error instead. Its absence is not proof the contract was kept.
- The same context governs the scan, so a context error alone does not say which phase ended. Pair it with `ErrLedgerHeightUnavailable` to tell a cancelled height read from a cancelled scan.

No in-tree caller inspects these sentinels yet — today the sole consumer logs the error and stops. Retry or restart around `ScanBlock` belongs to FSC's `events.ListenerManager` and is out of scope here: the wider budget shrinks that window but does not close it.

### DIFF
--- a/docs/services/network-fabric.md
+++ b/docs/services/network-fabric.md
@@ -294,6 +294,58 @@ sequenceDiagram
 - LRU cache for recent transactions
 - Automatic retry on connection failures
 
+#### Choosing the Starting Block
+
+The scan resumes at the peer's current ledger height, read via `GetLedgerInfo`
+([`delivery.go`](../../token/services/network/fabric/finality/delivery.go)). Because that
+one RPC decides the starting block, a failure is retried with an exponentially growing
+delay — 7 attempts spanning ~31.5s by default, aborting early if the context is cancelled.
+Both are configurable via `token.finality.delivery.ledgerInfoAttempts` and
+`ledgerInfoRetryDelay` (see [Finality Configuration](#finality-configuration)).
+
+That budget is deliberately long. Nothing retries `ScanBlock`: FSC's
+`events.ListenerManager` calls it once from a goroutine that only logs the result, so an
+error escaping the retry loop leaves the channel with **no block-based finality until the
+process restarts**. The retries therefore have to outlast a peer restart, not merely a
+dropped packet.
+
+If the height is still unavailable after the last attempt, `ScanBlock` returns the error
+rather than defaulting to block 0. Starting at genesis would rescan the entire chain and
+replay finality notifications for every historical transaction, while the caller would have
+no way to tell a transient RPC failure from a genuinely fresh chain. Block 0 is used only
+when no ledger is configured at all.
+
+The doubling has a ceiling of 30s, so raising `ledgerInfoAttempts` lengthens the budget
+without letting a single pause grow without bound. A `ledgerInfoRetryDelay` larger than the
+ceiling is honoured as configured — the cap limits growth, it does not shorten the delay you
+asked for. The backoff itself is `utils.RetryRunner`, shared with the rest of the SDK rather
+than reimplemented here.
+
+The returned error is classifiable with `errors.Is`, so a caller does not have to match on
+its message. `ErrLedgerHeightUnavailable` is the single test for "the starting block could
+not be resolved, so no scan started" — it accompanies every such failure, including a
+cancelled one:
+
+| Sentinel | Meaning |
+| --- | --- |
+| `finality.ErrLedgerHeightUnavailable` | the height could not be read, so no scan started |
+| `finality.ErrNoLedgerInfo` | **some** attempt saw the ledger return neither info nor an error — a driver contract violation |
+| `context.Canceled` / `context.DeadlineExceeded` | a wait between attempts was cut short, or the scan itself was cancelled |
+
+Every attempt's failure is reported, so `ErrNoLedgerInfo` is present whenever the contract
+was violated at least once, even intermittently. The context error is not a discriminator on
+its own: the same context governs the scan, so it does not say which phase ended — pair it
+with `ErrLedgerHeightUnavailable` to tell a cancelled height read from a cancelled scan.
+
+The underlying ledger errors are preserved in every case. An error that does not match
+`ErrLedgerHeightUnavailable` comes from the block scan itself rather than from resolving the
+starting block.
+
+There is no in-tree caller that inspects these sentinels yet — today the sole consumer logs
+the error and stops. Classifying the failure is what a future caller needs to react
+(fall back to query-based finality, retry, or restart the manager) rather than a description
+of current behaviour.
+
 ### Notification Mode
 
 Uses asynchronous event notifications from the FSC layer:
@@ -434,7 +486,14 @@ token:
       blockProcessParallelism: 10  # Parallel block processors
       lruSize: 30                  # Cache size for recent transactions
       listenerTimeout: 10s         # Timeout for listener notifications
+      ledgerInfoAttempts: 7        # Attempts at reading the starting ledger height
+      ledgerInfoRetryDelay: 500ms  # First retry pause; doubles each attempt (~31.5s total)
 ```
+
+`ledgerInfoAttempts` and `ledgerInfoRetryDelay` bound the ledger-height read that decides
+where a block scan starts — see [Choosing the Starting Block](#choosing-the-starting-block).
+Non-positive values are ignored in favour of the defaults: zero attempts would refuse every
+scan, and a non-positive delay would busy loop.
 
 ### Endorsement Configuration
 
--- a/token/services/network/fabric/finality/delivery.go
+++ b/token/services/network/fabric/finality/delivery.go
@@ -8,27 +8,201 @@ package finality
 
 import (
 	"context"
+	"time"
 
 	"github.com/LFDT-Panurus/panurus/token/services/logging"
+	"github.com/LFDT-Panurus/panurus/token/services/utils"
+	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"github.com/hyperledger-labs/fabric-smart-client/platform/fabric"
 )
 
-type Delivery struct {
-	*fabric.Delivery
-	*fabric.Ledger
-	Logger logging.Logger
+const (
+	// defaultLedgerInfoAttempts is how many times ScanBlock asks the ledger for
+	// its current height before giving up. A transient RPC hiccup must not
+	// decide the starting block, so the call is retried rather than trusted once.
+	//
+	// The budget is sized to outlast a peer restart, not just a dropped packet:
+	// nothing retries ScanBlock (see the ledgerHeight comment), so a failure that
+	// escapes here disables block-based finality until the process restarts.
+	defaultLedgerInfoAttempts = 7
+	// defaultLedgerInfoRetryDelay is the pause before the first retry. It doubles
+	// on each further attempt, so the defaults wait ~31.5s in total (0.5s + 1s +
+	// 2s + 4s + 8s + 16s over six retries).
+	defaultLedgerInfoRetryDelay = 500 * time.Millisecond
+	// defaultLedgerInfoMaxRetryDelay caps the doubling. Without a ceiling a large
+	// configured attempt budget grows the pause without bound — past ~35 doublings
+	// of the default delay a single wait exceeds a human lifetime, and the int64
+	// nanosecond duration eventually wraps negative — so the scan would never
+	// resume and never fail either. The cap only limits growth: a configured delay
+	// larger than this is honoured as-is (see ledgerHeight).
+	defaultLedgerInfoMaxRetryDelay = 30 * time.Second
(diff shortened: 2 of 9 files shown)


token/services/network/fabric/finality/delivery.go
```diff
@@ -8,27 +8,113 @@ package finality
...
+	var lastErr error
+	for attempt := 1; attempt <= attempts; attempt++ {
+		info, err := d.Ledger.GetLedgerInfo()
+		switch {
+		case err != nil:
+			lastErr = err
+		case info == nil:
+			// A driver that reports neither info nor error would otherwise nil
+			// deref below; treat it as a failure to read the height.
+			lastErr = errors.New("ledger returned no info")
```
adecaro: by introducing types errors, the caller can tell apart what happened without having to check the content of the error message.

AkramBitar: Good catch, thanks a lot. Fixed.

AkramBitar:
@adecaro 
Thanks a lot for the review. 

I flagging one change: the retry budget is now configurable instead of hardcoded.

Nothing retries ScanBlock — FSC's ListenerManager.start() calls it once and only logs the error — so an unreadable height costs the channel its block-based finality until restart. The old 3 attempts / ~1.5s didn't survive a peer restart, trading the genesis rescan for a silently dead listener.

Default is now 7 / ~31.5s, but that's a guess, not a measurement — hence:

token.finality.delivery.ledgerInfoAttempts: 7
token.finality.delivery.ledgerInfoRetryDelay: 500ms

Do you have a better sense of how long a peer usually takes to come back? 
Please let me know your thoughts on that change? 

Thanks a million,
Akram

docs/services/network-fabric.md
```diff
@@ -294,6 +294,58 @@ sequenceDiagram
...
+
+If the height is still unavailable after the last attempt, `ScanBlock` returns the error
+rather than defaulting to block 0. Starting at genesis would rescan the entire chain and
+replay finality notifications for every historical transaction, while the caller would have
+no way to tell a transient RPC failure from a genuinely fresh chain. Block 0 is used only
+when no ledger is configured at all.
+
+The budget is tunable through the `LedgerInfoAttempts` and `LedgerInfoRetryDelay` fields of
+`finality.Delivery`. These are Go fields with no YAML counterpart — neither construction site
+sets them, so deployments always run the defaults and only tests vary them.
```
SaidAltury-ibm: The new docs section first says ledgerInfoAttempts/ledgerInfoRetryDelay are configurable via YAML, then later claims those are dead Go fields with no wiring. That second claim is false — both finality/manager.go and lookup/manager.go do wire them through WithLedgerInfoRetry. This will mislead operators or a future maintainer into thinking there's a wiring gap that doesn't exist.

token/services/network/fabric/finality/delivery.go
```diff
@@ -8,27 +8,168 @@ package finality
...
+		if attempt == attempts {
+			break
+		}
+
+		select {
+		case <-ctx.Done():
+			return 0, errors.Wrap(errors.Join(ErrLedgerHeightUnavailable, ctx.Err(), lastErr), "cancelled while getting ledger info")
+		case <-time.After(delay):
 		}
+		delay *= 2
```
SaidAltury-ibm: no cap on delay?

token/services/network/fabric/finality/delivery.go
```diff
@@ -8,27 +8,168 @@ package finality
...
+// result, so an error escaping here leaves the channel without block-based
+// finality until the process restarts. Absorbing a peer restart here is therefore
+// worth more than failing fast.
+//
+// Failures are joined with a sentinel rather than described only in the message,
+// so callers classify them with errors.Is: ErrLedgerHeightUnavailable on both
+// exits, additionally ErrNoLedgerInfo for a driver returning (nil, nil) and the
+// context error for a cancelled wait. The underlying ledger error is preserved in
+// every case.
+func (d *Delivery) ledgerHeight(ctx context.Context) (uint64, error) {
```
SaidAltury-ibm, line 155: Duplicates existing RetryRunner utility, losing its safety cap. The hand-rolled retry/backoff loop reimplements token/services/utils.RetryRunner, whose maxDelay cap would have prevented the overflow bug in finding #1 for free. Now there are two retry implementations with diverging safety properties.

SaidAltury-ibm (changes requested):

token/services/network/fabric/finality/delivery.go
AkramBitar: Capped now — the backoff is `utils.RetryRunner`, ceiling 30s (raised to the configured delay when that is larger).

token/services/network/fabric/finality/delivery.go
AkramBitar: Agreed — swapped the hand-rolled loop for `utils.RetryRunner`, which brings its `maxDelay` cap with it.

docs/services/network-fabric.md
AkramBitar: Correct, that paragraph was stale — removed; both manager sites do wire the two knobs.

SaidAltury-ibm (approved):
there's some failing test

SaidAltury-ibm (approved):
LGTM

PR #2120, opened 2026-08-03
fix(auditdb): require matching anchor when refreshing an eid lease

### LINKED ISSUE #2033: auditdb: postgres locker's acquire upsert lets a node steal its own lock across unrelated anchors

AkramBitar:
Fixes #2033

## The problem

The lease row for an enrollment ID could be overwritten whenever `owner = excluded.owner`. `owner` is the replica ID — the same value for every audit on a node. So two concurrent audits on one node, for different transactions sharing an enrollment ID, both "acquired" it: the second overwrote the first's live row and `AcquireLocks` returned success to both. The mistake only surfaced later, at `AssertLocksHeld` just before the DB write.

## The fix

Also require the anchor to match, so a lease is only refreshed by the same owner *and* the same transaction:

```sql
-- before
WHERE expires_at < NOW() OR owner = excluded.owner

-- after
WHERE expires_at < NOW() OR (owner = excluded.owner AND anchor = excluded.anchor)
```

Same owner, different anchor is now ordinary contention, retried until `acquireDeadline`. Expired leases stay claimable, so crash recovery is unchanged.

## Tests

One query-shape test (no DB) and four against real Postgres: cross-anchor contention, same-anchor refresh, expired takeover, and four concurrent audits on one shared enrollment ID.

Against the old clause three fail — the concurrent one reporting all 4 audits as winners of the same lease. All pass with the fix under `-race`; `go vet` and `golangci-lint` clean.

### DIFF
--- a/docs/services/auditor.md
+++ b/docs/services/auditor.md
@@ -94,6 +94,8 @@ token:
 
 The Postgres backend creates an `eid_leases` table (prefixed per TMS persistence settings) and uses lease rows with heartbeat renewal.
 
+**Lease ownership:** each row is keyed by EID and carries the holding replica (`owner`) plus the request it was taken for (`anchor`). An acquisition may only take over an existing row when the lease has expired, or when the row is the same replica's lease for the *same* anchor — a re-acquisition, which just refreshes the deadline. Two different anchors therefore never hold the same EID at once, including two concurrent audits on a single replica; the second one is contended and retried until `acquireDeadline`.
+
 ### Replica owner identity
 
 Every lease row carries an `owner` column, and each replica scopes all of its lease
--- a/token/services/storage/auditdb/locker/postgres/postgres.go
+++ b/token/services/storage/auditdb/locker/postgres/postgres.go
@@ -198,10 +198,17 @@ func (p *Locker) tryAcquireAll(ctx context.Context, anchor string, eIDs []string
 // buildAcquireQuery builds the atomic acquisition statement: an INSERT of one
 // row per enrollment ID that, ON CONFLICT on the eid primary key, overwrites
 // the existing row only when it is safe to steal — the current lease has
-// expired (InPast) or is already owned by this replica. The WHERE clause on the
+// expired (InPast), or the row is this replica's own lease for this very anchor
+// (a re-acquisition, whose lease is simply refreshed). The WHERE clause on the
 // upsert enforces that condition, and RETURNING eid yields exactly the IDs that
 // were claimed, which tryAcquireAll counts. expires_at is set to now()+TTL via
 // an interval-bound parameter.
+//
+// The anchor must be compared as well as the owner: owner is a per-replica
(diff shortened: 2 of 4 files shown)


Effi-S:
LGTM. @AkramBitar. 
Notice you have a minor conflict

Effi-S (approved):

PR #2135, opened 2026-08-04
Github actions from forks

Effi-S:
1. Updated The report comment to only be on regression 
2. Made Permission Changes so PRs from forks won't always fail


Effi-S:
This fixes the problem 
image

Effi-S:
Notice: 

## These Fail because they are run from the base repo :

image

## And these are run from the Fork:

image

AkramBitar (approved):

PR #2137, opened 2026-08-04
fix(identity): make RecipientDataCache provisioning cancellable

### LINKED ISSUE #2080: identity/role: RecipientDataCache busy-loops with no backoff/logging, and an idemix identity cache is never closed
### Summary

`RecipientDataCache`'s background provisioning loop retries on every error with no backoff, no
logging, and no way to stop it, and a separate idemix identity cache with similar shape is started
but its `Close()` is never called because the cache reference is dropped after wiring up a single
method value. Together these mean a persistently failing backend can spin a goroutine at full CPU
indefinitely and silently, and identity-cache goroutines can accumulate with no way to tear them
down.

### Where

`token/services/identity/role/cache.go:89-99` `provisionIdentities` runs
`for { id, err := c.backed(ctx); if err != nil { continue }; ... }` against a `context.Background()`
with no cancellation hook. If the backend call keeps failing, this loop spins continuously with no
log line, no backoff, and no counter to signal the condition externally. It's started per
anonymous-owner-wallet via `sync.Once` (`cache.go:51-56`), and `Registry.Wallets` (where these
wallets accumulate) has no upper bound, so the number of such loops can grow with usage.
`Service.Done()` → `Registry.Done()` → `Role.Done()` doesn't reach this loop, so there's no shutdown
path for it either.

By contrast, `idemix/cache/cache.go:143-151`'s equivalent loop does log, back off for a second, and
check the context — but its `Close()` (`cache.go:58`) has no callers, because
`idemix/kmp.go:131-136` takes the cache's `Identity` method as a bare method value and discards the
cache object itself. Also worth a look while in this code: `cache.go:96` increments
`CacheLevelGauge` before the blocking send that would decrement it on the other side, so the gauge
can drift upward over time if sends are slow or blocked.

### Impact

Under a backend failure, the `RecipientDataCache` loop consumes a full core continuously with no
visible signal (no log, no metric) that anything is wrong — this is hard to diagnose in production
since there's nothing to alert on. The lack of a `Close()` call on the idemix side means that cache's
background goroutine outlives any expected owner and can't be cleanly stopped, which matters for
long-running processes that create and discard key managers over time.

### Suggested fix

Add the same treatment `idemix/cache/cache.go` already has — a short backoff and a log line on
error — to `RecipientDataCache.provisionIdentities`, and thread a cancellable context through so it
can be stopped from `Done()`. For the idemix side, keep a reference to the `IdentityCache` alongside
the method value in `kmp.go` and call `Close()` when the owning key manager is torn down. The gauge
increment/decrement ordering is a smaller follow-up worth fixing alongside this.

### Severity

MEDIUM-HIGH — primarily a local resource-usage and observability issue (busy CPU loop, leaked
goroutines) rather than a remote attack surface, but straightforward to fix and worth closing.

### LINKED ISSUE #2122: Fix Uncancellable Provisioning Goroutine in `RecipientDataCache` [mem]
**Document severity:** HIGH · **Verified priority:** High · **Effort:** Low · **Type:** Bug

### Old issue reference

> **Sub-Task 1 — Fix Uncancellable Provisioning Goroutine in `RecipientDataCache`**
>
> **Status:** `[ ] pending`

| 1 | `[RUN]` | **HIGH** | Goroutine leak | `identity/role/cache.go` | 54, 89–99 | Infinite provisioning goroutine, no cancellation path |

### Intent

`RecipientDataCache.provisionIdentities()` is started via `sync.Once` as a
bare `go` goroutine that runs an infinite `for` loop using a hardcoded
`context.Background()`. There is no termination mechanism: the goroutine
never exits, it cannot be stopped, and it blocks indefinitely on channel
send whenever the cache channel fills up.

### Expected Outcomes

- The cache has a cancellable background context and a `Stop()` / `Close()` method.
- The goroutine exits cleanly when the owning object is no longer needed.
- The goroutine does not block on channel send after the owner is stopped.

### Todo List

1. Add a `cancel context.CancelFunc` field to `RecipientDataCache`.
2. In `NewRecipientDataCache`, create a `context.WithCancel(context.Background())` and store the cancel.
3. Pass the cancellable context to `provisionIdentities(ctx context.Context)`.
4. Inside the goroutine's `for` loop, add a `select` on `ctx.Done()` before the blocking channel send.
5. Add a `Close()` method that calls `c.cancel()`.

### Relevant Context

- File: `token/services/identity/role/cache.go`
- Lines 35–99: constructor and `provisionIdentities` goroutine.
- `idemix/cache/cache.go` already implements this pattern correctly (it has a `cancel` field and `Close()`) — mirror that approach.

---

### Status in current code

**Confirmed — still present.**

`token/services/identity/role/cache.go:54` still starts `go c.provisionIdentities()`
under the `sync.Once` at `:27`, and `:89-90` still opens with a hardcoded
`ctx := context.Background()`. The struct has no `cancel` field and the type exposes no
`Stop()` or `Close()` method, so the goroutine has no termination path at all.

Two distinct costs, not one:

1. **Unbounded goroutine retention.** One goroutine per `RecipientDataCache` that lives
   for the process lifetime, pinning the cache, its channel and the backend closure.
2. **Unlogged CPU spin.** The error path at `:93-95` is a bare `continue`:

   ```go
   id, err := c.backed(ctx)
   if err != nil {
       continue
   }
   ```

   No backoff, no logging, no error counter. A persistently failing identity backend turns
   this into a tight busy loop that saturates a core with nothing in the logs to explain it.
   This is arguably worse operationally than the retained goroutine, and it is not called out
   in the source document.

Sole construction site is `token/services/identity/role/wallets.go:533`, so adding a
`Close()` has a small blast radius. The sibling `token/services/identity/idemix/cache/cache.go`
already implements exactly the cancellable pattern the document asks for, so the fix has a
working template in-tree.

This is the most serious of the 13 and a good place to start.

### Severity and priority

`origin/main` with `go 1.26.5` on 2026-08-04, the priority is **High** for the

### Effort

**Low**

AkramBitar:
Both issues are about the same background worker, so they are fixed together in one PR.

**Background.** An anonymous wallet needs a fresh pseudonym for every payment. Making one is slow, so each wallet keeps a small stock of them ready. A background worker refills that stock. There is one worker per wallet. Idemix identities are cached the same way, by a second worker of the same shape.

**#2122 — the worker cannot be stopped.**
Once started, the worker runs until the process exits. There is no way to shut it down. When the stock is full, the worker waits to hand over the next pseudonym, and that wait never ends. So the worker stays alive forever, and it keeps its wallet, its stock, and everything they point to alive with it. Because wallets are created as they are used and never removed, the number of stuck workers grows with usage.
This PR gives the worker a proper shutdown path, so it can be told to stop and it actually exits.

**#2080 — the same worker, plus three more problems.**
1. *A broken backend is invisible.* If pseudonym creation keeps failing, the worker retries immediately, forever. It writes nothing to the log and reports nothing to monitoring. The result is one CPU core at full load and no clue why. This PR makes it log the failure, count it on a new metric, and wait a second between attempts.
2. *A metric slowly becomes wrong.* The counter that reports how full the stock is counts a pseudonym before it is really added. Over time it reports more than the stock actually holds. This PR counts it only once it is really added.
3. *The second cache has the same problem, and its cleanup is unreachable.* The idemix cache already has a shutdown method, but the object holding it is thrown away as soon as it is created, so nothing can ever call that method. This PR keeps the object reachable and calls it during shutdown.

**A defect this PR had to fix in itself.** The idemix cache prepared its cancellation while serving the first request, and read it back while shutting down, with no synchronisation between the two. That was harmless while nothing could call its shutdown method. Making it reachable turned it into a real race, with a worse effect than the race itself: a shutdown that arrived at the wrong moment silently did nothing and left the worker running — the very thing this PR sets out to fix. It is now prepared up front, once, so the two can no longer overlap.

**New metrics.** `recipient_data_provision_failures_total` and `cache_provision_failures_total`, so a failing identity backend can be alerted on instead of only appearing in the logs. Both declare `network`/`channel`/`namespace`, and were checked against a real Prometheus provider behind the TMS wrapper, because a missing label there panics on first use.

**What the two issues share.** Only the first point: giving the worker a way to shut down. Everything else is unique to #2080.

**What is still not solved.** Wallets are never removed once created. So a worker is only released when the whole token service is unloaded, which happens on a public-parameter update. During normal operation nothing releases them. Fixing that means deciding when a wallet is no longer needed, which is a separate change.

**Testing.** 16 new tests. 8 of them were checked against the old code first and do fail there, so they really catch these bugs. Existing tests that used to leave workers running now release them. `make checks`, `make lint`, `./token/...`, `-race`, and repeated shuffled runs are all clean. Docs updated in `docs/services/identity.md`, including a table of the four cache metrics.

Fixes #2122
Fixes #2080

### DIFF
--- a/token/services/identity/idemix/cache/cache.go
+++ b/token/services/identity/idemix/cache/cache.go
@@ -36,29 +36,37 @@ type IdentityCache struct {
 	cacheTimeout time.Duration
 	// Cache performance metrics
 	metrics *Metrics
-	// Cancellation function for background provisioning
+	// provisionCtx governs the lifetime of the background provisioning goroutine.
+	// Created at construction time and cancelled by Close.
+	provisionCtx context.Context
+	// cancel cancels provisionCtx. Written once, at construction time, so that Close
+	// is safe to call concurrently with Identity.
 	cancel context.CancelFunc
 }
 
 // NewIdentityCache creates a new identity cache with specified size and backend.
 func NewIdentityCache(backed IdentityCacheBackendFunc, size int, auditInfo []byte, metrics *Metrics) *IdentityCache {
 	logger.Debugf("new identity cache with size [%d]", size)
+	// The provisioning goroutine must not inherit any caller's request context, but it
+	// must still be cancellable, hence a cancellable child of context.Background.
+	provisionCtx, cancel := context.WithCancel(context.Background())
 	ci := &IdentityCache{
 		backed:       backed,
 		cache:        make(chan *idriver.IdentityDescriptor, size),
 		auditInfo:    auditInfo,
 		cacheTimeout: 5 * time.Millisecond,
 		metrics:      metrics,
+		provisionCtx: provisionCtx,
+		cancel:       cancel,
 	}
 
 	return ci
 }
 
-// Close stops the background identity provisioning.
+// Close stops the background identity provisioning. It is idempotent and safe to call
+// even if provisioning was never started, or concurrently with Identity.
 func (c *IdentityCache) Close() {
-	if c.cancel != nil {
-		c.cancel()
-	}
+	c.cancel()
 }
 
 // Identity retrieves an identity from cache or generates on-demand.
@@ -70,10 +78,10 @@ func (c *IdentityCache) Identity(ctx context.Context, auditInfo []byte) (*idrive
 
 	c.once.Do(func() {
 		logger.DebugfContext(ctx, "provision identities with cache size [%d]", cap(c.cache))
-		if cap(c.cache) > 0 {
-			var backgroundCtx context.Context
-			backgroundCtx, c.cancel = context.WithCancel(context.Background())
-			go c.provisionIdentities(backgroundCtx)
+		// Do not spawn the goroutine if the cache has already been closed, otherwise a
+		// late first call would start a goroutine that nothing will ever stop.
+		if cap(c.cache) > 0 && c.provisionCtx.Err() == nil {
+			go c.provisionIdentities(c.provisionCtx)
 		}
 	})
 
@@ -143,6 +151,7 @@ func (c *IdentityCache) provisionIdentities(ctx context.Context) {
 	for {
 		identityDescriptor, err := c.backed(ctx, c.auditInfo)
 		if err != nil {
+			c.metrics.ProvisionFailuresCount.Add(1)
 			logger.Errorf("failed to provision identity [%s]", err)
 			select {
 			case <-ctx.Done():
@@ -153,9 +162,11 @@ func (c *IdentityCache) provisionIdentities(ctx context.Context) {
 			continue
 		}
 		logger.DebugfContext(ctx, "generated new idemix identity [%d]", count)
-		c.metrics.CacheLevelGauge.Add(1)
+		// The gauge is incremented only once the entry is actually in the channel, so a
+		// cancellation mid-send cannot leave the reported cache level skewed.
 		select {
 		case c.cache <- identityDescriptor:
+			c.metrics.CacheLevelGauge.Add(1)
 			count++
 		case <-ctx.Done():
 			return
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -106,6 +106,7 @@ linters:
         - go.opentelemetry.io/otel/trace.Span
         - github.com/hyperledger/fabric-lib-go/common/metrics.Gauge
         - github.com/hyperledger/fabric-lib-go/common/metrics.Histogram
+        - github.com/hyperledger-labs/fabric-smart-client/platform/view/services/metrics.Gauge
         - github.com/hyperledger-labs/fabric-smart-client/platform/common/driver.ConfigService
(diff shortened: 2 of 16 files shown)


token/services/identity/idemix/cache/cache.go
```diff
@@ -36,29 +36,37 @@ type IdentityCache struct {
...
 }
 
-// Close stops the background identity provisioning.
+// Close stops the background identity provisioning. It is idempotent and safe to call
+// even if provisioning was never started, or concurrently with Identity.
 func (c *IdentityCache) Close() {
-	if c.cancel != nil {
-		c.cancel()
-	}
+	c.cancel()
```
Effi-S, line 69: You sure about this?

Effi-S (approved):
Looks Good!

token/services/identity/idemix/cache/cache.go
AkramBitar: cancel can't be nil, it's set in the constructor, which is the only construction path. Adding the nil check back would only turn a loud panic into a silent "closed but still running" leak.

Issue #2054, opened 2026-07-30
network-driver: lookup fallback rescan starting block underflows on chains younger than 10 blocks


adecaro:
### Summary

The lookup package's fallback block-scan starting point is computed with an unsigned-integer subtraction that can underflow instead of clamping to the intended floor, on any chain younger than 10 blocks.

### Where

`token/services/network/fabric/lookup/deliveryqs.go:145`:

```go
startingBlock := max(FirstBlock, lastBlock-NumberPastBlocks)   // FirstBlock=1, NumberPastBlocks=10
```

All three values are `uint64`. When `lastBlock < NumberPastBlocks`, `lastBlock - NumberPastBlocks` wraps around to a value near `math.MaxUint64` instead of going negative — so `max()` never gets a chance to clamp it back down to `FirstBlock` as intended.

### Impact

On any chain younger than 10 blocks (freshly bootstrapped network, or a namespace/channel just created), a lookup query that falls back to the block scan (because the chaincode query failed or returned a value the lookup couldn't resolve) starts scanning from an astronomically high, nonexistent block number instead of genesis — silently finding nothing, ever, for otherwise valid data that is on-chain from block 1 onward. This is a correctness failure masquerading as "no results," with no error surfaced to the caller.

### Reproduction

Reproduced locally with a unit test (not yet committed):

```go
ch, err := newQuery(querier, scanner).QueryByID(t.Context(), 5, evictedFor(...))  // lastBlock=5 < 10
...
assert.Greater(t, scanner.startBlock, uint64(1)<<62, "startingBlock underflowed instead of clamping to FirstBlock")
```

Happy to include this regression test in the fix PR.

### Suggested fix

Compute the subtraction in a way that cannot underflow, e.g.:

```go
var startingBlock uint64
if lastBlock > NumberPastBlocks {
    startingBlock = lastBlock - NumberPastBlocks
}
startingBlock = max(FirstBlock, startingBlock)
```

The same fix is needed in the sibling `finality` package (#2056), which has the identical bug pattern.

### Severity

HIGH — silent, incorrect "not found" results on any young chain/channel, with no error surfaced.

PR #2138, opened 2026-08-04
fix(network-driver): clamp fallback scan starting block to the first block

### LINKED ISSUE #2056: network-driver: finality fallback rescan starting block underflows on chains younger than 10 blocks
### Summary

The finality package's fallback block-scan starting-block computation has the identical unsigned-underflow bug already filed for the `lookup` package (#2054), in its own independent `DeliveryScanQueryByID` implementation.

### Where

`token/services/network/fabric/finality/deliveryqs.go:101`:

```go
startingBlock := max(FirstBlock, lastBlock-NumberPastBlocks)   // finality/deliveryqs.go:101
```

`FirstBlock`/`NumberPastBlocks` are defined independently in this package (`finality/deliveryqs.go:20-21`) but with the same values (1 and 10) and the same unsigned-underflow exposure as the `lookup` package's copy of this logic.

### Impact

Same underlying bug as the `lookup` package (#2054): on a chain younger than 10 blocks, a finality lookup for a transaction not found on the local ledger view falls back to a block scan that starts from a wrapped-around near-`MaxUint64` block number instead of genesis — silently failing to find the transaction's finality outcome, with no error surfaced.

### Reproduction

```go
ch, err := q.QueryByID(ctx, 5, evicted("tx1"))  // lastBlock=5 < NumberPastBlocks=10
...
assert.Equal(t, uint64(18446744073709551611), scanner.startBlock,
    "startingBlock underflowed instead of clamping to FirstBlock (1)")
```

### Suggested fix

Same fix as the `lookup` package's copy of this bug (#2054): compute the subtraction in a way that cannot underflow, e.g.:

```go
var startingBlock uint64
if lastBlock > NumberPastBlocks {
    startingBlock = lastBlock - NumberPastBlocks
}
startingBlock = max(FirstBlock, startingBlock)
```

Given the two packages duplicate this logic identically, consider extracting a shared helper so the fix (and any future changes) only needs to be made once.

### Severity

HIGH — silent, incorrect finality-lookup failures on any young chain, with no error surfaced.

HayimShaul:
Fixes #2056

The finality and lookup services both computed the starting block of their fallback block scan as
`max(FirstBlock, lastBlock-NumberPastBlocks)`. `lastBlock` is a `uint64`, so on a chain younger than
`NumberPastBlocks` (10) the subtraction wrapped around and `max` picked a block number near
`MaxUint64`. The scan then found nothing and surfaced no error, silently losing the transaction's
finality outcome (or the looked-up key).

The computation now lives in a new `token/services/network/fabric/blockscan` package, where the
subtraction is guarded so a chain shorter than the rewind window is scanned from `FirstBlock`, and
both services call it instead of duplicating the expression. The same underflow existed in the
lookup copy of the logic (#2054), so extracting the helper removes it there too.

Tests: a table-driven test for `blockscan.StartingBlock` (including 0, 1, 5, 10, 11 and
`MaxUint64`), plus young-chain regression tests for the fallback scan in both `finality` and
`lookup` that assert the scan starts at block 1 rather than at the wrapped value. All three fail
against the previous arithmetic.

`make checks` and `make lint` pass.

### DIFF
--- a/docs/services/network-fabric.md
+++ b/docs/services/network-fabric.md
@@ -357,6 +357,13 @@ resolved independently: if its query cannot be built, does not reach the chainco
 response that cannot be decoded, only that namespace falls back to the block scan. The keys of the
 other namespaces in the same batch are still resolved and delivered.
 
+The block scan itself is the fallback shared by the lookup service and the
+[finality service](../../token/services/network/fabric/finality/deliveryqs.go): it rewinds
+`NumberPastBlocks` blocks behind the last known block and scans forward from there. The starting
+block is computed by [`blockscan.StartingBlock`](../../token/services/network/fabric/blockscan/blockscan.go),
+which clamps the result to `FirstBlock` so that a chain shorter than the rewind window is scanned
+from its first block rather than from an underflowed block number.
+
 ## State Queries
 
 The Fabric implementation provides efficient state querying through the chaincode:
--- a/token/services/network/fabric/blockscan/blockscan.go
+++ b/token/services/network/fabric/blockscan/blockscan.go
@@ -1,0 +1,28 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+// Package blockscan holds the parameters shared by the Fabric services that fall back to
+// scanning past blocks when a lookup misses on the local view of the ledger.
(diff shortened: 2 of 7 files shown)


SaidAltury-ibm (approved):
Thank you @HayimShaul , LGTM

PR #2139, opened 2026-08-04
fix(network): fill same-MSP endorser slots with distinct identities

### LINKED ISSUE #2048: network-driver: duplicate identity can fill two distinct required MSP slots in SelectEndorsersForMSPSets
### Summary

`SelectEndorsersForMSPSets`'s per-slot sampling never excludes identities already chosen earlier in the same candidate set. A policy requiring two distinct signers from the same MSP can select the *same* identity for both slots.

### Where

`token/services/network/fabric/endorsement/fsc/selection.go:42-50`:

```go
for _, mspID := range requiredMSPIDs {
    pool := byMSP[mspID]
    ...
    selected = append(selected, pool[rand.Intn(len(pool))])   // selection.go:49 — no dedup vs `selected`
}
```

### Impact

A namespace endorsement policy intended to require two *independent* signatures from the same organization (defense against a single compromised/misbehaving endorser within that org) can be satisfied by asking one endorser to sign twice, collapsing the intended 2-of-N-within-org security property down to a single point of trust — an unintended weakening of the endorsement policy's actual guarantee.

### Reproduction

Reproduced locally with a unit test (not yet committed) that runs the selection up to 2000 times (since selection is randomized) and asserts the duplicate is observed at least once:

```go
for range 2000 {
    selected, err := fsc.SelectEndorsersForMSPSets(configured, mspOf, candidates)
    require.NoError(t, err)
    require.Len(t, selected, 2)
    if selected[0].Equal(selected[1]) { duplicateObserved = true; break }
}
assert.True(t, duplicateObserved)
```

### Suggested fix

Exclude identities already present in `selected` from the sampling pool for subsequent slots in the same candidate set, so two required slots in the same MSP are guaranteed to be filled by two distinct identities.

### Severity

HIGH — silently weakens an intended multi-signer-within-org security property.

HayimShaul:
Fixes #2048

`SelectEndorsersForMSPSets` sampled one configured endorser per required MSP ID of a candidate set without excluding identities already picked for an earlier slot of the *same* set. A namespace endorsement policy requiring two signers from one MSP — e.g. `AND(Org1MSP.member, Org1MSP.member)`, whose principal set `inquire.SatisfiedBy()` reports as `["Org1MSP", "Org1MSP"]` — could therefore be satisfied by one endorser signing twice, collapsing the intended 2-of-N-within-org guarantee onto a single point of trust.

Changes:

- Per-slot sampling now excludes identities already selected for the same candidate set, so every returned identity is distinct and the result length always matches the chosen candidate set.
- A candidate set requiring more distinct signers from an MSP than that MSP has configured endorsers is no longer satisfiable: it is skipped so the remaining candidate sets are tried, and only a hard error is returned if none is coverable — correctness over availability, consistent with the rest of this selector.
- Duplicate entries in the configured endorser list collapse to a single candidate, since one endorser provides one endorsement; counting it twice skewed the uniform random pick and overstated how many distinct endorsers an MSP offers.
- Regression tests: the same-MSP-twice case never returns a duplicate over many randomized runs (and both endorsers remain reachable in both slots), plus non-coverable-set fallthrough, duplicate configured entries, and a mixed repeated/single MSP set. All five new subtests fail against the pre-fix code.
- Docs updated (`docs/services/network-fabricx.md`, `docs/services/network-fabric.md`, `docs/configuration.md`) to state that same-MSP signer slots are filled with distinct endorsers, and that an MSP with fewer distinct endorsers than required signers is a hard error.

`make checks` and `make lint` pass; the affected packages pass under `-race`.

**More Information about the Problem and the Fix**

**Problem (steps)**

1. Namespace endorsement policy requires 2 signers from the same MSP: AND(Org1MSP.member, Org1MSP.member)
2. Policy check reports required MSP set as ["Org1MSP", "Org1MSP"]
3. Selector loops over each slot independently, picking a random configured Org1 endorser each time
4. No exclusion of already-picked endorsers → same endorser can be drawn for both slots
5. Result: one endorser signs twice, satisfying a "2 distinct signers" policy with a single point of trust

**Solution (steps)**

1. Track used identities per candidate set being filled
2. For each required slot, pick randomly only from that MSP's pool minus already-used identities
3. If a slot's pool is empty (MSP has fewer distinct endorsers than required signers) → mark that candidate set unsatisfiable, skip it, try the next candidate set
4. Collapse duplicate entries in the configured endorser list to one (a duplicate is one endorser, not two)
5. If no candidate set is coverable, return a hard error naming the exact MSP and required-vs-available count
6. Add regression tests + update docs to state the new distinct-per-slot guarantee

### DIFF
--- a/token/services/network/fabric/endorsement/fsc/selection.go
+++ b/token/services/network/fabric/endorsement/fsc/selection.go
@@ -19,14 +19,27 @@ import (
 // that set, one random configured endorser belonging to it. mspOf resolves a configured
 // identity to the MSP ID it belongs to.
 //
-// It returns an error if none of the candidate sets can be fully covered by the
-// configured endorsers.
+// A candidate set may list the same MSP ID more than once: a policy such as
+// AND(Org1MSP.member, Org1MSP.member) requires that many *distinct* signers from that MSP,
+// which is exactly the property protecting against a single misbehaving endorser within the
+// organization. Every returned identity is therefore distinct by identity bytes: a slot is
+// never filled with an identity already selected for an earlier slot of the same set, and
+// duplicate entries in configured collapse to a single candidate endorser. The result always
+// has exactly as many elements as the chosen candidate set.
+//
+// A candidate set requiring more distinct signers from an MSP than there are distinct
+// configured endorsers in it cannot be satisfied and is skipped, like one naming an MSP with
+// no configured endorser at all.
+//
+// It returns an error if none of the candidate sets can be fully covered by distinct
+// configured endorsers. The error names the first MSP that blocked each candidate set.
 func SelectEndorsersForMSPSets(configured []view.Identity, mspOf func(view.Identity) (string, error), candidates [][]string) ([]view.Identity, error) {
 	if len(candidates) == 0 {
 		return nil, errors.Errorf("no candidate MSP set to satisfy the namespace endorsement policy")
 	}
 	var skipped []error
 	byMSP := make(map[string][]view.Identity)
+	seen := make(map[string]struct{}, len(configured))
 	for _, id := range configured {
 		mspID, err := mspOf(id)
 		if err != nil {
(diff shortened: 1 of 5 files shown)


AkramBitar (commented):
Three notes on the fix itself.

token/services/network/fabric/endorsement/fsc/selection.go
```diff
@@ -19,39 +19,77 @@ import (
...
 		mspID, err := mspOf(id)
 		if err != nil {
 			return nil, errors.WithMessagef(err, "failed to resolve MSP for configured endorser [%s]", id)
 		}
+		if _, ok := seen[string(id)]; ok {
+			// configured may legitimately list the same endorser twice; it still provides a
+			// single endorsement, so bucket it once.
+			continue
+		}
+		seen[string(id)] = struct{}{}
```
AkramBitar, line 56: Dedup is byte-exact, so two byte-distinct identities carrying the **same signing key** count as two endorsers and can fill both same-MSP slots — one physical signer, the collapse this PR closes.

The sibling `threshold_rule` path treats exactly that as an anomaly (`fabricx/endorsement/nspolicy.go:209-212`: two distinct identities with the same key "would mean two different endorsers share a private key"), so the two paths now disagree on the same question. Narrow in practice, but the godoc promises "never the same endorser twice", which byte equality does not quite deliver.

Either reuse `ecdsaPublicKeyOf` to close it, or add one godoc line saying distinctness is by identity bytes.

token/services/network/fabric/endorsement/fsc/selection.go
```diff
@@ -19,39 +19,77 @@ import (
...
-			if len(pool) == 0 {
-				satisfied = false
-
-				break
+		if selected, ok := selectDistinctForMSPSet(byMSP, candidates[idx]); ok {
+			return selected, nil
+		}
+	}
+
+	return nil, errors.Errorf("no configured endorser covers any of the [%d] policy-satisfying MSP set(s) with a distinct endorser per required signer", len(candidates))
```
AkramBitar: Both failure modes produce this one string, and neither is diagnosable: with `AND(Org1, Org1)` and one configured Org1 endorser, the operator learns only that [1] set failed — not which MSP was short, nor by how much. This PR’s docs make "fewer distinct endorsers than required signers" its own error class, so the error should be able to name it.

`selectDistinctForMSPSet` returns only `bool`. Returning the failing `mspID` plus required-vs-available would give: *"MSP [Org1MSP] requires 2 distinct signers but only 1 configured endorser belongs to it"*.

token/services/network/fabric/endorsement/fsc/selection.go
```diff
@@ -19,39 +19,77 @@ import (
...
+
+// selectDistinctForMSPSet fills one slot per entry of requiredMSPIDs with a random
+// configured endorser of that MSP, never reusing an identity already selected for an
+// earlier slot of the same set. It returns ok=false if some slot cannot be filled with a
+// still-unused endorser, meaning the set is not coverable by distinct endorsers.
+//
+// A greedy per-slot pick is complete here: every slot requiring a given MSP ID draws from
+// the same pool, so the only way this fails is that some MSP ID appears in requiredMSPIDs
+// more times than that MSP has distinct configured endorsers - genuinely unsatisfiable
+// whatever the order of the picks. No backtracking is needed.
```
AkramBitar, line 94: Confirming this argument rather than leaving the next reviewer to re-derive it: `byMSP` partitions the identities, so slots for MSP `m` draw only from `pool[m]` and the pools are disjoint. Feasibility is therefore just `count(m) <= |pool[m]|` per MSP, which greedy always attains — no ordering does better, so skipping backtracking is correct, not merely convenient.

Same for the test: it asserts both endorsers reach *both* slots, not just distinctness. The obvious "take the first unused" fix would pass the latter and fail the former.

AkramBitar (changes requested):

AkramBitar (dismissed):
@HayimShaul will be out of the office for the next two weeks. I'll be covering for him during that time, so I won't be able to continue reviewing this PR. Someone else will need to take over the review.

SaidAltury-ibm (approved):

Effi-S (approved):

AkramBitar (approved):
LGTM

PR #2140, opened 2026-08-04
Added Backoff

### LINKED ISSUE #2105: Auditor token validation can spuriously fail on recently-audited outputs (async commit-notification race)
## Problem

Auditor-side transaction validation can fail with a spurious error when a
transaction references (as inputs) outputs of a very recently-audited
transaction, because the local read used for audit token lookups is not
guaranteed to reflect a transaction's outputs by the time a dependent
transaction reaches audit validation.

## Where this happens

`AuditorCheck` (`token/core/fabtoken/v1/auditor.go`,
`token/core/zkatdlog/nogh/v1/auditor.go`) calls `common.RetrieveAuditTokens`
(`token/core/common/auditor.go`), which does a single, non-retrying call to
`driver.QueryEngine.ListAuditTokens`. That query engine
(`token/sdk/vault/vault.go`) reads directly from the SQL-backed token store
(`token/services/storage/db/sql/common/tokens.go`), which returns a hard
error (`"token not found for key [...]"`) the moment a requested token row is
missing — no retry, no backoff.

That SQL row for a transaction's outputs is only written by
`tokens.Service.AppendValid`, which itself only runs inside the async
finality-listener callback (`token/services/ttx/finality/listener.go`)
fired once the network layer reports the transaction as final. This is
fully decoupled — in timing — from anything the auditor did when it signed
off on that earlier transaction. A closely-following transaction that spends
or references those same outputs can therefore reach `AuditorCheck` before
the listener has finished persisting them, causing audit validation to fail
even though the referenced transaction was itself already validated and
signed off by the same auditor.

## Existing partial mitigation, and an inconsistency

A different call path — `auditor.Service.Audit` → `Request.AuditRecord`
(`token/request.go`) — uses a *different* query-engine wrapper
(`token.QueryEngine` in `token/vault.go`) that *does* retry on this exact
condition: it checks whether the owning transaction is `Pending` and retries
with backoff (3 retries × 3s) before giving up. That wrapper is a different
concrete type from `driver.QueryEngine` and cannot be substituted into the
driver-level `AuditorService.QueryEngine` field used by `AuditorCheck`, so
the retry logic never applies there.

The net effect is an inconsistency: `Validate()`/`AuditorCheck()` — which
runs *before* `Audit()` in the auditor flow — has no tolerance at all for
this race, while `Audit()`, which runs after, has a bounded (~9s) grace
window for the same underlying condition.

## Additional context: an unused signal

`tokens.Service` already maintains an in-memory `RequestsCache`, populated
at audit-approval time (`AuditApproveView.Call` in
`token/services/ttx/auditor.go`, via `CacheRequest`) — i.e. before the
transaction is even committed. That cache entry contains enough information
(`TxID`, `Index`, and the token itself) to resolve the exact token IDs that
`AuditorCheck` looks up. However, neither `RetrieveAuditTokens` nor either
driver's `AuditorService` currently has any reference to `tokens.Service` —
they're constructed with only a `driver.QueryEngine` — so this signal is
never consulted during audit validation today.

## Impact

Auditors can reject (or need to retry) valid, already-approved transactions
purely due to local read timing, not due to any actual problem with the
transaction. This is more likely under load or when transactions chain
quickly (e.g. an output spent shortly after being audited).

## Suggestions for investigation (not prescribing a fix)

- Whether `AuditorCheck`'s lookup path should gain the same
  pending-status retry/backoff tolerance that `Request.AuditRecord`'s path
  already has, to remove the inconsistency between the two gates.
- Whether the existing `tokens.Service` cache could serve as an additional
  fallback for resolving audit tokens that are momentarily missing from the
  DB but already known to be in flight, and if so what plumbing changes
  wiring `tokens.Service` into `AuditorService` construction would require
  (currently no such dependency exists in either the fabtoken or zkatdlog
  driver construction chains).

Effi-S:
Fixes #2105

Adds retry/backoff to the auditor validation lookup (`AuditorCheck` / `RetrieveAuditTokens`) so a quickly-chained transaction is no longer spuriously rejected when the async finality listener has not yet persisted a just-audited output. This mirrors the pending-status retry/backoff that the sibling `Audit()` path already has.

### DIFF
--- a/token/core/common/auditor.go
+++ b/token/core/common/auditor.go
@@ -8,6 +8,7 @@ package common
 
 import (
 	"context"
+	"time"
 
 	"github.com/LFDT-Panurus/panurus/token/driver"
 	"github.com/LFDT-Panurus/panurus/token/driver/protos-go/v1/request"
@@ -15,6 +16,16 @@ import (
 	"github.com/LFDT-Panurus/panurus/token/token"
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"go.opentelemetry.io/otel/trace"
+)
+
+// AuditTokensNumRetries and AuditTokensRetryDelay control how RetrieveAuditTokens
+// tolerates the read-timing race in which a referenced token's producing
+// transaction is still pending, so its outputs have not yet been persisted to the
+// token store by the asynchronous finality listener. They mirror the retry/backoff
+// already applied on the sibling token.QueryEngine audit path (see token/vault.go).
+var (
+	AuditTokensNumRetries = 3
+	AuditTokensRetryDelay = 3 * time.Second
 )
 
 // AuditContext contains the context for token request auditing.
@@ -267,7 +278,7 @@ func ExtractTokenIDsAndCheckDuplicates(
 // The returned map uses token ID pointers as keys, allowing callers to efficiently look up
 // tokens by their ID during validation.
 //
-// IMPORTANT: This function always returns a non-nil map (possibly empty) to ensure
+// This function always returns a non-nil map (possibly empty) to ensure
 // validation logic can distinguish between "no tokens requested" and "tokens not found".
 func RetrieveAuditTokens(
 	ctx context.Context,
@@ -288,7 +299,7 @@ func RetrieveAuditTokens(
 	}
 
 	logger.DebugfContext(ctx, "[%s] retrieving [%d] audit tokens...", anchor, len(tokenIDs))
-	tokens, err := queryEngine.ListAuditTokens(ctx, tokenIDs...)
+	tokens, err := listAuditTokensWithRetry(ctx, logger, queryEngine, tokenIDs, anchor)
 	if err != nil {
 		return nil, errors.WithMessagef(err, "failed to retrieve audit tokens for tx [%s]", anchor)
 	}
@@ -302,6 +313,59 @@ func RetrieveAuditTokens(
 	logger.DebugfContext(ctx, "[%s] retrieved [%d] audit tokens", anchor, len(auditTokens))
 
 	return auditTokens, nil
+}
+
+// listAuditTokensWithRetry calls queryEngine.ListAuditTokens, tolerating the
+// read-timing race where a referenced token is momentarily missing from the token
+// store because its producing transaction is still pending (its outputs are
+// persisted only later, by the asynchronous finality listener). On failure, it
+// checks whether any requested token belongs to a still-pending transaction and,
+// if so, waits AuditTokensRetryDelay and retries up to AuditTokensNumRetries times
+// before giving up. This mirrors the tolerance already implemented for the sibling
+// Audit() path in token/vault.go, so the earlier AuditorCheck gate no longer
+// spuriously rejects a validly-audited, quickly-chained transaction.
+func listAuditTokensWithRetry(
+	ctx context.Context,
+	logger logging.Logger,
+	queryEngine driver.QueryEngine,
+	tokenIDs []*token.ID,
+	anchor driver.TokenRequestAnchor,
+) ([]*token.Token, error) {
+	var tokens []*token.Token
+	var err error
+
+	for i := range AuditTokensNumRetries {
+		tokens, err = queryEngine.ListAuditTokens(ctx, tokenIDs...)
+		if err == nil {
+			return tokens, nil
+		}
+
+		// The lookup failed. Check whether any requested token belongs to a
+		// transaction that is still pending; if so, the row is expected to appear
+		// once the finality listener persists it, so wait a bit and retry.
+		retry := false
+		for _, id := range tokenIDs {
+			pending, pErr := queryEngine.IsPending(ctx, id)
+			if pending || pErr != nil {
+				logger.Warnf("[%s] cannot get audit token for id [%s] because the relative transaction is pending, retry [%d/%d]: with err [%v]", anchor, id, i+1, AuditTokensNumRetries, pErr)
+				if i == AuditTokensNumRetries-1 {
+					return nil, errors.Errorf("failed to get audit tokens, tx [%s] is still pending", id.TxId)
+				}
+				retry = true
+
+				break
+			}
+		}
+
+		if !retry {
+			// None of the tokens is pending: this is a genuine failure, do not retry.
+			return nil, err
+		}
+
+		time.Sleep(AuditTokensRetryDelay)
+	}
+
+	return tokens, err
 }
 
 // ValidateStructure ensures complete structural correspondence between TokenRequest and TokenRequestMetadata.
--- a/token/core/common/auditor.go
+++ b/token/core/common/auditor.go
@@ -18,14 +18,16 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
-// AuditTokensNumRetries and AuditTokensRetryDelay control how RetrieveAuditTokens
-// tolerates the read-timing race in which a referenced token's producing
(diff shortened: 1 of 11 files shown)


token/core/common/auditor.go
```diff
@@ -304,6 +315,59 @@ func RetrieveAuditTokens(
...
+
+		if !retry {
+			// None of the tokens is pending: this is a genuine failure, do not retry.
+			return nil, err
+		}
+
+		time.Sleep(AuditTokensRetryDelay)
+	}
+
+	return tokens, err
```
AkramBitar: **Panic (index out of range) when `AuditTokensNumRetries` is set to 0.**

`AuditTokensNumRetries` is an exported, documented tuning var. If it is set to `0` (the natural way to "disable retries") or to any value `<= 0`, `for i := range AuditTokensNumRetries` never executes its body and we fall through to this `return tokens, err` with **both** values `nil`.

`RetrieveAuditTokens` then sees `err == nil` and runs `auditTokens[id.String()] = tokens[i]` at line 311 over a nil slice:

```
panic: runtime error: index out of range [0] with length 0
    token/core/common/auditor.go:311
```

(verified with a throwaway test). Before this PR the setting was harmless — the old code always made exactly one attempt.

Suggest `attempts := max(1, AuditTokensNumRetries)` for the loop bound, and making this trailing return an explicit error so a nil/nil pair can never escape.

token/core/common/auditor.go
```diff
@@ -304,6 +315,59 @@ func RetrieveAuditTokens(
...
+				break
+			}
+		}
+
+		if !retry {
+			// None of the tokens is pending: this is a genuine failure, do not retry.
+			return nil, err
+		}
+
+		time.Sleep(AuditTokensRetryDelay)
```
AkramBitar: **`time.Sleep` ignores context cancellation, making the audit gate uninterruptible for ~6s on requester-controlled input.**

`ctx` is threaded into this function but never consulted; the loop sleeps 2 x 3s.

The token IDs come from `ExtractTokenIDsAndCheckDuplicates(metadata, ...)`, i.e. from the peer-supplied request metadata, and `AuditorCheck` runs *before* any lock acquisition or signature work. So a peer can reference inputs of a transaction that is (or stays) `Pending` in the auditor's DB and force each request to pin a goroutine for 6s plus `3 x N` `IsPending` calls — each `GetStatus` hits ttxdb then auditdb, so up to `6 x N` queries. Cancelling or timing out the client session frees nothing. Before this PR such a request failed immediately.

The ctx-aware helper is already used two frames up, at `token/services/auditor/auditor.go:189`:

```go
err := retryRunner.RunWithContext(ctx, func() error { ... })
```

Either use `utils.NewRetryRunnerWithJitter(...).RunWithContext`, or inline:

```go
select {
case <-ctx.Done():
    return nil, ctx.Err()
case <-time.After(AuditTokensRetryDelay):
}
```

token/core/common/auditor.go
```diff
@@ -304,6 +315,59 @@ func RetrieveAuditTokens(
...
+			return tokens, nil
+		}
+
+		// The lookup failed. Check whether any requested token belongs to a
+		// transaction that is still pending; if so, the row is expected to appear
+		// once the finality listener persists it, so wait a bit and retry.
+		retry := false
+		for _, id := range tokenIDs {
+			pending, pErr := queryEngine.IsPending(ctx, id)
+			if pending || pErr != nil {
```
AkramBitar: **`pErr != nil` is treated as "pending", so a store outage is misreported as a pending transaction and the real error is discarded.**

`if pending || pErr != nil` enters the retry path when `IsPending` *itself* fails. During a DB outage `ListAuditTokens` fails **and** `IsPending` fails, so every audit burns the full 6s window and then returns

```go
errors.Errorf("failed to get audit tokens, tx [%s] is still pending", id.TxId)  // line 352
```

dropping both the `ListAuditTokens` error and `pErr`. An operator debugging a connectivity or schema problem sees "still pending" with no underlying cause.

The same applies to a hard, non-retryable error such as `"retrieved wrong token [%s]"` from `tokens.go:681`: whenever any referenced tx happens to be pending, that error is swallowed and reported as pending too.

Suggest wrapping the original `err` (and `pErr`) into the returned error, and claiming "pending" only when `pending` is actually true.

token/core/common/auditor.go
```diff
@@ -17,6 +18,16 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
+// AuditTokensNumRetries and AuditTokensRetryDelay control how RetrieveAuditTokens
+// tolerates the read-timing race in which a referenced token's producing
+// transaction is still pending, so its outputs have not yet been persisted to the
+// token store by the asynchronous finality listener. They mirror the retry/backoff
+// already applied on the sibling token.QueryEngine audit path (see token/vault.go).
+var (
```
AkramBitar: **Retry parameters are unsynchronized exported package globals, rather than per-instance config.**

Unlike the sibling implementation these mirror — `token/vault.go:43-44`, where `NumRetries`/`RetryDelay` are struct fields — these are process-wide mutable vars read on every audit. Any consumer tuning them at runtime (e.g. a config-reload goroutine) races with concurrent `AuditorCheck` calls; `token/core/zkatdlog/nogh/v1/auditor_test.go:50` already exercises `AuditorCheck` from parallel goroutines.

The new tests themselves mutate the global (`withFastAuditRetries`), which is safe only because nothing in that package uses `t.Parallel()` today.

Prefer fields on `AuditorService`, or plumb them through `NewAuditorService`, so per-TMS tuning is possible and no shared mutable state is introduced.

token/core/common/auditor.go
```diff
@@ -304,6 +315,59 @@ func RetrieveAuditTokens(
 	return auditTokens, nil
 }
 
+// listAuditTokensWithRetry calls queryEngine.ListAuditTokens, tolerating the
+// read-timing race where a referenced token is momentarily missing from the token
+// store because its producing transaction is still pending (its outputs are
+// persisted only later, by the asynchronous finality listener). On failure, it
+// checks whether any requested token belongs to a still-pending transaction and,
+// if so, waits AuditTokensRetryDelay and retries up to AuditTokensNumRetries times
```
AkramBitar: **Doc comment overstates the retry budget (off-by-one).**

"retries up to `AuditTokensNumRetries` times" reads as 3 retries on top of the first attempt — 4 attempts, ~9s. The loop actually makes `AuditTokensNumRetries` *total attempts*: 3 attempts with only 2 sleeps, i.e. a ~6s grace window.

Issue #2105 describes the sibling path as a "~9s grace window" for the same reason, so the mistake is easy to inherit. Anyone sizing the window against observed commit latency will over-estimate it by one delay.

Suggest: "makes up to `AuditTokensNumRetries` attempts, sleeping `AuditTokensRetryDelay` between them (N-1 delays)".

AkramBitar (commented):
Review of the pending-status retry/backoff added to `RetrieveAuditTokens`.

The premise checks out: `finality.Commit` writes the token rows (`AppendValid`) and the `Confirmed` status in the same store transaction, so `IsPending == false` really does imply the rows are visible; the driver-level `qe` here is `token/sdk/vault.QueryEngine`, *not* the retrying `token.QueryEngine` wrapper, so there is no accidental 3x3 nesting; and the added latency fits inside the initiator's 1-minute `ReceiveTypedWithTimeout`.

`go vet`, `golangci-lint` (0 issues) and `go test ./token/core/common/` all pass on this head. Finding 1 below was reproduced empirically with a throwaway test.

Five findings inline, most severe first.

AkramBitar (changes requested):

AkramBitar (approved):

Issue #2128, opened 2026-08-04
Fix `time.After` in `StreamExternalWalletSignerClient.Respond` [mem]


AkramBitar:
**Document severity:** MEDIUM · **Verified priority:** Low · **Effort:** Low · **Type:** Task

### Old issue reference

> **Sub-Task 7 — Fix `time.After` in `StreamExternalWalletSignerClient.Respond`**
>
> **Status:** `[ ] pending`

| # | Tag | Severity | Category | File | Key Lines | Issue |
|---|-----|----------|----------|------|-----------|-------|
| 7 | `[RUN]` | **MEDIUM** | Timer leak | `ttx/external.go` | 188 | `time.After` created per request in `Respond()` |

### Intent

`Respond()` loops over incoming sign requests using `time.After(s.timeout)` in
a `select`. Since `s.timeout` defaults to 1 hour, every request that is
processed normally causes an extra 1-hour timer to linger until it fires.
Under steady load this keeps a large number of live timer objects in memory.

### Expected Outcomes

- A single timer (or ticker) is used across the entire lifetime of `Respond()`.
- The timer is reset to a fresh deadline only after a successful message is
  processed (i.e., a sliding timeout), or a single `time.NewTimer` wrapping
  the entire outer loop is used.

### Todo List

1. Move `timer := time.NewTimer(s.timeout)` outside the `for` loop.
2. Add `defer timer.Stop()` immediately after.
3. Use `timer.C` in the select.
4. After successfully receiving a `SigRequest`, call
   `timer.Reset(s.timeout)` (after draining `timer.C` if necessary) to
   restart the sliding window.

### Relevant Context

- File: `token/services/ttx/external.go`
- Lines 173–192: `Respond()` loop, line 188 is the offending call.

---

### Status in current code

**Not a leak on Go 1.26** — code is unchanged, but the reported defect does not occur on this toolchain. The real defect in this file is tracked separately as Sub-Task 11.

`token/services/ttx/external.go:188` still reads `case <-time.After(s.timeout):`
inside the `Respond()` loop, as documented.

This was the strongest timer argument in the source document — a 1-hour timer retained per
processed request, under steady load. It does not hold on this toolchain.

[...]

PR #2142, opened 2026-08-04
fix(ttx): give StreamExternalWalletSignerClient's receive goroutine a way out

### LINKED ISSUE #2132: Fix `err` Channel Never Read in `StreamExternalWalletSignerClient` [mem]
**Document severity:** HIGH · **Verified priority:** High · **Effort:** Low · **Type:** Bug

### Old issue reference

> **Sub-Task 11 — Fix `err` Channel Never Read in `StreamExternalWalletSignerClient`**
>
> **Status:** `[ ] pending`

| 11 | `[RUN]` | **HIGH** | Channel/goroutine leak | `ttx/external.go` | 134–192 | `err` channel never read; `init()` goroutine blocked |

### Intent

`StreamExternalWalletSignerClient` creates two channels: `input` and `err`.
The `init()` goroutine sends on `err` when a stream decode fails, then returns.
However, `Respond()` never selects on `err`. The `init()` goroutine will block
forever on the unbuffered `err` channel send and never exit, holding a
reference to the stream, the channels, and all captured state.

### Expected Outcomes

- `init()` goroutine always has a path to exit on stream errors.
- Errors from `init()` are surfaced to the caller of `Respond()`.

### Todo List

1. Make `err` a buffered channel of size 1: `make(chan error, 1)`.
   This allows `init()` to send its error and exit without blocking.
2. Add a `case err := <-s.err:` branch inside the `select` in `Respond()` so
   that the error is returned to the caller.
3. Ensure `Respond()` exits cleanly after receiving from `s.err`.

### Relevant Context

- File: `token/services/ttx/external.go`
- Lines 117–170: struct definition and `init()` goroutine.
- Lines 173–192: `Respond()` method.

---

### Status in current code

**Confirmed — still present.**

Verified by grepping every reference to `s.err` in the package. There are exactly
two, and both are sends:

- `token/services/ttx/external.go:149` — `s.err <- errors.Wrapf(err, "failed to receive signature request [%d]", i)`
- `token/services/ttx/external.go:157` — `s.err <- errors.Wrapf(err, "failed to get unmarshal msg type SigRequest")`

There is no receive on `s.err` anywhere. The channel is created unbuffered at `:135`
(`err: make(chan error)`), and `Respond()` (`:184-192`) selects only on `s.input` and
`time.After(s.timeout)`.

Consequences, both of which are real:

1. **Goroutine leak, unbounded in the number of failing streams.** On any `stream.Recv`
   failure or `json.Unmarshal` failure, `init()` blocks permanently on the send at `:149` or
   `:157`. It never reaches its `return`, so it pins the gRPC stream, both channels, and all
   captured state for the process lifetime. One leaked goroutine per failed stream, and
   nothing bounds how many streams fail.
2. **Silent error loss.** The error the goroutine was trying to report never reaches anyone.
   `Respond()` instead sits until `s.timeout` elapses and returns a timeout error, so an
   operator sees "timeout waiting for stream input" for what was actually a decode or
   transport failure. The misleading diagnostic is arguably worse than the leak.

The document's remediation is correct and minimal: buffer `err` to 1 so `init()` can always
complete its send and exit, and add a `case err := <-s.err:` arm to the `select` in `Respond()`
so the real error propagates. Both changes are confined to this file.

Note that Sub-Task 7 targets the same `select` in `Respond()` on timer grounds that do not
hold on Go 1.26. This issue is the one with a real defect behind it; if both are touched, do
this one and treat 7 as already addressed by proximity.

### Severity and priority

`origin/main` with `go 1.26.5` on 2026-08-04, the priority is **High** for the

### Effort

**Low**

### LINKED ISSUE #2141: Fix `input` Channel Handover With No Exit Path in `StreamExternalWalletSignerClient` [mem]
**Severity:** HIGH · **Verified priority:** Medium · **Effort:** Low · **Type:** Bug

### Origin

Found while verifying #2132 against this repository. #2132 covers the `err` channel of
`StreamExternalWalletSignerClient`; this issue covers the `input` channel of the same goroutine
pair. They are independent defects that happen to share a failure mode, and remediating #2132
does not remediate this one.

| — | `[RUN]` | **HIGH** | Channel/goroutine leak | `ttx/external.go` | 142–192 (send at 161) | unbuffered `input` handover has no exit path; `init()` goroutine blocked |

### Intent

`init()` decodes each incoming `SigRequest` and hands it to `Respond()` over the unbuffered
`input` channel (`s.input <- req`, `external.go:161`). That send has no exit path. If `Respond()`
has already returned, nothing will ever receive from `input` again, so `init()` parks on the send
permanently — pinning the FSC stream, both channels, and all captured state for the process
lifetime.

`Respond()` returns before the remote wallet is done in three places:

- `external.go:182` — `sign()` failed (no signer for the party, or the signer itself failed).
- `external.go:185` — `stream.Send` failed while returning the signature.
- `external.go:189` — the timeout elapsed.

Any `SigRequest` that `init()` decodes after `Respond()` has taken one of those exits strands the
goroutine.

The timeout exit can strand it even against a strictly lockstep peer: Go's `select` chooses at
random among ready arms, so the `time.After` arm at `external.go:188` can win while a decoded
request is already pending on the handover at `:161`. A peer that does not wait for each
`SignResponse` before sending the next `SigRequest` widens this considerably — the bundled
`StreamExternalWalletSignerServer` is strict request/response, but a third-party wallet
implementation of the same wire protocol need not be.

### Expected Outcomes

- `init()` always has a path to exit, including when no consumer remains for a decoded request.
- A request that can no longer be served is discarded rather than held on a blocked send.
- No goroutine, stream, or channel is retained after `Respond()` has returned.

### Todo List

1. Give the client a `done` channel that `Respond()` closes when it stops consuming input.
2. Wrap the handover in a `select` on `s.input <- req` and `<-s.done`, returning from `init()`
   on the latter.
3. Make the close idempotent (e.g. `sync.Once`) so a second `Respond()` call cannot panic on an
   already-closed channel.
4. Add a regression test that decodes a request after `Respond()` has returned and asserts the
   receive goroutine is no longer parked.

### Relevant Context

- File: `token/services/ttx/external.go`
- Lines 142–171: the `init()` goroutine; the handover is at 161.
- Lines 173–192: `Respond()`, including its three early-return paths and the timeout arm.
- Related: #2132 (`err` channel of the same goroutine), #2128 (`time.After` in the same `select`).

### Status in current code

**Confirmed present at `origin/main`.**

`input` is created unbuffered at `external.go:134` (`input: make(chan *StreamExternalWalletSignRequest)`).
The only receive is the `case req, done := <-s.input` arm of the `select` in `Respond()`
(`:176`), which is unreachable once `Respond()` has returned. The send at `:161` is bare — no
`select`, no default, no cancellation — so it blocks until a receiver appears, and after
`Respond()` returns none ever will.

Observable as a goroutine parked in `[chan send]` inside
`(*StreamExternalWalletSignerClient).init` in a stack dump taken after `Respond()` has returned.

Two consequences:

1. **Goroutine leak, one per affected stream.** The goroutine never reaches its `return`, so it
   retains the gRPC stream, both channels, and the captured `SignerProvider` for the process
   lifetime. Nothing bounds how many streams reach this state.
2. **Silently dropped signature request.** The decoded request is never served and never
   reported. Since `Respond()` has already returned its own error, the caller has no indication
   that a further request arrived and was lost.

Note that `init()` also waits in `s.stream.Recv` (`:148`), which is *not* part of this defect:
both `view2.Stream` implementations in FSC delegate to gRPC `RecvMsg`, which returns on a
message, `io.EOF`, a stream error, or context cancellation, so that wait is bounded by the
stream's lifetime rather than the process's.

### Severity and priority

Severity **HIGH** — an unbounded goroutine leak holding a network stream, identical in
consequence to #2132.

Verified priority **Medium**, one step below #2132. #2132 triggers on *any* `Recv` or
`Unmarshal` failure, which is ordinary operational noise. This defect additionally requires
`Respond()` to have already exited before a further request is decoded — either the
random-`select` timeout race, or a peer that pipelines requests. Real and worth fixing, but
materially harder to reach than #2132.

### Effort

**Low** — confined to `token/services/ttx/external.go`.

AkramBitar:
## The problem

`StreamExternalWalletSignerClient` starts a background goroutine. This goroutine reads signature requests from the stream and passes them to `Respond()`. It uses two channels to do this. On both channels, the goroutine could stop forever.

**1. Errors were lost (#2132).** The `err` channel had no buffer, and no code ever read from it. So when the stream failed, or when a request could not be decoded, the goroutine stopped while sending the error. The error never arrived. `Respond()` waited until its timeout (one hour by default), and then returned a timeout error. But the real cause was a broken stream, not a timeout.

**2. Late requests had no reader (#2141).** The `input` channel had the same problem. After `Respond()` returned, no code read from this channel. So if the goroutine decoded one more request, it stopped while passing it.

In both cases the goroutine never ends. It keeps the stream, the two channels and the signer in memory until the process stops. Each failed stream adds one more.

## The fix

- Give `err` a buffer of 1, and read it in `Respond()` with `case err := <-s.err:`. Now the goroutine can finish its send and end, and the caller gets the real error.
- `Respond()` closes a `done` channel when it stops reading. The goroutine uses `select` on `done` when it passes a request, so it can end instead of waiting.

The goroutine can now end at every point where it waits.

Fixes #2132
Fixes #2141

## Why one PR

Both problems are in the same goroutine: one blocked send on each of its two channels. Both fixes change the same `select` in `Respond()`. If we fix only #2132, the goroutine can still stop forever.

## Testing

The new file `token/services/ttx/external_internal_test.go` has 8 tests. If you undo the changes in `external.go`, 4 of them fail: two with `Respond did not return in time`, one on the channel buffer size, and one that finds the goroutine still waiting in `[chan send]`. So both fixes are needed. `make checks` passes, and `-race` passes.

**Not included:** the goroutine can still wait inside `stream.Recv`. This wait ends when the stream ends. We cannot stop it here, because the `view2.Stream` interface has only `Recv` and `Send` — no context and no `Close`. #2128 is not changed.

### DIFF
--- a/docs/services/ttx.md
+++ b/docs/services/ttx.md
@@ -513,6 +513,44 @@ The `CollectEndorsementsView` is responsible for gathering all signatures requir
 *   **Auditor Signatures**: If the TMS is configured with an auditor, the transaction must be approved via the `AuditApproveView`.
 *   **Network Endorsements**: The service delegates to the **Network Service** to obtain backend-specific endorsements (e.g., Fabric chaincode endorsements).
 
+### Streaming signer for external wallets (`external.go`)
+
+When the signing key for a wallet does not live on the node, `ttx.WithExternalWalletSigner(walletID, ews)`
+registers an `ExternalWalletSigner` (`Sign(party, message)` plus `Done()`) that
+`CollectEndorsementsView` calls instead of the local signature service. `external.go` provides an
+implementation of that interface on top of an FSC stream, so the wallet holder can be the
+application that invoked the view over `StreamCallView`:
+
+*   `StreamExternalWalletSignerServer` sits on the view's end of the stream. It is constructed
+    with the view's stream and passed to `WithExternalWalletSigner`; each `Sign` call sends a
+    `SigRequest` message and blocks for the matching `SignResponse`, and `Done` sends a `Done`
+    message once no further signatures are needed.
+*   `StreamExternalWalletSignerClient` sits on the application's end. It is constructed with a
+    `SignerProvider` over the local wallet, and its `Respond()` loop signs each incoming
+    `SigRequest` and writes back a `SignResponse` until the `Done` message arrives.
+
+All three message kinds travel inside a `StreamExternalWalletMsg` envelope (`Type` plus
+JSON-encoded `Raw`).
+
+`Respond()` returns `nil` once the peer signals `Done`. Otherwise it returns the failure that
+terminated the exchange:
+
+*   a signing failure (no signer for the party, or the signer itself failing) or a failure
+    sending the response back;
+*   a failure receiving or JSON-decoding a request. These are detected by the client's internal
+    receive goroutine and handed to `Respond()` over a buffered error channel, so the caller
+    gets the real transport or decoding error. The buffering matters: it lets that goroutine
+    complete its send and exit rather than blocking forever on a channel nobody reads;
+*   `Timeout waiting for stream input exceeded` if no message arrives within the configured
+    timeout. `NewStreamExternalWalletSignerClient` defaults to one hour; use
+    `NewStreamExternalWalletSignerClientWithTimeout` to choose another value. A timeout now means
+    what it says — the stream went quiet — rather than masking a stream error.
+
+However `Respond()` ends, it signals its receive goroutine to stop before returning, so a
+request decoded after the caller has given up is dropped rather than left waiting for a
+consumer that will never come. The client is therefore single-use: once `Respond()` has
+returned, no further signatures are served on that stream.
+
 ## Distribution and Ordering
 
 Once fully endorsed, the transaction metadata is distributed to all recipients so they can track and spend their new tokens. The initiator then uses the `OrderingView` to broadcast the transaction envelope to the network's ordering service.
--- a/token/services/ttx/external.go
+++ b/token/services/ttx/external.go
@@ -8,6 +8,7 @@ package ttx
 
 import (
 	"encoding/json"
+	"sync"
 	"time"
 
 	"github.com/LFDT-Panurus/panurus/token"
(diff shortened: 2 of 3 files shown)


emgapps (commented):
PR REVIEWER: NOT READY\n\nThe production fix looks correct: buffering err prevents the receive goroutine from blocking while reporting its terminal error, and closing done lets it abandon a signature request after Respond stops consuming. I found no security concern or apparent regression in those paths.\n\n[P2] external_internal_test.go:252: requireInitExited inspects a fixed-size process-wide goroutine dump for the method name. This can fail because an unrelated client instance is still running, or pass because the relevant stack was truncated. Since goroutine termination is the central regression being tested, please expose a per-client completion signal (for example, an initDone channel closed by defer in init) and wait on that instead.\n\nGitHub also currently reports this PR as non-mergeable, so the branch needs to be updated and conflicts resolved.\n\nCI is otherwise green, and the documented limitation around an indefinitely blocked stream.Recv is appropriately scoped out.

SaidAltury-ibm (approved):
LGTM

PR #2144, opened 2026-08-04
fix(idemix): reject out-of-range curve ids in audit info

### LINKED ISSUE #2078: identity/idemix: unguarded curve-ID index in mathlib during audit-info unmarshaling
### Summary

Idemix audit-info structures embed `mathlib` curve elements (`*math.G1`, `*math.Zr`) that are
JSON-unmarshalled directly from external byte streams. In the pinned version of `mathlib`, the
unmarshaling code indexes a fixed-size array of curves by an ID taken straight from the input, with
no bounds check — so a malformed or adversarial audit-info payload can cause an index-out-of-range
panic during deserialization, before any cryptographic validation runs.

### Where

`token/services/identity/idemix/crypto/audit.go` defines `AuditInfo` with `EidNymAuditData` and
`RhNymAuditData` fields of type `*csp.AttrNymAuditData` (from the `idemix` bccsp types package),
whose `Nym`, `Rand`, and `Attr` fields are `*math.G1`/`*math.Zr` and get populated by
`json.Unmarshal` on `FromBytes` (`audit.go:47-50`).

In the pinned `github.com/IBM/mathlib@v0.3.0` (`go.mod:8`, no `replace` directive), `marshaler.go`'s
`UnmarshalJSON` for these types does `Curves[z.curveID].NewZrFromBytes(...)` (and the `G1`/`G2`/`Gt`
equivalents) with no check that the decoded curve ID is within range of the `Curves` slice (length 8).
A curve ID outside that range causes an index panic during unmarshaling, i.e. before `Validate()` on
the containing struct gets a chance to run. This is the newest published version of `mathlib`
(confirmed via `go list -m -versions`), so there's no upgrade path — any fix needs to live at this
repo's boundary.

Worth noting: the two existing fuzz targets that touch this area seed with a zero-value
`&idemix.AttrNymAuditData{}`, which marshals these fields as JSON `null` — so the vulnerable
`UnmarshalJSON` path for a populated value is never exercised by the current seed corpus.

### Impact

An audit-info payload with an out-of-range curve ID panics during JSON unmarshaling rather than
being rejected by validation. Since audit info can originate from a counterparty (e.g. during
recipient registration or auditing flows), this is worth closing at the deserialization boundary
rather than relying on the upstream library, which has no available fix.

### Suggested fix

Add an explicit range check on the decoded curve ID immediately after unmarshaling
`AttrNymAuditData` (or its embedded fields) in this repo's own deserialization path, before passing
the value further into audit-info processing — rejecting anything outside `len(Curves)` with a
regular error rather than letting the library index out of bounds. It would also help to extend the
existing fuzz seed corpus (`f.Add(...)`) with non-null `AttrNymAuditData` values, including an
out-of-range curve ID, so this class of input is covered by the nightly fuzz run.

### Severity

HIGH — reachable via externally-sourced audit info, converts into a process-level panic during
deserialization.

AkramBitar:
Fixes #2078

### Problem

Idemix audit info is JSON from a counterparty. Its pseudonym audit data holds mathlib curve elements, which encode the curve as an index into mathlib's curve table:

```json
{"EidNymAuditData":{"Nym":{"curve":999999,"element":"AQID"}}}
```

mathlib's `UnmarshalJSON` does `Curves[ce.CurveID].NewG1FromBytes(...)` with no bounds check, so that payload crashes the process with `index out of range` — from *inside* `json.Unmarshal`, before any validation runs. `mathlib v0.3.0` is the latest release, so this has to be contained here.

Both deserializers were affected: `crypto.AuditInfo` and `nym.AuditInfo` (the latter embeds the former, so the curve elements inline into its own JSON and it decodes them itself).

### Fix

Both decodes now run through `crypto.UnmarshalAuditInfo`, which recovers the panic and returns it as an ordinary error — the same containment `FromG1Proto` already applies to this defect in `token/core/zkatdlog/nogh/protos-go/utils`.

The guard wraps the real decode rather than pre-validating the payload's curve IDs. That was the first attempt, and it was wrong: mathlib runs *during* `encoding/json`'s traversal, so a separate validation pass has to reproduce that traversal exactly. It didn't, five ways — trailing input after the first JSON value (found by the fuzz target), a duplicate key overwriting a bad element at either nesting level, an element following a type error, and case-insensitive key matching. All five are pinned in `TestDeserializeAuditInfoOutOfRangeCurveIDEvasions` and seeded into both fuzz corpora.

Those corpora previously seeded only zero-valued audit data, which marshals as JSON `null` and never entered the vulnerable path at all.

### DIFF
--- a/docs/services/identity.md
+++ b/docs/services/identity.md
@@ -296,6 +296,45 @@ An extension of Idemix that uses a **commitment to the Enrollment ID (EID)** as 
 | **Identity Size** | Large (~several KB) | Small (~32-64 bytes) |
 | **Storage Overhead** | High | Low |
 
+#### Audit Info Deserialization (Idemix and IdemixNym)
+
+Audit info is JSON and can arrive from a counterparty (recipient registration, auditing
+flows), so both `crypto.AuditInfo.FromBytes`
+(`token/services/identity/idemix/crypto/audit.go`) and `nym.AuditInfo.FromBytes`
+(`token/services/identity/idemixnym/nym/audit.go`) treat their input as untrusted and reject
+malformed payloads with an error.
+
+`EidNymAuditData` and `RhNymAuditData` embed `mathlib` curve elements, which JSON-encode as
+a curve ID plus the raw element bytes:
+
+```json
+{"EidNymAuditData":{"Nym":{"curve":3,"element":"..."},"Rand":{...},"Attr":{...}}}
+```
+
+`mathlib`'s `UnmarshalJSON` uses that curve ID to index its internal curve table **without a
+bounds check**, so an out-of-range ID raises an `index out of range` panic from inside
+`encoding/json`. Both `FromBytes` implementations therefore run their decode through
+`crypto.UnmarshalAuditInfo`, which recovers that panic and returns it as an ordinary error:
+
+```go
+return crypto.UnmarshalAuditInfo(func() error {
+    return json.Unmarshal(raw, a)
+})
+```
+
+The guard wraps the real decode rather than pre-validating the payload's curve IDs, because
+`mathlib` runs *during* `encoding/json`'s traversal: a separate validation pass has to
+reproduce that traversal exactly to see every curve element the decode reaches, including
+ones that never appear in the decoded result (a duplicate key overwriting an earlier value,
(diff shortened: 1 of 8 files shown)


SaidAltury-ibm (approved):
LGTM

PR #2146, opened 2026-08-05
fix(identity): derive the enrollment ID of a boolpolicy identity from its members

### LINKED ISSUE #2145: boolpolicy: policy identity AuditInfo always reports an empty enrollment ID
## Problem

The boolpolicy `AuditInfoDeserializer` never derives an enrollment ID: `AuditInfo.EnrollmentID()` returns `""` for every policy identity, even when all component identities belong to the same enrollment.

A policy-owned **input** therefore reaches the auditor with no enrollment attribution: the input row carries the composite identity itself, since both drivers put a single sender in `TransferInputMetadata`.

A policy-owned **output** is affected differently. Output rows are expanded one per component, and each row resolves its own member's enrollment ID, so outputs are attributed — but the same amount is counted once per member. That duplicate amount accounting is a separate defect, tracked in #2148.

## Concrete impact

The damage is clearest in the common case where **every component of the policy identity belongs to the same enrollment ID** — one wallet whose owner policy holds several keys of the same customer.

Setup: policy wallet P with two components, both enrolled as `alice`, holds 100 and pays 20 to a regular wallet enrolled as `bob`, taking 80 back as change.

What the audit record should book: `alice` spent 100 and received 80 change (net −20); `bob` received 20.

What is actually booked:

1. **P's input is booked to another party.** The input reaches the auditor with enrollment ID `""`, and the append-time gap filler (`completeInputsWithEmptyEID` in `token/services/auditor/auditor.go`) assigns unattributed inputs to `record.Outputs.EnrollmentIDs()[0]` — the first output row carrying a non-empty enrollment ID, which is the recipient whenever the payment output precedes the change. `bob` is then recorded as having spent the 100 input he never owned: his movement for this transaction is 20 received − 100 spent = **−80**.
2. **P's change output is counted twice.** The 80 change is owned by the policy identity again, which expands into one row per component, each carrying enrollment ID `alice` and the full 80. `alice`'s movement is 160 received − 0 spent = **+160**, where the truth is −20: the received amount is doubled and the spend is missing entirely.

The same double counting applies when the policy wallet is on the receiving side: a deposit of 100 into P books as 200 received for `alice`.

Consequences: sender-side balance and transfer limits are evaluated against audited balances that never decrease on a spend and grow by a factor of the member count on a receive; other parties' balances drift downward by inputs they never owned; per-enrollment holdings for policy wallets are wrong for every transaction. This affects every policy wallet, including the single-member case, where the doubling factor is 1 and only the input attribution is lost.

(The gap filler's choice of fallback target is itself questionable — `EnrollmentIDs()[0]` depends on output ordering — and will be raised separately; with a correct enrollment ID derived here it simply stops being exercised for policy identities.)

## Expected

A policy identity whose components all resolve to one enrollment ID should report that enrollment ID — the same way `htlc.NewAuditDeserializer` resolves the audit info of a script's participants through the parent multiplex deserializer. Amounts should then be aggregated once per enrollment ID rather than once per component row.

## Proposal

- Resolve each component's audit info through the parent multiplex deserializer and report the enrollment ID shared by all components.
- Components legitimately spanning enrollments keep the current empty value (there is no common enrollment ID).
- Components with no enrollment ID of their own (e.g. a nested composite spanning enrollments) also keep the empty value: composites resolve recursively, so a member may itself be a policy or multisig identity.
- A component whose audit info is missing is legal — `GetAuditInfo` returns nil for an identity that is not registered locally — and keeps the empty value. This takes precedence over subtype resolution, so an unknown component identity type carrying no audit info also yields the empty value rather than an error.
- Malformed audit info — a non-empty component audit info that cannot be resolved, an invalid component identity, or a component count mismatch — should surface as an error instead of being indistinguishable from the legal cross-enrollment case.
- The per-component duplication of output amounts is out of scope here and is addressed by #2148; only both changes together restore the correct net movement.

I have a fix ready and will open a PR.

EvanYan1024:
Fixes #2145

## What

Derive the enrollment ID of a boolpolicy identity from its member identities, mirroring the recursive `htlc.NewAuditDeserializer` pattern:

- `boolpolicy.NewAuditInfoDeserializer` now takes the parent multiplex deserializer and resolves each component's audit info through it; the policy identity reports the enrollment ID shared by all components. The fabtoken and zkatdlog drivers pass the multiplexer at registration.
- Components with no enrollment ID of their own (e.g. a nested composite spanning enrollments) or disagreeing components yield the legacy empty value — composites resolve recursively, so a member may itself be a policy or multisig identity.
- A component whose audit info is missing is legal: `GetAuditInfo` returns nil for an identity that is not registered locally and `WrapAuditInfo` preserves the empty entry, so such a component contributes no enrollment ID rather than failing the audit path. This exemption is applied before subtype resolution, so a component of an unknown identity type carrying no audit info also yields no enrollment ID rather than an error.
- Malformed audit info — a non-empty component audit info that cannot be resolved, or a component count mismatch — is an error instead of a silent empty value, and every component is resolved before the result is declared so corruption in a later component is never masked by an earlier "no common EID" outcome.
- The wire-supplied component identities are validated at this deserialization boundary, as `GetAuditInfoMatcher` and `DeserializeVerifier` already do. Nothing else on the audit path reaches that validation: `Request.IsValid` only checks action structure and metadata consistency, and owner verifiers are deserialized solely by the driver validator at commit.
- A component whose inner deserializer returns neither audit info nor an error contributes no enrollment ID instead of panicking.

## Why

`AuditInfo.EnrollmentID()` returned `""` for every policy identity, and the two sides of a transaction are affected differently.

The input side carries the composite identity itself — both drivers put a single sender in `TransferInputMetadata` — so a policy-owned input reached the auditor with no enrollment attribution at all, and `completeInputsWithEmptyEID` then reassigned it to whichever enrollment ID the output stream reports first. That is the gap this PR closes.

The output side is separate: output rows are expanded per component and already resolve each member's own enrollment ID, so they are attributed correctly but counted once per member. That duplicate amount accounting is fixed by #2148. Only the two together bring a policy wallet's audited movements back to the true net; see #2145 for a worked example.

In our test environments the auditor's audit balances for policy wallets were corrupted by this pair of defects.

## Testing

- Unit tests cover: common-enrollment derivation, single member, legal cross-enrollment components, nested policies (both the common-EID and cross-enrollment inner case), empty member enrollment IDs, missing component audit info (alone, alongside a resolvable component, followed by a malformed component, and on an unknown identity type), invalid component identities (duplicate and empty), a nil audit info returned by the inner deserializer, the malformed cases (unresolvable component, count mismatch including an empty audit-info blob against a non-empty policy, a policy identity with no components, garbage audit info), the masked-later-corruption case, and the zero-value deserializer.
- `FuzzDeserializeAuditInfoNoPanic` fuzzes the recursive deserialization path (seeds: valid single/two-member policies in both identity encodings, empty, truncated, empty audit-info blob, count mismatch, unknown member type, 5-deep nesting; inputs capped at 64 KiB like the neighbouring identity targets), wired into `.github/workflows/nightly-fuzz.yml`; 20s local run clean at ~50k execs/sec.

### DIFF
--- a/token/services/identity/boolpolicy/deserializer.go
+++ b/token/services/identity/boolpolicy/deserializer.go
@@ -183,8 +183,8 @@ func (a *AuditInfoDeserializer) DeserializeAuditInfo(ctx context.Context, id dri
 
 // commonEnrollmentID returns the enrollment ID shared by all components:
 // "" when a component has none (e.g. a nested composite), its audit info is
-// missing, or they disagree; an error on unresolvable audit info or a
-// component count mismatch.
+// missing, or they disagree; an error on an invalid component identity,
+// unresolvable audit info or a component count mismatch.
 func (a *AuditInfoDeserializer) commonEnrollmentID(ctx context.Context, id driver.Identity, ei *AuditInfo) (string, error) {
 	if a.inner == nil {
 		return "", nil
@@ -192,6 +192,9 @@ func (a *AuditInfoDeserializer) commonEnrollmentID(ctx context.Context, id drive
 	pi := PolicyIdentity{}
 	if err := pi.Deserialize(id); err != nil {
 		return "", errors.Wrapf(err, "failed to deserialize policy identity")
+	}
+	if err := validateComponentIdentities(pi.Identities); err != nil {
+		return "", errors.Wrap(err, "invalid policy identity")
 	}
 	if len(pi.Identities) != len(ei.IdentityAuditInfos) {
 		return "", errors.Errorf("expected %d component audit infos but received %d",
@@ -212,6 +215,11 @@ func (a *AuditInfoDeserializer) commonEnrollmentID(ctx context.Context, id drive
 		memberAuditInfo, err := a.inner.DeserializeAuditInfo(ctx, pi.Identities[k], info.AuditInfo)
 		if err != nil {
 			return "", errors.Wrapf(err, "failed to deserialize audit info of component [%d]", k)
+		}
+		if memberAuditInfo == nil {
+			// an inner deserializer may report neither audit info nor error:
+			// treat it like missing audit info rather than panicking
+			continue
 		}
 		eids[k] = memberAuditInfo.EnrollmentID()
 	}
--- a/token/services/identity/boolpolicy/deserializer.go
+++ b/token/services/identity/boolpolicy/deserializer.go
@@ -182,8 +182,9 @@ func (a *AuditInfoDeserializer) DeserializeAuditInfo(ctx context.Context, id dri
 }
 
 // commonEnrollmentID returns the enrollment ID shared by all components:
-// "" when a component has none (e.g. a nested composite) or they disagree,
-// an error on unresolvable audit info or a component count mismatch.
+// "" when a component has none (e.g. a nested composite), its audit info is
+// missing, or they disagree; an error on unresolvable audit info or a
+// component count mismatch.
 func (a *AuditInfoDeserializer) commonEnrollmentID(ctx context.Context, id driver.Identity, ei *AuditInfo) (string, error) {
 	if a.inner == nil {
 		return "", nil
@@ -203,6 +204,11 @@ func (a *AuditInfoDeserializer) commonEnrollmentID(ctx context.Context, id drive
 	// later component is not masked by an earlier "no common EID" outcome
 	eids := make([]string, len(ei.IdentityAuditInfos))
 	for k, info := range ei.IdentityAuditInfos {
+		if len(info.AuditInfo) == 0 {
+			// missing audit info is legal (e.g. an identity not registered
+			// locally): the component contributes no enrollment ID
+			continue
+		}
 		memberAuditInfo, err := a.inner.DeserializeAuditInfo(ctx, pi.Identities[k], info.AuditInfo)
 		if err != nil {
 			return "", errors.Wrapf(err, "failed to deserialize audit info of component [%d]", k)
--- a/token/services/identity/boolpolicy/deserializer.go
+++ b/token/services/identity/boolpolicy/deserializer.go
@@ -154,13 +154,68 @@ func (d *TypedIdentityDeserializer) Recipients(id driver.Identity, typ identity.
 
 // AuditInfoDeserializer deserialises raw audit info bytes into the AuditInfo
 // struct for the enrollment-ID / revocation-handle path.
-type AuditInfoDeserializer struct{}
+// It derives the policy identity's enrollment ID from its components.
+type AuditInfoDeserializer struct {
+	inner driver2.AuditInfoDeserializer
+}
 
-func (a *AuditInfoDeserializer) DeserializeAuditInfo(_ context.Context, _ driver.Identity, raw []byte) (driver2.AuditInfo, error) {
+// NewAuditInfoDeserializer constructs an AuditInfoDeserializer resolving
+// per-component audit infos through inner, typically the parent multiplex
+// deserializer.
+func NewAuditInfoDeserializer(inner driver2.AuditInfoDeserializer) *AuditInfoDeserializer {
+	return &AuditInfoDeserializer{inner: inner}
+}
+
+// DeserializeAuditInfo decodes raw policy audit info and derives its enrollment ID.
+func (a *AuditInfoDeserializer) DeserializeAuditInfo(ctx context.Context, id driver.Identity, raw []byte) (driver2.AuditInfo, error) {
 	ei := &AuditInfo{}
 	if err := json.Unmarshal(raw, ei); err != nil {
 		return nil, err
 	}
+	eid, err := a.commonEnrollmentID(ctx, id, ei)
+	if err != nil {
+		return nil, errors.WithMessagef(err, "failed deriving policy enrollment ID")
+	}
+	ei.eid = eid
 
 	return ei, nil
+}
+
+// commonEnrollmentID returns the enrollment ID shared by all components:
+// "" when a component has none (e.g. a nested composite) or they disagree,
+// an error on unresolvable audit info or a component count mismatch.
+func (a *AuditInfoDeserializer) commonEnrollmentID(ctx context.Context, id driver.Identity, ei *AuditInfo) (string, error) {
+	if a.inner == nil {
+		return "", nil
+	}
+	pi := PolicyIdentity{}
+	if err := pi.Deserialize(id); err != nil {
+		return "", errors.Wrapf(err, "failed to deserialize policy identity")
+	}
+	if len(pi.Identities) != len(ei.IdentityAuditInfos) {
+		return "", errors.Errorf("expected %d component audit infos but received %d",
+			len(pi.Identities), len(ei.IdentityAuditInfos))
+	}
+	if len(pi.Identities) == 0 {
+		return "", errors.New("policy identity has no components")
+	}
+	// resolve every component before declaring a result so corruption in a
+	// later component is not masked by an earlier "no common EID" outcome
+	eids := make([]string, len(ei.IdentityAuditInfos))
+	for k, info := range ei.IdentityAuditInfos {
+		memberAuditInfo, err := a.inner.DeserializeAuditInfo(ctx, pi.Identities[k], info.AuditInfo)
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to deserialize audit info of component [%d]", k)
+		}
+		eids[k] = memberAuditInfo.EnrollmentID()
+	}
+	for _, memberEID := range eids {
+		if memberEID == "" || memberEID != eids[0] {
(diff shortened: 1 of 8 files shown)


token/services/identity/boolpolicy/deserializer.go
```diff
@@ -154,13 +154,74 @@ func (d *TypedIdentityDeserializer) Recipients(id driver.Identity, typ identity.
...
+	}
+	for _, memberEID := range eids {
+		if memberEID == "" || memberEID != eids[0] {
+			// no common EID: a member has none (e.g. a nested composite
+			// spanning enrollments) or members belong to different ones
+			return "", nil
+		}
+	}
+
+	return eids[0], nil
```
AkramBitar, line 234: I think this fixes less of the auditor balance corruption than the PR description claims, and I'd like a second opinion on my reading.

As far as I can tell the *output* side still expands a policy owner into one row per component, each at the full quantity: `token/request.go:1435` calls `inputsAndOutputs(ctx, true, false, false)`, so `noOutputForRecipient` is false and the output loop iterates `recipients`, which for a policy owner comes from `Deobfuscate` → `TypedIdentityDeserializer.Recipients` → the component list.

So for a `$0 AND $1` policy whose two members share enrollment `wallet-42` receiving 100, I believe `OutputStream.ByEnrollmentID("wallet-42").Sum()` is 200. Spending it back produces a single `Input` row (both drivers set `Senders` to a 1-element slice — `fabtoken/v1/transfer.go:97`, `zkatdlog/nogh/v1/transfer.go:202`) contributing 100, so `ttxdb.Movements` records `+100` where the true net is 0.

If that's right, the drift goes from `+2q` to `+q` per round trip rather than to 0, and either the output-side expansion needs deduplicating by EID too, or the PR body should be softened. An integration assertion on `HoldingsFilter.ByEnrollmentId` for a multi-member same-enrollment policy would settle it either way — I haven't run one, so I could be wrong about the output loop.

token/services/identity/boolpolicy/deserializer.go
```diff
@@ -154,13 +154,74 @@ func (d *TypedIdentityDeserializer) Recipients(id driver.Identity, typ identity.
...
+// component count mismatch.
+func (a *AuditInfoDeserializer) commonEnrollmentID(ctx context.Context, id driver.Identity, ei *AuditInfo) (string, error) {
+	if a.inner == nil {
+		return "", nil
+	}
+	pi := PolicyIdentity{}
+	if err := pi.Deserialize(id); err != nil {
+		return "", errors.Wrapf(err, "failed to deserialize policy identity")
+	}
+	if len(pi.Identities) != len(ei.IdentityAuditInfos) {
```
AkramBitar, line 199: I'm wondering whether this should call `validateComponentIdentities(pi.Identities)` here. It's a new deserialization boundary over wire-supplied `pi.Identities`, and both siblings in this file do validate — `GetAuditInfoMatcher` (line 99) and `DeserializeVerifier` (line 124). The doc comment on the function in `identity.go:170-178` also says validation "must also happen at the deserialization boundary to actually close the gap."

Concretely: an AND policy with the same identity in both component slots is rejected by `DeserializeVerifier` but I think accepted here, yielding a derived enrollment ID. That seems to matter most on the auditor path, since the auditor never runs the verifier as far as I can see (`integration/token/fungible/views/auditor.go` goes `Audit` → business checks → `NewAuditApproveView` → `auditdb.Append`, no `IsValid`), so `Movements`/`TransactionRecords` rows would get attributed for an identity that only gets rejected later at commit.

Low severity, but it looks like a one-liner.

token/services/identity/boolpolicy/deserializer.go
```diff
@@ -154,13 +154,74 @@ func (d *TypedIdentityDeserializer) Recipients(id driver.Identity, typ identity.
...
+		if len(info.AuditInfo) == 0 {
+			// missing audit info is legal (e.g. an identity not registered
+			// locally): the component contributes no enrollment ID
+			continue
+		}
+		memberAuditInfo, err := a.inner.DeserializeAuditInfo(ctx, pi.Identities[k], info.AuditInfo)
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to deserialize audit info of component [%d]", k)
+		}
+		eids[k] = memberAuditInfo.EnrollmentID()
```
AkramBitar, line 224: `memberAuditInfo` is dereferenced without a nil check. `a.inner` is the interface injected by the exported constructor, and the counterfeiter fake for `driver2.AuditInfoDeserializer` returns `(nil, nil)` when unstubbed, so I think `NewAuditInfoDeserializer(defaultMock).DeserializeAuditInfo(...)` panics rather than returning an error. The analogous recursive code in `htlc/deserializer.go:177` sidesteps it by returning `ai` without dereferencing.

Probably not reachable through the real multiplexer wiring, so this is mostly about not handing callers a panic — `if memberAuditInfo == nil { continue }` above this line would do it.

AkramBitar (commented):
@EvanYan1024 

Thanks for the fixes that you added the empty-component case reads correctly now, and the three new regression tests cover it.

I left three comments while reading the rest. The first is the only one I'd like to discuss with you to see if we need to resolve it before merging; the other two are minor.

Regards,
Akram

AkramBitar (changes requested):

token/services/identity/boolpolicy/deserializer.go
EvanYan1024: You're right, and thanks for working through the arithmetic — I traced both sides and your reading of the output loop holds.

One detail worth adding, because it explains why the drift is +2q rather than +q before this PR: the two sides resolve enrollment IDs from different identities. Output rows are per component, so each already gets a real EID from GetEIDAndRH(recipient, ...). The input row carries the composite identity itself (sender.Identity is the token's owner, and both drivers set Senders to a single element), so its EID comes from the policy identity — which is precisely what returned "" before this change. For a two-member same-enrollment policy round-tripping q, where the correct movement is 0:

- before this PR: received 2q, sent 0 (the input is booked under ""), movement +2q
- with this PR: received 2q, sent q, movement +q
- with #2148 as well: received q, sent q, movement 0

So this PR fixes the attribution half, and the amount half is #2148, which is already open: ttxdb/store.go applies UniquePerOutput() to the received sums and UniquePerInput() to the sent sum, collapsing member rows only where amounts aggregate while leaving the identity and revocation consumers seeing every row. Since HoldingsFilter sums MovementRecords, the ByEnrollmentId assertion you suggest is settled by the pair rather than by either PR alone.

That cross-reference is missing from the PR body here, which is what makes the claim read as broader than it is — I'll scope the body to the attribution gap and point at #2148 for the amount correction. Happy to add the integration assertion on HoldingsFilter.ByEnrollmentId for a multi-member same-enrollment policy in #2148 if you'd like it landed with that half.

token/services/identity/boolpolicy/deserializer.go
EvanYan1024: Agreed, added in f0ac58c3.

On reachability: Audit reaches AuditRecord via inputsAndOutputs(ctx, true, false, false), so verifyActions is false there. The Validate path doesn't close the gap either — integration/token/fungible/views/auditor.go:42 does call auditor.Validate(tx) before Audit, which reaches Request.IsValid, but IsValid only verifies actions structurally: VerifyTransfer checks amounts, types and action shape and never deserializes an owner verifier. Owner verifiers are deserialized solely by the validator, at token/core/fabtoken/v1/validator/validator_transfer.go:48, which runs at commit time. So validateComponentIdentities is genuinely unreachable on the auditor path, as you say.

Worth flagging one behavioural consequence: such an identity now fails at audit time rather than being attributed and rejected later at commit. That's the opposite direction from the empty-component change earlier in this PR, but the two cases differ — an empty component audit info is legal, whereas a duplicate or empty component identity is already rejected by WrapPolicyIdentity and DeserializeVerifier, so no legitimate identity is affected. Covered by TestPolicyEnrollmentIDInvalidComponentIdentities.

token/services/identity/boolpolicy/deserializer.go
EvanYan1024: Confirmed and fixed in f0ac58c3. I reproduced it with a fake returning (nil, nil): SIGSEGV at deserializer.go:216, so it is a panic rather than an error as you expected. Guarded with continue rather than an error, which keeps it consistent with the missing-audit-info case just above — the component simply contributes no enrollment ID. TestPolicyEnrollmentIDNilMemberAuditInfo covers it.

You're right that it isn't reachable through the real multiplexer: EIDRHDeserializer.DeserializeAuditInfo either errors or returns the leaf result, and none of the x509, idemix, multisig or boolpolicy leaves return (nil, nil). The guard is there because NewAuditInfoDeserializer takes an exported interface.

AkramBitar (approved):

PR #2148, opened 2026-08-05
fix(token): count composite-owner outputs once per enrollment ID in amount sums

### LINKED ISSUE #2147: Composite-owner outputs are counted once per member in eid-keyed amount sums
## Problem

`extractIssueOutputs`/`extractTransferOutputs` expand a composite owner (multisig, boolpolicy) into one audit Output row per member, each row carrying the full output amount. The eid-keyed amount aggregations in ttxdb — `TransactionRecords` and `Movements`, shared by the owner and auditor stores — sum over rows, so members sharing an enrollment ID multiply the amount by the member count.

## Concrete impact

A wallet whose composite owner holds two keys of the same enrollment `alice` receives a 100 deposit:

- the Outputs stream carries two rows `(alice, 100)`;
- `Movements` and `TransactionRecords` book `alice` **+200** received.

Every eid-keyed sum over outputs is inflated by the member count: a wallet with N same-enrollment keys books N× every amount it receives. In our test environments a 100 deposit into a two-key wallet showed up as a 200 transaction in the owner's transaction records.

## Constraint

Deduplicating at extraction time is not an option: the per-member rows are what identity consumers read — `RevocationHandles()` must surface a revoked second key and `ByRecipient()` must see every member.

## Proposal

Collapse rows sharing `(Index, EnrollmentID)` only where amounts aggregate: an `OutputStream.UniquePerOutput` filter applied by the ttxdb sums.

I have a fix ready and will open a PR.

EvanYan1024:
Fixes #2147

## What

- Add `OutputStream.UniquePerOutput`, keeping for each `(Index, EnrollmentID)` pair only the first output, and apply it in the ttxdb `TransactionRecords` and `Movements` sums (shared by the owner and auditor stores).
- Add `InputStream.UniquePerInput`, keeping for each `(token ID value, EnrollmentID)` pair only the first input — rows with no token ID are all kept — and apply it in the `Movements` sent sum. The input side has the same shape (raised in review below): `extractTransferInputs` emits one row per member with the same token ID and `AuditRecord` fills every row with the token's full quantity.

## Why

A composite owner (multisig, boolpolicy) is expanded into one audit row per member on both sides of a transaction, each row carrying the full amount. Members sharing an enrollment ID therefore multiply the amount by the member count in every eid-keyed sum: in our test environments a 100 deposit into a wallet whose owner held two keys of the same enrollment booked as 200 received, and review reproduced the symmetric sent side — a 40 spend from a two-member wallet booked as -80.

Deduplicating at extraction time instead would drop data that identity consumers need — `RevocationHandles()` must surface a revoked second key and `ByRecipient()` must see every member — so the collapse happens only where amounts aggregate.

## Testing

- `token/request_composite_output_test.go`: composite owners with same-enrollment and cross-enrollment members, issue and transfer paths; asserts identity consumers still see every member row.
- `token/services/storage/ttxdb/store_test.go`: `TransactionRecords`/`Movements` book a composite-owner output once per enrollment ID; the composite-spend fixture expands the spent input into two member rows sharing one token ID, pinning the movement to -6 instead of -46.
- `token/stream_test.go`: table-driven contract tests for `UniquePerOutput` and `UniquePerInput` — the same `(Index, EnrollmentID)` / `(token ID, EnrollmentID)` pair collapses to the first row (for inputs via distinct pointers to the same token ID value); the same index or token ID with different enrollment IDs, and different indexes or token IDs with the same enrollment ID, all survive; inputs with no token ID are all kept.

## Docs

`docs/services/storage/ttxdb.md` gains a "Composite Owners" note: one audit row per member for identity/revocation visibility, amount aggregations count each `(output index, enrollment ID)` / `(token ID, enrollment ID)` pair once.

### DIFF
--- a/token/stream.go
+++ b/token/stream.go
@@ -91,6 +91,27 @@ func (o *OutputStream) ByType(typ token.Type) *OutputStream {
 	})
 }
 
+// UniquePerOutput returns a stream keeping, for each (Index, EnrollmentID)
+// pair, only the first output, so amount aggregation counts a composite
+// owner's members once. Identity consumers use the full stream instead.
+func (o *OutputStream) UniquePerOutput() *OutputStream {
+	type key struct {
+		index uint64
+		eID   string
+	}
+	seen := map[key]bool{}
+
+	return o.Filter(func(t *Output) bool {
+		k := key{index: t.Index, eID: t.EnrollmentID}
+		if seen[k] {
+			return false
+		}
+		seen[k] = true
+
+		return true
+	})
+}
+
 // IsRedeem returns true if this output is a redeem, i.e., it has no owner.
 func (o *Output) IsRedeem() bool {
 	return len(o.Owner) == 0
@@ -367,6 +388,31 @@ func (is *InputStream) ByEnrollmentID(id string) *InputStream {
 func (is *InputStream) ByType(tokenType token.Type) *InputStream {
 	return is.Filter(func(t *Input) bool {
 		return t.Type == tokenType
+	})
+}
+
+// UniquePerInput returns a stream keeping, for each (token ID, EnrollmentID)
+// pair, only the first input, so amount aggregation counts a composite
+// owner's members once. Inputs with no token ID are all kept. Identity
+// consumers use the full stream instead.
+func (is *InputStream) UniquePerInput() *InputStream {
+	type key struct {
+		id  token.ID
+		eID string
+	}
+	seen := map[key]bool{}
+
+	return is.Filter(func(t *Input) bool {
+		if t.Id == nil {
+			return true
+		}
+		k := key{id: *t.Id, eID: t.EnrollmentID}
+		if seen[k] {
+			return false
+		}
+		seen[k] = true
+
+		return true
 	})
 }
 
--- a/docs/services/storage/ttxdb.md
+++ b/docs/services/storage/ttxdb.md
@@ -89,3 +89,15 @@ The following example shows how to retrieve the total amount of transactions for
 		fmt.Printf("Transaction: %s\n", tx.ID())
 	}
 ```
+
+## Composite Owners
+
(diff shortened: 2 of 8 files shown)


AkramBitar (commented):
_Posted in error — retracted. Please disregard._

token/stream.go
```diff
@@ -371,6 +392,31 @@ func (is *InputStream) ByType(tokenType token.Type) *InputStream {
 }
 
 // Sum returns the sum of the quantities of the inputs.
```
AkramBitar: `UniquePerInput` was inserted between this comment and the `func` line it documents, so two things broke at once:

- `UniquePerInput`'s doc block now opens with "Sum returns the sum of the quantities of the inputs." — that first sentence is the synopsis `go doc`, pkg.go.dev and editor hovers show, so it describes a different function.
- `Sum` (line 420) is left with no Godoc at all, against the AGENTS.md rule that all exported functions have one.

Nothing in `.golangci.yml` catches it: the rule would be revive's `exported`, and while there is a full `revive:` settings block with `enable-all-rules: true`, `revive` is not in `linters.enable` — so that block is inert.

Comment move only, no code change:

```go
// UniquePerInput returns a stream keeping, for each (token ID, EnrollmentID)
// pair, only the first input, so amount aggregation counts a composite
// owner's members once. Inputs with no token ID are all kept. Identity
// consumers use the full stream instead.
func (is *InputStream) UniquePerInput() *InputStream {
	...
}

// Sum returns the sum of the quantities of the inputs.
func (is *InputStream) Sum() *big.Int {
```

AkramBitar (changes requested):

token/stream.go
EvanYan1024: Fixed, thanks

AkramBitar (approved):
LGTM

PR #2149, opened 2026-08-05
2047 - no fast fail on endorser MSP selection

### LINKED ISSUE #2047: network-driver: one unresolvable configured identity aborts all endorser selection (SelectEndorsersForMSPSets)
### Summary

`SelectEndorsersForMSPSets` eagerly resolves the MSP of *every* configured identity up front, before evaluating any candidate set. If even one configured identity fails to resolve, the whole call returns an error — even when the remaining, successfully-resolved identities would fully satisfy the required candidate MSP set on their own.

### Where

`token/services/network/fabric/endorsement/fsc/selection.go:29-36`:

```go
byMSP := make(map[string][]view.Identity)
for _, id := range configured {
    mspID, err := mspOf(id)
    if err != nil {
        return nil, errors.WithMessagef(err, "failed to resolve MSP for configured endorser [%s]", id)  // selection.go:33
    }
    byMSP[mspID] = append(byMSP[mspID], id)
}
```

### Impact

An operator adds a stale/decommissioned endorser identity to the configured list (or an MSP config update temporarily makes one identity unresolvable). Endorser selection for *every namespace and every endorsement request* now fails outright, even though the org's live, correctly-configured endorsers are more than sufficient to satisfy the policy — a single bad config entry produces total endorsement unavailability instead of graceful degradation.

### Reproduction

```go
configured := []view.Identity{org1Endorser1, org2Endorser1, staleEndorser}
_, err := fsc.SelectEndorsersForMSPSets(configured, mspOf, [][]string{{"Org1MSP", "Org2MSP"}})
require.Error(t, err)  // despite org1Endorser1 + org2Endorser1 alone satisfying the candidate
```

### Suggested fix

Skip (and log) individual identities that fail MSP resolution instead of aborting the whole call, so selection can still succeed on the remaining, resolvable identities when they satisfy the candidate set.

### Severity

HIGH — single bad config entry causes total endorsement unavailability.

Effi-S:
Fixes #2047

## Summary

`SelectEndorsersForMSPSets` resolved the MSP of **every** configured endorser
identity up front, and a single resolution failure aborted the whole function.
Because this runs on the submit path of every token transaction with the full
configured endorser list, one unresolvable/stale entry made endorser selection
fail for **every namespace and every endorsement request** — total endorsement
unavailability instead of graceful degradation. Rated HIGH.

## Fix

A broken identity is now **skipped and logged**, not fatal. We only need enough
healthy endorsers to cover one candidate MSP set:

Phase 2 is unchanged. Its coverage error is now the **only** failure path, and
it fires only when the surviving healthy endorsers genuinely cannot satisfy any
candidate MSP set.

## Tests

- Added a regression subtest in `selection_test.go`: with a stale endorser
  present alongside a sufficient set (`org1Endorser1 + org2Endorser1`),
  selection now succeeds; with the stale endorser but an insufficient set, it
  fails with the coverage error rather than a resolve error.

Run: `go test ./token/services/network/fabric/endorsement/fsc/...`

## Files changed

- `token/services/network/fabric/endorsement/fsc/selection.go`
- `token/services/network/fabric/endorsement/fsc/selection_test.go

### DIFF
--- a/token/services/network/fabric/endorsement/fsc/selection.go
+++ b/token/services/network/fabric/endorsement/fsc/selection.go
@@ -25,13 +25,17 @@ func SelectEndorsersForMSPSets(configured []view.Identity, mspOf func(view.Ident
 	if len(candidates) == 0 {
 		return nil, errors.Errorf("no candidate MSP set to satisfy the namespace endorsement policy")
 	}
-
+	var skipped []error
 	byMSP := make(map[string][]view.Identity)
 	for _, id := range configured {
 		mspID, err := mspOf(id)
 		if err != nil {
-			return nil, errors.WithMessagef(err, "failed to resolve MSP for configured endorser [%s]", id)
+			skipped = append(skipped, err)
+			logger.Warnf("skipping unresolvable MSP for configured endorser [%s]: %v", id, err)
+
+			continue
 		}
+
 		byMSP[mspID] = append(byMSP[mspID], id)
 	}
 
@@ -53,5 +57,7 @@ func SelectEndorsersForMSPSets(configured []view.Identity, mspOf func(view.Ident
 		}
 	}
 
-	return nil, errors.Errorf("no configured endorser covers any of the [%d] policy-satisfying MSP set(s)", len(candidates))
+	return nil, errors.Join(errors.Errorf("no configured endorser covers any of the [%d] policy-satisfying MSP set(s)", len(candidates)),
+		errors.Join(skipped...),
+		errors.Errorf("failed to resolve MSP"))
 }
--- a/token/services/network/fabric/endorsement/fsc/selection_test.go
+++ b/token/services/network/fabric/endorsement/fsc/selection_test.go
@@ -21,6 +21,7 @@ func TestSelectEndorsersForMSPSets(t *testing.T) {
 	org1Endorser2 := view.Identity("org1-endorser2")
 	org2Endorser1 := view.Identity("org2-endorser1")
 	org3Endorser1 := view.Identity("org3-endorser1")
+	staleEndorser := view.Identity("stale-endorser")
 
 	mspOf := func(id view.Identity) (string, error) {
 		switch id.String() {
@@ -98,6 +99,18 @@ func TestSelectEndorsersForMSPSets(t *testing.T) {
 
 		require.Error(t, err)
 		assert.Contains(t, err.Error(), "no candidate MSP set")
+	})
+
+	t.Run("unsatisfiable - stale endorser check", func(t *testing.T) {
+		configured := []view.Identity{org1Endorser1, staleEndorser}
+		_, err := fsc.SelectEndorsersForMSPSets(configured, mspOf, [][]string{{"Org1MSP", "Org2MSP"}})
+		require.Error(t, err)
+	})
+
+	t.Run("satisfiable - stale endorser skipped check", func(t *testing.T) {
+		configured := []view.Identity{org1Endorser1, org2Endorser1, staleEndorser}
+		_, err := fsc.SelectEndorsersForMSPSets(configured, mspOf, [][]string{{"Org1MSP", "Org2MSP"}})
+		require.NoError(t, err)
 	})
 
 	t.Run("mspOf error surfaces", func(t *testing.T) {


token/services/network/fabric/endorsement/fsc/selection_test.go
```diff
@@ -30,6 +31,8 @@ func TestSelectEndorsersForMSPSets(t *testing.T) {
 			return "Org2MSP", nil
 		case org3Endorser1.String():
 			return "Org3MSP", nil
+		case staleEndorser.String():
+			return "StaleMSP", nil
```
AkramBitar: **This is what makes the new test vacuous — please delete these two lines.**

The bug is about endorsers whose MSP *fails* to resolve. Returning `"StaleMSP", nil` here means `staleEndorser` resolves fine — it's a perfectly healthy endorser that just sits in an MSP outside the candidate set. The new `continue` branch in `selection.go` is never reached.

The `default:` arm two lines below already returns an error for any unknown identity, which is exactly the unresolvable case needed. So just drop the case:

```suggestion
		default:
```

(i.e. remove `case staleEndorser.String(): return "StaleMSP", nil`). With it gone, `staleEndorser` falls through to `default`, the skip path is genuinely covered, and the subtest below fails on `main` and passes here — which is the point of a regression test.

token/services/network/fabric/endorsement/fsc/selection_test.go
```diff
@@ -99,6 +102,15 @@ func TestSelectEndorsersForMSPSets(t *testing.T) {
...
 	})
+	t.Run("stale endorser check", func(t *testing.T) {
+		configured := []view.Identity{org1Endorser1, staleEndorser}
+		_, err := fsc.SelectEndorsersForMSPSets(configured, mspOf, [][]string{{"Org1MSP", "Org2MSP"}})
+		require.Error(t, err)
+
+		configured = []view.Identity{org1Endorser1, org2Endorser1, staleEndorser}
+		_, err = fsc.SelectEndorsersForMSPSets(configured, mspOf, [][]string{{"Org1MSP", "Org2MSP"}})
+		require.NoError(t, err)
+	})
```
AkramBitar: **This subtest passes against unpatched `main`** — verified by reverting only `selection.go` to `origin/main`. Both assertions pass for reasons unrelated to the fix:

- Case 1 errors because **`Org2MSP` is uncovered**, not because anything was skipped.
- Case 2 succeeds because all three identities resolve and coverage is met.

Once the `StaleMSP` case above is removed this actually tests the fix. Two smaller things while you're here — please split the two scenarios and assert the message like every sibling subtest does, so a failure says which half broke and can't silently start passing/failing for a different reason:

```go
t.Run("unresolvable endorser skipped when the rest suffice", func(t *testing.T) {
	configured := []view.Identity{org1Endorser1, org2Endorser1, staleEndorser}

	selected, err := fsc.SelectEndorsersForMSPSets(configured, mspOf, [][]string{{"Org1MSP", "Org2MSP"}})

	require.NoError(t, err)
	assert.ElementsMatch(t, []string{"Org1MSP", "Org2MSP"}, mspIDsOf(t, mspOf, selected))
})

t.Run("unresolvable endorser skipped, remainder insufficient", func(t *testing.T) {
	configured := []view.Identity{org1Endorser1, staleEndorser}

	_, err := fsc.SelectEndorsersForMSPSets(configured, mspOf, [][]string{{"Org1MSP", "Org2MSP"}})

	require.Error(t, err)
	assert.Contains(t, err.Error(), "no configured endorser covers")
})
```

Also missing a blank line before `t.Run("stale endorser check"` — every other subtest in this file is separated by one.

token/services/network/fabric/endorsement/fsc/selection_test.go
```diff
@@ -99,6 +102,15 @@ func TestSelectEndorsersForMSPSets(t *testing.T) {
...
 
 	t.Run("mspOf error surfaces", func(t *testing.T) {
```
AkramBitar, line 116: **This subtest is what breaks CI** — line 123 below asserts `"failed to resolve MSP"`, the exact string this PR removes from `selection.go`:

```
selection_test.go:123:
    Error: "no configured endorser covers any of the [1] policy-satisfying MSP set(s)"
           does not contain "failed to resolve MSP"
--- FAIL: TestSelectEndorsersForMSPSets/mspOf_error_surfaces
```

It encodes the *old* fail-fast contract, so it needs rewriting rather than deleting — the all-unresolvable case is still worth pinning, just with the new expected outcome (skip everything, then fall through to the coverage error). Also worth renaming, since nothing "surfaces" any more:

```go
t.Run("all identities unresolvable - falls through to coverage error", func(t *testing.T) {
	badMspOf := func(view.Identity) (string, error) { return "", errors.Errorf("boom") }

	_, err := fsc.SelectEndorsersForMSPSets([]view.Identity{org1Endorser1}, badMspOf, [][]string{{"Org1MSP"}})

	require.Error(t, err)
	assert.Contains(t, err.Error(), "no configured endorser covers")
})
```

AkramBitar (commented):
The one-line fix in `selection.go` is correct and matches what #2047 prescribed. Two problems in the tests though, both from the same root cause: `mspOf` is the fixture that decides whether an identity resolves, and `staleEndorser` was made to resolve **successfully** — so the fixture can no longer express the bug being fixed.

1. **CI is red.** `utest (unit-tests-regression)` fails: the pre-existing `mspOf error surfaces` subtest asserts the `"failed to resolve MSP"` string that this PR deletes.
2. **The new test passes without the fix.** I reverted only `selection.go` to `origin/main` and re-ran it: `--- PASS: stale_endorser_check`. It never reaches the new `continue` branch.

Details inline. Nothing wrong with the production change itself — it fails closed, since phase 2 still requires full coverage of a candidate set, so skipping an endorser can't produce an under-endorsed selection.

token/services/network/fabric/endorsement/fsc/selection.go
```diff
@@ -30,7 +30,9 @@ func SelectEndorsersForMSPSets(configured []view.Identity, mspOf func(view.Ident
 	for _, id := range configured {
 		mspID, err := mspOf(id)
 		if err != nil {
-			return nil, errors.WithMessagef(err, "failed to resolve MSP for configured endorser [%s]", id)
+			logger.Warnf("skipping unresolvable cMSP for configured endorser [%s]: %v", id, err)
```
AkramBitar: The change itself is right. Three non-blocking notes on this line:

1. **Typo**: `cMSP` — stray `c`. Suggest `"skipping configured endorser [%s]: unresolvable MSP: %v"`.
2. **Root cause is lost from the error.** When *all* endorsers fail to resolve, the caller gets `"no configured endorser covers any of the [1] policy-satisfying MSP set(s)"`, which reads as a policy/config mismatch and points an operator at the wrong thing. The real cause (MSP resolution failing, e.g. mid-MSP-update) survives only in a warn log that may be filtered or far from the returned error — and this is a submit-path failure people will triage from the error string. Consider carrying the skips through with `errors.Join` (already imported, and the helper AGENTS.md mandates):

```go
var skipped []error
...
	if err != nil {
		skipped = append(skipped, err)
		continue
	}
...
return nil, errors.Join(
	errors.Errorf("no configured endorser covers any of the [%d] policy-satisfying MSP set(s) ([%d] skipped: unresolvable MSP)", len(candidates), len(skipped)),
	errors.Join(skipped...),
)
```

3. **Unbounded warn logging on a hot path.** A permanently stale config entry now logs once per configured endorser, per endorsement request, per transaction — sustained volume for a condition that's static until an operator intervenes. Either drop the per-identity line to `Debugf`, or log once per call with an aggregated count (pairs naturally with the `skipped` slice above).

While this file is open: the godoc still documents only the coverage error. The skip-and-log behaviour is a contract change for both callers (`fabric/endorsement/nspolicy.go:97` and `fabricx/endorsement/nspolicy.go:90`) and belongs in the comment — e.g. *"Configured identities whose MSP cannot be resolved are skipped and logged rather than treated as fatal."*

token/services/network/fabric/endorsement/fsc/selection_test.go
Effi-S: OK

token/services/network/fabric/endorsement/fsc/selection.go
Effi-S: Good point. Done

Effi-S: Done

AkramBitar (approved):
LGTM

PR #2150, opened 2026-08-05
test: stop two timing-dependent tests from flaking

### LINKED ISSUE #2121: Flaky unit tests fail utest (unit-tests-race) on unrelated PRs
## Problem

Two unit tests fail intermittently in `utest (unit-tests-race)` on PRs that do not touch the code
they exercise. Both have shown up in the last week on unrelated changes:

- `TestSufficientTokensBigDenominationsManyReplicas` failed on #2098, a PR that only wraps long lines
- `TestNextDelay_MaxDelayCap` failed on #2085, a PR that only touches recovery lock ids

Both pass locally, including under `-race`.

## Impact

A red `utest` on an unrelated PR costs every contributor the same work: pull the job log, find the
assertion, check whether their diff could plausibly have caused it, and re-run tests locally to
convince themselves. It also trains people to assume red CI is noise, which is the expensive part.

## 1. `TestSufficientTokensBigDenominationsManyReplicas`

`token/services/selector/sherdlock/manager_test.go:53`, body in
`token/services/selector/testutils/test_cases.go:67`.

The setup is 3 replicas, 2 tokens of CHF150, and each replica selecting 100 times, with
`startManagers(t, 3, 2*time.Second, 10)` giving a budget of 10 retries. The assertion is
`assert.Empty(t, errs)`, so it requires that **no** replica ever exhausts its retries.

With 3 replicas contending for 2 tokens, whether a given replica gets a token within 10 retries
depends on scheduling. Under `-race` on a loaded runner it does not always, and the run fails with:

```
Should be empty, but was [aborted too many times and no other process unlocked or added tokens: insufficient funds]
```

The funds are sufficient in total, so this is contention timing rather than a real insufficiency.
The retry budget is a timing assumption, not an invariant, but the assertion treats it as one.

## 2. `TestNextDelay_MaxDelayCap`

`token/services/utils/retry_test.go:291`.

It measures real wall-clock gaps between retries and asserts each one roughly doubles, with
`assert.InDelta(t, 2.0, ratio, 1.0)`, so a ratio outside 1.0 to 3.0 fails. The early delays are
10ms and 20ms, where ordinary scheduling jitter is a large fraction of the interval. The observed
failure:

```
delay 0 (10.185452ms) should roughly double to delay 1 (44.320386ms)   ratio 4.35
delay 1 (44.320386ms) should roughly double to delay 2 (40.17014ms)    ratio 0.91
```

A ~34ms late wake-up on a 10ms sleep is unremarkable on a busy CI runner. The test comment already
says the tolerance exists "to account for CI scheduling jitter", so the intent is there, it just is
not wide enough at the small end.

Worth noting separately: this test also takes about 131s, since it walks the backoff up to the 30s
cap 15 times. That is a meaningful slice of `unit-tests-race` for one assertion about a ratio.

## Possible directions

Not prescribing, but for discussion:

- For the selector test, either assert on the outcome that matters (all requested value eventually
  selected) rather than on no replica ever retrying out, or give the retry budget enough headroom
  that scheduling cannot decide the result.
- For the backoff test, either check the cap and the growth direction rather than per-step ratios,
  or drive it with an injected clock so it asserts on the computed delays instead of measured
  sleeps. The latter would also remove most of the 131s.

atharrva01:
Fixes #2121.

### `TestNextDelay_MaxDelayCap`

It measured wall clock gaps between retries and asserted each one roughly doubled. The first gaps
are 10ms and 20ms, where scheduler jitter on a loaded runner is a large fraction of the interval, so
a ~34ms late wake up was enough to fail it:

```
delay 0 (10.185452ms) should roughly double to delay 1 (44.320386ms)   ratio 4.35
```

`nextDelay` is a pure function though, so the policy can be asserted directly rather than inferred
from how long the runner actually slept. The new `retry_internal_test.go` covers exact exponential
growth, the 30s cap, the no-backoff case, the jitter band and the non-negative guard. With
`jitterFactor` at 0 the values are exact, so there is nothing left to be flaky about.

It also runs in **0.086s** against 131s for the test it replaces. Worth being clear that this does
not make the package fast: `token/services/utils` still takes around 378s, dominated by other
sleep-based tests I have not touched here.

### `TestSufficientTokensBigDenominationsManyReplicas`

It asserted `assert.Empty(t, errs)`, so it required that no replica ever exhausts its retry budget.
That is a scheduling outcome, not an invariant: three replicas compete for the same two tokens, so a
replica can legitimately lose several races before it wins one.

The budgets look like an oversight rather than a deliberate choice:

| test | replicas | retries |
| --- | --- | --- |
| `...BigDenominationsOneReplica` | 1 | 20 |
| `...BigDenominationsManyReplicas` | 3 | 10 |

Three times the contention on half the budget. Raised to 60 so contention cannot decide the result.
The assertion is unchanged, so the test still fails if selection genuinely cannot be satisfied.

Ran the selector test with `-count=3 -race` and the new retry tests repeatedly, all green.

### `TestGetRandomBytes_Success/single_byte`

Added after the fact, it turned up in the CI on #1982. The table includes a one byte case and every
non-zero length asserts the buffer is not all zeros. The comment says that is "extremely unlikely
with crypto/rand", which holds at 16 bytes and up, but at one byte all-zeros just means the byte is
0, so it fails about 1 run in 256.

Now only asserted once the buffer is wide enough for an all-zero draw to be implausible.
`TestGetRandomBytes_Uniqueness` already covers whether the output is actually random.

### DIFF
--- a/token/services/utils/retry_internal_test.go
+++ b/token/services/utils/retry_internal_test.go
@@ -1,0 +1,137 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package utils
+
+import (
+	"testing"
+	"time"
+
+	"github.com/LFDT-Panurus/panurus/token/services/logging"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// internalTestLogger mirrors the helper in the external test package, which this file cannot
+// reach from inside package utils.
+func internalTestLogger() logging.Logger {
+	return logging.DriverLogger("test", "n1", "c1", "ns1")
+}
+
+// nextDelay is the whole of the backoff policy, so it can be checked directly rather than by
+// measuring how long RunWithContext actually sleeps. Asserting on measured sleeps made this
+// flaky: at a 10ms initial delay, ordinary scheduler jitter on a loaded CI runner is a large
+// fraction of the interval, and it cost around 131s per run to walk the backoff to the cap.
+
+// With the default constructor the jitter factor is zero, so every delay is exact.
+func TestNextDelay_ExponentialUntilCapped(t *testing.T) {
+	r := NewRetryRunner(internalTestLogger(), 15, 10*time.Millisecond, true)
+
+	// 10ms doubling: the 13th attempt (index 12) is the first that would exceed the 30s cap.
+	for attempt := range 12 {
+		want := 10 * time.Millisecond << attempt
+		require.Less(t, want, defaultMaxDelay, "attempt %d should still be under the cap", attempt)
+		assert.Equal(t, want, r.nextDelay(attempt), "attempt %d", attempt)
+	}
+
+	for _, attempt := range []int{12, 13, 20, 100} {
+		assert.Equal(t, defaultMaxDelay, r.nextDelay(attempt),
+			"attempt %d must be capped at %s", attempt, defaultMaxDelay)
+	}
+}
+
+// Without exponential backoff every attempt waits the initial delay.
+func TestNextDelay_ConstantWhenBackoffDisabled(t *testing.T) {
+	r := NewRetryRunner(internalTestLogger(), 5, 25*time.Millisecond, false)
+
+	for _, attempt := range []int{0, 1, 7, 50} {
+		assert.Equal(t, 25*time.Millisecond, r.nextDelay(attempt), "attempt %d", attempt)
+	}
+}
+
+// A custom cap is honoured, and growth stops there rather than at the default.
+func TestNextDelay_RespectsCustomMaxDelay(t *testing.T) {
+	r := NewRetryRunnerWithJitter(internalTestLogger(), 10, 100*time.Millisecond, time.Second, 2.0, 0.0)
+
+	assert.Equal(t, 100*time.Millisecond, r.nextDelay(0))
+	assert.Equal(t, 200*time.Millisecond, r.nextDelay(1))
+	assert.Equal(t, 400*time.Millisecond, r.nextDelay(2))
+	assert.Equal(t, 800*time.Millisecond, r.nextDelay(3))
+	// 1.6s would exceed the 1s cap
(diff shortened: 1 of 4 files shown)


token/services/utils/retry_internal_test.go
AkramBitar, line 1: `TestNextDelay_NeverNegative` can't reach the branch it targets. `NewRetryRunnerWithJitter` clamps `jitterFactor` to <= 1.0, and `nextDelay` applies `(1 +/- jf/2)`, so the floor is 0.5x base — I ran 200k draws at jf=1.0, 2.0 and 3.0 and the `finalDelay < 0` branch never fired, so the assertion passes trivially.

Since this file is inside package `utils`, building the struct directly bypasses the clamp and does hit it (16.7% of draws):

```go
r := &retryRunner{initialDelay: 10 * time.Millisecond, maxDelay: time.Second,
	expBackoff: true, backoffMultiplier: 2.0, jitterFactor: 3.0, ...}
```

Worth doing that, or renaming to what it actually asserts? What do you think?

AkramBitar (changes requested):

token/services/utils/retry_internal_test.go
atharrva01: good catch, you're right and i checked it: through the constructor the factor comes back clamped to 1.00 every time, and 600k draws across jf=1.0, 2.0 and 3.0 gave zero negatives. so the assertion was passing on a branch it could never reach.

split it in two rather than just renaming:

- `TestNextDelay_ConstructorClampsJitter` pins the clamp itself, since that is the actual reason the guard is unreachable from outside and worth stating.
- `TestNextDelay_FallsBackWhenJitterWouldGoNegative` builds the struct directly with `jitterFactor: 3.0` as you suggested, and asserts both that nothing comes back non-positive and that the guard fired often enough to have really been exercised.

i got 16.6% fallbacks at jf=3.0, matching your 16.7%. also put the factor back to 1.0 to check the new test fails there, it does, 0/20000, so it isn't passing trivially this time.

AkramBitar (approved):
LGTM

PR #2151, opened 2026-08-05
test(fungible): await co-owner commit in CheckCoOwnedBalance

### LINKED ISSUE #2152: TestMultiSig (dlog-fabric-t12) flakes: CheckCoOwnedBalanceForTMSID races against co-owner block delivery
TestMultiSig (suite dlog-fabric-t12) fails intermittently with:

[50]!=[&{0}] ... CheckCoOwnedBalanceForTMSID support.go:354

CheckCoOwnedBalanceForTMSID issues a single, one-shot query against the co-owner node with no retry. This races against the co-owner's block delivery pipeline.

The sender's MultiSigLockView returns as soon as its own NewOrderingAndFinalityView sees the transaction committed. The co-owners (bob, charlie) observe the same block through their own independent delivery pipelines and complete their AcceptCashView asynchronously. In the failing run the block committed on the peers at 13:03:19.197, the sender returned, and the one-shot co-owned balance query ran at 13:03:19.442 — before bob had applied the block — so it read 0 instead of 50.

By contrast, the sibling helpers CheckBalanceForTMSID and CheckHoldingForTMSID both wrap the assertion in gomega.Eventually to tolerate exactly this asynchrony. CheckCoOwnedBalanceForTMSID was the only one that did not.

Impact
Sporadic CI failures in the dlog-fabric-t12 integration-test suite, specifically in the TestMultiSig test case. The failure is a false negative — the token state is correct, but the assertion fires before the co-owner has applied the block.

Expected behaviour
CheckCoOwnedBalanceForTMSID should retry the balance query with the same eventualCheckTimeout / eventualCheckPolling values used by the other balance helpers in support.go, matching the eventual-consistency model of the delivery pipeline.

HayimShaul:
TestMultiSig (dlog-fabric-t12) failed with

  [50]!=[&{0}] ... CheckCoOwnedBalanceForTMSID support.go:354

CheckCoOwnedBalanceForTMSID queried the co-owner once, with no retry, unlike CheckBalanceForTMSID and CheckHoldingForTMSID which wrap the same assertion in gomega.Eventually. The sender's MultiSigLockView returns as soon as its own NewOrderingAndFinalityView sees the commit; bob and charlie observe the same block through their own delivery pipelines and finish their AcceptCashView independently. In the failing run the block committed on the peers at 13:03:19.197, the sender returned, and the one-shot query for bob's co-owned balance ran at 13:03:19.442 before bob had applied the block, so it read 0 instead of 50.

Wrap the check in gomega.Eventually with the shared eventualCheckTimeout / eventualCheckPolling, matching the other balance checks in this file.

### DIFF
--- a/integration/token/fungible/support.go
+++ b/integration/token/fungible/support.go
@@ -379,20 +379,26 @@ func CheckCoOwnedBalance(network *integration.Infrastructure, ref *token3.NodeRe
 	CheckCoOwnedBalanceForTMSID(network, ref, wallet, typ, expected, nil)
 }
 
+// CheckCoOwnedBalanceForTMSID asserts that ref reports expected as the balance of the
+// tokens it co-owns. The check retries: the co-owners of a multisig transaction observe
+// its commit independently of the node that assembled and ordered it, so a one-shot query
+// right after the sender's view returns races with the co-owner's own commit.
 func CheckCoOwnedBalanceForTMSID(network *integration.Infrastructure, ref *token3.NodeReference, wallet string, typ token.Type, expected uint64, tmsID *token2.TMSID) {
-	res, err := network.Client(ref.ReplicaName()).CallView("CoOwnedBalance", common.JSONMarshall(&views.CoOwnedBalanceQuery{
-		Wallet: wallet,
-		Type:   typ,
-		TMSID:  tmsID,
-	}))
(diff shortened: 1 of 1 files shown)


AkramBitar (commented):
LGTM

AkramBitar (approved):

PR #2153, opened 2026-08-05
Guard queryByID against oversized peer responses to prevent node crash

### LINKED ISSUE #2055: network-driver: lookup queryByID panics the goroutine (crashing the process) on an oversized peer response
### Summary

`queryByID` unmarshals a single peer's raw chaincode-query response directly into `values`, then ranges over it while indexing the original `keys` slice by position — with nothing validating that the response returned exactly `len(keys)` elements. This function runs inside a goroutine with no `recover()` anywhere in the call chain, so an oversized response crashes the whole process, not just the request.

### Where

`token/services/network/fabric/lookup/deliveryqs.go:109-132`:

```go
values := make([][]byte, 0, len(keys))
err = json.Unmarshal(res, &values)
...
for i, value := range values {
    ...
    notFound = append(notFound, keys[i])   // deliveryqs.go — panics if len(values) > len(keys)
    ...
}
```

`res` is the raw response from a *single peer's* chaincode query (`ChannelStateQuerier.QueryStates` → `Channel.Chaincode(ns).Query(...).Query()`).

### Impact

`queryByID` runs inside a goroutine spawned by `QueryByID` (`go q.queryByID(...)`) with **no `recover()` anywhere in the call chain**. A byzantine, buggy, or simply out-of-sync peer that answers a `QueryStates` request with more elements than were requested for a given namespace drives `i` past the end of `keys`, panicking that goroutine — which is unrecovered and therefore **crashes the entire process**, not just the one request. This is a full-availability attack surface reachable by anything capable of influencing or replacing a single peer's chaincode-query response.

### Reproduction

Reproduced locally with a white-box unit test (not yet committed) calling the unexported `queryByID` synchronously (rather than through the goroutine-spawning `QueryByID`, since an unrecovered panic there would kill the whole test binary rather than just fail an assertion):

```go
raw, _ := json.Marshal([][]byte{[]byte("v1"), []byte("v2-unexpected")})  // 2 values, 1 requested key
q := &DeliveryScanQueryByID{Delivery: internalTestScanner{}, Querier: &internalTestQuerier{raw: raw}}
require.Panics(t, func() {
    q.queryByID(t.Context(), []driver.PKey{"k1"}, ch, 100, evicted)
})
```

### Suggested fix

Validate `len(values) == len(keys)` before the loop (returning/logging an error on mismatch instead of indexing blindly), and add a `recover()` around the goroutine body in `QueryByID` so any remaining panic degrades to a failed request rather than crashing the process.

### Severity

HIGH — unrecovered goroutine panic reachable via a single misbehaving/byzantine peer response, crashing the whole process.

Effi-S:
Fixes #2055

### Summary

`queryByID` unmarshals a single peer's raw chaincode-query response directly into `values`, then ranges over it while indexing the original `keys` slice by position — with nothing validating that the response returned exactly `len(keys)` elements. This function runs inside a goroutine with no `recover()` anywhere in the call chain, so an oversized response crashes the whole process, not just the request.

### Where

`token/services/network/fabric/lookup/deliveryqs.go:109-132`:

```go
values := make([][]byte, 0, len(keys))
err = json.Unmarshal(res, &values)
...
for i, value := range values {
    ...
    notFound = append(notFound, keys[i])   // deliveryqs.go — panics if len(values) > len(keys)
    ...
}
```

`res` is the raw response from a *single peer's* chaincode query (`ChannelStateQuerier.QueryStates` → `Channel.Chaincode(ns).Query(...).Query()`).

### Impact

`queryByID` runs inside a goroutine spawned by `QueryByID` (`go q.queryByID(...)`) with **no `recover()` anywhere in the call chain**. A byzantine, buggy, or simply out-of-sync peer that answers a `QueryStates` request with more elements than were requested for a given namespace drives `i` past the end of `keys`, panicking that goroutine — which is unrecovered and therefore **crashes the entire process**, not just the one request. This is a full-availability attack surface reachable by anything capable of influencing or replacing a single peer's chaincode-query response.

## Fix

Two independent, complementary defenses:

**A — Validate the response length before the loop.** Never trust the peer to return the right number of values. A mismatch is treated like the other failure cases already handled in this function (bad marshal, query error): log it and fall back to the slower block scan instead of trusting the response.

```go
if len(values) != len(keys) {
    logger.Errorf("peer returned %d values for %d keys in ns [%s]; falling back to block scan",
        len(values), len(keys), ns)
    startDelivery = true
    continue // treat as a per-namespace failure (=> fall back to the slow block scan)
}
```

**B — Wrap the goroutine body in `recover()` (defense in depth).** Even after Fix A, any future unforeseen panic in this background path must degrade to a failed request rather than crashing the node.

```go
go func() {
    defer func() {
        if r := recover(); r != nil {
            logger.Errorf("recovered from panic in queryByID: %v", r)
        }
    }()
    q.queryByID(ctx, keys, ch, startingBlock, evicted)
}()
```

## Tests

`TestQueryByID_OversizedResponse` in `deliveryqs_test.go` feeds a crafted response with more values (2) than keys requested (1). Before the fix this panics; after Fix A it delivers nothing for the oversized namespace and falls back to the block scan:

```go
querier := &fakeQuerier{results: map[driver.Namespace]querierResult{
    "ns1": {raw: values(t, []byte("v1"), []byte("v2-unexpected"))},
}}
scanner := &fakeScanner{}

ch, err := newQuery(querier, scanner).QueryByID(t.Context(), 100, evictedFor(map[driver.Namespace]driver.PKey{
    "ns1": "k1",
}))
require.NoError(t, err)
assert.Empty(t, drain(ch))
assert.True(t, scanner.called, "the oversized response must fall back to the block scan")
```

### DIFF
--- a/token/services/network/fabric/lookup/deliveryqs.go
+++ b/token/services/network/fabric/lookup/deliveryqs.go
@@ -66,6 +66,7 @@ func (q *DeliveryScanQueryByID) QueryByID(ctx context.Context, startingBlock dri
 	// Keys are supposed to be unique
 	keys := collections.Keys(evicted) // These are the state keys we are looking for
 	ch := make(chan []KeyInfo, len(keys))
+
 	go q.queryByID(ctx, keys, ch, startingBlock, evicted)
 
 	return ch, nil
@@ -116,6 +117,13 @@ func (q *DeliveryScanQueryByID) queryByID(ctx context.Context, keys []driver.PKe
 			startDelivery = true
 
 			continue
+		}
+		if len(values) != len(keys) {
+			logger.Errorf("peer returned %d values for %d keys in ns [%s]; falling back to block scan",
+				len(values), len(keys), ns)
+			startDelivery = true
+
+			continue // treat as a per-namespace failure (=> fall back to the slow block scan)
 		}
 		found := make([]KeyInfo, 0, len(values))
 		var notFound []string
--- a/token/services/network/fabric/lookup/deliveryqs_test.go
+++ b/token/services/network/fabric/lookup/deliveryqs_test.go
@@ -194,6 +194,68 @@ func TestQueryByID_MissingValueTriggersScan(t *testing.T) {
 	assert.True(t, scanner.called)
 }
 
+// A peer that returns more values than keys requested
+// must not index the keys slice out of range and crash the process.
+// (From Issue #2055.)
+func TestQueryByID_OversizedResponse(t *testing.T) {
+	// 2 values returned, but only 1 key requested.
+	querier := &fakeQuerier{results: map[driver.Namespace]querierResult{
+		"ns1": {raw: values(t, []byte("v1"), []byte("v2-unexpected"))},
+	}}
+	scanner := &fakeScanner{}
+
+	ch, err := newQuery(querier, scanner).QueryByID(t.Context(), 100, evictedFor(map[driver.Namespace]driver.PKey{
+		"ns1": "k1",
+	}))
+	require.NoError(t, err)
+
+	// Nothing is delivered for the oversized namespace.
+	// It should fall back to the block scan rather than crashing.
+	assert.Empty(t, drain(ch))
+	assert.True(t, scanner.called, "the oversized response must fall back to the block scan")
+}
+
+// A namespace holding several keys where only some resolve must deliver the resolved ones and
+// fall back to the block scan for the rest, keeping the still-missing keys grouped under their
+// namespace (the notFound branch that rewrites keysByNS rather than deleting it).
+func TestQueryByID_PartialResolutionWithinNamespace(t *testing.T) {
+	// One value present, one absent. Key<->value pairing follows the (map-random) request order,
+	// so assert on the shape rather than a fixed key.
+	querier := &fakeQuerier{results: map[driver.Namespace]querierResult{
+		"ns1": {raw: values(t, []byte("v"), []byte{})},
+	}}
+	scanner := &fakeScanner{}
+
+	evicted := map[driver.PKey][]events.ListenerEntry[lookup.KeyInfo]{
+		"k1": {&fakeEntry{ns: "ns1"}},
+		"k2": {&fakeEntry{ns: "ns1"}},
+	}
+
+	ch, err := newQuery(querier, scanner).QueryByID(t.Context(), 100, evicted)
+	require.NoError(t, err)
+
+	got := drain(ch)
+	require.Len(t, got, 1, "exactly the resolved key must be delivered")
+	assert.Equal(t, driver.Namespace("ns1"), got[0].Namespace)
+	assert.Contains(t, []driver.PKey{"k1", "k2"}, got[0].Key)
+	assert.Equal(t, []byte("v"), got[0].Value)
+	assert.True(t, scanner.called, "the unresolved key must fall back to the block scan")
+	assert.Equal(t, uint64(90), scanner.startBlock)
+}
+
+// An empty batch has nothing to query and nothing to scan: the channel simply closes empty.
+func TestQueryByID_EmptyEvicted(t *testing.T) {
+	querier := &fakeQuerier{results: map[driver.Namespace]querierResult{}}
+	scanner := &fakeScanner{}
+
+	ch, err := newQuery(querier, scanner).QueryByID(t.Context(), 100, nil)
+	require.NoError(t, err)
+
+	assert.Empty(t, drain(ch))
+	assert.Empty(t, querier.queried, "no namespace should be queried")
+	assert.False(t, scanner.called, "an empty batch must not start a block scan")
+}
+
 // The fallback block scan must never start from an underflowed block number: on a chain younger
 // than NumberPastBlocks it starts from FirstBlock instead of wrapping around to a block near
 // MaxUint64, which would silently find nothing and surface no error.


token/services/network/fabric/lookup/deliveryqs.go
```diff
@@ -66,7 +66,18 @@ func (q *DeliveryScanQueryByID) QueryByID(ctx context.Context, startingBlock dri
...
 	ch := make(chan []KeyInfo, len(keys))
-	go q.queryByID(ctx, keys, ch, startingBlock, evicted)
+
+	go func() {
+		// Defense in depth: any unforeseen panic in this background path must degrade to a
+		// failed request rather than crashing the whole node process, since nothing else in
+		// this goroutine's call chain has a recover().
+		defer func() {
+			if r := recover(); r != nil {
+				logger.Errorf("recovered from panic in queryByID: %v", r)
```
atharrva01: A question rather than a request, since I might be weighing this differently to you.

When the recover fires, `queryByID`'s own `defer close(ch)` has already run during unwinding, so the caller gets a closed empty channel and no error. `ScanFromBlock` is never reached, so those keys get no block scan fallback either. As far as I can tell the listener then just never fires for them, which looks like a silent permanent non-resolution rather than a failed request.

Your own test pins this, which is what made me notice: `TestQueryByID_PanicIsRecovered` asserts `assert.False(t, scanner.called)` (line 274).

A crash is clearly worse, so this is still an improvement. But the Fix A path is careful to set `startDelivery` and degrade to the scan, and it might be nice if the recovered path landed somewhere similar rather than closing empty. I appreciate that is awkward given the inner defer has already closed the channel, so possibly not worth it here.

token/services/network/fabric/lookup/deliveryqs.go
```diff
@@ -66,7 +66,18 @@ func (q *DeliveryScanQueryByID) QueryByID(ctx context.Context, startingBlock dri
 	// Keys are supposed to be unique
 	keys := collections.Keys(evicted) // These are the state keys we are looking for
 	ch := make(chan []KeyInfo, len(keys))
-	go q.queryByID(ctx, keys, ch, startingBlock, evicted)
+
+	go func() {
```
atharrva01: Not for this PR, just noting it while it is in view: the finality side has the same bare `go q.queryByID(...)` with no recover (`network/fabric/finality/deliveryqs.go:57`). Fix A's bug genuinely is not there, but if an unrecovered panic in one of these background goroutines counts as a node-crash surface, the argument for Fix B seems to apply to that one too.

Happy to open an issue for it so it does not widen this PR, if you think it is worth tracking.

token/services/network/fabric/lookup/deliveryqs_test.go
```diff
@@ -194,6 +194,86 @@ func TestQueryByID_MissingValueTriggersScan(t *testing.T) {
...
+	assert.False(t, scanner.called, "an empty batch must not start a block scan")
+}
+
+// A panic in the background goroutine (here: a listener entry with no namespace source) must be
+// recovered so it degrades to an empty, closed channel instead of crashing the node process.
+func TestQueryByID_PanicIsRecovered(t *testing.T) {
+	querier := &fakeQuerier{results: map[driver.Namespace]querierResult{}}
+	scanner := &fakeScanner{}
+
+	// An empty listener slice makes the namespace lookup index out of range and panic.
```
atharrva01: Minor, and possibly not worth changing: this leans on `slices2.GetAny` panicking on an empty slice, so the test is tied to how that FSC helper behaves on empty input rather than to anything in this file.

A `fakeQuerier` whose `QueryStates` panics would pin Fix B through one of the interfaces the test already controls, and would keep working regardless of what the helper does. The current version does exercise the recover today, so this is only about how it ages.

atharrva01 (commented):
hi @Effi-S , I spent some time in this function recently (#1426, #1999) so I read this one closely. The panic looks real to me: the loop ranges `values` but indexes `keys`, and since `for ns, keys := range keysByNS` shadows the outer parameter, the check is comparing against that namespace's keys, which is the right thing to compare against. Using `!=` rather than `>` also seems right, since a short response would index in range but pair values with the wrong keys.

One thing I checked in case it was worth widening the PR: the finality twin at `network/fabric/finality/deliveryqs.go` does not have this bug, it walks the key set and fetches per txID with no positional indexing. So Fix A really is specific to the lookup path.

Ran the package at `-count=2 -race`, green. Three notes below, all non-blocking, and the first is more of a question than a suggestion.

token/services/network/fabric/lookup/deliveryqs.go
Effi-S: Yes, let's not widen  this PR 
Please open and assign to me,
Thanks :)

token/services/network/fabric/lookup/deliveryqs.go
Effi-S: Yes. 
@AkramBitar and I have been discussing this for the past few days. 

The problem is that there is no receiving channel for errors on FSC side. 

Let's lay down the options to make a clear picture:
|     | outcome |
|---|---|
|1. don't recover | node can crash for some new unforeseen bug|
|2. current solution| early/in-loop panic is silently abandoned| 
|3. A more Structured solution on our side recovering in-frame | scan-phase  panic will still be silent|
|4. our side and FSC solution| will solve the issue but is a large cross-repo PR| 


I suggest the current solution and then create  a new Issue for implementing 4.

token/services/network/fabric/lookup/deliveryqs.go
```diff
@@ -66,7 +66,18 @@ func (q *DeliveryScanQueryByID) QueryByID(ctx context.Context, startingBlock dri
...
+		// Defense in depth: any unforeseen panic in this background path must degrade to a
```
AkramBitar: The comment says "any unforeseen panic in this background path", but the guard is narrower than that: the fallback block scan's callback is not invoked on this goroutine. FSC runs it on its own (`platform/fabric/core/generic/delivery/delivery.go`: `go d.readBlocks(ch)` → `d.callback(...)`), so a panic inside the closure passed to `q.Delivery.ScanFromBlock` — `InspectRWSet`, `rws.GetWriteAt`, the `keysByNS` reads, `ch <- txInfos` — escapes this `recover()` entirely.

A concrete path: on `ctx.Done()`, `runReceiver` calls `d.Stop(...)`, `untilStop()` returns, `ScanFromBlock` returns, and `queryByID`'s `defer close(ch)` fires — while `readBlocks` is still draining its buffered block channel and invokes the callback, which then does `ch <- txInfos` on a closed channel. That panic is on FSC's goroutine and still takes down the process.

Either move/duplicate the guard into the scan callback, or reword the comment so it doesn't imply coverage it doesn't have — otherwise this reads as "already handled" to the next person who looks.

token/services/network/fabric/lookup/deliveryqs.go
AkramBitar: Could you add `string(debug.Stack())` here? This is an explicit catch-all for *unknown* panics, and it removes the crash dump that was previously the only diagnostic for them. The single line has no location — in a local run it came out as:

```
[panurus.services.network.fabric.lookup] 1 -> recovered from panic in queryByID: runtime error: index out of range [0] with length 0
```

The caller field is the anonymous func (`1`), so an operator can't tell where in `queryByID` it happened. Without a stack, a panic that this handler catches in production is effectively undiagnosable.

token/services/network/fabric/lookup/deliveryqs.go
```diff
@@ -66,7 +66,18 @@ func (q *DeliveryScanQueryByID) QueryByID(ctx context.Context, startingBlock dri
...
+		// failed request rather than crashing the whole node process, since nothing else in
+		// this goroutine's call chain has a recover().
+		defer func() {
+			if r := recover(); r != nil {
+				logger.Errorf("recovered from panic in queryByID: %v", r)
+			}
+		}()
+		q.queryByID(ctx, keys, ch, startingBlock, evicted)
```
AkramBitar: Worth considering what this `recover()` does to the batch rather than just to the process.

The only panic actually reachable on this path today — and the one `TestQueryByID_PanicIsRecovered` injects — is `slices2.GetAny(v)` → `v[0]` on an empty listener slice (`token/services/utils/slices/slices.go:16`), reached from the group-by-namespace loop around line 93. That fires *before* any query is issued, so the recover closes `ch` empty and no error is surfaced anywhere: FSC's consumer (`listenermanager.go`, `fetchTxs`) only ranges the channel and has no error path. The result is that every listener in the eviction batch is never notified and each caller blocks until its own timeout, with one log line to show for it.

A targeted guard in that loop instead — e.g. `if len(v) == 0 { logger.Errorf(...); continue }` — would lose only the offending key while the rest of the batch resolves normally. As written, the blanket recover trades a crash for silent, timeout-delayed loss of the whole batch, and the new test currently encodes that as the intended behaviour.

(Related, minor: because `TestQueryByID_PanicIsRecovered` induces the panic through the unguarded `v[0]`, adding that bounds check would make the test pass without any panic occurring — no panic, channel still closes empty, `scanner.called` still false. It would silently stop exercising the recover. Injecting via something like a `fakeEntry` whose `Namespace()` panics would keep it honest.)

AkramBitar (commented):
@Effi-S, few comments

AkramBitar (approved):
LGTM

PR #1912, opened 2026-07-14
Perf/zkp ipa optimization

SurbhiAgarwal1:
## Description

This PR introduces native `gnark-crypto` field arithmetic for the Inner Product Argument (IPA) prover to eliminate the massive allocation overhead of the generic `mathlib` wrapper in the tight reduction loops.

As discussed in #1432 and the parent performance epic (#954), the core curve layer heavily relies on `mathlib` interface allocations, which becomes a major bottleneck during Bulletproof/IPA Proof generation. 

**Changes Made:**
* **Added Native IPA Execution**: Created `ipa_native.go` containing `nativeIPAReduce` to execute round reductions natively.
* **Dynamic Curve Dispatch**: Updated `ipaProver.reduce` to dynamically dispatch execution based on the underlying curve via `math.DispatchCurve`. If the curve is `BLS12-381` or `BN254`, the execution seamlessly hands off to the native implementation.
* **In-Place Generic Arithmetic**: In `nativeIPAReduce`, vectors are converted into `gnark-crypto` native array types (`math.GnarkFr[T]`). All scalar computations now happen strictly in-place, completely bypassing dynamic `mathlib.Zr` allocations until absolutely necessary for Multi-Scalar Multiplication (`MultiScalarMul`).

## Related Issues

* Fixes #1432 
* Part of Epic #954

## Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Performance Improvement (optimizing code execution and memory allocations)

## Testing Done

- [x] Unit tests passed (`make unit-tests`)
- [x] Native `bulletproof` cryptography logic verified against all bit-lengths and tampering failure cases.
- [x] Benchmark testing shows significant reduction in heap allocation count and garbage collector pausing inside inner loops.

Issue #954, opened 2025-03-12
Performance improvements


alexandrosfilios:
The performance of the system measured in txs/sec and end-to-end latency must be improved. There is no predefined set of tasks to do to achieve this, but it is more an iterative process of running experiments and pinpointing and tackling the next bottleneck

PR #2154, opened 2026-08-06
perf(zkatdlog): optimize IPA prover using native gnark-crypto arithmetic

### LINKED ISSUE #1432: Optimize ZKP Computation Using CPU SIMD and Specialized Instructions

SurbhiAgarwal1:
### Description
This PR introduces native gnark-crypto field arithmetic for the Inner Product Argument (IPA) prover to eliminate the allocation overhead of the generic mathlib wrapper in tight reduction loops. Supersedes #1912.

As discussed in #1432 and the parent performance epic (#954), the core curve layer heavily relies on mathlib interface allocations, which becomes a bottleneck during Bulletproof/IPA proof generation.

### Changes Made
- **Native IPA Execution**: Created `ipa_native.go` containing `nativeIPAReduce` to execute round reductions natively.
- **Dynamic Curve Dispatch**: Updated `ipaProver.reduce` to dynamically dispatch execution based on the underlying curve via `math.DispatchCurve` (BLS12-381 and BN254).
- **In-Place Generic Arithmetic**: Converted vectors into gnark-crypto native array types (`math.GnarkFr[T]`) for in-place scalar computations.

### Related Issues
- Supersedes #1912
- Fixes #1432
- Part of Epic #954

### Type of Change
- [x] Performance Improvement (optimizing code execution and memory allocations)

### Testing Done
- [x] Unit tests passed (`make unit-tests`)
- [x] Verified against all bit-lengths and tampering test cases.

### DIFF
--- a/token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof/ipa.go
+++ b/token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof/ipa.go
@@ -12,10 +12,14 @@ import (
 	"github.com/LFDT-Panurus/panurus/token/core/zkatdlog/nogh/v1/crypto/common"
 	"github.com/LFDT-Panurus/panurus/token/core/zkatdlog/nogh/v1/crypto/math"
 	executor "github.com/LFDT-Panurus/panurus/token/core/zkatdlog/nogh/v1/crypto/rp/executor"
-	bls12381fr "github.com/consensys/gnark-crypto/ecc/bls12-381/fr"
 	bn254fr "github.com/consensys/gnark-crypto/ecc/bn254/fr"
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 )
+
+// DisableNativeIPA when set to true bypasses the gnark-crypto fast path in ipaProver.reduce,
+// forcing execution of the standard fallback mathlib path.
+// This is primarily used for testing and benchmarking to compare the two paths side by side.
+var DisableNativeIPA bool
 
 // IPA contains the proof for the inner product argument.
 type IPA struct {
@@ -186,10 +190,11 @@ func (p *ipaProver) Prove() (*IPA, error) {
 // of the left vector and right is a function of right vector.
 // Both vectors are committed in com which is passed as a parameter to reduce
 func (p *ipaProver) reduce(X, com *mathlib.G1) (*mathlib.Zr, *mathlib.Zr, []*mathlib.G1, []*mathlib.G1, error) {
-	isBLS, isBN254 := math.DispatchCurve(p.Curve)
-	if isBLS {
-		return nativeIPAReduce[bls12381fr.Element, *bls12381fr.Element](p, X, com)
-	} else if isBN254 {
+	_, isBN254 := math.DispatchCurve(p.Curve)
+	// The native gnark-crypto path is only beneficial for BN254. Benchmarking showed
+	// that for BLS12-381 the 48-byte field element size makes BigInt conversion overhead
+	// outweigh the arithmetic savings, yielding no speedup and more allocations.
+	if isBN254 && !DisableNativeIPA {
 		return nativeIPAReduce[bn254fr.Element, *bn254fr.Element](p, X, com)
 	}
 
--- a/token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof/ipa.go
+++ b/token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof/ipa.go
@@ -520,8 +520,9 @@ func CloneGenerators(LeftGenerators, RightGenerators []*mathlib.G1) ([]*mathlib.
 //	  sInv[i + 2^r] = sInv[i] · x_{k-1-r}^{-1}   (swapped)
 //	  sInv[i]       = sInv[i] · x_{k-1-r}         (swapped)
 //
-// This replaces the previous O(n·log n) nested-loop implementation and
-// eliminates the final BatchInverse call for sInv.
+// This replaces the previous O(n·log n) nested-loop implementation.
+// One BatchInverse call computes challenge inverses up front; the butterfly
+// then builds s and sInv together without a second BatchInverse pass.
 //
 // Input: n, challenges = [x_0, …, x_{k-1}] where n = 2^k.
 // Returns (s, sInv) where sInv[i] = s[i]^{-1}.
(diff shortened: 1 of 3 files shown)


AkramBitar (changes requested):
@SurbhiAgarwal1 

Please see my comment regarding the benchmark compare.

Regards,
Akram

token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof/rp_native_test.go
```diff
@@ -198,59 +198,6 @@ func TestNativeRPTamperedProofRejected(t *testing.T) {
 				bad.Data.T2 = ctx.curve.GenG1.Mul(ctx.curve.NewRandomZr(rand))
 				assert.Error(t, ctx.verify(t, bad))
 			})
-
-			t.Run("tampered_C", func(t *testing.T) {
```
Effi-S, line 202: @SurbhiAgarwal1, 
Please revert this. 
we need `tampered_C` and `tampered_D` for a security item.

token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof/ipa_native.go
```diff
@@ -8,115 +8,152 @@ package bulletproof
...
+		E(&xSquareE).Mul(xE, xE)
+		xSquare := math.NativeToZr[T, E](E(&xSquareE), p.Curve)
+
+		var xSquareInvE T
+		E(&xSquareInvE).Inverse(E(&xSquareE))
+		xSquareInv := math.NativeToZr[T, E](E(&xSquareInvE), p.Curve)
+
+		CPrime := LArray[i].Mul2(xSquare, RArray[i], xSquareInv)
+		CPrime.Add(com)
+		com = CPrime
```
Effi-S: @SurbhiAgarwal1,
Please check weather we can remove this: 

```go 
 var xSquareE T
E(&xSquareE).Mul(xE, xE)
xSquare := math.NativeToZr[T, E](E(&xSquareE), p.Curve)

var xSquareInvE T
E(&xSquareInvE).Inverse(E(&xSquareE))          // per-round field inversion
xSquareInv := math.NativeToZr[T, E](E(&xSquareInvE), p.Curve)

CPrime := LArray[i].Mul2(xSquare, RArray[i], xSquareInv)
CPrime.Add(com)
com = CPrime
```
Please correct me if I'm wrong, but I don't think`com` feeds any challenge

token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof/ipa.go
```diff
@@ -525,31 +520,19 @@ func CloneGenerators(LeftGenerators, RightGenerators []*mathlib.G1) ([]*mathlib.
 //	  sInv[i + 2^r] = sInv[i] · x_{k-1-r}^{-1}   (swapped)
 //	  sInv[i]       = sInv[i] · x_{k-1-r}         (swapped)
 //
-// For BLS12-381 and BN254 curves the inner loop is executed using native
-// gnark-crypto field elements (nativeComputeSVector), which eliminates the
-// big.Int allocation overhead of the mathlib.Zr wrapper on every multiply.
-// For all other curves the pure-mathlib path is used as a fallback.
+// This replaces the previous O(n·log n) nested-loop implementation and
+// eliminates the final BatchInverse call for sInv.
 //
```
Effi-S, line 521: comment is misleading, we still call BatchInverse on line 555: 
```go 
challengeInvs := math.BatchInverse(challenges, curve)
```

Effi-S (changes requested):

token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof/ipa_native.go
SurbhiAgarwal1: Hi @Effi-S, you're right - com is never hashed inside the prover's reduce loop, so updating it was dead computation. I've removed the xSquare/xSquareInv/CPrime/com block from both nativeIPAReduce (ipa_native.go) and the generic reduce loop (ipa.go). The verifier reconstructs the commitment independently from LArray and RArray without depending on the prover's com accumulation. All tests including tampered_C and tampered_D still pass. Thanks for catching this.

token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof/ipa.go
SurbhiAgarwal1: Hi @Effi-S, good catch! The comment was misleading - BatchInverse is still called once up front for the challenge inverses. What the butterfly eliminates is the second BatchInverse over the full s vector to derive sInv. Godoc updated to clarify this. Thanks!

token/core/zkatdlog/nogh/v1/crypto/rp/bulletproof/rp_native_test.go
SurbhiAgarwal1: Hi @Effi-S, tampered_C and tampered_D have been restored , both subtests are present in TestNativeRPTamperedProofRejected and pass for both BLS12-381 and BN254 (commit 99020c5). Thanks!

Effi-S: Hi @SurbhiAgarwal1, 
Thanks! 
Please revert `TestNativeRPForeignCommitmentsRejected ` as well.

SurbhiAgarwal1: Hi @Effi-S, TestNativeRPForeignCommitmentsRejected (F-09) has been restored as well. Thanks!

Effi-S (commented):
Thanks!

Effi-S:
Hi @SurbhiAgarwal1,

Thank you so much for all the hard work here. 
The port was clean and you were responsive and patient as we were deliberating. 🙏 

After a lot of discussion on our side, including running the branch multiple times on our performance environment, we've decided not to merge this one, and I want to be open about why.

The BN254 micro-benchmark win is real, but it didn't carry through end-to-end: TPS benchmark was too similar to `main` and memory improvement was 4% for IPA only.
Proof generation is precomputed, so the IPA reduce loop just isn't the bottleneck in the overall path. And to keep even that win we'd have to restrict the native path to a single curve and add curve-specific dispatch.
This adds extra complexity and is a step away from the clean mathlib abstraction.


The investigation was genuinely useful and helps point us at where the real costs are for the performance epic (#954).

Thank you again,
Effi

PR #2157, opened 2026-08-06
feat(observability): implement injectable Token-API decorators and circuit breaker (#1644)

### LINKED ISSUE #1644: Observability, Alerting, and Circuit Breaking [LOW]
**Description**
Instrument all wallet service entry points with metrics (request rate, latency percentiles, error rate, cache level, registry size) and expose them through the existing `metrics.Provider` infrastructure already present in the codebase. Define alert thresholds that trigger when request rates or resource utilisation approach saturation. Pair this with a circuit-breaker pattern on the wallet service: when error rates or latency exceed configured thresholds, the service temporarily rejects new requests and returns a back-pressure signal to callers. This limits blast radius during an active DoS attempt and gives operators visibility to detect and respond before full exhaustion occurs.

SurbhiAgarwal1:
Fixes #1644

### Summary
This PR implements a clean, injectable decorator architecture for Token-API entry points (`driver.WalletService`, `driver.OwnerWallet`, `driver.IssuerWallet`) incorporating Prometheus-compatible metrics and circuit breaker back-pressure protection, following @adecaro's recommendation to use `gowrap` generator templates.

#### 1. Circuit Breaker (`token/services/observability/circuitbreaker.go`)
- Implements thread-safe state machine (`StateClosed`, `StateOpen`, `StateHalfOpen`).
- Configurable consecutive error limit and cooldown timeout.
- Fast-fails with back-pressure error `ErrCircuitOpen` when open to limit blast radius during DoS attacks.

#### 2. Metrics Instrumentation (`token/services/observability/metrics.go`)
- Integrated with `metrics.Provider`.
- Exposes `requests_total` (Counter), `errors_total` (Counter), `request_duration_seconds` (Histogram), and `inflight_requests` (Gauge).

#### 3. Injectable Decorators (`token/services/observability/decorator.go`)
- `NewWalletServiceDecorator`, `NewOwnerWalletDecorator`, `NewIssuerWalletDecorator` wrap services non-intrusively without modifying underlying concrete driver logic.

#### 4. GoWrap Template (`token/services/observability/templates/gowrap_prometheus_cb.tmpl`)
- Template for automated decorator generation using `gowrap`.

#### 5. Unit Tests & Race Detection (`token/services/observability/observability_test.go`)
- Comprehensive unit test suite covering metrics observation, circuit breaker state transitions, back-pressure rejection, and recovery.
- Tested and verified cleanly with Go race detector (`go test -race`).

### Verification & Compliance
- `go test -v ./token/services/observability/...` 
- `go test -v -race ./token/services/observability/...`
- `gofmt -l -s ./token/services/observability` 
- Error handling complies with codebase guidelines using `github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors`.
- Commit is signed off with DCO (`Signed-off-by: Surbhi Agarwal <SurbhiAgarwal1@users.noreply.github.com>`).


plan.md
```diff
@@ -0,0 +1,24 @@
+# Observability, Alerting, and Circuit Breaking for Token-API (#1644)
```
AkramBitar: Need to delete this file

AkramBitar (changes requested):

AkramBitar (changes requested):
All 7 original comments are addressed — split PR, wired in both drivers, per-wallet independent breakers, all four wallet types covered, plan.md removed, disabled.Provider used. Three new issues below.

**1. Metrics are dead code at runtime**
Both driver call sites pass `&disabled.Provider{}`, so every counter/histogram/gauge is a no-op. The decorator is wired but the metrics half of the feature never reports anything. If this is intentional and a real provider will be injected in a follow-up, please add a comment or open a tracking issue — as-is it reads as an oversight.

**2. `CircuitBreaker.State()` is advisory, not authoritative**
`State()` computes the HalfOpen transition under RLock without updating `cb.state`. It can return `StateHalfOpen` while `cb.state` is still `StateOpen`, so the two disagree. The doc says it "returns the current State" — that claim is inaccurate. Either update `cb.state` inside `State()` (needs a write lock) or document explicitly that `State()` is advisory-only and `Allow()` is the authoritative gating call.

**3. Inconsistent circuit breaker coverage on `WalletServiceDecorator`**
`GetEnrollmentID`, `GetRevocationHandle`, and `GetEIDAndRH` observe latency but bypass `d.cb.Execute`, so repeated failures on those paths never trip the breaker. Meanwhile `RegisterRecipientIdentity` and `RegisterOwnerIdentity` do go through it. The asymmetry should either be made consistent or documented as intentional.

AkramBitar (changes requested):
Hello @SurbhiAgarwal1 

Please see my comments. 

Regards,
Akram

Effi-S (changes requested):
Hi @SurbhiAgarwal1,
Please resolve conflicts and Rebase

PR #2158, opened 2026-08-06
fix(selector): keep the status lookup out of the shard lock, and bound the scanner test waits

### LINKED ISSUE #2156: TestScannerDoesNotDeleteReclaimed intermittently hangs until the 10m test timeout
## Problem

`TestScannerDoesNotDeleteReclaimed` in `token/services/selector/simple/inmemory/locker_test.go:98`
intermittently hangs until the package-level timeout. Seen on #2114, a `.golangci.yml` only change
with no Go code in it:

```
FAIL  github.com/LFDT-Panurus/panurus/token/services/selector/simple/inmemory   600.288s
panic: test timed out after 10m0s
  running tests:
    TestScannerDoesNotDeleteReclaimed (9m50s)
```

## Impact

Worse than an ordinary flaky assertion. A wrong assertion fails in seconds; this one holds the whole
`utest (unit-tests-race)` job for the full ten-minute timeout before failing, and it takes the rest of
the package's results with it.

## What the dump shows

Two goroutines matter:

```
goroutine 6 [chan receive, 9 minutes]:
  ...inmemory.TestScannerDoesNotDeleteReclaimed(...)

goroutine 7 [select]:
  ...inmemory.(*locker).scan(...)
      token/services/selector/simple/inmemory/locker.go:429
```

`locker.go:429` is the idle branch of the collector loop, the
`select { case <-time.After(d.sleepTimeout): case <-ctx.Done(): }`. So the scanner is not stuck, it is
sleeping normally. The test is the one blocked, and the only unguarded receive it has is `<-entered`.

That points at the one-shot hook never firing for the scanner:

```go
entered := make(chan struct{})
release := make(chan struct{})
var once sync.Once
mock.setGetStatusHook(func(txID string) {
    if txID == txA && mock.status(txA) == ttxdb.Deleted {
        first := false
        once.Do(func() { first = true })
        if !first {
            return
        }
        close(entered)
        <-release
    }
})
mock.setStatus(txA, ttxdb.Deleted)
...
<-entered
```

`entered` is only closed by whichever caller first observes tx-A as `Deleted`. If that never happens,
`<-entered` blocks forever, which matches goroutine 6 sitting on a channel receive for nine minutes
while the scanner idles.

The `scan goroutine did not stop within timeout` warning logged just before the panic is consistent
with that: `t.Cleanup` calls `Stop()`, which waits on `scanDone`, and by then the test itself is
already wedged.

## Reproduction

Not reproducible locally so far:

- clean on current `main`
- 40 runs with `-count=40 -race` all passed, 13s total

So it is rare rather than reliable, and probably needs the scheduling pressure of the full CI run.

## Notes

The test coordinates with its own background scanner through a `sync.Once` hook, so it depends on the
scanner being the first caller to observe the new status, within a 20ms scan interval
(`NewLocker(mock, 20*time.Millisecond, time.Minute)`). That ordering is an assumption about
scheduling rather than something the test enforces, which is the same shape as the cases in #2121.

Worth confirming the exact interleaving before changing anything, since the property it is testing
(the scanner must not delete an entry that was reclaimed while it was mid-evaluation) is a real one
worth keeping, and it guards the re-validation added in `locker.go` around line 420.

atharrva01:
Refs #2156.

I could not reproduce the hang, so this does not claim to fix its cause. It makes the next
occurrence diagnosable instead of silent.

### What this changes

The test coordinates with the locker's scan goroutine over two unbuffered channels, `entered` and
`release`, and receives on both without a bound. If the interleaving it assumes does not happen, the
receive never returns and the package sits until the 10 minute timeout, which hides which wait failed
and takes the rest of the package's results with it.

Both receives are now bounded, well above the 20ms scan interval the test drives, so they only fire
when something is genuinely wrong. The one waiting on the scanner reports what did not happen.

Verified by disabling the status hook so the wait can never be satisfied: it now fails in seconds
with `scanner never observed tx-A as Deleted: the status hook did not fire within 2s`, rather than
hanging.

### On the cause

Not reproducible locally: clean on `main`, and ~260 runs across `-count=120`, `-cpu=1` and whole
package runs under `-race` all passed.

From the CI dump, the scanner was idle in the sleep branch of its collector loop
(`locker.go:429` is the `time.After`/`ctx.Done()` select), not blocked in the hook, while the test
sat on a channel receive for nine minutes. That points at the hook never firing for the scanner
rather than the two goroutines deadlocking against each other, but I could not pin the interleaving
that produces it.

One thing worth a separate look, found while reading rather than from the failure:
`lockInShard` holds the shard lock across `reclaim`, which calls `GetStatus`
(`locker.go:172-185`, `306-307`). An external status lookup under a shard lock means a slow or
blocking provider stalls every `Lock` and `Unlock` for that owner. It also means a hook that blocks
there blocks while holding the lock. That is a latent hazard regardless of this test, and I have
noted it on #2156 rather than changing production code in a test-only PR.

---

## Added: the production fix

Went back through the locking path properly rather than leaving this test-only.

`Lock(reclaim=true)` called `GetStatus` from inside `lockInShard` **while holding the shard write
lock** (`locker.go:172-185` and the old `reclaim` at `306-307`). `GetStatus` goes to the transaction
store, so a slow or stuck store blocked every `Lock`, `UnlockIDs` and `IsLocked` for that owner, and
the collector deleting phase with them, for the duration of the query.

The collector already avoids precisely this, and states the rule:

> Copy the entries and release the shard lock before looking their status up: the lookups may be
> slow, and no Lock/UnlockIDs of this owner must ever wait behind the collector on the status
> provider.

The reclaim path inverted that: it made the collector, and everything else on the shard, wait behind
*the locker* on the status provider.

**Fix**: resolve the holder status before taking the write lock, then re-validate under it, the same
phase split the collector uses. If the entry changed hands while the lock was released, the reclaim
is refused and the caller retries, rather than taking a token away from its new holder. The old
`reclaim` helper had one caller and is gone.

**Regression test**: blocks a status lookup and asserts the shard is still usable while it is in
flight. On the old code it fails with `IsLocked blocked while a reclaim status lookup was in flight`,
after it passes.

This is a real hazard on its own merits and I am not claiming it is the cause of the intermittent
hang, which I still could not reproduce. It does remove the mechanism by which a blocking status
provider can wedge a shard, which is the closest thing to that shape I could find in the code.

### DIFF
--- a/token/services/selector/simple/inmemory/locker.go
+++ b/token/services/selector/simple/inmemory/locker.go
@@ -169,6 +169,38 @@ func (d *locker) lockInShard(ctx context.Context, s *shard, owner string, id *to
 	}
 	s.mu.RUnlock()
 
+	// When reclaiming, resolve the current holder's status before taking the write
+	// lock. GetStatus goes to the transaction store, and holding the shard lock
+	// across it would make every Lock, UnlockIDs, IsLocked and the collector's own
+	// delete phase for this owner wait on that query. The collector already splits
+	// its cycle this way for exactly that reason; doing the lookup here under the
+	// lock inverted it, so a slow or stuck status provider stalled the whole shard.
+	//
+	// The holder observed here is re-validated under the lock below, since it may
+	// change while the lock is not held.
+	var (
+		observedTxID   string
+		observedStatus int
+		statusResolved bool
+	)
+	if reclaim {
+		s.mu.RLock()
+		e, held := s.locked[k]
+		if held {
+			observedTxID = e.TxID
+		}
+		s.mu.RUnlock()
+
+		if held {
+			status, _, err := d.ttxdb.GetStatus(ctx, observedTxID)
+			if err != nil {
+				logger.DebugfContext(ctx, "failed getting status of [%s] while reclaiming [%s]: [%s]", observedTxID, id, err)
+			} else {
+				observedStatus, statusResolved = status, true
+			}
+		}
+	}
+
 	// it is either not locked or we are reclaiming
 	s.mu.Lock()
 	defer s.mu.Unlock()
@@ -180,9 +212,18 @@ func (d *locker) lockInShard(ctx context.Context, s *shard, owner string, id *to
 		e.LastAccess = time.Now()
 
 		if reclaim {
-			// Second chance
+			// Second chance. Only act on the status resolved above if the entry is
+			// still the one it was resolved for: a concurrent Lock(reclaim=true) may
+			// have handed the token to another transaction in the meantime, and
+			// reclaiming on a stale status would take a token away from its new
+			// holder. When it no longer matches, report the token as locked and let
+			// the caller retry.
 			logger.DebugfContext(ctx, "[%s] already locked by [%s], try to reclaim...", id, e)
-			reclaimed, status := d.reclaim(ctx, s, id, e.TxID)
+			status := observedStatus
+			reclaimed := statusResolved && e.TxID == observedTxID && status == ttxdb.Deleted
+			if reclaimed {
+				delete(s.locked, k)
+			}
 			if !reclaimed {
 				logger.DebugfContext(ctx, "[%s] already locked by [%s], reclaim failed, tx status [%s]", id, e, ttxdb.TxStatusMessage[status])
 				if logger.IsEnabledFor(zapcore.DebugLevel) {
@@ -298,24 +339,6 @@ func (d *locker) IsLocked(id *token2.ID) bool {
 	}
 
 	return false
-}
-
-// reclaim checks the tx status for id inside shard s and deletes the entry
-// if the holding transaction is finalized (Deleted). The caller must hold
-// s.mu (write lock).
-func (d *locker) reclaim(ctx context.Context, s *shard, id *token2.ID, txID string) (bool, int) {
-	status, _, err := d.ttxdb.GetStatus(ctx, txID)
-	if err != nil {
-		return false, status
-	}
-	switch status {
-	case ttxdb.Deleted:
-		delete(s.locked, *id)
-
-		return true, status
-	default:
-		return false, status
-	}
 }
 
 func (d *locker) start(ctx context.Context) {
--- a/token/services/selector/simple/inmemory/locker.go
+++ b/token/services/selector/simple/inmemory/locker.go
@@ -179,15 +179,16 @@ func (d *locker) lockInShard(ctx context.Context, s *shard, owner string, id *to
 	// The holder observed here is re-validated under the lock below, since it may
 	// change while the lock is not held.
 	var (
-		observedTxID   string
-		observedStatus int
-		statusResolved bool
+		observedTxID       string
+		observedLastAccess time.Time
+		observedStatus     int
+		statusResolved     bool
 	)
 	if reclaim {
 		s.mu.RLock()
 		e, held := s.locked[k]
 		if held {
-			observedTxID = e.TxID
+			observedTxID, observedLastAccess = e.TxID, e.LastAccess
 		}
 		s.mu.RUnlock()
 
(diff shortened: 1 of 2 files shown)


token/services/selector/simple/inmemory/locker.go
```diff
@@ -180,9 +212,18 @@ func (d *locker) lockInShard(ctx context.Context, s *shard, owner string, id *to
...
+			// Second chance. Only act on the status resolved above if the entry is
+			// still the one it was resolved for: a concurrent Lock(reclaim=true) may
+			// have handed the token to another transaction in the meantime, and
+			// reclaiming on a stale status would take a token away from its new
+			// holder. When it no longer matches, report the token as locked and let
+			// the caller retry.
 			logger.DebugfContext(ctx, "[%s] already locked by [%s], try to reclaim...", id, e)
-			reclaimed, status := d.reclaim(ctx, s, id, e.TxID)
+			status := observedStatus
+			reclaimed := statusResolved && e.TxID == observedTxID && status == ttxdb.Deleted
```
AkramBitar: The re-validation is weaker than the collector's, which this comment claims parity with. Here it is `e.TxID == observedTxID`; the delete phase at `locker.go:442` compares `e.TxID == entry.txID && e.LastAccess.Equal(entry.lastAccess)`.

Concrete gap: entry for token T is held by `txA`; a reclaim for `txB` observes `(txA, Deleted)`; while the lookup is in flight T is unlocked (`UnlockIDs`/`UnlockByTxID`, shard non-empty so not pruned) and re-locked for `txA` again; the reclaim then takes the write lock, sees a matching `TxID`, and deletes a **fresh** entry on a stale verdict. The selector's own retry loop does unlock-then-relock the same token under the same `txID`, so the ABA shape exists — it only bites when that tx's stored status is already `Deleted`, which makes it unlikely rather than impossible.

Capturing `LastAccess` in the observe phase and comparing it here would close it — but note it has to be read before the `e.LastAccess = time.Now()` write at line 212 clobbers it.

token/services/selector/simple/inmemory/locker.go
```diff
@@ -180,9 +212,18 @@ func (d *locker) lockInShard(ctx context.Context, s *shard, owner string, id *to
...
+			if reclaimed {
+				delete(s.locked, k)
+			}
 			if !reclaimed {
 				logger.DebugfContext(ctx, "[%s] already locked by [%s], reclaim failed, tx status [%s]", id, e, ttxdb.TxStatusMessage[status])
```
AkramBitar: This pairs the **current** holder `e` with `TxStatusMessage[status]`, where `status = observedStatus` is the status of the *previously observed* holder. In exactly the changed-hands case the new guard exists to catch, it emits self-contradictory diagnostics:

```
[tok] already locked by [[tx-C]…], reclaim failed, tx status [Deleted]
```

…which sends anyone debugging a reclaim failure after the wrong transaction. Logging `observedTxID` alongside the status, or suppressing the status when `e.TxID != observedTxID`, would keep the two halves consistent.

token/services/selector/simple/inmemory/locker_test.go
```diff
@@ -170,3 +195,71 @@ func TestScannerDeletesStaleEntry(t *testing.T) {
...
+	// the locking path, not the scanner.
+	d := NewLocker(mock, time.Hour, time.Hour).(*locker)
+	t.Cleanup(func() { _ = d.Stop() })
+	_, err := d.Lock(context.Background(), "w1", tokenID, txA, false)
+	require.NoError(t, err)
+
+	// Block the next status lookup for tx-A, standing in for a slow transaction store.
+	inLookup := make(chan struct{})
+	unblock := make(chan struct{})
+	var once sync.Once
```
AkramBitar, line 229: This one-shot `once` can be consumed by the wrong goroutine, which silently disarms the regression guard.

`scan` performs a full pass *before* its first sleep, so despite `sleepTimeout = time.Hour` the collector races the initial `Lock`. If its shard snapshot lands after the `tx-A` entry is registered and its `GetStatus` lands after `setGetStatusHook`, then the **collector** closes `inLookup` and parks on `<-unblock`, the reclaim's own lookup passes straight through, and the test passes having never held a lookup open inside `Lock`. In that interleaving it would pass identically on the pre-fix code.

Calling `d.Stop()` before arming the hook (or gating the hook on the reclaiming txID) makes it deterministic.

token/services/selector/simple/inmemory/locker_test.go
```diff
@@ -95,6 +95,14 @@ func (m *mockTXStatusProvider) GetStatus(_ context.Context, txID string) (ttxdb.
 // transaction before the scanner deletes, the scanner must NOT delete the
 // new entry. The test drives the real scan loop and blocks the scanner's
 // status lookup to open the race window deterministically.
+// Bounds for the coordination in TestScannerDoesNotDeleteReclaimed. Both are far longer than the
+// 20ms scan interval the test drives, so they only fire when something is genuinely wrong, but they
+// keep a missed interleaving from turning into a 10 minute package timeout. See #2156.
+const (
+	scannerObserveTimeout = 30 * time.Second
+	hookReleaseTimeout    = 30 * time.Second
```
AkramBitar: `hookReleaseTimeout = 30s` exceeds `stopTimeout = 10s` (`locker.go:36`), which works against the PR's stated goal of bounding these waits.

On the failure path the test `t.Fatal`s before `close(release)`, leaving the scan goroutine parked inside the hook — it cannot observe `ctx.Done()` from there. So `t.Cleanup`'s `d.Stop()` always burns its full 10s and returns an `ErrTimeout` that is discarded, and the goroutine then runs its delete phase up to 20s later, during other tests in the package. Combined with `scannerObserveTimeout = 30s`, a broken interleaving costs ~40s rather than failing in seconds.

Setting both bounds below `stopTimeout` would let `Stop` actually join the goroutine.

token/services/selector/simple/inmemory/locker_test.go
```diff
@@ -124,14 +132,31 @@ func TestScannerDoesNotDeleteReclaimed(t *testing.T) {
 			// only the first observer (the scanner) blocks; later lookups
 			// (the reclaim's) must pass through or they would deadlock
 			close(entered)
-			<-release
+			// Bounded: if the test fails before reaching close(release), this
+			// goroutine must not sit here forever. Note reclaim() calls
```
AkramBitar: This comment is stale as of the next commit in this same PR: `reclaim()` is deleted and `GetStatus` is moved out of the shard lock. As written it names a function that no longer exists and asserts the exact opposite of the invariant the PR establishes, so a future reader will believe the hazard is still present and reason about the hook's blocking incorrectly.

Worth rewording to say the hook may block the *collector's* lookup phase, which is where the shard lock is genuinely not held.

token/services/selector/simple/inmemory/locker_test.go
```diff
@@ -95,6 +95,14 @@ func (m *mockTXStatusProvider) GetStatus(_ context.Context, txID string) (ttxdb.
 // transaction before the scanner deletes, the scanner must NOT delete the
 // new entry. The test drives the real scan loop and blocks the scanner's
 // status lookup to open the race window deterministically.
+// Bounds for the coordination in TestScannerDoesNotDeleteReclaimed. Both are far longer than the
```
AkramBitar: This `const` block landed **between** `TestScannerDoesNotDeleteReclaimed`'s existing doc comment and the function itself. The paragraph about verifying the TOCTOU protection in the scanner now documents `scannerObserveTimeout`/`hookReleaseTimeout`, and the test is left undocumented — godoc and revive will both attribute it to the constants. Moving the const block above the original comment fixes it.

AkramBitar (commented):
@atharrva01 

Great work thanks a lot.

Left some comments.

AkramBitar (changes requested):

AkramBitar (approved):

PR #2159, opened 2026-08-07
feat(evm): Besu test network and the fungible suite running on it

atharrva01:
Stands up an EVM-backed test network and runs the existing fungible suite against it, with the shared test bodies unchanged.

The network handler boots a Besu node in a container, deploys the contracts through the project's own forge script, generates an endorser identity per endorsing node plus a funded submitter, and writes each node a token configuration pointing at what it just deployed. The rendered config is parsed back by the driver's own `LoadConfig` in a test, so the harness and the driver cannot drift apart.

On the driver side this fills in what the suite needed and was missing:

- local membership, so the token drivers can read a default identity while building a TMS
- a per-TMS endorsement service, resolved lazily from the TMS the network is asked to approve for, since it needs that TMS's validator
- the public-params fetcher, so a node that starts after the network can catch up
- endorser registration at startup rather than on first approval, because an endorser answers requests without ever making one
- a watcher for public-parameters updates: they arrive as an endorsed setup delta somebody else submitted, so there is nothing local to trigger off and a node kept serving what it started with

A few things only showed up once real nodes were talking to each other:

- the endorsement config names nodes, but a session is opened to an identity and the allowlist is compared against the identity a session authenticated. Names are now resolved through the identity provider, the way the fabric endorsement service does.
- every node was being handed the whole network's wallets instead of the ones it was issued, which left it without a usable identity of its own
- Besu does not implement `eth_maxPriorityFeePerGas`. It is an extension rather than part of the JSON-RPC spec, so the submitter falls back to the part of `eth_gasPrice` above the base fee.
- contracts are compiled for paris. Besu's dev network does not support PUSH0, so a shanghai build reverts every contract creation. Golden digests and the forge tests are unchanged by the switch.

Public parameters are updated through an endorsed setup delta, since the TokenState contract has no administrative setter. A factory creates and seeds a clone in one transaction, so there is no window between the two for anyone else to initialize it.

There are two suites, zkatdlog and fabtoken, over the same driver. That is worth the duplication: the two token layers put very different work through one network driver, so a failure under one and not the other says where the problem is not. `make integration-tests-evm` and `make integration-tests-evm-fabtoken` run them. They need docker and forge, and pull the besu image if missing. No FAB_BINS.

The fabtoken suite passes end to end on a real Besu node with `fungible.TestAll` unmodified. The zkatdlog one still stops on a time-window assertion in `CheckAuditedTransactions`; details are in a comment below, and it is being tracked separately rather than held against this branch.

There is also a deployment runbook under `docs/services/network-ethereum-deployment.md`.


Makefile
```diff
@@ -111,6 +111,23 @@ monitoring-docker-images:
...
 
+# The EVM integration suite runs against this Besu image.
+BESU_IMAGE ?= hyperledger/besu:24.3.0
+
+.PHONY: integration-tests-evm
+# run the fungible integration tests against an EVM backend (Besu).
+# Unlike the fabric suites this needs no FAB_BINS, but it does need docker and the besu image, which
+# it pulls if missing, plus forge to deploy the contracts.
+integration-tests-evm:
+	docker image inspect $(BESU_IMAGE) >/dev/null 2>&1 || docker pull $(BESU_IMAGE)
```
adecaro: introduce an entry to just fetch this docker imager: besu-docker-images.
You can then call it from `.PHONY: docker-images`.

adecaro (approved):
LGTM

PR #2160, opened 2026-08-07
docs(htlc): document the clock-synchronisation assumption behind HTLC deadlines

### LINKED ISSUE #1750: HTLC claim/reclaim deadline is validated against local time.Now(), making validation non-deterministic
### Summary

  The HTLC claim-vs-reclaim decision is evaluated during transaction validation using the validating node's local wall clock (`time.Now()`), instead of the deterministic ledger/block timestamp. The same transaction can therefore be judged a Claim on one validation and a Reclaim on another, depending on when and where validation runs.

  ### Where
  
  Both validators compute `now` from the wall clock and feed it into the HTLC deadline check:

  - `token/core/fabtoken/v1/validator/validator_transfer.go:143` — `now := time.Now()`
  - `token/core/zkatdlog/nogh/v1/validator/validator_transfer.go:151` — `now := time.Now()`

  That `now` flows into `htlc.VerifyOwner(...)` (`token/services/identity/interop/htlc/validator.go`), which decides the operation from `now.Before(script.Deadline)`:

  - before the deadline → must be a **Claim** (output owner must be the recipient),
  - at/after the deadline → must be a **Reclaim** (output owner must be the sender).

  It also reaches `script.Validate(now)` in the same files.

## Example

Token-request validation runs on multiple peers and at more than one point in a transaction's life:
first during endorsement, then again during commit-time re-validation.

Example:

```
Alice locks 100 TOKEN in an HTLC for Bob.
Deadline: 3:00:00
        ↓
Bob submits the Claim near the deadline
        ↓
Peer 1: clock = 2:59:59
→ Before deadline → Claim is valid ✅
        ↓
Peer 2: clock = 3:00:01
→ After deadline → Claim is expired ❌
        ↓
❌ Same transaction → different validation results

The same problem can also happen on the same peer:
        ↓
Endorsement at 2:59:59 → Claim valid ✅
        ↓
Commit-time re-validation at 3:00:01 → Claim expired ❌
```

  ### Why this matters

  Token-request validation runs on multiple peers and at more than one point in a transaction's life (endorsement, then commit-time re-validation). Using `time.Now()` makes the Claim/Reclaim verdict depend on *which node* and *what moment* the check runs:

  - **Cross-peer clock skew:** for a spend submitted near the deadline, one endorser (clock just before) validates it as a Claim and accepts, while another (clock just after) expects a Reclaim and rejects. Endorsement becomes inconsistent.
  - **Endorse vs commit on the same peer:** a spend valid as a Claim at endorsement time can flip to "should have been a Reclaim" by commit-time re-validation, or vice versa.

  For HTLC-based atomic swaps this is a safety problem, not just a robustness one: the deadline is the hinge of the swap, and a fuzzy/non-deterministic deadline can let a claim through past the intended cutoff or let a claim and a reclaim race.

  ### Note: the fix hook already exists but is unused

  `htlc.Verifier` has a `ClockFunc func() time.Time` field (`token/services/interop/htlc/signer.go:96-100`) whose own doc says callers with the Fabric block timestamp should inject it "to ensure deterministic, clock-skew-safe deadline enforcement." `ClockFunc` is currently never set anywhere, and the validators bypass it by calling `time.Now()` directly.

AkramBitar:
Fixes #1750

## Issue

An HTLC is claimable before its deadline and reclaimable after it, and every validating node decides which side a spend falls on by reading its own clock. Endorsers therefore agree only insofar as their clocks agree — and that assumption was nowhere written down. Nothing in Panurus or Fabric Smart Client verifies, enforces, or monitors clock synchronisation (FSC has no NTP mechanism at all), so an operator had no way to know the requirement existed.

## Fix (docs only)

We assume endorsing nodes keep synchronised clocks, so no code change is needed — the earlier transaction-timestamp approach has been reverted. New [`docs/security/htlc_deadline_clock_assumptions.md`](docs/security/htlc_deadline_clock_assumptions.md) states the assumption, gives NTP/chrony as the example mechanism with typical accuracy figures, explains why it suffices and what a violation costs (endorsement liveness, not token safety), and gives the deadline-margin guidance that follows.

Also re-points `docs/README.md` and `docs/services/interop.md`, and adds a Security section to the mkdocs nav, which previously omitted `docs/security/` entirely.

**125 lines added, docs only. No code changes.**

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -13,6 +13,11 @@ Welcome to Panurus documentation.
 *   [**Upgradability**](upgradability.md): How to upgrade tokens, drivers, and storage.
 *   [**Public Parameters Lifecycle**](public_parameters.md): How public parameters are generated, published, and updated across the network.
 
+## Security
+
+*   [**HTLC Deadlines and Clock Synchronisation**](security/htlc_deadline_clock_assumptions.md): The clock-synchronisation assumption that the HTLC claim/reclaim deadline rests on, and the deadline margin it requires of a deployment.
+*   [**Selector Resource Limits**](security/selector_resource_limits.md): How to throttle token selection by supplying a custom `Locker`.
+
 ## Command-Line Tools
 
 Panurus ships several standalone CLI tools, each living in its own Go module under `cmd/`.
--- a/docs/security/htlc_deadline_clock_assumptions.md
+++ b/docs/security/htlc_deadline_clock_assumptions.md
@@ -1,0 +1,54 @@
+# HTLC Deadlines and Clock Synchronisation
+
+An HTLC-locked token can be spent in one of two ways, decided by a deadline:
+
+* **before** the deadline — only a **claim** is legal (tokens to the recipient, who must reveal the
+  hash preimage);
+* **at or after** it — only a **reclaim** is legal (tokens back to the sender).
+
+Every validating node decides which side of the deadline a spend falls on **by reading its own
+clock**. Nothing in a Panurus transaction carries a time that validation could use instead, so
+nodes agree on the verdict only insofar as their clocks agree.
+
+## Requirements
+
+1. **Every validating node MUST run clock synchronisation.** This covers every endorser and the
+   token chaincode on every endorsing peer. NTP (`ntpd`) or chrony is the usual choice; PTP or a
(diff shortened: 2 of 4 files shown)


Effi-S (approved):
LGTM

PR #2161, opened 2026-08-07
fix(identity): bound nesting depth and fan-out when deserializing composite identities

### LINKED ISSUE #2076: identity/deserializer: unbounded recursion in the verifier multiplex on untrusted identity input
### Summary

The verifier deserializer multiplex is wired to itself for every composite identity type
(`multisig`, `boolpolicy`, `htlc`), and none of the recursive deserializers track or cap recursion
depth. A deeply nested or widely branching composite identity can therefore drive an unbounded
amount of recursive work during deserialization, before any signature is checked.

### Where

`token/core/zkatdlog/nogh/v1/driver/deserializer.go:48-50` (and the equivalent in
`token/core/fabtoken/v1/driver/deserializer.go:33`):

```go
des := deserializer.NewTypedVerifierDeserializerMultiplex()
...
des.AddTypedVerifierDeserializer(htlc2.ScriptType, htlc.NewTypedIdentityDeserializer(des))
des.AddTypedVerifierDeserializer(multisig.Multisig, multisig.NewTypedIdentityDeserializer(des, des))
des.AddTypedVerifierDeserializer(boolpolicy.Policy, boolpolicy.NewTypedIdentityDeserializer(des, des))
```

The multiplex is registered as the component deserializer for all three composite types, including
for itself. `multisig/deserializer.go:115-137` loops over `multisigIdentity.Identities` and calls back
into the multiplex recursively; `boolpolicy` and `interop/htlc` follow the same shape. There is no
depth counter anywhere in this call chain (`grep -rn "depth"` over `deserializer/`, `multisig/`,
`interop/htlc/` returns nothing), and `validateComponentIdentities` only rejects duplicate identities
within a single nesting level, so a composite identity with distinct components at each level is not
caught by that check.

### Impact

`GetOwnerVerifier` is called from the transfer validator once per input token
(`token/core/zkatdlog/nogh/v1/validator/validator_transfer.go:64`), which means a maliciously
constructed owner identity reaches this recursive deserialization before signature verification. In
Go, sufficiently deep recursion exhausts the goroutine stack and terminates the process with an
unrecoverable fatal error rather than a panic that `recover()` can catch. This makes the lack of a
depth/fan-out bound worth closing as a hardening item, since the deserialization path is on the
validation critical path and takes untrusted input.

### Suggested fix

Thread a bounded recursion depth through `context.Context` (already the first parameter on every
method in this call chain — `DeserializeVerifier`, `GetAuditInfo`, `GetAuditInfoMatcher`, `Match`).
A small helper, e.g. `deserializer.EnterDeserialization(ctx) (context.Context, error)`, could
increment an ambient counter and return an error past a reasonable cap (real deployments nest at most
2-3 levels, e.g. a policy over multisig over x509). Call it at the top of each recursive step in the
`multisig`, `boolpolicy`, and `htlc` deserializers, and at the returned matcher's `Match` (which
recurses independently after deserialization). It would also help to cap the number of components per
composite identity in `validateComponentIdentities` for both `multisig` and `boolpolicy`, since depth
alone doesn't bound fan-out. This can be done without changing the `TypedVerifierDeserializer`
interface signature, keeping the change local to the affected implementations.

### Severity

HIGH — reachable on the validation path with untrusted input, worth hardening even though it
requires a specifically constructed identity to exercise.

AkramBitar:
Fixes #2076

## Problem

A token's owner field can be a *composite* identity whose components are themselves identities:
`multisig` (N co-owners), `boolpolicy` (a boolean expression over N identities), `htlc` (a script's
sender and recipient). Each driver's `NewDeserializer` registers the verifier multiplex as the
component deserializer for all three — including for itself:

```go
des := deserializer.NewTypedVerifierDeserializerMultiplex()
...
des.AddTypedVerifierDeserializer(htlc2.ScriptType, htlc.NewTypedIdentityDeserializer(des))
des.AddTypedVerifierDeserializer(multisig.Multisig, multisig.NewTypedIdentityDeserializer(des, des))
des.AddTypedVerifierDeserializer(boolpolicy.Policy, boolpolicy.NewTypedIdentityDeserializer(des, des))
```

That self-registration is what makes composite identities compose. It is also what makes the
recursion unbounded, in four independent chains, with nothing counting depth or fan-out:

| Chain | Entry point |
| --- | --- |
| Verifier deserialization | `DeserializeVerifier` in all three packages |
| Matcher construction | `GetAuditInfoMatcher` (`multisig`, `boolpolicy`) |
| Matcher evaluation | `InfoMatcher.Match` (`multisig`, `boolpolicy`), `AuditInfoMatcher.Match` (`htlc`) |
| Audit-info collection | `GetAuditInfo` in all three packages |

Step by step, as an attacker:

1. Own any token, any amount.
2. Build a transfer spending it, with an owner identity that is a `multisig` nested inside a
   `multisig` inside a `multisig`, as deep as you care to type.
3. Submit it. No permission needed — anyone can submit.
4. A validating peer calls `GetOwnerVerifier` once per input token
   (`validator_transfer.go`) to work out *whose signature to demand*.
5. Answering that means opening the identity, which means descending every level.
6. `MultiIdentity.Deserialize` is `asn1.Unmarshal` over the whole remaining subtree at every
   level, so a depth-`d`, size-`n` identity costs **O(n²)** — times up to `MaxInputs` tokens.
7. All of it lands **before any signature is checked**, so no authorization gates it.

`validateComponentIdentities` only rejected empties and intra-level duplicates, so distinct
components at each level walked straight through, and fan-out was uncapped.

### On severity

The issue describes an unrecoverable Go stack overflow. That part does not reproduce:
`MaxRequestBytes` (256 KiB) already caps nesting at roughly 8k levels, well short of exhausting a
1 GiB goroutine stack. What *is* reachable is the quadratic re-parsing above — CPU amplification
ahead of signature verification, not a crash. So this reads closer to **MEDIUM** than HIGH. Still
worth closing, and the existing size limit should not be the only thing standing in the way.

The recursion being genuinely unbounded is not theoretical: with the new guards stubbed out, the
end-to-end test package below does not fail — it **hangs**.

## Example 

Composite identities can be nested in two main forms:

```
1. Multisig — N co-owners

Multisig(Alice, Bob, Multisig(Carol, Dave, Multisig(Eve, Frank)))

Multisig
├── Alice
├── Bob
└── Multisig
    ├── Carol
    ├── Dave
    └── Multisig
        ├── Eve
        └── Frank

2. BoolPolicy — Boolean expression over N identities

BoolPolicy((Alice AND Bob) OR (Carol AND Dave))

          OR
         /  \
       AND  AND
      / \    / \
   Alice Bob Carol Dave

Both structures can be nested recursively, so validation needs limits on:
- MaxIdentityDepth: maximum nesting depth.
- MaxIdentityComponents: maximum components per composite identity.

Depth limits the vertical nesting; component limits the fan-out at each level.
```

## Fix

Two bounds, added to the existing `driver.ResourceLimits`, which already exists for exactly this
class of limit ("resources a validator will spend deserializing and validating an untrusted token
request, before any cryptographic work is performed"):

| Field | Default | Bounds |
| --- | --- | --- |
| `MaxIdentityDepth` | 5 | How deeply composite identities may nest |
| `MaxIdentityComponents` | 16 | How many components one composite identity may carry |

1. Depth rides in `context.Context` (`token/driver/identity_nesting.go`), already the first
   parameter of every method in all four chains — no interface or constructor signature changes.
   This follows `token/driver/validation_time.go`, which threads the validation reference time the
   same way. `driver.EnterCompositeIdentity(ctx)` accounts for one level and returns the context for
   the components; past the limit it errors with `driver.ErrIdentityNestingTooDeep`.
2. Called at the top of every recursive step in all four chains. The matcher paths recurse
   *separately* from the deserialization that built them, so each accounts for its own depth rather
   than inheriting a budget already spent.
3. Fan-out capped in both `validateComponentIdentities`, returning
   `driver.ErrTooManyIdentityComponents`, checked before any component is touched. Both bounds are
   needed: depth does not bound fan-out (one level with thousands of components is a single
   recursive step doing unbounded work), and fan-out does not bound depth.
4. Validators seed the configured limits at each public entry point
   (`(*Validator).withIdentityNestingLimits`).
5. Also applied in `WrapIdentities` / `WrapPolicyIdentity`, so an identity built in-process cannot
   exceed what a validator will later accept.

Real deployments nest 2–3 levels — a policy over a multisig over x509 — comfortably inside the
default, and a test asserts that shape is *not* rejected.

**Depth is per-path, not global.** `context.WithValue` returns a new context per descent, so sibling
components each start from their parent's depth instead of sharing a running total. That is the
correct semantics for a depth bound, and it is precisely why the fan-out bound is needed too.

**An unseeded context gets the defaults, not "unlimited."** Composite identity deserialization is
also reachable from paths with no `ResourceLimits` in scope — a wallet resolving a recipient, an
auditor inspecting a request, tests. Defaulting means a seeding site added later and forgotten
weakens the bound to the default rather than switching it off.

## Two adjacent defects the tests surfaced

- `htlc.DeserializeVerifier` replaced its component error with `errors.Errorf` instead of wrapping
  it, so **nothing** underneath an htlc level was testable with `errors.Is` — including the new
  sentinel. Both sites now `errors.Wrap`.
- `ResourceLimits` is only reachable through two providers that map fields **explicitly**
  (`services/config`, `network/fabric/tcc`), so a new field is silently unconfigurable until added
  to both. Both now carry `maxIdentityDepth` / `maxIdentityComponents` and
  `TOKEN_VALIDATION_MAX_IDENTITY_{DEPTH,COMPONENTS}`.

## Consensus note

These bounds change which requests a validator accepts, so they are consensus-relevant. No public
parameters version bump: `driver.ResourceLimits` already established the convention — configurable,
defaults every peer shares, and the keep-your-peers-in-sync requirement documented on the struct.
This follows that precedent rather than adding a new mechanism.

## Tests

- `nesting_test.go` in `multisig` and `boolpolicy` — each of the four chains at `limit` and
  `limit+1`, depth-is-per-path, and unseeded-context-still-bounded.
- `deserializer_nesting_test.go` (`fabtoken/v1/driver`) — the same against the **real assembled
  multiplex**, including composite types alternating so no type gets a fresh budget, plus the
  realistic three-level shape passing the bound.
- Every new test was confirmed to fail with the guards stubbed out; the end-to-end package hangs.
- `FuzzOwnerVerifierNoPanic` — fuzzes the owner deserialization path itself, seeded with identities
  nested 1 to 600 levels deep. 400k+ execs clean, and **wired into
  `.github/workflows/nightly-fuzz.yml`**.
- `make checks` and `make lint` clean; `-race` clean on the touched packages.

Pre-existing and unrelated: `TestTranslatePath`
(`token/services/identity/config`) asserts an absolute path contains `"panurus"` and fails in any
checkout whose directory is named otherwise.

## Docs

`docs/drivers/validation-resource-limits.md` gains an enforcement-points section covering the
composite types, the four chains, the per-path semantics and the unseeded-context behaviour;
`docs/configuration.md` gains both keys with defaults.

## Note on the branch

Cut from `origin/main` rather than the local `2076` branch, which carries the commit behind #2160
(htlc deterministic deadlines) — unrelated to this fix, so it is kept out and this reviews on its
own. One consequence: the `fabric-fsc-channel-header-timestamp` nightly-fuzz entry belongs to #2160
and is deliberately not duplicated here.

### DIFF
--- a/.github/workflows/nightly-fuzz.yml
+++ b/.github/workflows/nightly-fuzz.yml
@@ -99,6 +99,9 @@ jobs:
           - name: fabricx-endorser-for-threshold-rule
             pkg: ./token/services/network/fabricx/endorsement
             func: FuzzEndorserForThresholdRuleNoPanic
+          - name: fabtoken-owner-verifier
+            pkg: ./token/core/fabtoken/v1/driver
+            func: FuzzOwnerVerifierNoPanic
 
     steps:
       - name: Checkout code
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -129,6 +129,8 @@ token:
       maxMetadataKeyBytes: 256
       maxMetadataValueBytes: 4096
       maxProofBytes: 131072
+      maxIdentityDepth: 5
+      maxIdentityComponents: 16
 
   # optional global SQL table name overrides (applied to all TMS instances).
   # The value replaces the short code; the FSC-generated prefix and params still wrap it.
@@ -512,6 +514,8 @@ token:
       maxMetadataKeyBytes: 256
       maxMetadataValueBytes: 4096
       maxProofBytes: 131072
+      maxIdentityDepth: 5
+      maxIdentityComponents: 16
 ```
 
 Default values:
@@ -527,6 +531,10 @@ Default values:
 - maxMetadataKeyBytes: 256
 - maxMetadataValueBytes: 4096 (4 KiB)
 - maxProofBytes: 131072 (128 KiB) — ignored by drivers without a zero-knowledge proof (fabtoken)
+- maxIdentityDepth: 5 — how deeply composite owner identities (multisig, policy, HTLC script) may
+  nest inside one another. Real deployments nest 2–3 levels, e.g. a policy over a multisig over x509
+- maxIdentityComponents: 16 — how many component identities a single composite identity may carry.
+  Bounds fan-out, which maxIdentityDepth does not
 
 Every field is optional; any field omitted (or the whole `token.validation.limits` key omitted)
 resolves to its default. Read via the config service, so this key applies only to the FSC/DI
--- a/docs/drivers/validation-resource-limits.md
+++ b/docs/drivers/validation-resource-limits.md
@@ -10,11 +10,11 @@ The token request validators (`token/core/common`, and the fabtoken/zkatdlog dri
 of it) accept raw, attacker-controlled bytes over the network. Aside from the signing anchor
 (`driver.MaxAnchorSize`), nothing else bounds the size of the raw request, the number of actions or
 signatures, the size of an individual action or signature, the number of inputs/outputs/metadata
-entries in an action, or the length of a zero-knowledge proof — unless these limits are enforced.
-Without them, an attacker could force unbounded allocations
-(`make([]..., len(attackerControlledCount))`) and expensive cryptographic work (proof
-deserialization, ZK verification) purely by shaping the wire bytes, without needing any valid
-signature.
+entries in an action, the length of a zero-knowledge proof, or how deeply a composite owner identity
+nests — unless these limits are enforced. Without them, an attacker could force unbounded allocations
+(`make([]..., len(attackerControlledCount))`), unbounded recursion (a composite identity nested
+inside itself), and expensive cryptographic work (proof deserialization, ZK verification) purely by
+shaping the wire bytes, without needing any valid signature.
 
 ## Configuration mechanism
 
@@ -37,6 +37,8 @@ Two sources resolve a `driver.ResourceLimits` value at composition-root time, bo
       limits:
         maxActions: 128
         maxProofBytes: 65536
+        maxIdentityDepth: 5
+        maxIdentityComponents: 16
   ```
 
   Every field is optional; an entirely absent `token.validation.limits` key resolves to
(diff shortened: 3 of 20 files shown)


Effi-S (approved):
LGTM

PR #2165, opened 2026-08-09
identity/deserializer: unsynchronized map access in the signer multiplex can crash the process

### LINKED ISSUE #2079: identity/deserializer: unsynchronized map access in the signer multiplex can crash the process
### Summary

The signer deserializer multiplex keeps its per-type deserializer lists in a plain Go map that is
written under one mutex and read without any lock, so a concurrent registration and lookup can race
and trigger Go's "concurrent map read and map write" fatal error.

### Where

`token/services/identity/deserializer/signer.go:23` declares
`deserializers map[IdentityType][]TypedSignerDeserializer`. Writes happen via
`AddTypedSignerDeserializer` (`signer.go:30-38`), called from `membership/lm.go:818` while holding
`localIdentitiesMutex`. Reads happen unlocked at `signer.go:45`, from
`Provider.getSignerAndCache` (`provider.go:325`). These are two different mutexes, so the lock held
during registration provides no protection for the concurrent read path. Registration can happen at
runtime via `refreshAndGet` (`lm.go:889`), so a live `GetSigner` call can genuinely race a new
registration. `verifier.go:26` has the same shape, and `eidrh.go:26`'s `AddDeserializer` has no
mutex at all.

### Impact

A registration happening concurrently with a lookup on the same multiplex can trigger Go's runtime
map-race detector, which terminates the process — this is a stability concern under concurrent
identity-registration workloads, independent of any race detector build (the runtime check is always
active for maps, not just under `-race`).

### Suggested fix

Guard map access consistently — either take the same mutex used for writes on the read path (a small
read-lock around the lookup), or switch to a concurrency-safe map type if lock contention on the read
path is a concern given how frequently `getSignerAndCache` is called. Apply the same fix to
`verifier.go` and add a mutex to `eidrh.go`'s `AddDeserializer`, since all three share the same
pattern.

### Severity

HIGH — a data-race that can crash the process, though it requires registration and lookup to overlap
in time to manifest.

Effi-S:
Fixes #2079

### Summary

The signer deserializer multiplex keeps its per-type deserializer lists in a plain Go map that is
written under one mutex and read without any lock, so a concurrent registration and lookup can race
and trigger Go's "concurrent map read and map write" fatal error.

### Where

`token/services/identity/deserializer/signer.go:23` declares
`deserializers map[IdentityType][]TypedSignerDeserializer`. Writes happen via
`AddTypedSignerDeserializer` (`signer.go:30-38`), called from `membership/lm.go:818` while holding
`localIdentitiesMutex`. Reads happen unlocked at `signer.go:45`, from
`Provider.getSignerAndCache` (`provider.go:325`). These are two different mutexes, so the lock held
during registration provides no protection for the concurrent read path. Registration can happen at
runtime via `refreshAndGet` (`lm.go:889`), so a live `GetSigner` call can genuinely race a new
registration. `verifier.go:26` has the same shape, and `eidrh.go:26`'s `AddDeserializer` has no
mutex at all.

### Impact

A registration happening concurrently with a lookup on the same multiplex can trigger Go's runtime
map-race detector, which terminates the process — this is a stability concern under concurrent
identity-registration workloads, independent of any race detector build (the runtime check is always
active for maps, not just under `-race`).

### Suggested fix

Guard map access consistently — either take the same mutex used for writes on the read path (a small
read-lock around the lookup), or switch to a concurrency-safe map type if lock contention on the read
path is a concern given how frequently `getSignerAndCache` is called. Apply the same fix to
`verifier.go` and add a mutex to `eidrh.go`'s `AddDeserializer`, since all three share the same
pattern.

### Severity

HIGH — a data-race that can crash the process, though it requires registration and lookup to overlap
in time to manifest.

### DIFF
--- a/token/services/identity/deserializer/signer.go
+++ b/token/services/identity/deserializer/signer.go
@@ -9,6 +9,8 @@ package deserializer
 import (
 	"context"
 	errors2 "errors"
+	"slices"
+	"sync"
 
 	"github.com/LFDT-Panurus/panurus/token/driver"
 	"github.com/LFDT-Panurus/panurus/token/services/identity"
@@ -20,6 +22,7 @@ import (
 type TypedSignerDeserializer = idriver.TypedSignerDeserializer
 
 type TypedSignerDeserializerMultiplex struct {
+	mutex         sync.RWMutex
 	deserializers map[idriver.IdentityType][]TypedSignerDeserializer
 }
 
@@ -28,6 +31,8 @@ func NewTypedSignerDeserializerMultiplex() *TypedSignerDeserializerMultiplex {
 }
 
 func (v *TypedSignerDeserializerMultiplex) AddTypedSignerDeserializer(typ idriver.IdentityType, d idriver.TypedSignerDeserializer) {
+	v.mutex.Lock()
+	defer v.mutex.Unlock()
 	_, ok := v.deserializers[typ]
 	if !ok {
 		v.deserializers[typ] = []TypedSignerDeserializer{d}
@@ -42,8 +47,10 @@ func (v *TypedSignerDeserializerMultiplex) DeserializeSigner(ctx context.Context
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to unmarshal to TypedIdentity")
 	}
-	dess, ok := v.deserializers[si.Type]
-	if !ok {
+	v.mutex.RLock()
+	dess := slices.Clone(v.deserializers[si.Type])
+	v.mutex.RUnlock()
+	if dess == nil {
 		return nil, errors.Errorf("no deserializer found for [%v]", si.Type)
 	}
 	logger.DebugfContext(ctx, "deserializing [%s] with type [%v]", logging.Base64(id), si.Type)
--- a/token/services/identity/deserializer/deserializer_test.go
+++ b/token/services/identity/deserializer/deserializer_test.go
@@ -9,6 +9,7 @@ package deserializer
 import (
 	"context"
 	"errors"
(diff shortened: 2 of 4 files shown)


token/services/identity/deserializer/signer.go
```diff
@@ -20,6 +21,7 @@ import (
 type TypedSignerDeserializer = idriver.TypedSignerDeserializer
 
 type TypedSignerDeserializerMultiplex struct {
+	mutex         sync.RWMutex
 	deserializers map[idriver.IdentityType][]TypedSignerDeserializer
 }
```
AkramBitar, line 27: Lock is released before `dess` is iterated. A concurrent `append` on this type's slice may allocate a new backing array; this reader holds the old slice header and will miss the new entry. Consider holding the lock through the iteration or copying the slice under the lock before releasing.

AkramBitar (commented):
The read lock is released before the slice is iterated in three places (signer.go, verifier.go, eidrh.go). A concurrent writer calling `append` may allocate a new backing array at that moment; the reader holds a copy of the old slice header and misses the new entry. This is safe (no crash) but creates a stale-read window.

Consider either holding the lock through the iteration, or copying the slice under the lock before releasing:
```go
v.mutex.RLock()
dess := make([]TypedSignerDeserializer, len(v.deserializers[si.Type]))
copy(dess, v.deserializers[si.Type])
v.mutex.RUnlock()
```

AkramBitar (changes requested):

AkramBitar (approved):
LGTM

Issue #2143, opened 2026-08-04
Audit recovery sweep runs on every replica: leadership always granted and claims not persisted


atharrva01:
## Problem

The recovery sweep for audit storage runs on every replica at once, with neither of the two mechanisms that protect the owner transaction path.

A recovery manager is started per TMS for audit storage in `token/services/network/fabric/network.go:471-478`, alongside the one for owner storage. It goes through `auditdb.StoreService.AcquireRecoveryLeadership` and then `ClaimPendingTransactions`. On the audit path both of those are no-ops:

**1. Leader election is always granted.**
`postgres.NewAuditTransactionStore` builds its base through
`sqlcommon.NewAuditTransactionStore` -> `NewOwnerTransactionStore`, which calls
`newTransactionStore(..., nil, nil)`. The last argument is the recovery leader factory, so it is
nil. `AcquireRecoveryLeadership` then takes this branch every time
(`db/sql/common/transactions.go:375`):

```go
if db.recoveryLeaderFactory == nil {
    return noopRecoveryLeadership{}, true, nil
}
```

Every replica is told it is the leader. By contrast `NewTransactionStoreWithNotifier`, used for owner storage, passes a real advisory-lock factory.

**2. Claims are not persisted.**
`postgres.TransactionStore` overrides `ClaimPendingTransactions` with the atomic
`UPDATE ... RETURNING` implementation (`db/sql/postgres/transactions.go:127`).
`postgres.AuditTransactionStore` does not override it, it only overrides `WriteDB`, `GetSchema`
and `CreateSchema`, so it inherits the common implementation
(`db/sql/common/transactions.go:389`), which is a plain `SELECT` whose own comment describes it as
permissive and not persisting recovery claims.

## Impact

With N replicas, each audit recovery sweep selects the same set of pending audit transactions and every replica processes all of them, on every tick. The work is duplicated N times, and the lease and owner fields the recovery manager passes are not doing anything on this path.

It also reads as though it is protected. `AcquireRecoveryLeadership` is called and its result checked, so someone reading `recovery/manager.go` would reasonably assume leader election is in force for audit as well as owner storage.

## Notes

- The owner path is unaffected, it has both the advisory lock and the atomic claim.
[...]

PR #2167, opened 2026-08-10
fix(recovery): persist audit recovery claims

atharrva01:
Fixes the second half of #2143, the one that causes the duplicated work.

`postgres.AuditTransactionStore` only overrode `WriteDB`, `GetSchema` and `CreateSchema`, so it
inherited `ClaimPendingTransactions` and `ReleaseRecoveryClaim` from `sqlcommon`. Those are a
plain `SELECT` and a no-op, so no claim is ever persisted and every replica selects the same
pending audit transactions on every tick and processes all of them. The audit store now uses the
same atomic `UPDATE ... RETURNING` claim the owner store has had since it was introduced.

The SQL is identical for the two stores and only the requests table differs, so it moves into a
shared `recoveryClaimStore` that both hold as a field. It is a named field and not an embedded
one on purpose: `sqlcommon.TransactionStore` provides the same two methods, and embedding both at
the same depth makes the selectors ambiguous, which drops them from the method set and quietly
stops the store satisfying `driver.AuditTransactionStore`. Explicit forwarding keeps the override
visible where it is called.

No schema change. `recovery_claimed_by`, `recovery_claim_expires_at` and the two supporting
indexes already come from the shared `sqlcommon` schema that audit storage uses.

### Left for a follow-up

The other half of #2143, leadership always being granted on the audit path, is not fixed here.
That needs a lock id derived per store, which is exactly what #2085 introduces, and doing it on
top of `main` would mean inventing a second scheme that competes with the one already in review.
Worth noting that both recovery managers currently read the same `AdvisoryLockID` from config, so
handing audit a real factory today would make owner and audit recovery contend on one lock and
starve each other.

This is still worth landing on its own. With the atomic claim in place, each pending audit
transaction goes to exactly one replica even while every replica believes it is the leader, so
the duplicated processing is gone. What remains is N cheap claim queries per tick instead of one.

### Question

`CleanupExpiredClaims` is not exposed on the audit store. Nothing calls it on either path, and
the claim query already reclaims expired leases inline through `recovery_claim_expires_at < NOW()`,
so adding it looked like new dead code. Happy to add it for symmetry if you would rather have it.

### Tests

Five new tests in `audit_recovery_claim_test.go` mirroring the owner ones: claim exclusivity
across two replicas, lease expiry, release making a row immediately available again, release
under the wrong owner being ignored, and one pinning that audit and owner claims stay in their
own requests tables. The existing owner tests pass unchanged, `ageRequests` just takes the claim
store now so both suites can share it.


AkramBitar (changes requested):
@atharrva01 

Please see my comments. 

Regards,
Akram

token/services/storage/db/sql/postgres/transactions.go
```diff
@@ -49,10 +50,64 @@ func (s *AuditTransactionStore) CreateSchema() error {
...
+// same pending rows on every sweep and processes all of them. The owner path has had the
+// atomic claim since it was introduced; audit was simply never given it.
+func (s *AuditTransactionStore) ClaimPendingTransactions(ctx context.Context, params tokensdriver.RecoveryClaimParams) ([]*tokensdriver.RecoveryClaim, error) {
+	return s.claims.claimPending(ctx, params)
+}
+
+// ReleaseRecoveryClaim releases the recovery claim on an audit transaction.
+// The inherited implementation is a no-op, which would leave the claim set until its lease
+// expired even after the sweep finished with the row.
+func (s *AuditTransactionStore) ReleaseRecoveryClaim(ctx context.Context, txID string, owner string, message string) error {
```
AkramBitar: **Address before merge.** Giving the audit store a real `ReleaseRecoveryClaim` makes the recovery loop overwrite `status_message` on the audit requests row, destroying the reason an auditor actually cares about.

Mechanism:
- `recoverTransaction` always passes a non-empty message (`manager.go:404-416`): `"recovered successfully"`, `"recovery failed: %v"`, or the Orphan message.
- `releaseClaim` (`transactions.go:259-278`) matches only on `tx_id` + `recovery_claimed_by` and, when `message != ""`, does `Set("status_message", message)`. Nothing re-checks the status.
- `sqlcommon.TransactionStore.SetStatus` (`token/services/storage/db/sql/common/transactions.go:494-519`) never clears `recovery_claimed_by`, so the claim is still ours when the release runs.
- `TTXRecoveryHandler.applyFinalityLogic` (`token/services/ttx/finality/recovery.go:146-153`) writes the ledger's reason **synchronously, before** the release.

Verified against a live PostgreSQL 16 instance using this branch: claim `tx1` as `i1`, then `SetStatus(tx1, Deleted, "ledger rejected: bad signature")`, then `ReleaseRecoveryClaim(tx1, "i1", "recovered successfully")` → `GetStatus` returns status `Deleted` with message `"recovered successfully"`. The rejection reason is gone from the audit trail.

The more common case is worse: `Recover` returns nil for a Busy/Unknown tx (`token/services/ttx/finality/recovery.go:136-143`), so **every still-pending audit row gets stamped `"recovered successfully"` on every sweep while remaining `Pending`**. `TestAuditReleaseRecoveryClaim:167-168` pins exactly that as the expected state.

This was harmless before this PR precisely because the inherited audit release was a no-op. Suggest not writing `status_message` from the audit release path at all, or gating it on the row still being `Pending` and the message not being a success placeholder.

token/services/storage/db/sql/postgres/transactions.go
```diff
@@ -156,8 +211,9 @@ func (db *TransactionStore) ClaimPendingTransactions(ctx context.Context, params
 		db.tables.Requests,
 	)
 
-	// Convert lease duration to PostgreSQL interval format
-	leaseInterval := fmt.Sprintf("%d seconds", int(params.LeaseDuration.Seconds()))
+	// Convert lease duration to PostgreSQL interval format. Milliseconds, not seconds:
```
AkramBitar, line 229: **Address before merge.** *(This comment is about line 203 — the self-reclaim term in the `claimPending` WHERE clause. That line is unchanged context, so GitHub only lets me anchor to the nearest diffed line.)*

Treating a row as free when it is already claimed by the calling replica's own id has no remaining purpose, and it is the one thing that can break exclusivity on the leaderless audit path this PR is enabling.

Mechanism:
- Nothing in the tree renews a lease — there is no renew or extend call anywhere — so the "that last case is what makes a lease renewal safe" rationale in `docs/services/storage/recovery.md:64-66` describes a code path that does not exist.
- `runSweep` is strictly sequential within one manager (`manager.go:174-186` drives it off a single ticker in one goroutine), so a manager never legitimately re-claims a row it already holds.
- On a fresh `Start()` the instance id is a new UUID, so a restarted replica does not need it either.

Verified against a live PostgreSQL 16 instance on this branch: two consecutive `claimPending` calls with the same owner both return the row (the second call returned 1 row, not 0).

Concrete failure: an operator sets `instanceID` explicitly and ships it in a shared ConfigMap, so every replica presents the same owner string. The audit store is built without a leader factory (`NewAuditTransactionStore` passes no `recoveryLeaderFactory`, so `AcquireRecoveryLeadership` grants leadership to everyone), so there is no advisory-lock backstop — every replica claims and concurrently processes the same rows, which is the original #2143 behaviour, now silent. `docs/configuration.md:914` actively recommends that configuration.

Two ways to close it: remove the self-reclaim term, or append a per-process random suffix to a configured `instanceID` in `manager.go:108-114` so a shared config value can never collide.

token/services/storage/db/sql/postgres/transactions.go
```diff
@@ -156,8 +211,9 @@ func (db *TransactionStore) ClaimPendingTransactions(ctx context.Context, params
...
+	// truncating a sub-second lease to "0 seconds" claims a row that is already expired.
+	leaseInterval := fmt.Sprintf("%d milliseconds", params.LeaseDuration.Milliseconds())
```
AkramBitar, line 231: Lower severity. This narrows the truncation window but does not close it.

`validateConfig` (`manager.go:204-219`) only requires that the lease duration be greater than zero. A lease shorter than a millisecond therefore still renders as a zero-length interval, the expiry is written as already-elapsed, and the lease-expired term in the claim predicate makes every row claimable by every replica on every tick — the exact failure this commit set out to fix, three orders of magnitude further down.

Either move to microsecond resolution, or reject sub-second lease durations in `validateConfig` so the configuration cannot express one.

docs/services/storage/recovery.md
```diff
@@ -48,6 +48,27 @@ PostgreSQL is the recommended database for production multi-instance deployments
 - Supports horizontal scaling with multiple replicas
 - Leader election prevents conflicting recovery attempts
 
+A recovery manager is started per TMS for both owner and audit storage. Both stores use the
+same atomic claim: a pending row is handed to exactly one replica for the duration of its
```
AkramBitar: Lower severity, but this overstates what the claim buys on the audit path, and the "duplicated work" half of #2143 is not actually reduced in the common case.

"a pending row is handed to exactly one replica for the duration of its lease" does not hold, because the lease is cut short: `recoverTransaction` releases the claim as soon as `Recover` returns (`manager.go:414`), and `Recover` returns **nil** for a tx that is not yet final (`token/services/ttx/finality/recovery.go:136-143`, the Busy/Unknown default branch) without changing the row's status. Those are precisely the rows the sweep keeps re-encountering.

Concrete scenario — 3 leaderless audit replicas, one pending tx, 5s `scanInterval`:
1. Replica A's tick claims the row, does its `GetTransactionStatus` round-trip, releases the claim; the row is `Pending` and free again after a few ms.
2. Replica B's tick re-claims the same row and repeats the round-trip.
3. Replica C likewise.

Total network round-trips per scan interval is still N, exactly as before this change. What the claim removes is the *concurrency* — no two replicas run `Commit()` for the same newly-valid tx at the same moment — which is a real and worthwhile win, and worth saying instead. Genuine duplicate-work suppression only kicks in once a sweep's processing time exceeds the scan interval, i.e. for large backlogs.

Suggest rewording lines 51-56 and 61-62 to claim exclusivity *at any instant* rather than for the duration of a lease, and noting that a still-pending row is re-processed once per replica per tick.

docs/services/storage/recovery.md
```diff
@@ -48,6 +48,27 @@ PostgreSQL is the recommended database for production multi-instance deployments
...
+`AcquireRecoveryLeadership` grants leadership locally to every replica. Audit sweeps therefore
+run everywhere at once, and it is the atomic claim rather than leader election that keeps each
+pending audit transaction from being claimed by more than one replica at a time.
+
+The claim query treats a row as free if it is unclaimed, its lease has expired, or it is
+already claimed by the calling replica's own `instanceID` (`recovery_claimed_by`). That last
+case is what makes a lease renewal safe, but it also means exclusivity depends on every replica
+having a distinct `instanceID`. The owner path has the advisory lock as a backstop if two
+replicas ever collided on the same id; the audit path does not, so a collision there would let
+both replicas claim the same rows. Leave `instanceID` unset and let it auto-generate unless you
```
AkramBitar: Lower severity, docs consistency. *(This comment is about `docs/configuration.md:914`, which this PR does not touch and so has no diff line to anchor to; I am commenting here because it is the line it contradicts.)*

In the recovery section of `docs/configuration.md`, the "4. For Multi-Instance Deployments" list still carries the bullet "Consider setting explicit `instanceID` values for easier debugging and monitoring".

That directly contradicts the guidance added here, and it recommends exactly the configuration that breaks audit-side claim exclusivity — see my comment on the self-reclaim term in `claimPending`. An operator reading the config reference rather than this page will do the unsafe thing, most plausibly by stamping one shared `instanceID` across a whole Deployment.

Worth either dropping that bullet or amending it to say the values must be unique per replica (derived from the pod name, say) and never a single shared value.

AkramBitar (changes requested):

AkramBitar (approved):
LGTM

PR #2170, opened 2026-08-10
identity/role: GetWalletID swallows storage errors, so a transient DB blip creates a duplicate wallet

### LINKED ISSUE #2063: identity/role: GetWalletID swallows storage errors, so a transient DB blip creates a duplicate wallet
### Summary

`Registry.GetWalletID` swallows every storage error and returns `("", nil)`, which its callers treat
identically to "no wallet is bound to this identity." A transient storage error (DB blip, timeout)
therefore looks exactly like "not registered yet," and the wallet-lookup fallback chain in `Lookup`
creates a brand-new wallet and a second identity binding for an identity that already has one.

### Where

`token/services/identity/role/registry.go:237-246`:

```go
func (r *Registry) GetWalletID(ctx context.Context, identity driver.Identity) (string, error) {
	wID, err := r.Storage.GetWalletID(ctx, identity, int(r.Role.ID()))
	if err != nil {
		//nolint:nilerr
		return "", nil
	}
	...
	return wID, nil
}
```

All three call sites treat `err == nil && len(wID) == 0` as "not found, try the next fallback":
`registry.go:85-92`, `registry.go:114-117`, `registry.go:131-133`.

### Impact

`Lookup` (`registry.go:72-166`) is the fallback path used by `WalletByID` (`registry.go:271-281`)
whenever the in-memory cache misses. If the storage-level `GetWalletID` call fails transiently (not
"no row found," but an actual error — timeout, connection reset, etc.), the registry cannot
distinguish that from "this identity has never been registered." It falls through to
`WalletFactory.NewWallet` (`registry.go:293`), creating a second wallet and a second identity→wallet
binding for the same identity. This duplicates wallet state and, depending on the wallet
implementation, can duplicate key material bookkeeping.

### Reproduction

Not yet committed. Mock `idriver.WalletStoreService.GetWalletID` to return a non-nil error (e.g. a
simulated transient DB error) on the first call for an identity that has a real binding, and a
successful lookup on a second, independent call. Assert `Registry.GetWalletID` propagates the error
rather than returning `("", nil)`, and that `WalletByID` does not create a duplicate wallet when the
underlying storage error is transient.

### Suggested fix

Propagate the error and let callers decide:

```go
func (r *Registry) GetWalletID(ctx context.Context, identity driver.Identity) (string, error) {
	wID, err := r.Storage.GetWalletID(ctx, identity, int(r.Role.ID()))
	if err != nil {
		return "", errors.Wrapf(err, "failed to get wallet id for identity [%s]", identity)
	}
	...
	return wID, nil
}
```

This requires updating the three call sites in `Lookup` to distinguish "storage error" (abort/retry)
from "no row found" (continue to the next fallback) — today they're merged into one case. This
should land together with #8 (`WalletByID` write-lock scope), since both touch the same
creation path and fixing one without the other leaves the duplicate-wallet risk only partially
addressed.

### Severity

MEDIUM — requires a transient storage failure to trigger, but results in persisted duplicate wallet
state, which is hard to reconcile after the fact.

Effi-S:
Fixes #2063

### Summary

`Registry.GetWalletID` swallows every storage error and returns `("", nil)`, which its callers treat
identically to "no wallet is bound to this identity." A transient storage error (DB blip, timeout)
therefore looks exactly like "not registered yet," and the wallet-lookup fallback chain in `Lookup`
creates a brand-new wallet and a second identity binding for an identity that already has one.

### Where

`token/services/identity/role/registry.go:237-246`:

```go
func (r *Registry) GetWalletID(ctx context.Context, identity driver.Identity) (string, error) {
	wID, err := r.Storage.GetWalletID(ctx, identity, int(r.Role.ID()))
	if err != nil {
		//nolint:nilerr
		return "", nil
	}
	...
	return wID, nil
}
```

All three call sites treat `err == nil && len(wID) == 0` as "not found, try the next fallback":
`registry.go:85-92`, `registry.go:114-117`, `registry.go:131-133`.

### Impact

`Lookup` (`registry.go:72-166`) is the fallback path used by `WalletByID` (`registry.go:271-281`)
whenever the in-memory cache misses. If the storage-level `GetWalletID` call fails transiently (not
"no row found," but an actual error — timeout, connection reset, etc.), the registry cannot
distinguish that from "this identity has never been registered." It falls through to
`WalletFactory.NewWallet` (`registry.go:293`), creating a second wallet and a second identity→wallet
binding for the same identity. This duplicates wallet state and, depending on the wallet
implementation, can duplicate key material bookkeeping.

### Reproduction

Not yet committed. Mock `idriver.WalletStoreService.GetWalletID` to return a non-nil error (e.g. a
simulated transient DB error) on the first call for an identity that has a real binding, and a
successful lookup on a second, independent call. Assert `Registry.GetWalletID` propagates the error
rather than returning `("", nil)`, and that `WalletByID` does not create a duplicate wallet when the
underlying storage error is transient.

### Suggested fix

Propagate the error and let callers decide:

```go
func (r *Registry) GetWalletID(ctx context.Context, identity driver.Identity) (string, error) {
	wID, err := r.Storage.GetWalletID(ctx, identity, int(r.Role.ID()))
	if err != nil {
		return "", errors.Wrapf(err, "failed to get wallet id for identity [%s]", identity)
	}
	...
	return wID, nil
}
```

This requires updating the three call sites in `Lookup` to distinguish "storage error" (abort/retry)
from "no row found" (continue to the next fallback) — today they're merged into one case. This
should land together with #8 (`WalletByID` write-lock scope), since both touch the same
creation path and fixing one without the other leaves the duplicate-wallet risk only partially
addressed.

### Severity

MEDIUM — requires a transient storage failure to trigger, but results in persisted duplicate wallet
state, which is hard to reconcile after the fact.

### DIFF
--- a/token/services/identity/role/registry.go
+++ b/token/services/identity/role/registry.go
@@ -24,6 +24,87 @@ type WalletFactory interface {
 	NewWallet(ctx context.Context, id idriver.WalletID, role idriver.IdentityRoleType, is IdentitySupport, info idriver.IdentityInfo) (driver.Wallet, error)
 }
 
+// WalletIDStatus classifies the outcome of resolving an identity to the wallet id
+// bound to it. It exists so callers branch on an explicit, named state instead of
+// re-deriving intent from the ambiguous "(string, error)" shape — where ("", nil),
+// ("", err) and ("id", nil) each mean something different and the difference is easy
+// to get wrong (see issue #2063).
+type WalletIDStatus int
+
+const (
+	// WalletIDUnknown is the zero value and never returned by GetWalletID. It is the
+	// guard against a WalletIDResolution constructed without going through GetWalletID
+	// (a mock, or a future constructor that forgets to set Status): because callers act
+	// on a fallthrough only when authoritative() is true — Bound or Unbound — this zero
+	// value is treated as a lookup failure, not as a safe "no binding" that would create
+	// a duplicate wallet.
+	WalletIDUnknown WalletIDStatus = iota
+	// WalletIDBound means storage holds a wallet id for the identity. WalletID is set.
+	WalletIDBound
+	// WalletIDUnbound means storage answered authoritatively that the identity has no
+	// wallet binding. This is a definitive, successful miss: it is safe to fall through
+	// to the next resolution step and, ultimately, to create a wallet.
+	WalletIDUnbound
+	// WalletIDFailed means the storage lookup itself failed (timeout, connection reset,
+	// ...), so whether a binding exists is UNKNOWN. Err carries the cause. Callers MUST
+	// NOT treat this as WalletIDUnbound: doing so lets a transient blip masquerade as an
+	// unregistered identity and triggers the creation of a duplicate wallet.
+	WalletIDFailed
+)
+
+// WalletIDResolution is the explicit result of resolving an identity to its bound
+// wallet id. It is the single shared value every wallet-lookup fallback branches on,
+// so the "not found" vs "could not check" distinction is decided once (in GetWalletID)
+// rather than re-inferred at each call site.
+type WalletIDResolution struct {
+	// Status is the outcome of the lookup. Always inspect it via Bound/Unbound/Failed
+	// before reading the other fields, and branch exhaustively: any status that is
+	// neither Bound nor Unbound (Failed, or the WalletIDUnknown zero value) is not
+	// authoritative and must abort the lookup rather than fall through to creation.
+	Status WalletIDStatus
+	// WalletID is the bound wallet id; meaningful only when Status is WalletIDBound.
+	WalletID idriver.WalletID
+	// Err is the underlying storage failure; set only when Status is WalletIDFailed.
+	Err error
+}
+
+// Bound reports whether the identity has a wallet id bound to it.
+func (r WalletIDResolution) Bound() bool { return r.Status == WalletIDBound }
+
+// Unbound reports whether storage answered authoritatively that the identity has no
+// wallet binding. This is the ONLY non-Bound state that a caller may act on by falling
+// through to the next resolution step and, ultimately, wallet creation. Every other
+// state — Failed, or the zero value produced by a WalletIDResolution built without
+// going through GetWalletID — leaves the binding unknown and must abort the lookup.
+func (r WalletIDResolution) Unbound() bool { return r.Status == WalletIDUnbound }
+
+// Failed reports whether the storage lookup failed, leaving the binding unknown.
+// A failed resolution must abort the enclosing lookup, never fall through to creation.
+func (r WalletIDResolution) Failed() bool { return r.Status == WalletIDFailed }
+
+// authoritative reports whether the resolution definitively answers whether a wallet is
+// bound — i.e. it came back from GetWalletID as Bound or Unbound. Any other status
+// (Failed, or the zero-value WalletIDUnknown of a resolution built outside GetWalletID)
+// is non-authoritative: the binding is unknown and the caller MUST abort rather than
+// fall through to wallet creation. This is the guard the WalletIDUnknown zero value was
+// introduced to provide.
+func (r WalletIDResolution) authoritative() bool { return r.Bound() || r.Unbound() }
+
+// abortError returns the error a non-authoritative resolution must abort a wallet lookup
+// with. It preserves the storage cause for a WalletIDFailed and synthesizes one for a
+// zero-value / unknown resolution (whose Err is nil), so an unknown status can never
+// collapse into a nil error — via errors.WithMessagef(nil, ...) returning nil — and be
+// silently mistaken for a successful lookup. It must only be called once Bound and
+// Unbound have been ruled out (i.e. authoritative() is false).
+func (r WalletIDResolution) abortError(id driver.WalletLookupID) error {
+	cause := r.Err
+	if cause == nil {
+		cause = errors.Errorf("non-authoritative wallet id resolution status [%d]", r.Status)
+	}
+
+	return errors.WithMessagef(cause, "failed to lookup wallet [%s]", id)
+}
+
 // Registry manages wallets whose long-term identities have a given role.
 //
 // Concurrency and invariants:
@@ -108,11 +189,17 @@ func (r *Registry) Lookup(ctx context.Context, id driver.WalletLookupID) (driver
 		if ok {
 			r.Logger.DebugfContext(ctx, "lookup failed, check if there is a wallet for identity [%s]", passedIdentity)
 			// is this identity registered
-			wID, err := r.GetWalletID(ctx, passedIdentity)
-			if err == nil && len(wID) != 0 {
-				r.Logger.DebugfContext(ctx, "lookup failed, there is a wallet for identity [%s]: [%s]", passedIdentity, wID)
+			res := r.GetWalletID(ctx, passedIdentity)
+			if !res.authoritative() {
+				// A storage failure — or a resolution that never went through GetWalletID —
+				// leaves the binding unknown; it must not be treated as "not registered", or
+				// a transient blip would fall through to wallet creation and duplicate state.
+				return nil, nil, "", res.abortError(id)
+			}
+			if res.Bound() {
+				r.Logger.DebugfContext(ctx, "lookup failed, there is a wallet for identity [%s]: [%s]", passedIdentity, res.WalletID)
 				// we got a hit
-				walletID = wID
+				walletID = res.WalletID
 				ident = passedIdentity
 				fail = false
 			}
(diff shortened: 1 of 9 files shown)


AkramBitar (approved):
e

AkramBitar (commented):
Still need to review

AkramBitar (changes requested):

token/services/storage/db/kvs/walletdb.go
```diff
@@ -85,9 +85,16 @@ func (s *WalletStore) GetWalletID(ctx context.Context, identity driver2.Identity
 	if err != nil {
 		return "", errors.Wrapf(err, "failed to create key")
 	}
+	// kvs.Get returns an error both for a missing key and for a real store failure.
+	// The WalletStoreService contract requires that "no binding" be reported as ("", nil)
+	// so that callers can distinguish it from a transient storage error; probe with Exists
+	// first (as IdentityExists does) and only Get when a binding is present.
+	if !s.kvs.Exists(ctx, k) {
```
AkramBitar: `kvs.Exists` can't give us the contract this comment promises. It is a bare `bool`, implemented upstream as `len(GetExisting(...)) > 0`, and `GetExisting` swallows store failures:

```go
// fabric-smart-client@v0.17.0/platform/view/services/storage/kvs/kvs.go:110-113
it, err := o.store.GetStateSetIterator(ctx, o.namespace, ids...)
if err != nil {
    return result   // <- error dropped, empty result
}
```

So on a genuine failure (connection reset, timeout, `no such table`) we get `Exists → false` → `GetWalletID → ("", nil)` → `Registry.GetWalletID → WalletIDUnbound`, which this PR documents as "a definitive, successful miss: it is safe to fall through … to create a wallet". That is the #2063 failure mode unchanged — and now harder to spot, because the error is dropped inside FSC instead of in our registry. It is a live path, not hypothetical: the integration identity SDK wires the KVS wallet store (`integration/token/common/sdk/identity/provider.go:25`).

The awkward part is that `kvs.Get` has no sentinel for absence either — it returns `errors.Errorf("state [%s,%s] does not exist", ...)` (kvs.go:172-174). Options as I see them, best first:

- Add a `kvs.ErrNotFound` (or an error-returning `Exists`/`GetIfExists`) upstream in FSC and use it here. Cleanest, and other FSC consumers have the same problem.
- Interim: call `Get` and classify only the "does not exist" error as unbound, propagating everything else. Works today, but matching on an upstream message is fragile — worth a `TODO` pointing at the FSC issue.
- If neither fits in this PR, then this store cannot satisfy the contract yet, and we should say so explicitly (in the `WalletStoreService` doc and in #2063) rather than leave a comment claiming it does.

Either way, a KVS test that makes the underlying store fail and asserts the error propagates would lock this down — right now `TestGetWalletIDNotFound` only covers real absence.

token/services/identity/role/registry.go
```diff
@@ -111,36 +162,48 @@ func (r *Registry) Lookup(ctx context.Context, id driver.WalletLookupID) (driver
 	if ok {
 		r.Logger.DebugfContext(ctx, "no wallet found, check if there is a wallet for identity [%s]", passedIdentity)
 		// is this identity registered
-		passedWalletID, err := r.GetWalletID(ctx, passedIdentity)
-		if err == nil && len(passedWalletID) != 0 {
-			r.Logger.DebugfContext(ctx, "no wallet found, there is a wallet for identity [%s]: [%s]", passedIdentity, passedWalletID)
+		res := r.GetWalletID(ctx, passedIdentity)
+		if res.Failed() {
```
AkramBitar, line 228: Failing closed is right at line 131, but I don't think it is here. At this point `MapToIdentity` has already **succeeded**: `walletID` and `ident` are authoritative, and `wID` is already in `walletIdentifiers`. The `GetWalletID(passedIdentity)` call here is only an optimization — "is this identity already bound to a wallet we can reuse?" — so a failed probe leaves nothing in doubt.

Concretely, after a restart: `WalletByID(ctx, role, aliceIdentityBytes)` misses the cache, `MapToIdentity` returns `("alice")`, this probe hits a transient DB error → we now return an error and the caller's transaction fails. Before this change we fell through to `Role.GetIdentityInfo("alice")` and correctly returned/created wallet `"alice"` — with **no** duplicate risk, because the id came from the role mapping and `WalletByID` re-checks the cache under the write lock before creating anything.

Suggest logging at warn and continuing here (and in the same branch at :190), keeping the mapped candidate:

```go
res := r.GetWalletID(ctx, passedIdentity)
if res.Failed() {
    // Optional probe: the mapped wallet id is already authoritative, so a failed
    // lookup only costs us a reuse opportunity, not correctness.
    r.Logger.Warnf("failed to check wallet binding for identity [%s], continuing with mapped wallet [%s]: %v", passedIdentity, wID, res.Err)
} else if res.Bound() {
    ...
}
```

Line 131 is the one branch where the probe's answer decides whether we create a wallet — that is where fail-closed belongs, and it is correct as written.

token/services/identity/role/registry.go
```diff
@@ -111,36 +162,48 @@ func (r *Registry) Lookup(ctx context.Context, id driver.WalletLookupID) (driver
...
 	}
 
 	r.Logger.DebugfContext(ctx, "no wallet found for [%s] at [%s]", passedIdentity, logging.Prefix(wID))
 	if len(ident) != 0 {
-		identityWID, err := r.GetWalletID(ctx, ident)
-		r.Logger.DebugfContext(ctx, "wallet for identity [%s] -> [%s:%s]", ident, identityWID, err)
-		if err == nil && len(identityWID) != 0 {
+		res := r.GetWalletID(ctx, ident)
+		r.Logger.DebugfContext(ctx, "wallet for identity [%s] -> [%s:%d]", ident, res.WalletID, res.Status)
+		if res.Failed() {
```
AkramBitar, line 257: Same as the comment on line 166: by here `MapToIdentity` has succeeded and `ident` is authoritative, so this probe is an optimization and a storage failure should be logged and skipped rather than aborting the lookup.

token/services/identity/role/registry.go
```diff
@@ -111,36 +162,48 @@ func (r *Registry) Lookup(ctx context.Context, id driver.WalletLookupID) (driver
...
 			r.WalletMu.RUnlock()
 			if ok {
-				return walletEntry, nil, passedWalletID, nil
+				return walletEntry, nil, res.WalletID, nil
 			}
-			r.Logger.DebugfContext(ctx, "no wallet found, there is a wallet for identity [%s]: [%s] but it has not been recreated yet", passedIdentity, passedWalletID)
+			r.Logger.DebugfContext(ctx, "no wallet found, there is a wallet for identity [%s]: [%s] but it has not been recreated yet", passedIdentity, res.WalletID)
 		}
-		walletIdentifiers = append(walletIdentifiers, passedWalletID)
+		walletIdentifiers = append(walletIdentifiers, res.WalletID)
```
AkramBitar: Minor, and I think pre-existing: when `res` is `WalletIDUnbound`, `res.WalletID` is `""`, so this appends an empty id to `walletIdentifiers` (same at :206). The old code appended the same empty `wID`, but now that the status is explicit it is easy to guard — `if res.Bound() { walletIdentifiers = append(...) }`.

token/services/identity/role/registry.go
```diff
@@ -22,6 +22,51 @@ type WalletFactory interface {
...
+// bound to it. It exists so callers branch on an explicit, named state instead of
+// re-deriving intent from the ambiguous "(string, error)" shape — where ("", nil),
+// ("", err) and ("id", nil) each mean something different and the difference is easy
+// to get wrong (see issue #2063).
+type WalletIDStatus int
+
+const (
+	// WalletIDUnknown is the zero value and never returned by GetWalletID; it guards
+	// against a WalletIDResolution that was constructed without going through GetWalletID.
+	WalletIDUnknown WalletIDStatus = iota
```
AkramBitar, line 41: The doc says `WalletIDUnknown` "guards against a `WalletIDResolution` that was constructed without going through `GetWalletID`", but nothing actually guards. With only `Bound()` and `Failed()` defined, a zero-value `WalletIDResolution{}` is indistinguishable from `WalletIDUnbound` at every call site — i.e. the zero value silently means "authoritative miss, safe to create a wallet", which is precisely the interpretation this type was introduced to make impossible. A mock, or a future constructor that forgets to set `Status`, gets the dangerous default.

Related: the field comment below tells callers to inspect via `Bound/Unbound/Failed`, but there is no `Unbound()` method. Adding it and branching exhaustively fixes both:

```go
func (r WalletIDResolution) Unbound() bool { return r.Status == WalletIDUnbound }
```

…then at each call site treat "neither `Bound()` nor `Unbound()`" as a failure.

token/services/identity/role/registry.go
```diff
@@ -233,16 +296,31 @@ func (r *Registry) GetIdentityMetadata(ctx context.Context, identity driver.Iden
...
+// GetWalletID resolves the wallet identifier bound to the passed identity.
+//
+// It is the single point that translates the storage layer's (WalletID, error)
+// convention into an explicit WalletIDResolution, so no caller has to re-derive the
+// meaning of ("", nil) vs ("", err). The storage contract reports an unbound identity
+// as ("", nil); a non-nil error is a genuine storage failure (timeout, connection
+// reset, ...) whose result is therefore WalletIDFailed, never WalletIDUnbound. Keeping
+// the two apart here is what prevents a transient blip from looking like an
+// unregistered identity and triggering the creation of a duplicate wallet (issue #2063).
+func (r *Registry) GetWalletID(ctx context.Context, identity driver.Identity) WalletIDResolution {
```
AkramBitar, line 375: `Registry.GetWalletID` changes from `(string, error)` to `WalletIDResolution`. Nothing in-tree breaks (the only callers are the three sites in `Lookup`; `wallet.RoleRegistry` does not include the method), but `Registry` is exported API of a released SDK, so downstream code compiled against the old signature will not build.

Either keep `GetWalletID(ctx, identity) (string, error)` — now returning the honest error instead of swallowing it — and expose the resolution under a new name (`ResolveWalletID`?), or make sure this lands in the release notes as a breaking change.

token/services/identity/role/registry_test.go
```diff
@@ -135,17 +135,81 @@ func TestBindIdentityAndContainsAndMetadataAndGetWalletID(t *testing.T) {
...
+	t.Run("MapToIdentity fails, view-identity fallback hits storage error", func(t *testing.T) {
+		reg, storage, role, wf := newRegistryWithFakes()
+		ctx := t.Context()
+
+		role.MapToIdentityReturns(nil, "", errors.New("no mapping"))
+		storage.GetWalletIDReturns("", errors.New("transient DB blip"))
+
+		_, _, _, err := reg.Lookup(ctx, []byte("id-with-binding"))
+		require.Error(t, err)
+		require.Equal(t, 0, wf.NewWalletCallCount())
```
AkramBitar: `require.Equal(t, 0, wf.NewWalletCallCount())` can't fail in these `Lookup` subtests (same at :195) — `Lookup` never touches `WalletFactory`; only `WalletByID` does (registry.go:371). These assertions would still pass with the old swallow-and-fall-through behaviour restored inside `Lookup`, so they don't guard the regression. Asserting on the returned error/values is the real check here; `TestWalletByID_StorageErrorDoesNotCreateDuplicate` is doing the actual work.

Worth adding one more `WalletByID` case where `MapToIdentity` **succeeds** and the probe fails — that is the common path, and the one affected by my comment on registry.go:166.

AkramBitar (commented):
Thanks for taking this on — replacing the `(string, error)` ambiguity with an explicit `WalletIDResolution` is the right shape for #2063, and `GetWalletID` itself now does exactly the right thing.

Two things I think need another pass before this lands, one of which means the bug is still reachable:

1. **The KVS store still can't distinguish "no binding" from "lookup failed"** (`walletdb.go:92`), so for a KVS-backed wallet store the duplicate-wallet path from #2063 survives — the swallowed error just moved one layer down, where it is no longer visible.
2. **Two of the three new fail-closed branches are on an *optional* probe** (`registry.go:166` and `:190`), which turns a transient storage blip into a failed transaction on a path that previously succeeded with no duplicate risk.

Details inline. The remaining comments are small: a missing `Unbound()`, an exported-API break worth calling out, and a test assertion that can't fail.

Two repo conventions to tick off as well (AGENTS.md): there is no `plan.md` for this change, and no `docs/` update — the error semantics of `Lookup`/`WalletByID` change in a user-visible way (they now fail where they previously fell back), which the wallet/identity docs should reflect.

AkramBitar (changes requested):

AkramBitar (approved):
LGTM

PR #2172, opened 2026-08-10
identity: areMe converts storage errors into false-negative ownership answers

### LINKED ISSUE #2066: identity: areMe converts storage errors into false-negative ownership answers
### Summary

`Provider.areMe` (backing both `AreMe` and `IsMe`) returns whatever partial result it has accumulated
when the storage lookup for signer existence errors, rather than propagating the error. This means a
transient storage failure makes `IsMe` report `false` ("not mine") for an identity that actually is
ours, which is on the token-ownership decision path.

### Where

`token/services/identity/provider.go:270-279`:

```go
// check Storage
found, err := p.storage.GetExistingSignerInfo(ctx, notFound...)
if err != nil {
	p.Logger.Errorf("failed checking if a signer exists [%s]", err)
	return result.ToSlice()
}
result.Add(found...)
return result.ToSlice()
```

On error, `result` contains only whatever was already resolved from the in-memory cache
(`provider.go:257-264`) before the storage call — everything not in cache is silently reported as
"not mine," logged only at `Errorf` with no error propagated to the caller.

Related: `token/services/identity/wallet/service.go:138-149` `Wallet()` similarly discards two errors
from `OwnerWallet`/`IssuerWallet` and returns `nil` (a caller cannot distinguish "no wallet" from
"lookup failed"):

```go
func (s *Service) Wallet(ctx context.Context, identity tdriver.Identity) tdriver.Wallet {
	w, _ := s.OwnerWallet(ctx, identity)
	if w != nil {
		return w
	}
	iw, _ := s.IssuerWallet(ctx, identity)
	if iw != nil {
		return iw
	}
	return nil
}
```

### Impact

`IsMe` is used to decide ownership-related behavior (e.g. whether to react to a token as an owned
token). A transient storage error causing a false negative means an owned token can be treated as
not-owned for that call, with no visible error — the caller has no signal that the answer is
unreliable rather than authoritative.

### Reproduction

Not yet committed. Mock `Storage.GetExistingSignerInfo` to return an error for a set of identities
that includes at least one genuinely-owned identity not already warm in the in-memory cache; assert
`IsMe`/`AreMe` either propagates the error or is documented as best-effort at the call sites that rely
on it for correctness-sensitive decisions.

### Suggested fix

`AreMe`/`IsMe`'s current signatures (`[]string` / `bool`, no error return) make it structurally unable
to distinguish "confirmed not mine" from "couldn't check." Changing the signature ripples through the
`driver.IdentityProvider` interface and every caller, so the minimal fix here is: keep the signatures,
but do not silently swallow storage errors into a negative result for the *specific identities the
storage call was about* — instead of returning early, consider retrying once, or making the
error-swallowing explicit and loud enough (metric increment, not just a log line) that an operator can
detect ownership answers were degraded. If callers need a hard guarantee, a follow-up to thread an
error return through `AreMe` is worth considering separately.

### Severity

MEDIUM — requires a transient storage failure to trigger, and produces an unsignaled false negative
on an ownership decision.

Effi-S:
Fixes #2066

### Summary

`Provider.areMe` (backing both `AreMe` and `IsMe`) returns whatever partial result it has accumulated
when the storage lookup for signer existence errors, rather than propagating the error. This means a
transient storage failure makes `IsMe` report `false` ("not mine") for an identity that actually is
ours, which is on the token-ownership decision path.

### Where

`token/services/identity/provider.go:270-279`:

```go
// check Storage
found, err := p.storage.GetExistingSignerInfo(ctx, notFound...)
if err != nil {
	p.Logger.Errorf("failed checking if a signer exists [%s]", err)
	return result.ToSlice()
}
result.Add(found...)
return result.ToSlice()
```

On error, `result` contains only whatever was already resolved from the in-memory cache
(`provider.go:257-264`) before the storage call — everything not in cache is silently reported as
"not mine," logged only at `Errorf` with no error propagated to the caller.

Related: `token/services/identity/wallet/service.go:138-149` `Wallet()` similarly discards two errors
from `OwnerWallet`/`IssuerWallet` and returns `nil` (a caller cannot distinguish "no wallet" from
"lookup failed"):

```go
func (s *Service) Wallet(ctx context.Context, identity tdriver.Identity) tdriver.Wallet {
	w, _ := s.OwnerWallet(ctx, identity)
	if w != nil {
		return w
	}
	iw, _ := s.IssuerWallet(ctx, identity)
	if iw != nil {
		return iw
	}
	return nil
}
```

### Impact

`IsMe` is used to decide ownership-related behavior (e.g. whether to react to a token as an owned
token). A transient storage error causing a false negative means an owned token can be treated as
not-owned for that call, with no visible error — the caller has no signal that the answer is
unreliable rather than authoritative.

### Reproduction

Not yet committed. Mock `Storage.GetExistingSignerInfo` to return an error for a set of identities
that includes at least one genuinely-owned identity not already warm in the in-memory cache; assert
`IsMe`/`AreMe` either propagates the error or is documented as best-effort at the call sites that rely
on it for correctness-sensitive decisions.

### Suggested fix

`AreMe`/`IsMe`'s current signatures (`[]string` / `bool`, no error return) make it structurally unable
to distinguish "confirmed not mine" from "couldn't check." Changing the signature ripples through the
`driver.IdentityProvider` interface and every caller, so the minimal fix here is: keep the signatures,
but do not silently swallow storage errors into a negative result for the *specific identities the
storage call was about* — instead of returning early, consider retrying once, or making the
error-swallowing explicit and loud enough (metric increment, not just a log line) that an operator can
detect ownership answers were degraded. If callers need a hard guarantee, a follow-up to thread an
error return through `AreMe` is worth considering separately.

### Severity

MEDIUM — requires a transient storage failure to trigger, and produces an unsignaled false negative
on an ownership decision.


token/services/identity/role/role.go
```diff
@@ -170,7 +170,15 @@ func (r *Role) mapStringToID(ctx context.Context, v string) (driver.Identity, st
 		r.logger.DebugfContext(ctx, "passed node identity as view identity")
 
 		return nil, defaultIdentifier, nil
-	case r.localMembership.IsMe(ctx, labelAsIdentity):
+	}
+
+	// IsMe can fail (e.g. a storage error); a failure must not be silently treated as
+	// "not a local member", so propagate it instead of folding it into the fallthrough.
+	isMe, err := r.localMembership.IsMe(ctx, labelAsIdentity)
```
AkramBitar, line 177: Making an `IsMe` failure fatal here breaks label-based wallet lookup that never needed storage in the first place.

Every other step in `mapStringToID` is in-memory only — `LocalMembership.GetIdentifier` just reads `localIdentitiesByName` under a mutex — so `IsMe` is the sole storage-touching step, and for a *label* it is purely speculative: "do these label bytes happen to be a registered signer?".

Concretely: `WalletManager().OwnerWallet(ctx, "alice")` where a wallet was registered in `Registry.Wallets["alice"]` but `"alice"` is not a configured local-identity name.

- Before: `IsMe` → `false` → `return nil, "alice", nil` → `Registry.Lookup` hits `r.Wallets["alice"]` and returns the wallet with zero DB access.
- After: a transient signer-store error makes `MapToIdentity` return an error, and `Registry.Lookup`'s recovery path (registry.go:79-86) calls `toViewIdentity(id)`, which returns `ok=false` for a `string` (registry.go:335-344) — so `fail` stays true and the whole lookup errors out.

Suggestion: keep the resolution heuristics fail-open (or teach `Lookup` to distinguish "couldn't check" from "not found"), and reserve fail-closed for the security-relevant decisions — `extractRequiredSigners` and `prepareDistributionList`.

token/services/identity/role/role.go
```diff
@@ -220,7 +228,15 @@ func (r *Role) mapIdentityToID(ctx context.Context, v driver.Identity) (driver.I
 		r.logger.DebugfContext(ctx, "passed identity is the node identity (same bytes)")
 
 		return nil, defaultIdentifier, nil
-	case r.localMembership.IsMe(ctx, id):
+	}
+
+	// IsMe can fail (e.g. a storage error); a failure must not be silently treated as
+	// "not me", so propagate it instead of falling through to label resolution.
+	isMe, err := r.localMembership.IsMe(ctx, id)
```
AkramBitar, line 235: Same class of change as in `mapStringToID`: the previous fall-through to `GetIdentityInfo(label, nil)` / `GetIdentifier` is now skipped on error.

This is partly mitigated because `Registry.Lookup`'s `GetWalletID(passedIdentity)` fallback does work for `driver.Identity`/`[]byte` lookups — but that fallback reads the same database that just failed, so in practice both fail together, and a lookup that used to be satisfied straight from the in-memory `Registry.Wallets` cache now errors out.

integration/token/fungible/views/utils.go
```diff
@@ -29,7 +29,9 @@ func AssertTokens(sp token.ServiceProvider, tx *ttx.Transaction, outputs *token.
 	assert.NoError(err, "failed to get token db for [%s]", tx.TokenService().ID())
 	for _, output := range outputs.Outputs() {
 		tokenID := output.ID(token.RequestAnchor(tx.ID()))
-		if output.Owner.Equal(id) || tx.TokenService().SigService().IsMe(ctx, output.Owner) {
+		isMe, err := tx.TokenService().SigService().IsMe(ctx, output.Owner)
```
AkramBitar: This loses the `||` short-circuit: `IsMe` is now called for every output, including the ones whose owner is literally the `id` argument, where the answer is irrelevant.

Two consequences:

1. The added `assert.NoError` aborts `AssertTokens` on an ownership-check error even for outputs that were already matched by `output.Owner.Equal(id)`.
2. One extra signer-store round-trip per output.

Computing `isMe` only inside an `if !output.Owner.Equal(id)` branch preserves the original behaviour.

token/driver/wallet.go
```diff
@@ -39,10 +39,14 @@ type IdentityProvider interface {
 	RegisterSigner(ctx context.Context, identity Identity, signer Signer, verifier Verifier, signerInfo []byte, ephemeral bool) error
 
 	// AreMe checks a list of identities and returns those that have signers registered with this provider.
-	AreMe(ctx context.Context, identities ...Identity) []string
+	// A non-nil error means the ownership check could not be completed (for example a storage failure);
+	// in that case the returned slice must not be treated as an authoritative answer.
+	AreMe(ctx context.Context, identities ...Identity) ([]string, error)
```
AkramBitar, line 45: Contract inconsistency between this doc and the implementation.

The doc says that on a non-nil error "the returned slice must not be treated as an authoritative answer", which implies a partial slice may still come back. But `Provider.areMe` (token/services/identity/provider.go:286) returns `nil` on error, discarding the cache-confirmed positives it had already collected in `result`.

Either return `result.ToSlice()` alongside the error so callers can salvage the confirmed hits, or tighten the wording here (and on `IsMe`) to state that the slice is always nil when the error is non-nil.

AkramBitar (dismissed):

token/services/identity/provider.go
```diff
@@ -271,19 +277,20 @@ func (p *Provider) areMe(ctx context.Context, identities ...driver.Identity) []s
...
 	if err != nil {
-		p.Logger.Errorf("failed checking if a signer exists [%s]", err)
-
-		return result.ToSlice()
+		return result.ToSlice(), errors.Wrapf(err, "failed checking if a signer exists")
 	}
 	result.Add(found...)
 
-	return result.ToSlice()
+	return result.ToSlice(), nil
```
AkramBitar, line 306: **Blocking:** returning `result.ToSlice()` alongside the error leaks a cache-only partial slice to the caller. The contract documented in the PR — "must not be trusted when err is non-nil" — is enforced only by convention, which is exactly the kind of footgun this PR is trying to eliminate. Return `nil, err` instead so there is nothing to accidentally use:

```go
if err != nil {
    return nil, errors.Wrapf(err, "failed checking if a signer exists")
}
```

The test assertion `assert.Equal(t, []string{cached.UniqueID()}, me)` in `TestProvider_AreMe_StorageErrorIsPropagated` should then become `assert.Nil(t, me)`.

AkramBitar (changes requested):
The fix is correct — propagating the error rather than silently returning a false-negative is the right call. One blocking issue on the partial-slice contract, plus minor nits inline.

AkramBitar (changes requested):
@Effi-S 

See my comments and could you please squash the commits to one?

Thanks a lot,
Akram

token/services/identity/provider.go
Effi-S: Done

AkramBitar (commented):
Reviewed `origin/main...fix-2066` at `03597e04` in a clean worktree. **The direction and the interface shape are right** — `(bool, error)` / `([]string, error)` is exactly what lets a caller tell "confirmed not mine" from "couldn't check", and the doc comments you added on each interface are the kind of thing that keeps that distinction alive. Two things I think need another iteration before this lands, then some smaller notes.

## 1. CI is red from this PR's own test

`TestProvider_AreMe_StorageErrorIsPropagated` asserts `AreMe` returns the cache-only partial slice alongside the error, but `areMe` returns `nil`:

```
$ go test ./token/services/identity/
--- FAIL: TestProvider_AreMe_StorageErrorIsPropagated (0.00s)
    provider_test.go:131:
        	Error:      	Not equal:
        	            	expected: []string{"NnMBTnK2c4O+MCSFaUVVpXrTk6/euu1t7REKd1vQVW0="}
        	            	actual  : []string(nil)
FAIL	github.com/LFDT-Panurus/panurus/token/services/identity	0.085s
```

Worth deciding deliberately rather than just making the test pass: the docs on all three interfaces describe a partial slice, the code returns `nil`. Details inline on `provider.go:152`. (For the record, the contract is stated two ways — docs vs. implementation — not three; `driver/wallet.go:44`, `provider.go:152` and `token/sig.go:67` all agree with each other.)

## 2. The propagated error is swallowed one layer up, at the site where #2066 actually loses tokens

This is the one I'd most want your read on. #2066's concrete damage is an owned token being silently dropped, and that happens in `token/core/common/authorization.go:61`:

```go
func (w *WalletBasedAuthorization) IsMine(ctx context.Context, tok *token2.Token) (string, []string, bool) {
	wallet, err := w.WalletService.OwnerWallet(ctx, tok.Owner)
	if err != nil {
		return "", nil, false
	}
	...
}
```

and then `token/services/tokens/tokens.go:497`:

```go
...
	logger.DebugfContext(ctx, "transaction [%s], discarding token, not mine, not an auditor, not an issuer", requestAnchor)
	continue
}
```

So the error this PR now correctly propagates out of `areMe` → `IsMe` → `mapIdentityToID` → `Registry.Lookup` → `OwnerWallet` arrives at `IsMine`, is turned straight back into `mine = false`, and the output is discarded and never recorded as owned — the exact symptom, unchanged. Same shape at four sibling sites:

- `token/services/interop/htlc/script.go:144` and `:151` — `if wallet, err := s.WalletService.OwnerWallet(ctx, script.Sender); err == nil`
- `token/services/ttx/boolpolicy/auth.go:58`
- `token/services/ttx/multisig/auth.go:59`
- `token/services/identity/wallet/service.go:139` — `w, _ := s.OwnerWallet(ctx, identity)` (you already flag this one as "Related" in the PR description, so this may be half-acknowledged already)

The structural blocker is that `driver.Authorization.IsMine` (`token/driver/wallet.go:332`) has no error return, so there is nowhere for the distinction to go:

```go
IsMine(ctx context.Context, tok *token.Token) (walletID string, additionalOwners []string, mine bool)
```

Extending `Authorization` is a bigger change than this PR, and splitting it out is completely reasonable — but if it isn't in scope here, it'd help to say so in the PR description and open the follow-up, because as it stands `Fixes #2066` overstates what merges.

## 3. `docs/upgradability.md` needs an "SDK API Changes (Go)" entry

That section exists precisely for this (it was added for #2063's `GetWalletID` return-type change). This PR breaks these exported symbols:

- `driver.IdentityProvider.IsMe` / `.AreMe` (`token/driver/wallet.go:44,48`)
- `token.SignatureService.IsMe` / `.AreMe` (`token/sig.go:70,77`) — the public facade, so any application calling `tms.SigService().IsMe(...)` stops compiling
- `role.LocalMembership.IsMe` (`token/services/identity/role/role.go:29`)
- `membership.IdentityProvider.IsMe` (`token/services/identity/membership/lm.go:103`)
- `ttx/dep.SignatureService.IsMe` (`token/services/ttx/dep/providers.go:62`)

AGENTS.md's "Documentation Updates" rule makes this required before the task counts as done. (`docs/services/identity.md:45` only carries a bare `+IsMe()` in a mermaid class diagram, so that one is *not* stale — no change needed there.)

## 4. None of the new error paths are exercised

```
$ grep -rn "IsMeReturns\|AreMeReturns" --include=*_test.go .
token/sig_test.go:128:	ip.IsMeReturns(true, nil)
token/sig_test.go:185:	ip.AreMeReturns(expectedHashes, nil)
token/services/identity/membership/lm_security_test.go:191:	ip.IsMeReturns(false, nil)
token/services/identity/membership/lm_test.go:37:	ip.IsMeReturns(true, nil)
token/services/ttx/endorse_test.go:73:	tokenIP.IsMeReturns(true, nil)
token/services/ttx/boolpolicy/spend_test.go:74:	ip.AreMeReturns(areMe, nil)
token/services/ttx/multisig/spend_test.go:74:	ip.AreMeReturns(areMe, nil)
```

Every stub passes `nil`. Uncovered: `role/role.go:177`, `role/role.go:235`, `membership/lm.go:298`, `token/sig.go:70`, `token/sig.go:77`, `ttx/endorse.go:257`, `ttx/collectendorsements.go:694`, `ttx/boolpolicy/spend.go:153`, `ttx/multisig/spend.go:138`. The two new `registry_test.go` cases stub `MapToIdentityReturns(…, err)` directly, so they cover `Registry.Lookup`'s fallback but never `role.go`'s propagation. A single error-returning stub per package would lock in the intent.

## Smaller notes

Left inline: a nil-cache-probe bug at `registry.go:210` (with a reproduction), a `[]byte`-vs-`string` lookup asymmetry, the dropped storage cause in the string-label branch, and a context-cancellation conflation at `provider.go:289`.

Also minor: the comments at `provider_test.go:118` and `:129-130` describe the partial-slice behaviour that the implementation doesn't have, and `assert.Empty(t, me)` at `:96` passes vacuously against `nil` — both will need a touch-up once #1 is settled either way.

## Things I checked that are fine

- Build (`go build ./...` across all 9 modules), `go vet`, and `golangci-lint run` are all clean.
- Counterfeiter mocks are in sync — all four regenerated fakes still carry their `var _ Interface = new(Fake)` assertions and compile.
- The `switch` → `if` refactors in `role.go` preserve the original short-circuit ordering exactly (`IsMe` is still only reached when every earlier case is false).
- No goroutine-lifetime regressions: `boolpolicy/spend.go` returns before any `go collectAnswers`, and `multisig/spend.go` constructs the `AnswersCollector` first but it holds only a buffered channel — no goroutine, no timer — so nothing leaks on the new early return.
- `collectendorsements.go:687` (FSC's `AreMe`, left swallowing) is benign as written: a false negative there just pushes the identity into `remainingIds`, which the token-level `AreMe` at `:694` re-checks — so no change needed.
- The string-label fallback at `registry.go:228` is fail-*closed*, not fail-open — I traced it end-to-end and when nothing resolves it still errors. My note there is only about the lost cause and the unchanged semantics.
- FYI, `TestTranslatePath` in `token/services/identity/config` fails in any worktree whose path doesn't contain "panurus". It reproduces on `origin/main` and is nothing to do with this PR.

Out of scope, for a possible follow-up issue rather than this PR: `Registry.ContainsIdentity` (`role/registry.go:355`) → `storage/db/sql/common/wallet.go:143` still logs-and-returns `result != ""` on a query failure, and `collectendorsements.go:462` uses FSC's `sig.Service.IsMe`, which still has no error return and decides self-vs-remote auditing. Both are the same bug class, both pre-existing and untouched here.

token/services/identity/provider_test.go
```diff
@@ -72,10 +72,65 @@ func TestProvider_RegisterSigner_And_IsMe(t *testing.T) {
...
+	require.NoError(t, p.RegisterSigner(t.Context(), cached, &drvmock.Signer{}, &drvmock.Verifier{}, nil, false))
+
+	// The storage lookup for the remaining (uncached) identity fails.
+	storage.GetExistingSignerInfoReturns(nil, errors.New("boom"))
+
+	me, err := p.AreMe(t.Context(), cached, driver.Identity("uncached"))
+	require.Error(t, err, "a storage failure must be propagated so the cache-only partial is not trusted")
+	// The returned slice is the non-authoritative cache-only partial (just the cached hit) and
+	// must be ignored precisely because err is non-nil.
+	assert.Equal(t, []string{cached.UniqueID()}, me)
```
AkramBitar: **Blocker: this assertion fails as committed, so the PR is red on `make unit-tests`.**

```
$ go test ./token/services/identity/
--- FAIL: TestProvider_AreMe_StorageErrorIsPropagated (0.00s)
    provider_test.go:131:
        	Error:      	Not equal:
        	            	expected: []string{"NnMBTnK2c4O+MCSFaUVVpXrTk6/euu1t7REKd1vQVW0="}
        	            	actual  : []string(nil)
FAIL	github.com/LFDT-Panurus/panurus/token/services/identity	0.085s
```

The test encodes the contract the doc comments describe (a non-authoritative cache-only partial), but `areMe` returns `nil, err` at `provider.go:289` and throws the cached hits away. Both behaviours are defensible — see my note on `provider.go:152` — but code, docs and test have to pick the same one.

token/services/identity/provider.go
```diff
@@ -148,15 +148,25 @@ func (p *Provider) RegisterSigner(ctx context.Context, identity driver.Identity,
 // Each identity is resolved via the signer cache and configured storage.
 // There is no secondary "is me" cache: a real cache would need careful handling for
 // single-use identities (for example Idemix nyms) and is intentionally omitted here.
-func (p *Provider) AreMe(ctx context.Context, identities ...driver.Identity) []string {
+// A non-nil error means the storage lookup failed and the returned slice is not
+// authoritative: identities that could not be checked are absent from it precisely because
```
AkramBitar: This doc block, `token/driver/wallet.go:44` ("the returned slice must not be treated as an authoritative answer") and `token/sig.go:67` all describe **a partial slice that happens to be untrustworthy**. The implementation at `provider.go:289` returns `nil`. The new test at `provider_test.go:131` asserts the partial. So the contract is stated consistently in three places and then not implemented — worth choosing on purpose:

- **Return `result.ToSlice(), err`** — matches these docs and the test as written, and gives a caller that wants to be lenient something to work with. Costs nothing since every in-tree caller checks `err` first.
- **Keep `nil`** — arguably safer (a caller that forgets to check `err` gets an empty set rather than a plausible-looking subset), but then these three doc blocks and `provider_test.go:129-131` need rewriting to say "on error the slice is nil and carries no information".

I'd lean toward `nil` for the defensive reason and fix the docs, but either is fine as long as all four sites agree.

token/services/identity/provider.go
```diff
@@ -271,19 +277,20 @@ func (p *Provider) areMe(ctx context.Context, identities ...driver.Identity) []s
...
+	// check Storage. A failure here must not be flattened into a negative answer: the
+	// identities we could not check are exactly the ones in notFound, and silently dropping
+	// them would report an owned identity as not-owned. Propagate instead so callers can tell
+	// "confirmed not mine" from "couldn't check".
 	found, err := p.storage.GetExistingSignerInfo(ctx, notFound...)
 	if err != nil {
-		p.Logger.Errorf("failed checking if a signer exists [%s]", err)
-
-		return result.ToSlice()
+		return nil, errors.Wrapf(err, "failed checking if a signer exists")
```
AkramBitar: Minor, and not a reason to hold the PR: `GetExistingSignerInfo` wraps every `QueryContext` failure the same way (`storage/db/sql/common/identity.go:403`, `"error querying db"`), so `context.Canceled` and `context.DeadlineExceeded` now surface here as an ownership-check error indistinguishable from a broken store.

Where that shows up concretely is the three integration views this PR touches — `views/boolpolicy.go:221`, `views/multisig.go:210`, `views/utils.go:36` — which all `assert.NoError` on the result, and FSC's `assert.NoError` panics. A view whose context is cancelled mid-check now panics the responder instead of unwinding cleanly.

If you want to be strict about it, `errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded)` could be passed through unwrapped so callers can distinguish "we gave up" from "storage is broken". Equally fine to leave and note it.

token/services/identity/role/registry.go
```diff
@@ -184,25 +184,55 @@ func (r *Registry) Lookup(ctx context.Context, id driver.WalletLookupID) (driver
...
 				fail = false
+			} else {
+				// Bound is ruled out: GetWalletID either failed (a storage blip, the
+				// same failure that produced the mapping error) or answered Unbound.
+				// Either way, before giving up try the in-memory Wallets cache directly
+				// under the raw identity string — it does not touch the failed storage,
+				// exactly as the pre-error-propagation fall-through did.
+				candidate := string(passedIdentity)
+				r.WalletMu.RLock()
+				_, cached := r.Wallets[candidate]
```
AkramBitar: This probe tests **key presence**, not a non-nil wallet — and `RegisterWallet` accepts a nil wallet, which `TestWalletByID_NilCachedWalletIsRebuilt` (`registry_test.go:820`) exists specifically to pin as "must count as absent", matching the comment at `registry.go:552`:

> An entry holding a nil wallet counts as absent, exactly as the fast path in `WalletByID` treats it.

So a nil-valued entry under `string(passedIdentity)` satisfies `cached`, skips the `!res.authoritative()` abort below, and hands the lookup back as a success on an unknown binding. I confirmed it with a throwaway test — nil wallet registered under `"idnil"`, `MapToIdentity` failing with the newly-propagated `IsMe` error, `GetWalletID` also failing:

```
Lookup     -> w=<nil> idInfo=<nil> wID="idnil" err=<nil>     // expected res.abortError(id)
WalletByID -> err=<nil> newWalletCalled=true idInfo=<nil>
```

Two consequences: the #2063 guard is bypassed (creation proceeds while the binding is unknown, which is what `abortError` is there to stop), and `NewWallet` is called with a nil `IdentityInfo` — which with the real factory reaches `info.Anonymous()` at `wfactory.go:76` and panics on the nil interface.

```suggestion
				cached := r.Wallets[candidate] != nil
```

token/services/identity/role/registry.go
```diff
@@ -184,25 +184,55 @@ func (r *Registry) Lookup(ctx context.Context, id driver.WalletLookupID) (driver
...
+				} else if !res.authoritative() {
+					// A storage failure — or a resolution that never went through
+					// GetWalletID — leaves the binding unknown, and nothing is cached to
+					// recover from. It must not be treated as "not registered", or a
+					// transient blip would fall through to wallet creation and duplicate
+					// state (issue #2063).
+					return nil, nil, "", res.abortError(id)
+				}
+				// Unbound and nothing cached: keep fail=true so the original mapping
+				// error is surfaced below.
```
AkramBitar: This fallback recovers from `Bound` and from the in-memory cache, but it deliberately doesn't retry the `GetIdentityInfo` / `GetIdentifier` resolution that `mapIdentityToID` used to reach after a false `IsMe` (`role.go:245-251`). For identities that are genuinely bound that's fine — `GetWalletID` returns `Bound` and line 196 catches it. But for a **label-shaped** lookup it turns a previously-successful resolution into a hard failure, and makes the two lookup shapes disagree.

Probe: `MapToIdentity` failing with the propagated `IsMe` error, wallet store healthy (`Unbound`), `role.GetIdentityInfo("alice")` succeeding, cold cache:

```
WalletByID(ctx, 0, []byte("alice"))  -> err = failed to lookup wallet [alice]: failed checking if identity is me
WalletByID(ctx, 0, "alice")          -> resolves, err = <nil>
```

Pre-PR both succeeded, because `mapIdentityToID` fell through to `if info, err := GetIdentityInfo(label, nil); err == nil { return nil, info.ID(), nil }`. The string label still works only because of the new branch at line 228; the `[]byte` shape has no equivalent.

Failing closed is defensible in itself, but the asymmetry looks unintended, and `Unbound`-and-not-cached is the case where you *know* the wallet store answered authoritatively — so retrying the identity-info resolution there (rather than surfacing the mapping error) would both restore the old behaviour and stay honest about the storage failure.

token/services/identity/role/registry.go
```diff
@@ -184,25 +184,55 @@ func (r *Registry) Lookup(ctx context.Context, id driver.WalletLookupID) (driver
...
 			}
+		} else if label, isString := id.(string); isString && len(label) != 0 {
```
AkramBitar, line 246: Two notes on this branch, neither a correctness blocker.

First, to be clear about what I did *not* find: this is fail-**closed**. I traced it end to end — when the cache misses, `wID = label` flows to `walletIdentifiers`, `GetIdentityInfo(label)` decides, and a failure ends at line 322. It doesn't fabricate ownership, and it faithfully reproduces the pre-PR "not a local member" fall-through, as the comment says.

But `err` is dropped on the floor here, and the terminal error at line 322 doesn't wrap it:

```go
return nil, nil, "", errors.Errorf("failed to get wallet info for [%s]", logging.Prefix(walletID))
```

So the operator sees "failed to get wallet info for [alice]" and never "signer store unavailable" — the root cause this PR exists to surface is lost precisely on the path where it mattered. `errors.WithMessagef(err, ...)` at line 322, or stashing `err` and wrapping it in this branch, would fix that.

Second, worth stating in the comment: because this restores the pre-PR fall-through exactly, an `IsMe` storage failure during `Wallet(ctx, "<label>")` is still silently ignored. That's a reasonable trade (it avoids breaking every label lookup on a transient blip), but a reader will otherwise assume the PR's propagation covers this path too.

AkramBitar (changes requested):

token/services/identity/provider_test.go
Effi-S: will update the test to reflect the new behavior (nil and error returned instead of slice and error)

token/services/identity/provider.go
Effi-S: Will update docs as well to reflect most recent change

token/services/identity/role/registry.go
Effi-S: Will Do

token/services/identity/role/registry.go
Effi-S: Replaced the if cached / else if !res.authoritative() chain with a three-way switch, adding a case res.Unbound(): branch.

token/services/identity/role/registry.go
Effi-S: renamed the loop error, join+wrap the cause at the terminal return, and add the documenting comment to the label / Unbound branches

token/services/identity/provider.go
Effi-S: OK, will do

AkramBitar (commented):
**Approve after 3 fixes:**

**1. Test is broken** — `TestProvider_AreMe_StorageErrorIsPropagated` asserts the cached partial slice is returned on error, but the code returns `nil`. Fix the assertion to `assert.Nil(t, me)`. While there, the comment block at `provider_test.go:118,129-130` also describes partial-slice behaviour the implementation doesn't have, and `assert.Empty(t, me)` at `:96` passes vacuously against `nil`.

**2. Missing docs** — `docs/upgradability.md` needs a breaking-change table for the 6 renamed symbols. AGENTS.md's "Documentation Updates" rule requires this before the task counts as done:

| Symbol | Before | After |
| :--- | :--- | :--- |
| `driver.IdentityProvider.IsMe` | `(ctx, id) bool` | `(ctx, id) (bool, error)` |
| `driver.IdentityProvider.AreMe` | `(ctx, ids...) []string` | `(ctx, ids...) ([]string, error)` |
| `token.SignatureService.IsMe` / `.AreMe` | same | same |
| `role.LocalMembership.IsMe` | `(ctx, id) bool` | `(ctx, id) (bool, error)` |
| `membership.LocalMembership.IsMe` | `(ctx, id) bool` | `(ctx, id) (bool, error)` |
| `ttx/dep.SignatureService.IsMe` | `(ctx, id) bool` | `(ctx, id) (bool, error)` |

**3. Error paths untested** — every mock stub passes `nil` for the new error return. These 8 production call sites have no test exercising the error branch: `role/role.go:177,235`, `membership/lm.go:298`, `token/sig.go:70,77`, `ttx/endorse.go:257`, `ttx/collectendorsements.go:694`, `ttx/boolpolicy/spend.go:153`, `ttx/multisig/spend.go:138`. One error-returning stub per package is sufficient.

---

The core fix is correct — `(bool, error)` / `([]string, error)` is the right shape and the `registry.go` fallback logic (cache → Unbound → non-authoritative) is sound and well-covered by the new regression tests.

AkramBitar (changes requested):

AkramBitar (changes requested):
Approve after these 3 fixes:

**1. `TestProvider_AreMe_StorageErrorIsPropagated` is broken — CI is red from the PR's own test.**

The test asserts the cache-only partial slice is returned on error, but `areMe` returns `nil`. Fix: change `assert.Equal(t, cachedHash, me)` → `assert.Nil(t, me)` and update the comment block around `provider_test.go:118,129-130` that still describes partial-slice semantics. Also `assert.Empty(t, me)` at ~:96 passes vacuously against `nil` — change to `assert.Nil` to make intent explicit.

**2. `docs/upgradability.md` is missing the breaking-change entry.**

Six exported symbols changed signatures — any downstream application calling `tms.SigService().IsMe(...)` stops compiling. AGENTS.md's Documentation Updates rule requires this before the task counts as done. Add:

| Symbol | Before | After |
| :--- | :--- | :--- |
| `driver.IdentityProvider.IsMe` | `(ctx, id) bool` | `(ctx, id) (bool, error)` |
| `driver.IdentityProvider.AreMe` | `(ctx, ids...) []string` | `(ctx, ids...) ([]string, error)` |
| `token.SignatureService.IsMe` / `.AreMe` | same | same |
| `role.LocalMembership.IsMe` | `(ctx, id) bool` | `(ctx, id) (bool, error)` |
| `membership.LocalMembership.IsMe` | `(ctx, id) bool` | `(ctx, id) (bool, error)` |
| `ttx/dep.SignatureService.IsMe` | `(ctx, id) bool` | `(ctx, id) (bool, error)` |

**3. New error paths are untested — every mock stub passes `nil` for the new error return.**

These 8 production call sites have no test exercising the error branch: `role/role.go:177,235`, `membership/lm.go:298`, `token/sig.go:70,77`, `ttx/endorse.go:257`, `ttx/collectendorsements.go:694`, `ttx/boolpolicy/spend.go:153`, `ttx/multisig/spend.go:138`. One error-returning stub per package is sufficient.

---

The core fix and interface shape are correct. The `registry.go` fallback logic (in-memory cache → Unbound → non-authoritative error) is sound and well-covered by the new regression tests.

AkramBitar (approved):
LGTM

PR #2177, opened 2026-08-10
fix(network): fix goroutine leak in waitTimeout

### LINKED ISSUE #2124: Fix Goroutine Leak in `waitTimeout` (network/fabric) [mem]
**Document severity:** HIGH · **Verified priority:** Medium · **Effort:** Low · **Type:** Bug

### Old issue reference

> **Sub-Task 3 — Fix Goroutine Leak in `waitTimeout` (network/fabric)**
>
> **Status:** `[ ] pending`

| 3 | `[RUN]` | **HIGH** | Goroutine leak | `network/fabric/network.go` | 574–586 | `waitTimeout` orphans goroutine on timeout |

### Intent

`waitTimeout` spawns a goroutine that calls `wg.Wait()`. If the `time.After`
branch fires first (timeout), the function returns but the goroutine remains
blocked on `wg.Wait()` forever, holding a reference to the `WaitGroup` and
keeping the goroutine alive for the lifetime of the process.

Additionally, when the timeout branch is taken, the `time.After` timer has
already fired so there is no timer leak — but there is an orphaned goroutine.

### Expected Outcomes

- No goroutine is left blocked on `wg.Wait()` after `waitTimeout` returns.
- The helper uses the existing `context`-based cancellation that callers already have.

### Todo List

1. Accept a `context.Context` parameter in `waitTimeout`.
2. Replace the internal goroutine + `time.After` with a `context.WithTimeout`
   derived from the caller's context.
3. Return the context error when timeout/cancellation fires.
4. If `wg.Wait()` cannot be made context-aware, document the residual goroutine
   as acceptable and add a `//nolint:revive` comment with a clear explanation.

### Relevant Context

- File: `token/services/network/fabric/network.go`
- Lines 574–586: `waitTimeout` helper.

---

### Status in current code

**Confirmed — still present.** Line numbers have shifted: the helper is now at `token/services/network/fabric/network.go:582-594`, not 574–586.

The helper reads, verbatim at HEAD:

```go
func waitTimeout(wg *sync.WaitGroup, timeout time.Duration) error {
	c := make(chan struct{})
	go func() {
		defer close(c)
		wg.Wait()
	}()
	select {
	case <-c:
		return nil
	case <-time.After(timeout):
		return errors.Errorf("context done")
	}
}
```

When the `time.After` branch wins, `waitTimeout` returns but the goroutine stays parked on
`wg.Wait()` for the rest of the process lifetime, pinning the `WaitGroup` and the unbuffered
`c`. Nothing will ever call `Done()` enough times to release it, because the caller has already
given up.

The document is right about the mechanism and right that the timer is *not* the problem here —
this is the one item in the timer-flavoured group whose defect is real and independent of Go
version.

Downgraded from HIGH to **Medium** on blast radius, not on validity: the leak is one goroutine
per timeout *event*, not one per request, so it grows only as fast as operations actually time
out. A node that never times out never leaks. That still accumulates on an unhealthy network,
which is why it is worth fixing rather than documenting away.

### Severity and priority

`origin/main` with `go 1.26.5` on 2026-08-04, the priority is **Medium** for the

### Effort

**Low**

AkramBitar:
Fixes #2124

### Problem

`LookupTransferMetadataKey` waited for its `lookupListener` by spawning a goroutine parked on `sync.WaitGroup.Wait()`. On the timeout path the function returns and its deferred `RemoveLookupListener` tears the listener down, so `OnStatus`/`OnError` — the only callers of `wg.Done()` — never fire, and the goroutine stays parked for the life of the process.

`ScanForPreImage` uses a 5 minute timeout, so every HTLC pre-image that is never revealed stranded one goroutine.

Note that buffering the channel does **not** fix this: the goroutine was blocked on `wg.Wait()`, not on a channel send, and the send it would replace was a `close()`, which never blocks.

### Fix

Remove the goroutine instead of trying to unblock it. `lookupListener` now owns a `done` channel that the first matching notification closes, and the waiter selects on it against a timer. With nothing parked on the notification there is nothing to leak.

Two things fall out of the same change:

- Closing exactly once (via `sync.Once`) removes the `negative WaitGroup counter` panic that a duplicate notification could trigger.
- The timer is explicitly stopped, so a lookup that completes early no longer retains one for the rest of its timeout.

The timeout is reported as a plain error naming the key rather than a bare context sentinel: a lookup that does not complete in time is an expected outcome here, and surfacing `context.DeadlineExceeded` would make it indistinguishable from caller cancellation to `errors.Is`-based retry logic.

The `context` parameter is dropped rather than left dead — neither `driver.Network.LookupTransferMetadataKey` nor `htlc.ScanForPreImage` carries a context, so the only caller could pass nothing but `context.Background()`.

### Tests

Seven tests in the new `token/services/network/fabric/lookup_listener_test.go`:

- Happy path — the value reported by the lookup manager reaches the waiter
- Error path — a reported failure is surfaced to the waiter
- Notifications for a different key do not release the waiter
- First notification wins; duplicates neither panic nor overwrite the result
- 16 concurrent notifications are safe under `-race`
- **Regression test for #2124** — `goleak` after a single abandoned timeout
- **Regression test for #2124** — 32 successive timeouts leave nothing parked

The tests are written to actually reproduce the bug: they never release the listener, since releasing it also let the leaked goroutine exit — which is why a test that releases it passes against the unfixed code. Verified that `goleak` flags the pre-fix implementation under these assertions and passes the new one.

### DIFF
--- a/token/services/network/fabric/network.go
+++ b/token/services/network/fabric/network.go
@@ -407,9 +407,7 @@ func (n *Network) LookupTransferMetadataKey(namespace string, key string, timeou
 		return nil, errors.Wrapf(err, "failed to generate transfer action metadata key from [%s]", key)
 	}
 	logger.Debugf("lookup transfer metadata key [%s] from [%s] in namespace [%s]", key, transferMetadataKey, namespace)
-	wg := &sync.WaitGroup{}
-	wg.Add(1)
-	l := &lookupListener{wg: wg, key: transferMetadataKey}
+	l := newLookupListener(transferMetadataKey)
 	if err := n.llm.AddLookupListener(namespace, transferMetadataKey, l); err != nil {
 		return nil, errors.Wrapf(err, "failed to add lookup listener")
 	}
@@ -418,12 +416,10 @@ func (n *Network) LookupTransferMetadataKey(namespace string, key string, timeou
 			logger.Debugf("failed to remove lookup listener [%s]: %v", transferMetadataKey, err)
 		}
 	}()
-	if err := waitTimeout(wg, timeout); err != nil {
-		return nil, err
-	}
-	logger.Debugf("lookup transfer metadata key [%s] from [%s] in namespace [%s], done, result [%s][%s]", key, transferMetadataKey, namespace, l.value, l.err)
+	value, err := l.wait(timeout)
+	logger.Debugf("lookup transfer metadata key [%s] from [%s] in namespace [%s], done, result [%s][%v]", key, transferMetadataKey, namespace, value, err)
 
-	return l.value, l.err
+	return value, err
 }
 
 // Ledger returns direct access to the ledger querying layer.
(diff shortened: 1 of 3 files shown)


token/services/network/fabric/network.go
```diff
@@ -407,9 +407,7 @@ func (n *Network) LookupTransferMetadataKey(namespace string, key string, timeou
 		return nil, errors.Wrapf(err, "failed to generate transfer action metadata key from [%s]", key)
 	}
 	logger.Debugf("lookup transfer metadata key [%s] from [%s] in namespace [%s]", key, transferMetadataKey, namespace)
-	wg := &sync.WaitGroup{}
-	wg.Add(1)
-	l := &lookupListener{wg: wg, key: transferMetadataKey}
+	l := newLookupListener(transferMetadataKey)
```
Effi-S, line 410: Maybe better name than `l`

Effi-S (approved):

PR #2180, opened 2026-08-10
fix(ttx): remove context.Context field from Transaction struct

### LINKED ISSUE #2178: ttx: Transaction stores context.Context on the struct, causing stale context propagation and a live bug in htlc.Transaction.Lock
`token/services/ttx/transaction.go` stores a `context.Context` as a field on `Transaction`, captured once at construction time from `view.Context.Context()` in `NewTransaction` and `NewTransactionFromBytes`. Every action-building method (`Issue`, `Transfer`, `Redeem`, `Upgrade`, `Outputs`, `Inputs`, `Bytes`) then reads `t.Context` instead of accepting a context from its caller.

This is the Go context-in-struct anti-pattern, and it has concrete consequences here:

- **Stale/wrong context propagation.** The captured context belongs to the view that *created* the transaction. A transaction that is serialized, shipped over a session, and rebuilt via `NewTransactionFromBytes` in a different view still carries whatever context that originating view had — cancellation, deadlines, and tracing spans no longer correspond to the operation actually running against the rebuilt transaction.
- **A live bug in `htlc.Transaction.Lock`** (`token/services/interop/htlc/transaction.go`): the method already accepts a `ctx context.Context` parameter, and every caller already passes a correct one, but the method body silently ignores it and uses `t.Context` instead — so the caller's deadline/cancellation/tracing context is dropped on the floor without any indication to the caller.
- **Untestable methods.** `Issue`/`Transfer`/`Outputs`/`Inputs` cannot be exercised against a hand-built `Transaction` in a unit test without also populating an unexported context field, since there is no constructor path that sets it outside of `NewTransaction`/`NewTransactionFromBytes`.

Impact: incorrect trace spans and missed cancellation/timeout propagation for transactions that cross session/view boundaries, plus reduced unit-test coverage for the affected transaction-building methods.

### LINKED ISSUE #2179: lint: containedctx enabled — audit structs that store context.Context as a field
`.golangci.yml` newly enables the `containedctx` linter (flags any struct with a `context.Context` field, the context-in-struct anti-pattern). A full repo-wide scan across every Go module (root, `integration/`, and the `cmd/*` modules — the default `golangci-lint` output truncates same-message findings via `--max-same-issues`, so a full `--max-same-issues=0 --max-issues-per-linter=0` scan is needed to see everything) surfaces roughly a dozen structs holding a `context.Context` field, most pre-dating this linter change.

They fall into a few shapes:

- Long-lived background-worker/service lifecycle structs that set `ctx`/`cancel` once via `context.WithCancel(context.Background())` in a `Start()`/`NewXxx()` method, scope a goroutine loop with it, and cancel it in `Stop()` (e.g. recovery/cleanup managers, notifier, event queue, certifier client, a stress-test worker pool).
- A per-request context override wrapper.
- A per-event/message struct carrying a context through a channel for logging/cancellation.
- Session-wrapper types that capture a default `ctx` at construction for `Send`/`Receive`-style convenience methods, alongside explicit `...WithContext` variants that take a context explicitly.
- A test fake mirroring an existing `Context()` accessor pattern.

None of these were flagged before `containedctx` was enabled, and enabling the linter with no remediation plan will break `make lint-auto-fix`/CI repo-wide. This issue tracks reviewing each finding to decide, case by case, whether it represents a real bug (stale/wrong context reused across requests) or an accepted, justified pattern (e.g. genuine service-lifetime scoping) that should be explicitly suppressed with a documented `//nolint:containedctx` rather than silently left to accumulate.

adecaro:
## Summary
- Removes the `Context context.Context` field from `ttx.Transaction`; `Bytes`, `Issue`, `Transfer`, `Redeem`, `Upgrade`, `Outputs`, and `Inputs` now take `ctx context.Context` explicitly from the caller instead of reading a value captured once at construction time.
- Fixes `htlc.Transaction.Lock`, which already accepted `ctx` but silently used the struct field instead — the caller's deadline/cancellation/tracing context was being dropped.
- Threads `ctx` through every same-package caller (`marshaller.go`, `auditor.go`, `collectactions.go`, `collectendorsements.go`), the wrapper transaction types (`htlc`, `nfttx`, `boolpolicy`, `multisig`), ~45 integration view call sites, and doc code samples.
- Enables the `containedctx` linter and resolves every resulting finding repo-wide (root + `integration` + `cmd/*` modules) — either by threading `ctx` through as above, or with a `//nolint:containedctx` plus justification for the reviewed, accepted patterns (long-lived worker/service lifecycles, a per-event struct, session-wrapper convenience defaults, and one test fake).

## Test plan
- [x] `go build ./...` (root and `integration` module) clean
- [x] `go vet ./...` (root and `integration` module) clean
- [x] `make unit-tests` passing
- [x] `make unit-tests-race` passing, no new races
- [x] `make checks` clean across all modules
- [x] `make lint-auto-fix` clean across all modules (0 `containedctx` issues)
- [x] Confirmed no remaining references to the removed field via targeted grep

Fixes #2178
Fixes #2179

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,7 @@
 version: "2"
 linters:
   enable:
+    - containedctx
     - asasalint # Check for pass []any as any in variadic func(...any).
     - canonicalheader # Canonicalheader checks whether net/http.Header uses canonical header. [auto-fix]
     - copyloopvar # A linter detects places where loop variables are copied. [fast, auto-fix]
--- a/docs/token_sdk_usage.md
+++ b/docs/token_sdk_usage.md
@@ -56,7 +56,7 @@ if err != nil {
 // 3. Issue Tokens
 // Use the issuer wallet to issue 'quantity' of 'tokenType' to 'recipient'.
 wallet := ttx.GetIssuerWallet(context, issuerWalletID)
-err = tx.Issue(wallet, recipient, tokenType, quantity)
+err = tx.Issue(context.Context(), wallet, recipient, tokenType, quantity)
 if err != nil {
     return nil, err
 }
@@ -99,6 +99,7 @@ if err != nil {
 // Sender wallet is used to select input tokens.
 senderWallet := ttx.GetWallet(context, senderWalletID)
 err = tx.Transfer(
+    context.Context(),
     senderWallet,
     tokenType,
     []uint64{amount},
@@ -147,6 +148,7 @@ if err != nil {
 // If needed, also pin the issuer signing key expected by public parameters.
 senderWallet := ttx.GetWallet(context, senderWalletID)
 err = tx.Redeem(
+    context.Context(),
     senderWallet,
     tokenType,
     amount,
@@ -195,7 +197,7 @@ if err != nil {
 
 // 3. Add Alice's Transfer
 senderWallet := ttx.GetWallet(context, aliceWallet)
-err = tx.Transfer(senderWallet, aliceTokenType, []uint64{aliceAmount}, []view.Identity{other})
+err = tx.Transfer(context.Context(), senderWallet, aliceTokenType, []uint64{aliceAmount}, []view.Identity{other})
 if err != nil {
     return nil, err
 }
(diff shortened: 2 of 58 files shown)


AkramBitar (approved):
LGTM

PR #2181, opened 2026-08-11
docs(evm): explain how the contracts process a transaction

atharrva01:
The implementation guide described the contracts as they were sketched rather than as they were built. It named an entry point that does not exist (`applyStateUpdate`), emitted an event that does not exist (`StateUpdateEvent`), and drew one contract doing work that is actually split across two.

This replaces that with a walkthrough aimed at somebody meeting the contracts for the first time:

- it opens with what usually confuses people, that the chain validates nothing and verifies no proofs, since everything else follows from that
- a field by field tour of the `StateDelta`, which the guide never showed even though every diagram referred to it
- the six sources in `contracts/src`, and which of them actually reach an address
- deployment and transaction processing as separate phases, since a different set of contracts takes part in each
- the sequence of one transaction, with a note on why the contract recomputes the digest itself instead of being handed one
- the checks in the order they run

Failures are a table instead of a diagram: thirteen errors, what each means and what usually causes it. Somebody holding a reverted receipt wants to look up a name, not trace a flowchart.

The phase split came out of Angelo's question on which contracts are involved in processing versus deployment, and whether `Clones` is only there for the tests. It is not: `TokenStateFactory.create` calls it for every clone.

Writing that up caught two things the first draft had wrong. The clone is an EIP-1167 proxy, so it `delegatecall`s the implementation on every call and three addresses take part in processing rather than two. And `StateCommitted` is emitted at the clone's address, which is the address the finality layer filters `eth_getLogs` on.

Design section 3 gets a pointer to it, per rule R5.

The diffs show mermaid as source, so the diagrams are easier to judge rendered:

https://github.com/atharrva01/panurus/blob/evm-contract-flow-docs/docs/services/network-ethereum.md#how-the-contracts-interact

### DIFF
--- a/docs/services/network-ethereum.md
+++ b/docs/services/network-ethereum.md
@@ -596,7 +596,7 @@ the table to reach for when a transaction reverts.
 |---|---|---|
 | `NotInitialized` | The clone was never seeded | Deployment did not go through the factory |
 | `AnchorAlreadyProcessed` | This anchor was applied before | A resubmitted transaction, or a genuine replay attempt |
-| `MetadataLengthMismatch` | Key and value lists differ in length | A malformed delta; the driver builds these aligned |
+| `MetadataLengthMismatch` | Key and value lists differ in length | A malformed delta; the endorser's translator builds these aligned |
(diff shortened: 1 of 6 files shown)


adecaro (approved):
LGTM

PR #2191, opened 2026-08-11
test(ttx): fix flaky TestStatusBatcher_CoalescesConcurrentLookups

AkramBitar:
300µs batch window caused flakes on loaded CI: the first goroutine's flush fired before the others joined the batch. Use 100ms window in this test so all goroutines register before the flush.


Effi-S (approved):
Hi @AkramBitar, 
Interesting find.. 
I wonder if other tests in `token/services/ttx/dep/wrapper/status_batcher_test.go` can benifit from the 100 milliseconds delay rather than 300 nanoseconds

PR #2199, opened 2026-08-12
fix(auditor): attribute empty-EID inputs to their own token owner

### LINKED ISSUE #2198: Auditor books inputs with an unresolved enrollment ID against the first output's enrollment ID
## What happens

`requestWrapper.completeInputsWithEmptyEID` (`token/services/auditor/auditor.go`) fills every input whose enrollment ID could not be resolved with the *first output's* enrollment ID:

```go
// TODO: extract from the audit tokens
targetEID := record.Outputs.EnrollmentIDs()[0]
...
for i := range filter.Count() {
    item := filter.At(i)
    item.EnrollmentID = targetEID
```

In a payment the first output is the recipient, so the payer's spent input is booked as spent *by the counterparty*.

## Why it matters

A single transfer corrupts both parties' eid-keyed audit balances at once:

- the recipient is charged a spend it never made;
- the payer's spend is never recorded, so its balance reads higher than it is.

Nothing fails or warns — the audit record is written and looks well-formed. It affects `Movements`, per-enrollment-id balance queries, and every auditor limit check evaluated against those sums.

An input reaches this path with an empty enrollment ID whenever `extractTransferInputs` cannot resolve the transfer's sender locally, which is routine for a counterparty's identity.

Two smaller issues sit in the same function:

- `record.Outputs.EnrollmentIDs()[0]` panics with an index-out-of-range on a record that has no outputs.
- The resolved revocation handle is never filled in, so `RevocationHandler` stays empty on these inputs.

## Expected

Each input should be attributed to the enrollment ID of *its own* spent token's owner, resolved from the audit info the record carries (falling back to the locally stored one).

An owner that maps to no single enrollment ID — a composite owner such as a multisig — should leave its input unattributed, with an empty enrollment ID, rather than borrow one. Amount aggregations skip empty enrollment IDs, so such an input is counted for nobody; that case is routine rather than a failure, so it should not fail the audit record.

A token upgrade is the one case where the current guess is right and must be kept: `extractIssueInputs` fills only the token id, and the pre-upgrade owner resolves to nothing, but the upgrade re-issues the spent tokens to their owner. Such an input — one the request describes no sender for — should take the enrollment ID of what the request issues, when the request issues to exactly one party.

The `TODO` in that line suggests the current behaviour was always meant to be temporary, but I would like to confirm before proposing a change: is attributing to the first output's enrollment ID intentional for some driver, or is this the placeholder it looks like?

I have a fix ready and will open a PR referencing this issue.

EvanYan1024:
Fixes #2198

## What

- Attribute each input with an unresolved enrollment ID to its **own** spent token's owner, instead of to the first output's enrollment ID. The revocation handle is filled in from the same resolution.
- An owner that maps to no single enrollment ID — a composite owner such as a multisig — leaves its input **unattributed**, with an empty enrollment ID.
- A **token upgrade** is the exception: its issue metadata describes no sender for the inputs, and the pre-upgrade owner resolves to nothing here, so such an input takes the enrollment ID of the outputs issued by *its own action*, when every one of them resolves to the same party. The revocation handle comes from the same output.
- Add `WalletManager.GetEIDAndRH(ctx, identity, auditInfo)`, which resolves through the audit info attached to the input — the locally stored audit info of the spent token's owner where present (`Request.AuditRecord` fills it in), the request-carried one otherwise; when neither is available it returns empty values rather than an error.
- Run the gap filling in `Audit`, before the enrollment IDs are collected, so the locks cover the enrollment ID each input is finally booked under, and return a record cached by `Audit` from `Append` as it stands. `Append` still attributes the record when called without a preceding `Audit`.
- Guard the vault result the way `Request.AuditRecord` does — a length check plus a per-token nil check — since this code now runs on the audit approval path.

## Why

In a payment the first output is the recipient, so the old `targetEID := record.Outputs.EnrollmentIDs()[0]` recorded the payer's spend against the counterparty: the recipient is charged a spend it never made, and the payer's spend is never booked. Neither side errors or warns. The `TODO` above that line marked it as a placeholder.

Leaving an unattributable input empty rather than guessing is the point of the change: `EnrollmentIDs()` skips empty enrollment IDs, so such an input is counted for nobody, which is the honest outcome when the owner genuinely maps to no single enrollment ID. A multisig input is exactly that — several members, no single one to charge — so it is a normal state to represent, not a failure to raise.

The upgrade case is the one where the same guess was right, and dropping it showed up as a red `TokensUpgrade` integration test (alice's holding 220 instead of 110). `extractIssueInputs` fills only the token id, and the pre-upgrade owner predates the current driver, so nothing in the record resolves it. But an upgrade re-issues the spent tokens to their owner under a fresh identity, so what the request issues to *is* the input's enrollment ID — leaving it empty credits the upgraded amount without ever debiting it. The fallback is confined to inputs the request describes no sender for, and scoped to the outputs of the input's own action: a second issue action neither suppresses the attribution nor lends its enrollment ID to it. Every issued output of that action must resolve to the same party — one resolving to none cannot be shown to belong to the others — and the handle is kept only while it stays paired with that enrollment ID.

The lifecycle matters as much as the attribution. `Audit` collects the enrollment IDs it locks from the record and `Append` writes that record, so filling the gaps only in `Append` would leave the final record booked against a payer that was never locked — an exact spend has no change output, so the payer's enrollment ID appears nowhere in what `Audit` saw.

The same reasoning applies in reverse to the cached record: re-running the gap filling in `Append` could attribute an input that `Audit` deliberately left empty, storing it under an enrollment ID that was never locked. `Append` therefore returns the cached record as it stands. Together these keep the locked set and the stored record in agreement, and remove the second vault read for records carrying an unattributed input.

Resolving from the input's own owner also drops the `record.Outputs.EnrollmentIDs()[0]` index-out-of-range on a record with no outputs.

## Testing

- `token/services/auditor/auditor_internal_test.go`: attribution from the token owner; record-carried audit info preferred over a local lookup (asserting no local lookup happens); a composite owner and a missing-audit-info owner both staying unattributed without an error, with the remaining fields still filled from the spent token; a genuine resolution error still propagating; a short vault answer and a nil vault token each erroring instead of panicking.
- The upgrade fallback: an input with no sender and an unresolvable owner takes the issued enrollment ID and revocation handle; a transfer input whose owner maps to nothing keeps none even when the same request issues tokens; issuing to two parties, or to one party plus an output resolving to none, leaves the input unattributed; two handles under one enrollment ID keep the ID and drop the handle; a two-action record attributes each input from its own action.
- `TestRequestWrapper_AuditRecord_UpgradeInputAttributedToReceiver` runs an upgrade-shaped request through the real `Request.AuditRecord` pipeline rather than a hand-written record, so what `extractIssueInputs` and `extractIssueOutputs` actually produce — the input arriving without an owner, the issued output carrying issuer, owner and enrollment ID under the same action index — is exercised instead of assumed. It also asserts the record's inputs and outputs sum equal, which is the holding that went to 220 in `update-t1`.
- `TestRequestWrapper_AuditRecord_CachedKeepsUnattributedInput`: a record cached by `Audit` with an unattributed input is returned untouched even though the wallet service would now resolve it, with no further vault or identity lookup.
- `TestService_Audit_AttributesAndLocksEmptyEIDInput` pins the lifecycle over an exact spend whose sender does not resolve locally: `Audit` returns the resolved payer enrollment ID and revocation handle, the acquired locks contain that payer and no empty ID, and `Append` reuses the record without a second vault read or identity lookup. Moving the gap filling back after lock acquisition fails this test.
- Verified locally on both modules: `gofmt`, `go build`, `go vet`, `go test -race`, and golangci-lint v2.12.2 with the repo config.

## Docs

`docs/services/auditor.md` gains an "Input Attribution" section: the per-input resolution order, the upgrade fallback, what happens to an owner with no single enrollment ID, and where in the `Audit`/`Append` lifecycle it runs relative to the EID locking documented just below it.

## Review rounds

- An action spending tokens of multiple enrollment IDs, a composite owner spanning enrollment IDs, and issued outputs that only partly resolve now fail the audit with errors naming the cause; representing multi-sender actions is #2242.
- The identity layer marks its decoding failures with `identity.ErrUnresolvableIdentity`, keeping the cause in the chain; the gap filling treats only that class as unresolvable, while storage failures and context cancellation fail the audit.
- `Audit` and `Append` rebind the request to the provider-resolved TMS before the record is computed, so the request cannot influence which TMS attributes the record.
- `docs/services/auditor.md` gained an Input Attribution section describing the behavior, including the audit-info preference order (the local store wins; flipping it is #2251).
- Follow-ups: #2249 (action-scoped output filter), #2250 (cache entry lifetime), #2251 (preference order).

### DIFF
--- a/token/services/auditor/auditor.go
+++ b/token/services/auditor/auditor.go
@@ -14,6 +14,7 @@ import (
 
 	"github.com/LFDT-Panurus/panurus/token"
 	"github.com/LFDT-Panurus/panurus/token/core/common/metrics"
+	"github.com/LFDT-Panurus/panurus/token/services/identity"
 	"github.com/LFDT-Panurus/panurus/token/services/logging"
 	"github.com/LFDT-Panurus/panurus/token/services/network"
 	"github.com/LFDT-Panurus/panurus/token/services/network/driver"
@@ -159,7 +160,13 @@ func (a *Service) Audit(ctx context.Context, tx Transaction) (*token.InputStream
 	start := time.Now()
 	logger.DebugfContext(ctx, "audit transaction [%s]....", tx.ID())
 	request := tx.Request()
-	record, err := request.AuditRecord(ctx)
+	tms, err := a.bindProviderTMS(request)
+	if err != nil {
+		return nil, nil, err
+	}
+	// the record is completed before the enrollment IDs are collected, so that
+	// the locks cover the enrollment ID every input is finally booked under
+	record, err := newRequestWrapper(request, tms).AuditRecord(ctx)
 	if err != nil {
 		return nil, nil, errors.WithMessagef(err, "failed getting transaction audit record")
 	}
@@ -217,7 +224,7 @@ func (a *Service) Append(ctx context.Context, tx Transaction) error {
 	defer func() { a.metrics.AppendDuration.Observe(time.Since(start).Seconds()) }()
 	defer a.Release(ctx, tx)
 
-	tms, err := a.tmsProvider.TokenManagementService(token.WithTMSID(a.tmsID))
+	tms, err := a.bindProviderTMS(tx.Request())
 	if err != nil {
 		return err
 	}
@@ -252,6 +259,22 @@ func (a *Service) Append(ctx context.Context, tx Transaction) error {
 	logger.DebugfContext(ctx, "append done for request [%s]", tx.ID())
 
 	return nil
+}
+
+// bindProviderTMS resolves the TMS for the service's TMS ID through the
+// provider and rebinds the request to it. The record computation runs through
+// request.TokenService, so this is what keeps the request from influencing
+// which TMS computes and attributes the record.
+func (a *Service) bindProviderTMS(request *token.Request) (dep.TokenManagementServiceWithExtensions, error) {
+	tms, err := a.tmsProvider.TokenManagementService(token.WithTMSID(a.tmsID))
+	if err != nil {
+		return nil, err
+	}
+	if err := tms.SetTokenManagementService(request); err != nil {
+		return nil, err
+	}
+
+	return tms, nil
 }
 
 // Release releases the lock acquired of the passed transaction and drops the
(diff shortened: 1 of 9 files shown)


AkramBitar (commented):
@EvanYan1024 

Thanks a lot for all the efforts that you put on that PR. 

I did additional review on what we have. 

Regarding your question: "That leaves a real question: reject such an action earlier with a clear error, or teach the store to represent a multi-sender action. I would rather not settle it inside this PR — happy to open a separate issue unless you prefer it handled here." 

I preferer to handle it in different issue, however I leave this to you. 

Thanks a million, 
Akram

**Optional / follow-up:**

- `token/services/storage/ttxdb/store.go:376` — happy to defer the multi-owner case as agreed. Two small asks: open the follow-up issue as part of this PR, and reject at `Audit` time with a message that names the multi-owner cause. Today it surfaces as `expected at most 1 input enrollment id` from `Append`, which is easy to misdiagnose. *Optional for this PR.*
- **Documented preference order is inverted.** `docs/services/auditor.md:62` and `WalletManager.GetEIDAndRH`'s godoc (`token/wallet.go:169`) both say the record-carried audit info is preferred over the locally stored one, but `Request.AuditRecord` has already overwritten `in.OwnerAuditInfo` with the local lookup of `toks[i].Owner` before `completeInputsWithEmptyEID` runs — so in practice the local one wins wherever it exists. `TestRequestWrapper_CompleteInputsWithEmptyEID_UsesRecordAuditInfo` hand-builds the record, so it doesn't catch this. *Optional.*
- `auditor.go:437` re-reads audit tokens that `Request.AuditRecord` already fetched, only to add `Owner`. Having `AuditRecord` set `in.Owner = toks[i].Owner` would remove a vault query per unattributed input. *Optional, nit.*

token/services/auditor/auditor.go
```diff
@@ -415,36 +420,102 @@ func (r *requestWrapper) AuditRecord(ctx context.Context) (*token.AuditRecord, e
...
 }
 
+// issuedToEIDAndRH returns the enrollment ID and revocation handle the given issue
+// action issues to, and empty values when it does not issue to exactly one party.
+// An issued output carries both an issuer and an owner; a redeem output carries an
+// issuer but no owner. Every issued output of the action must resolve to the same
+// enrollment ID: one resolving to none cannot be shown to belong to the same party
+// as the others. The handle is kept only while it stays paired with that ID.
+func issuedToEIDAndRH(outputs *token.OutputStream, actionIndex int) (string, string) {
+	issued := outputs.Filter(func(o *token.Output) bool {
```
AkramBitar, line 565: **Should fix before merge.** `extractIssueOutputs` emits one output per recipient of a composite owner, each with its own enrollment ID. Upgrading a token owned by a multisig therefore yields several issued outputs with differing EIDs, `issuedToEIDAndRH` returns empty, and the input stays unattributed — the upgraded amount is credited to the members but never debited, which is the same credit-without-debit this PR set out to fix.

Deduplicating by output `Index` before comparing EIDs (as `UniquePerOutput` does elsewhere) would cover it. Could we also add a test for the composite-owner upgrade case? I don't see one.

token/services/auditor/auditor.go
```diff
@@ -415,36 +420,102 @@ func (r *requestWrapper) AuditRecord(ctx context.Context) (*token.AuditRecord, e
...
 		item.Type = tokens[i].Type
 		q, err := token2.ToQuantity(tokens[i].Quantity, precision)
 		if err != nil {
 			return errors.WithMessagef(err, "failed converting token quantity [%s]", tokens[i].Quantity)
 		}
 		item.Quantity = q
+
+		eID, rID, err := wm.GetEIDAndRH(ctx, item.Owner, item.OwnerAuditInfo)
+		if err != nil {
+			return errors.WithMessagef(err, "failed resolving enrollment id for input [%v]", item.Id)
```
AkramBitar: **Should fix before merge.** The design here is "unresolvable owner ⇒ leave unattributed", but the identity layer signals unresolvable by returning an *error* in several cases — `UnmarshalTypedIdentity` failure, `no deserializer found for [type]`, audit-info deserialization failure.

A pre-upgrade owner from an older driver is exactly such an identity: `GetEIDAndRH` errors and we return here, before reaching `if eID == "" && upgraded`. The whole audit then fails and the upgrade can never be approved, where an empty result would have been handled cleanly. Suggest treating "cannot resolve" errors as an empty EID on the fallback/unattributed paths.

token/services/auditor/auditor.go
```diff
@@ -415,36 +420,102 @@ func (r *requestWrapper) AuditRecord(ctx context.Context) (*token.AuditRecord, e
...
 
+// issuedToEIDAndRH returns the enrollment ID and revocation handle the given issue
+// action issues to, and empty values when it does not issue to exactly one party.
+// An issued output carries both an issuer and an owner; a redeem output carries an
+// issuer but no owner. Every issued output of the action must resolve to the same
+// enrollment ID: one resolving to none cannot be shown to belong to the same party
+// as the others. The handle is kept only while it stays paired with that ID.
+func issuedToEIDAndRH(outputs *token.OutputStream, actionIndex int) (string, string) {
+	issued := outputs.Filter(func(o *token.Output) bool {
+		return o.ActionIndex == actionIndex && len(o.Issuer) != 0 && len(o.Owner) != 0
```
AkramBitar, line 566: *Optional — hardening.* `extractIssueOutputs` and `extractTransferOutputs` each number outputs by their own loop index, so this filter separates issue from transfer only via the `len(o.Issuer) != 0 && len(o.Owner) != 0` heuristic. A request combining an upgrade issue at index 0 with a transfer at index 0 already mixes both actions' outputs in here; it is correct today only because no driver attaches an issuer to an *owned* transfer output. Scoping the filter by the action itself rather than by `ActionIndex` + issuer presence would make it correct by construction.

token/services/auditor/auditor.go
```diff
@@ -395,16 +397,19 @@ func (r *requestWrapper) PublicParamsHash() token.PPHash { return r.r.PublicPara
...
-		var err error
-		record, err = r.r.AuditRecord(ctx)
-		if err != nil {
-			return nil, err
-		}
+	// re-running the gap filling on a cached record could attribute an input
+	// Audit deliberately left empty, booking it under an enrollment ID that
+	// was never locked
+	if r.cached != nil {
+		return r.cached, nil
```
AkramBitar, line 428: *Optional — follow-up.* This invariant only holds while the cache entry lives: `Append` defers `Release`, which drops the cached record. A second `Append` for the same transaction (a retry, or an append after an explicit `Release`) finds no cached record, re-runs the gap filling, and can store an enrollment ID that was never locked — with the locks already gone. Worth either keeping the entry until finality or rejecting a second `Append`.

token/services/auditor/auditor.go
```diff
@@ -159,7 +159,9 @@ func (a *Service) Audit(ctx context.Context, tx Transaction) (*token.InputStream
 	start := time.Now()
 	logger.DebugfContext(ctx, "audit transaction [%s]....", tx.ID())
 	request := tx.Request()
-	record, err := request.AuditRecord(ctx)
+	// the record is completed before the enrollment IDs are collected, so that
+	// the locks cover the enrollment ID every input is finally booked under
+	record, err := newRequestWrapper(request, request.TokenService).AuditRecord(ctx)
```
AkramBitar: *Optional.* On which TMS source to standardise on: I'd suggest `tmsProvider.TokenManagementService(WithTMSID(a.tmsID))` in both places. Taking the service from the request here means a caller can influence which wallet service attributes the record.

AkramBitar (approved):
@EvanYan1024 

Thanks a million for all the efforts that you put on this PR. I appreciate your support a lot.

The PR LGTM.

One last thing could you please rebase on top of latest main and squash all the commits to one.

Regards,
Akram

PR #2201, opened 2026-08-12
[BUG]: Flaky test  TestTokens/TimeFiltering    Messages: recent token should not be in results

### LINKED ISSUE #2200: [BUG]: Flaky test  TestTokens/TimeFiltering    Messages: recent token should not be in results
### 👾 Description

```bash
--- FAIL: TestTokens (8.33s)
    --- FAIL: TestTokens/TimeFiltering (0.17s)
        tokens.go:1427: Should be false
        Test:     TestTokens/TimeFiltering
        Messages: recent token should not be in results
```

Caused by flaky timing test

The test (token/services/storage/db/dbtest/tokens.go:1399) does:

```go
createAndDeleteToken("old_token", 0, "idemix")
time.Sleep(100 * time.Millisecond)
createAndDeleteToken("recent_token", 0, "idemix")
// return tokens deleted MORE than 50ms ago
tokens, err := db.GetDeletedTokensPendingSKICleanup(ctx, 50*time.Millisecond, 100)
...
assert.False(t, foundRecent, "recent token should not be in results")   // line 1427
```

The intent: recent_token was deleted only moments ago, so with a 50 ms cutoff it should be too young to appear. The problem is the margin is  **wall-clock dependent**

- recent_token needs to still be younger than 50 ms at the instant the query's now() is evaluated.
- Between the createAndDeleteToken("recent_token") call (which itself does INSERT + DELETE round-trips to Postgres) and the GetDeletedTokensPendingSKICleanup query executing, more than 50 ms of real time can easily elapse on a loaded CI runner.

When that happens, recent_token's deletion age crosses 50 ms before the query runs, so it's returned, foundRecent is true, and the assertion fails. The test passes locally and usually in CI, but the runner here was slow enough (note it was also fighting Postgres container churn) to blow the 50 ms budget. 

### 🛠️ Steps to reproduce

Flaky test - hard to reproduce. 
Loaded CI or hardware limitations might reproduce this.

### 📋 Logs and Screenshots

Failed Job: [job #94021328014](https://github.com/LFDT-Panurus/panurus/actions/runs/31566632721/job/94021328014)

Effi-S:
Fixes #2200

### 👾 Description

```bash
--- FAIL: TestTokens (8.33s)
    --- FAIL: TestTokens/TimeFiltering (0.17s)
        tokens.go:1427: Should be false
        Test:     TestTokens/TimeFiltering
        Messages: recent token should not be in results
```

Caused by flaky timing test

The test (token/services/storage/db/dbtest/tokens.go:1399) does:

```go
createAndDeleteToken("old_token", 0, "idemix")
time.Sleep(100 * time.Millisecond)
createAndDeleteToken("recent_token", 0, "idemix")
// return tokens deleted MORE than 50ms ago
tokens, err := db.GetDeletedTokensPendingSKICleanup(ctx, 50*time.Millisecond, 100)
...
assert.False(t, foundRecent, "recent token should not be in results")   // line 1427
```

The intent: recent_token was deleted only moments ago, so with a 50 ms cutoff it should be too young to appear. The problem is the margin is  **wall-clock dependent**

- recent_token needs to still be younger than 50 ms at the instant the query's now() is evaluated.
- Between the createAndDeleteToken("recent_token") call (which itself does INSERT + DELETE round-trips to Postgres) and the GetDeletedTokensPendingSKICleanup query executing, more than 50 ms of real time can easily elapse on a loaded CI runner.

When that happens, recent_token's deletion age crosses 50 ms before the query runs, so it's returned, foundRecent is true, and the assertion fails. The test passes locally and usually in CI, but the runner here was slow enough (note it was also fighting Postgres container churn) to blow the 50 ms budget. 

### 🛠️ Steps to reproduce

Flaky test - hard to reproduce. 
Loaded CI or hardware limitations might reproduce this.

### 📋 Logs and Screenshots

Failed Job: [job #94021328014](https://github.com/LFDT-Panurus/panurus/actions/runs/31566632721/job/94021328014)

### DIFF
--- a/token/services/storage/db/dbtest/tokens.go
+++ b/token/services/storage/db/dbtest/tokens.go
@@ -1398,17 +1398,20 @@ func TGetDeletedTokensPendingSKICleanup(t *testing.T, db TestTokenDB) {
 
 	// Test 5: Time filtering - only tokens older than duration
 	t.Run("TimeFiltering", func(t *testing.T) {
+		const gap = 1 * time.Second
+		const duration = 500 * time.Millisecond
+
 		// Create an old token
 		createAndDeleteToken("old_token", 0, "idemix")
 
-		// Wait a bit to ensure time difference
-		time.Sleep(100 * time.Millisecond)
+		// Wait long enough that old_token is comfortably older than `duration`.
+		time.Sleep(gap)
 
 		// Create a recent token
 		createAndDeleteToken("recent_token", 0, "idemix")
 
 		// Query with duration that should exclude the recent token
-		tokens, err := db.GetDeletedTokensPendingSKICleanup(ctx, 50*time.Millisecond, 100)
+		tokens, err := db.GetDeletedTokensPendingSKICleanup(ctx, duration, 100)
 		require.NoError(t, err, "query should not error")
 
 		// Verify old token is in results


AkramBitar (approved):
LGTM

PR #2205, opened 2026-08-12
identity: secret material (PKCS11 PIN, idemix Sk) and unbounded attacker bytes reach error strings and logs

### LINKED ISSUE #2069: identity: secret material (PKCS11 PIN, idemix Sk) and unbounded attacker bytes reach error strings and logs
### Summary

Several code paths in the identity stack put secret material (PKCS11 PINs, idemix user secret keys)
into error strings or debug logs, and several others echo unbounded attacker-controlled bytes
directly into log/error messages instead of hashing or truncating them — both are logging-hygiene
issues that this tree otherwise avoids (it already uses `utils.Hashable(...)` in several places for
exactly this reason).

### Where

Secret material in errors/logs:
- `token/services/identity/idemix/crypto/config.go:176` — wraps an error with `string(configRaw)`,
  the full serialized `IdemixConfig`, whose `Signer.Sk` field is the idemix user secret key.
- `token/services/identity/x509/crypto/pkcs11/pkcs11.go:35` — logs `opts` directly, where
  `PKCS11Opts` contains a `Pin` field.
- `token/services/identity/x509/crypto/setup.go:64` — logs `bccspConfig`, which embeds `*PKCS11`
  (same `Pin` field).
- `token/services/identity/x509/kmp.go:89` — logs `opts.PKCS11` twice at debug level.

Unbounded attacker-byte echo into logs/errors (should use `utils.Hashable(raw)` or lengths, as done
elsewhere in this tree):
- `token/services/identity/interop/htlc/deserializer.go:172`
- `token/services/identity/interop/htlc/validator.go:90`
- `token/services/identity/idemix/crypto/deserializer.go:115`
- `token/services/identity/idemix/deserializer.go:184`
- `token/services/identity/idemixnym/deserializer.go:106`
- `token/services/identity/x509/crypto/msp.go:245`

### Impact

- PIN/secret-key exposure: if `LogLevel` is set to debug in a deployment (a plausible operational
  choice for troubleshooting), the idemix user secret key or the PKCS11 PIN can land in log files,
  which are frequently shipped to less-trusted aggregation systems than the node itself, or retained
  longer than the credential's intended lifetime.
- Unbounded byte echo: logs/error messages become arbitrarily large under attacker control (a crafted
  large or adversarial identity blob), and error strings containing raw untrusted bytes can pollute
  structured-logging pipelines or make log injection easier.

### Reproduction

Not yet committed. For the secret-exposure items: set debug logging, trigger the corresponding error
path (e.g. malformed idemix config, PKCS11 setup with a bad slot), and grep the resulting log output
for the PIN/secret-key value. For the byte-echo items: pass an oversized or binary-garbage identity
through the affected deserializer/validator and confirm the raw bytes appear verbatim in the emitted
log/error.

### Suggested fix

- For secret-bearing structs, add a custom `String()`/`MarshalLogObject` that redacts `Sk`/`Pin`
  before it's ever passed to a logger or error constructor, or explicitly select only the
  non-secret fields to log (e.g. log `opts.PKCS11.Label`/`.Slot` but never `.Pin`).
- For attacker-controlled byte fields, replace direct interpolation with
  `utils.Hashable(raw)` (already imported and used for this exact purpose elsewhere in this package,
  e.g. `wallet/service.go:115,118`) or log only `len(raw)`.

### Severity

MEDIUM — secret exposure requires debug logging to be enabled; byte-echo is a hygiene/DoS-adjacent
issue rather than directly exploitable, but both are inconsistent with the pattern already used
elsewhere in this codebase.

Effi-S:
Fixes #2069

### Summary

Several code paths in the identity stack put secret material (PKCS11 PINs, idemix user secret keys)
into error strings or debug logs, and several others echo unbounded attacker-controlled bytes
directly into log/error messages instead of hashing or truncating them — both are logging-hygiene
issues that this tree otherwise avoids (it already uses `utils.Hashable(...)` in several places for
exactly this reason).

### Where

Secret material in errors/logs:
- `token/services/identity/idemix/crypto/config.go:176` — wraps an error with `string(configRaw)`,
  the full serialized `IdemixConfig`, whose `Signer.Sk` field is the idemix user secret key.
- `token/services/identity/x509/crypto/pkcs11/pkcs11.go:35` — logs `opts` directly, where
  `PKCS11Opts` contains a `Pin` field.
- `token/services/identity/x509/crypto/setup.go:64` — logs `bccspConfig`, which embeds `*PKCS11`
  (same `Pin` field).
- `token/services/identity/x509/kmp.go:89` — logs `opts.PKCS11` twice at debug level.

Unbounded attacker-byte echo into logs/errors (should use `utils.Hashable(raw)` or lengths, as done
elsewhere in this tree):
- `token/services/identity/interop/htlc/deserializer.go:172`
- `token/services/identity/interop/htlc/validator.go:90`
- `token/services/identity/idemix/crypto/deserializer.go:115`
- `token/services/identity/idemix/deserializer.go:184`
- `token/services/identity/idemixnym/deserializer.go:106`
- `token/services/identity/x509/crypto/msp.go:245`

### Impact

- PIN/secret-key exposure: if `LogLevel` is set to debug in a deployment (a plausible operational
  choice for troubleshooting), the idemix user secret key or the PKCS11 PIN can land in log files,
  which are frequently shipped to less-trusted aggregation systems than the node itself, or retained
  longer than the credential's intended lifetime.
- Unbounded byte echo: logs/error messages become arbitrarily large under attacker control (a crafted
  large or adversarial identity blob), and error strings containing raw untrusted bytes can pollute
  structured-logging pipelines or make log injection easier.

### Reproduction

Not yet committed. For the secret-exposure items: set debug logging, trigger the corresponding error
path (e.g. malformed idemix config, PKCS11 setup with a bad slot), and grep the resulting log output
for the PIN/secret-key value. For the byte-echo items: pass an oversized or binary-garbage identity
through the affected deserializer/validator and confirm the raw bytes appear verbatim in the emitted
log/error.

### Suggested fix

- For secret-bearing structs, add a custom `String()`/`MarshalLogObject` that redacts `Sk`/`Pin`
  before it's ever passed to a logger or error constructor, or explicitly select only the
  non-secret fields to log (e.g. log `opts.PKCS11.Label`/`.Slot` but never `.Pin`).
- For attacker-controlled byte fields, replace direct interpolation with
  `utils.Hashable(raw)` (already imported and used for this exact purpose elsewhere in this package,
  e.g. `wallet/service.go:115,118`) or log only `len(raw)`.

### Severity

MEDIUM — secret exposure requires debug logging to be enabled; byte-echo is a hygiene/DoS-adjacent
issue rather than directly exploitable, but both are inconsistent with the pattern already used
elsewhere in this codebase.

### DIFF
--- a/token/services/identity/idemix/crypto/config.go
+++ b/token/services/identity/idemix/crypto/config.go
@@ -173,7 +173,7 @@ func NewFabricCAIdemixConfig(issuerPublicKey []byte, dir string) (*Config, error
 func NewConfigFromRaw(issuerPublicKey []byte, configRaw []byte) (*Config, error) {
 	config := &config.IdemixConfig{}
 	if err := proto.Unmarshal(configRaw, config); err != nil {
-		return nil, errors.Wrapf(err, "failed to unmarshal idemix config at [%s]", string(configRaw))
+		return nil, errors.Wrapf(err, "failed to unmarshal idemix config of PK [%s]", utils.Hashable(issuerPublicKey))
 	}
 	// match public keys
 	if !bytes.Equal(issuerPublicKey, config.Ipk) {
--- a/token/services/identity/x509/crypto/config.go
+++ b/token/services/identity/x509/crypto/config.go
@@ -7,6 +7,8 @@ SPDX-License-Identifier: Apache-2.0
 package crypto
 
 import (
+	"fmt"
+
 	"github.com/LFDT-Panurus/panurus/token/services/identity/x509/crypto/pkcs11"
 	"github.com/LFDT-Panurus/panurus/token/services/identity/x509/crypto/protos-go/v1/config"
 	"github.com/go-viper/mapstructure/v2"
@@ -73,6 +75,20 @@ type PKCS11 struct {
 type KeyIDMapping struct {
 	SKI string `yaml:"SKI,omitempty"`
 	ID  string `yaml:"ID,omitempty"`
+}
+
+// String renders the PKCS11 configuration for logging with the PIN redacted, so
+// the secret can never reach a log line or error string, even when a *PKCS11 is
+// interpolated directly or as a field of a *BCCSP. It is nil-safe.
+func (p *PKCS11) String() string {
+	if p == nil {
+		return "<nil>"
+	}
+
+	return fmt.Sprintf(
+		"{Security:%d Hash:%s Library:%s Label:%s Pin:[REDACTED] SoftwareVerify:%t Immutable:%t AltID:%s KeyIDs:%v SessionCacheSize:%d}",
+		p.Security, p.Hash, p.Library, p.Label, p.SoftwareVerify, p.Immutable, p.AltID, p.KeyIDs, p.SessionCacheSize,
+	)
 }
 
 // ToBCCSPOpts converts the passed opts to `config.BCCSP`
--- a/token/services/identity/x509/crypto/redact_test.go
+++ b/token/services/identity/x509/crypto/redact_test.go
@@ -1,0 +1,103 @@
+/*
+Copyright IBM Corp. All Rights Reserved.
+
+SPDX-License-Identifier: Apache-2.0
+*/
+
+package crypto
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/hyperledger/fabric-lib-go/bccsp"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+const testPIN = "s3cr3t-pin-9999"
+
+// PKCS11.String must redact the PIN and never emit it verbatim, whether the
+// *PKCS11 is logged directly or as a field of a *BCCSP.
+func TestPKCS11_String_RedactsPin(t *testing.T) {
+	p := &PKCS11{Library: "/usr/lib/libpkcs11.so", Label: "tok", Pin: testPIN}
+
+	s := p.String()
+	assert.Contains(t, s, "[REDACTED]")
+	assert.NotContains(t, s, testPIN)
+	// Non-secret fields remain visible for debugging.
+	assert.Contains(t, s, "/usr/lib/libpkcs11.so")
+	assert.Contains(t, s, "tok")
+
+	// %v / %s go through the Stringer as well.
+	assert.NotContains(t, fmt.Sprintf("%s", p), testPIN) //nolint:staticcheck // Explicitly checking fmt.Sprintf
+	assert.NotContains(t, fmt.Sprintf("%+v", p), testPIN)
+}
+
+// A nil *PKCS11 must format without panicking (SW-only configs have a nil PKCS11).
+func TestPKCS11_String_Nil(t *testing.T) {
+	var p *PKCS11
+	assert.NotPanics(t, func() { _ = p.String() })
+	assert.Equal(t, "<nil>", p.String())
+	assert.NotPanics(t, func() { _ = fmt.Sprintf("%v", p) })
+}
+
+// Formatting a *BCCSP (as setup.go and kmp.go do) must redact the nested PKCS11 PIN,
+// and must not panic when the PKCS11 field is nil.
+func TestBCCSP_Format_RedactsPin(t *testing.T) {
+	t.Run("WithPKCS11", func(t *testing.T) {
+		cfg := &BCCSP{Default: "PKCS11", PKCS11: &PKCS11{Label: "tok", Pin: testPIN}}
+		out := fmt.Sprintf("%v", cfg)
+		assert.Contains(t, out, "[REDACTED]")
+		assert.NotContains(t, out, testPIN)
+	})
+	t.Run("NilPKCS11", func(t *testing.T) {
+		cfg := &BCCSP{Default: "SW", SW: &SoftwareProvider{Hash: "SHA2", Security: 256}}
+		assert.NotPanics(t, func() { _ = fmt.Sprintf("%v", cfg) })
+	})
+	t.Run("NilBCCSP", func(t *testing.T) {
+		var cfg *BCCSP
+		assert.NotPanics(t, func() { _ = fmt.Sprintf("%v", cfg) })
+	})
+}
+
+// getIdentityFactory must never mutate the caller's live config and must never panic,
+// across the nil, SW-only, and error paths. Regression test for the pointer-aliasing
+// redaction bug (Issue #2069) that overwrote the real PIN and dereferenced nil.
+func TestGetIdentityFactory_NoMutationNoPanic(t *testing.T) {
+	conf := &Config{CryptoConfig: &CryptoConfig{SignatureHashFamily: bccsp.SHA2}}
+
+	t.Run("NilBccspConfig", func(t *testing.T) {
+		// Default SW provider; nil bccspConfig is a supported, common path.
+		assert.NotPanics(t, func() {
+			_, err := getIdentityFactory(conf, nil, nil)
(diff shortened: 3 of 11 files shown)


token/services/identity/idemix/crypto/config.go
```diff
@@ -173,7 +173,7 @@ func NewFabricCAIdemixConfig(issuerPublicKey []byte, dir string) (*Config, error
 func NewConfigFromRaw(issuerPublicKey []byte, configRaw []byte) (*Config, error) {
 	config := &config.IdemixConfig{}
 	if err := proto.Unmarshal(configRaw, config); err != nil {
-		return nil, errors.Wrapf(err, "failed to unmarshal idemix config at [%s]", string(configRaw))
+		return nil, errors.Wrapf(err, "failed to unmarshal idemix config of PK [%s]", string(issuerPublicKey))
```
AkramBitar: **P1 — raw bytes are still interpolated, on the very line this PR edits.**

This swaps *which* bytes get dumped rather than removing the pattern: `string(issuerPublicKey)` interpolates a few hundred bytes of raw binary protobuf (containing NUL and non-UTF-8 bytes) into the error string — the same "unbounded raw bytes in messages" problem this PR fixes in the six deserializer/validator files.

Three lines below, this same function already does it correctly with `utils.Hashable(issuerPublicKey)`.

It also drops the only identifier for the payload that actually failed to parse, so this error can no longer be correlated with the `utils.Hashable(identityConfig.Raw)` log line in `token/services/identity/idemix/kmp.go:98` that covers the same blob.

```go
return nil, errors.Wrapf(err, "failed to unmarshal idemix config [%s] of PK [%s]", utils.Hashable(configRaw), utils.Hashable(issuerPublicKey))
```

Effort: one line.

token/services/identity/x509/crypto/redact_test.go
```diff
@@ -0,0 +1,103 @@
...
+	s := p.String()
+	assert.Contains(t, s, "[REDACTED]")
+	assert.NotContains(t, s, testPIN)
+	// Non-secret fields remain visible for debugging.
+	assert.Contains(t, s, "/usr/lib/libpkcs11.so")
+	assert.Contains(t, s, "tok")
+
+	// %v / %s go through the Stringer as well.
+	assert.NotContains(t, p.String(), testPIN)
+	assert.NotContains(t, p.String(), testPIN)
```
AkramBitar: **P2 — these two assertions don't test what the comment above them says.**

Both lines re-assert on `p.String()`, so the `%s` / `%+v` formatting paths have no coverage at all. `%+v` is the one that matters most here — it's the verb the patched `pkcs11/pkcs11.go` actually uses.

```go
assert.NotContains(t, fmt.Sprintf("%s", p), testPIN)
assert.NotContains(t, fmt.Sprintf("%+v", p), testPIN)
```

(The nested `*BCCSP` case is genuinely covered by `TestBCCSP_Format_RedactsPin` below — this is just the direct-formatting case.)

Effort: two lines.

token/services/identity/x509/crypto/config.go
```diff
@@ -75,6 +77,20 @@ type KeyIDMapping struct {
 	ID  string `yaml:"ID,omitempty"`
 }
 
+// String renders the PKCS11 configuration for logging with the PIN redacted, so
+// the secret can never reach a log line or error string, even when a *PKCS11 is
+// interpolated directly or as a field of a *BCCSP. It is nil-safe.
+func (p *PKCS11) String() string {
```
AkramBitar, line 83: **P3 (defer) — the pointer receiver means the value form still leaks.**

Redaction applies only to `*PKCS11`; a `PKCS11` **value** still formats the PIN verbatim (I confirmed `%+v` on a value prints `Pin:<the pin>`).

No caller logs a value today — `bccsp.go:56` passes `*p11Opts` by value into `skiMapper`, and `bccsp.go:50` converts to `pkcs11.PKCS11Opts`, a distinct type with no redacting `String()` — so this is a latent robustness gap rather than a live leak.

Worth *not* switching to a value receiver: the pointer form is what makes the nil case safe (and `TestPKCS11_String_Nil` depends on it). A small value-form wrapper, or just a test pinning the current behaviour so a future caller doesn't get surprised, would close it. Fine to leave for a follow-up.

Effort: ~20 min.

AkramBitar (commented):
**Problem:** error and log messages were dumping whole config objects and raw network data, leaking private keys and HSM PINs into log files and letting attacker-supplied bytes bloat them

**Fix:** the PIN is now printed as [REDACTED] and untrusted bytes are printed as a short SHA-256 fingerprint instead of the bytes themselves.


### Review: logging hygiene for secrets and untrusted bytes

Direction is right and the change is low-risk. I verified the head commit (`1097bc3`) compiles — including the `-tags pkcs11` path — and that the affected packages' tests pass. I also confirmed experimentally that the new `String()` fixes the `setup.go` and `kmp.go` leak sites *without* those files being edited, and that nested/nil cases don't panic.

Four findings, none blocking on their own; the two P1s are worth clearing before merge.

| # | Where | Priority | Effort |
|---|---|---|---|
| 1 | `idemix/crypto/config.go:176` | **P1** | 1 line |
| 2 | `x509/crypto/msp.go:245` (not in diff) | **P1** | 1 line, or 1 line of PR text |
| 3 | `x509/crypto/redact_test.go:33-34` | P2 | 2 lines |
| 4 | `x509/crypto/config.go:83` | P3 — defer | ~20 min |

Findings 1, 3 and 4 are inline. The remaining two can't be anchored inline:

---

**2. P1 — `token/services/identity/x509/crypto/msp.go:245` is listed in the PR description but not fixed.**

```go
errors.Errorf("could not decode pem bytes [%v]", idBytes)
```

This is still present, and it's the highest-impact of the six byte-echo sites the description enumerates: `%v` on a `[]byte` renders a decimal array (`[45 45 45 ...]`), roughly 4x the input size, and `getCertFromPem` is reachable from `IdentityFactory.DeserializeFullIdentity` with bytes supplied by a counterparty — so a 1 MB identity blob becomes a ~4 MB error string that then gets logged.

Severity-wise this is pre-existing rather than a regression, but it's P1 because the PR description claims it's handled: merging as-is closes #2069 with a documented site still open, and the next reader trusts the description. `utils.Hashable(idBytes)` matches the fix applied elsewhere in this PR. Either fix it or drop it from the description.

---

**Out of scope, but relevant to #2069:** `token/services/identity/x509/kmp.go` yaml-marshals the raw opts map into `idConfig.Config`, persisting the HSM PIN in cleartext in the identity store. That outlives any log rotation, so on severity it's the largest exposure in this area — but it needs its own issue and some design, not a bolt-on to a logging PR.

Also worth noting: the `pkcs11/pkcs11.go` redaction sits behind `//go:build pkcs11`, so it compiles only in tagged builds and the new test file never exercises it.

AkramBitar (changes requested):

token/services/identity/idemix/crypto/config.go
Effi-S: Yeah good point. 
Not a big issue since it's just the publicKey and metadata but fixed it anyway

token/services/identity/x509/crypto/redact_test.go
Effi-S: Done

token/services/identity/x509/crypto/config.go
Effi-S: Done

AkramBitar (approved):
LGTM

PR #2208, opened 2026-08-12
fix(tokengen): inject the public params into the chaincode package

### LINKED ISSUE #2202: [BUG]: `tokengen --cc` produces a chaincode package with empty public parameters
### 👾 Description

`cc.GeneratePackage` matches the wrong path when injecting the public parameters
into the token chaincode package. The replacer never fires, so `tcc.tar` ships
the placeholder `var Params = ``` instead of the generated public parameters.

In more detail:

`cmd/tokengen/cobra/pp/cc/cc.go:41` tests:

```go
if strings.HasSuffix(s, "github.com/LFDT-Panurus/panurus/token/tcc/params.go") {
    return "", paramsFile.Bytes()
}
```

but the file that holds the parameters is `token/services/network/fabric/tcc/params.go`
(the `token/tcc/...` layout is a leftover from the token-sdk → panurus move).

`s` is the *absolute filesystem path* of the file being packaged (from
`filepath.Walk` over the module root in the FSC packager), so the
`github.com/LFDT-Panurus/panurus/` prefix only matches at all in a GOPATH-style
checkout — and the tail is wrong in every layout. Result: no file is replaced
and the package contains the unmodified

```go
package tcc

var Params = ``
```

The integration harness gets this right and can be used as the reference:
`integration/nwo/token/fabric/cc/tcc.go:33` uses
`DefaultTokenChaincodeParamsReplaceSuffix = "/token/services/network/fabric/tcc/params.go"`.

### 🛠️ Steps to reproduce

```bash
tokengen gen fabtokenv1 --cc -o /tmp/out    # same for zkatdlognoghv1
tar xOzf /tmp/out/tcc.tar code.tar.gz | tar xOz | grep -a 'var Params'
# => var Params = ``   (expected: the base64 public parameters)
```

### 📋 Logs and Screenshots

_No response_

### 🌐 Network Environment

### 📦 Panurus Version

main

### 🐹 Go Version

go1.26.5

### 💻 Operating System

Linux (Ubuntu/Debian)

## Impact

Any chaincode package built with `tokengen ... --cc` is unusable: the deployed
TCC has no public parameters. Both `fabtokenv1`
(`cmd/tokengen/cobra/pp/fabtokenv1/gen.go:77`) and `zkatdlognoghv1`
(`cmd/tokengen/cobra/pp/zkatdlognoghv1/gen.go:112`) are affected. The
integration tests do not catch it because they use `tcc.GenericBackend`, which
has the correct suffix.

## Suggested fix

1. Export the suffix from one place (e.g. reuse
   `cc.DefaultTokenChaincodeParamsReplaceSuffix`) and use it in both the CLI and
   the integration backend, so the two cannot drift again.
2. Extend `cc_test.go` beyond the current error-path-only test: package into a
   temp dir and assert that the resulting `code.tar.gz` contains the expected
   `var Params = "<base64>"`. That assertion would have caught this.

mbrandenburger:
GeneratePackage matched the suffix

    github.com/LFDT-Panurus/panurus/token/tcc/params.go

but the file holding the public parameters lives at

    token/services/network/fabric/tcc/params.go

(the token/tcc layout is a leftover of the token-sdk move). The packager passes the absolute path of every file it packages, so the replacer never fired and `tokengen ... --cc` produced a package still carrying the placeholder `var Params = ``', i.e. a token chaincode without public parameters.

The integration harness got this right, so the two definitions had silently drifted apart. Move the chaincode import path, the params file suffix and the params template into a single package in the root module, shared by tokengen and by the harness, and use it on both sides.

Cover the replacer with a test that fails on the old suffix.

Fixes #2202

### DIFF
--- a/cmd/tokengen/cobra/pp/cc/cc.go
+++ b/cmd/tokengen/cobra/pp/cc/cc.go
@@ -14,7 +14,9 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/LFDT-Panurus/panurus/token/services/network/fabric/tcc/ccpackage"
 	"github.com/hyperledger-labs/fabric-smart-client/integration/nwo/fabric/packager"
+	"github.com/hyperledger-labs/fabric-smart-client/integration/nwo/fabric/packager/replacer"
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 )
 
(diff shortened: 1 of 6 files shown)


AkramBitar (approved):
LGTM

Issue #2053, opened 2026-07-30
network-driver: tcc AreTokensSpent dereferences a nil PublicParameters, panicking


adecaro:
### Summary

`AreTokensSpent` calls `cc.PublicParameters.GraphHiding()` without ever checking that `PublicParameters` is non-nil. If `TokenServicesFactory` succeeds (returns a `nil` error, so `Initialize` does not fail) but yields a `nil` `PublicParameters` value — a state production code never guards against — the very next `areTokensSpent` request panics on a method call through a nil interface.

### Where

`token/services/network/fabric/tcc/tcc.go:328`

### Impact

Any code path where public-parameters initialization can report success while leaving `PublicParameters` nil (e.g. a driver bug, or a factory implementation that forgets to populate it in some branch) turns every subsequent `areTokensSpent` call into a guaranteed panic. Combined with #2052 (`GetValidator`'s poisoned `sync.Once`), there is also no way to recover from this state at runtime.

### Reproduction

Reproduced locally with a unit test (not yet committed):

```go
cc := &chaincode2.TokenChaincode{
    TokenServicesFactory: func(_ []byte) (PublicParameters, Validator, error) {
        return nil, &mock.Validator{}, nil   // "successful" init, nil PublicParameters
    },
}
require.Panics(t, func() { cc.AreTokensSpent([]byte(`["id1"]`), fakestub) })
```

Happy to include this regression test in the fix PR.

### Suggested fix

Add a nil check on `cc.PublicParameters` before calling `GraphHiding()`, returning an error (e.g. "chaincode not properly initialized") instead of panicking.

### Severity

MEDIUM (dampened by the top-level `recover()` in `Invoke`, but leaves the chaincode permanently non-functional for this call with no diagnosable error)

PR #2209, opened 2026-08-12
fix(tcc): re-surface chaincode initialization errors and allow retry

### LINKED ISSUE #2052: network-driver: tcc GetValidator's poisoned sync.Once silently swallows the initialization error after the first failed call
### Summary

`GetValidator` uses `cc.initOnce.Do(func() {...})`, capturing the `TokenServicesFactory` error in a closure-local variable. `sync.Once` guarantees the closure body runs at most once *regardless of whether it succeeds or fails*. If the very first call fails (e.g. transient error reading/parsing public params), every subsequent call to `GetValidator` — and therefore every subsequent `AreTokensSpent`/`ProcessRequest` invocation that depends on it — silently returns `(nil, nil)` as if initialization had succeeded, with no way to retry.

### Where

`token/services/network/fabric/tcc/tcc.go:172-186`

### Impact

A transient failure during the chaincode's first initialization attempt (e.g. a momentarily-unreadable public-params file, or any other one-off error in `TokenServicesFactory`) permanently and silently disables validation for the lifetime of the chaincode container: later calls report no error at all, masking the fact that the chaincode was never actually initialized, rather than surfacing the original failure or allowing a retry.

### Reproduction

```go
v1, err1 := cc.GetValidator("dGVzdA==")
require.Error(t, err1)               // first call observes the failure
v2, err2 := cc.GetValidator("dGVzdA==")
require.NoError(t, err2)             // second call silently swallows the cached failure
require.Equal(t, 1, calls)           // factory only ever ran once
```

### Suggested fix

Track the cached error explicitly and re-surface it on every call after a failed initialization (or reset `sync.Once` so a later call can retry initialization) instead of silently reporting success once the closure has run.

### Severity

MEDIUM — combined with #2053 (`AreTokensSpent` nil `PublicParameters` dereference), there is no runtime recovery path once this state is reached.

AkramBitar:
## Problem

The token chaincode builds its validator and public-parameters manager lazily, on the first request that needs them, guarded by `sync.Once`. `sync.Once` runs its closure once **whether it succeeds or fails**, and the error was kept in a call-local variable:

```go
func (cc *TokenChaincode) GetValidator(builtInParams string) (Validator, error) {
	var firstInitError error          // a FRESH local on every call
	cc.initOnce.Do(func() {
		if err := cc.Initialize(builtInParams); err != nil {
			firstInitError = err      // only assigned by the caller that ran Do
		}
	})
	if firstInitError != nil {
		return nil, firstInitError
	}
	return cc.Validator, nil          // (nil, nil) on calls 2..N
}
```

The error lives in **per-call** scope, while the state it describes (the `sync.Once` outcome) is **per-object**. The one caller that ran the closure learns the truth; every caller after it inherits the assumption of success.

## Scenario

Chaincode container just started: `initOnce` unused, `cc.Validator == nil`, `cc.PublicParameters == nil`.

### Request 1 — `invoke` (a transfer), and the public-params file happens to be unreadable right now

1. `ProcessRequest` → `cc.GetValidator(Params)`
2. `var firstInitError error` → nil *(local #1)*
3. `initOnce.Do(...)` → has not run yet, so it **runs** the closure
4. `cc.Initialize()` → file read fails → returns an error
5. closure assigns `firstInitError = err` → local #1 holds the error
6. `firstInitError != nil` → `return nil, err`
7. `ProcessRequest` → `shim.Error(...)` → **status 500 with the correct message** ✅
8. Side effect: `initOnce` is now **permanently spent**, yet `cc.Validator` and `cc.PublicParameters` are still nil

### Request 2 — `invoke` again a minute later; the file is readable now, and it makes no difference

1. `ProcessRequest` → `cc.GetValidator(Params)`
2. `var firstInitError error` → nil *(local #2 — a brand-new variable, unrelated to local #1)*
3. `initOnce.Do(...)` → already ran → **skipped**. `Initialize` is never called again, the file is never re-read
4. nothing assigns to local #2 → it is **still nil**
5. `firstInitError != nil` → **false** → falls through
6. `return cc.Validator, nil` → returns **`(nil, nil)`** — "here is your validator, no error"
7. `ProcessRequest` checks `if err != nil` → nil → proceeds
8. `validator.UnmarshallAndVerifyWithMetadata(...)` → method call on a **nil interface → panic**
9. `Invoke`'s `defer recover()` catches it → `shim.Error("failed responding [runtime error: invalid memory address ...]")` → status 500 with a message that says nothing about public parameters ❌

### Requests 3, 4, 5, … for the life of the container

Identical to request 2: same panic, same misleading message, `Initialize` never attempted again. `areTokensSpent` takes the same path and panics one line later, on `cc.PublicParameters.GraphHiding()` (#2053).

Net effect: one transient failure — e.g. a momentarily unreadable `PUBLIC_PARAMS_FILE_PATH` — is reported once and then silently disables validation until the container is restarted, with error messages that point nowhere near the cause.

## Fix

In short GetValidator now remembers whether initialization succeeded (a mutex-guarded initialized flag) instead of merely whether it ran (sync.Once), so a failure returns the real error to every caller and the next request retries, and it can never report success with a nil validator.

Track whether initialization **succeeded**, not merely whether it **ran**. A `sync.Mutex` plus an `initialized` flag replace `sync.Once`:

- **failure** → the wrapped error is returned to the caller and nothing is cached, so the next request **retries**;
- **success** → cached once; the factory is never called again;
- `GetValidator` can no longer return "no validator and no error".

The mutex also provides the happens-before edge for the exported `Validator` / `PublicParameters` fields that `sync.Once` used to give.

Request 2 above now goes: `initialized` is still `false` → `Initialize` runs again → file readable → validator cached → transfer proceeds. And while the file stays unreadable, every request returns the real error instead of `(nil, nil)`.

## Changes

- `token/services/network/fabric/tcc/tcc.go` — the fix.
- `token/services/network/fabric/tcc/tcc_init_test.go` — regression tests: failure-then-retry, persistent failure (never `(nil, nil)`), single init on success, `Invoke` reporting the error on *every* call for `invoke` and `areTokensSpent`, and concurrent init. The three failure-path tests fail against the pre-fix code.
- `docs/services/network-fabric.md` — new `### Validator Initialization` subsection.

## Verification

`go test ./token/services/network/fabric/tcc/...`, `go test -race` on the package, `go build ./...`, `go vet`, `golangci-lint run token/services/network/fabric/tcc/...` → all clean.

## Notes

Retry-on-next-call was chosen over permanently caching the error (the issue allows either): a container that can never recover is the worse failure mode.

Out of scope: #2053 (`AreTokensSpent` nil `PublicParameters` dereference). This change removes its trigger, but the defensive nil check belongs to that issue.

Fixes #2052


SaidAltury-ibm (approved):
LGTM

PR #2210, opened 2026-08-12
fix(identity): avoid nil notifier dereference in LocalMembership.Close

### LINKED ISSUE #2067: identity/membership: nil-pointer dereference in LocalMembership.Close
### Summary

`LocalMembership.Close` falls through to calling `UnsubscribeAll()` on a nil `notifier` when
`identityDB.Notifier()` returns an error other than "not supported" — it logs the error but doesn't
return, so the nil-pointer dereference happens right after, on the shutdown path.

### Where

`token/services/identity/membership/lm.go:253-262` (paraphrased from the read source structure):

```go
notifier, err := identityDB.Notifier()
if err != nil {
	if !errors.Is(err, ErrNotSupported) {
		logger.Errorf("failed to get notifier: %s", err)
	}
	// falls through instead of returning here
}
...
notifier.UnsubscribeAll()
```

When `err` is a real (non-"not supported") error, `notifier` is nil, and the code proceeds to call a
method on it anyway.

### Impact

This is on the shutdown/`Close` path. A storage backend that returns an unexpected error from
`Notifier()` — as opposed to the expected "not supported" sentinel — turns an already-degraded
shutdown into a panic instead of a clean (if incomplete) teardown. This is likely to surface during
operational incidents (a backend already failing is more likely to also fail `Notifier()`), which is
the worst time for `Close` itself to crash.

### Reproduction

Not yet committed. Mock the identity DB's `Notifier()` to return a non-`ErrNotSupported` error and
call `LocalMembership.Close()`; assert it returns an error (or otherwise handles the nil notifier)
rather than panicking.

### Suggested fix

Return (or otherwise skip the `UnsubscribeAll` call) when the error isn't the expected sentinel:

```go
notifier, err := identityDB.Notifier()
if err != nil {
	if !errors.Is(err, ErrNotSupported) {
		logger.Errorf("failed to get notifier: %s", err)
		return errors.Wrap(err, "failed to get notifier during close")
	}
} else {
	notifier.UnsubscribeAll()
}
```

### Severity

MEDIUM — shutdown-path panic, triggered by a backend error condition that's plausible during
operational incidents.

AkramBitar:
Fixes #2067

## What was wrong

When the identity component shuts down, it tries to unhook itself from the storage layer's
**notifier** — the mechanism the store uses to tell it "an identity record just changed".

There are two possible answers to "give me your notifier":

1. *"I don't have one"* (`storage.ErrNotSupported`) — expected, and handled: nothing to unhook.
2. *"Something went wrong"* — a real error. Here the code logged the error and then carried on
   to call `UnsubscribeAll()` on the notifier it never received, i.e. on `nil`. That panics.

The timing is what makes it matter: a storage backend is most likely to fail this call when it
is already unhealthy, so shutdown would crash exactly during an incident — turning an
incomplete-but-clean teardown into a panic.

## The fix

Return on *every* error, not just the expected one. Unexpected errors are still logged; either
way the unsubscribe step is skipped and shutdown completes cleanly. This is the same handling
`subscribeNotifier` already uses on the startup side.

```go
token/services/identity/membership/lm.go, inside Close():

Before — the bug
notifier, err := l.identityDB.Notifier()
if err != nil {
    if errors.Is(err, storage.ErrNotSupported) {
        return                  // ← only THIS error path returns
    }
    logger.Errorf("failed to get identity notifier: [%s]", err)
}                               // ← any other error falls through, notifier == nil
notifier.UnsubscribeAll()       // ← panic

After — the fix
notifier, err := l.identityDB.Notifier()
if err != nil {
    if !errors.Is(err, storage.ErrNotSupported) {
        logger.Errorf("failed to get identity notifier: [%s]", err)
    }
    // no notifier, nothing to close
    return                      // ← now covers EVERY error
}
notifier.UnsubscribeAll()       // ← only reached when notifier is non-nil

The check is inverted so the log is the conditional part and the return is unconditional — same shape subscribeNotifier already uses at lm.go:490.
```

Note `Close()` has no return value, so the error is logged rather than propagated.

## Verification

`TestLocalMembership_Close_NotifierError` mocks the store to return a non-sentinel error and
asserts `Close()` does not panic (twice, to also cover idempotency). It fails with
`invalid memory address or nil pointer dereference` against the old code and passes with the fix.

```
go test ./token/services/identity/membership/... -run TestLocalMembership_Close
```

`make lint-auto-fix` and `make checks` are clean.

### DIFF
--- a/docs/services/identity.md
+++ b/docs/services/identity.md
@@ -222,6 +222,12 @@ core.TMSProvider.Update            (token/core/tms.go)
                     └── Role.Done() → LocalMembership.Close()
 ```
 
+`LocalMembership.Close()` is best-effort: it releases its key managers, then unsubscribes
+from the identity store's change notifier. If the store cannot supply a notifier — because
+it does not support one (`storage.ErrNotSupported`) or because it fails outright — the
+unsubscribe step is skipped and, in the failure case, logged. `Close()` returns no error and
+must never panic, since it runs on the shutdown path of an already-degraded node.
+
 `role.Registry.Done()` closes wallets through a local `interface{ Close() }` assertion
 rather than through `driver.Wallet`, so wallet types with nothing to release need not
 implement a no-op `Close()`. If you add a wallet type that owns a goroutine, a ticker or
--- a/token/services/identity/membership/lm.go
+++ b/token/services/identity/membership/lm.go
@@ -276,11 +276,11 @@ func (l *LocalMembership) Close() {
 
 		notifier, err := l.identityDB.Notifier()
 		if err != nil {
-			if errors.Is(err, storage.ErrNotSupported) {
-				// notithing to close
-				return
+			if !errors.Is(err, storage.ErrNotSupported) {
+				logger.Errorf("failed to get identity notifier: [%s]", err)
 			}
-			logger.Errorf("failed to get identity notifier: [%s]", err)
+			// no notifier, nothing to close
+			return
 		}
 		if err := notifier.UnsubscribeAll(); err != nil {
 			logger.Errorf("failed to unsubscribe [%s]: [%s]", l.IdentityType, err)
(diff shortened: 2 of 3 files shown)


Effi-S (approved):
LGTM

PR #2211, opened 2026-08-12
fix(auditdb): align locker semantics, split key namespaces, stop nesting retries

### LINKED ISSUE #2040: auditdb/locker: misc medium/low findings (empty-lock semantics, key-namespace collision, retry storms)

AkramBitar:
## In one paragraph

The auditor briefly "locks" each person involved in a transaction, so that two auditors can never write the same person's records at the same time. There are two interchangeable implementations of that lock, chosen by configuration: `memory` for a single node, `postgres` for a cluster. They did not behave the same way, which meant the same auditor code was correct on one deployment and broken on the other. This PR makes them behave identically, and adds a test suite that runs the same expectations against both so they cannot drift apart again.

Fixes #2040

## What was broken

1. **A transaction with nobody to lock could not be written at all (Postgres only).** Asked afterwards "do you still hold your locks?", the Postgres locker answered "no, you lost them" when the truth was "you never took any". That failed two legitimate flows: a transaction whose inputs and outputs yield no enrollment IDs, and an auditor that validates and appends without calling `Audit`.
2. **The auditor could crash (memory only).** Two unrelated kinds of key — transaction IDs and person IDs — shared a single map. A transaction ID that happened to equal a person ID made the code read one kind of value as the other and panic.
3. **One audit could block for ~10 minutes.** The auditor retried lock acquisition 10 times *around* the Postgres locker's own 1-minute internal wait, so the two waiting loops multiplied — and each of those attempts polled the database on a fixed interval, identical on every replica, so contenders stayed in lockstep and kept colliding.

**In short** 

Think of two security guards (memory and postgres) with the same job: "make sure nobody edits Alice's records at the same time." They each had their own rulebook, and the rulebooks disagreed in 3 ways — one would crash if Alice's ID matched a transaction ID, one would falsely say "you lost access" for empty transactions, and both would keep each other waiting 10x longer than needed. This PR gives them one shared rulebook and fixes all three inconsistencies.

## What this PR changes

- **"No locks" is success, on both backends.** `AssertLocksHeld` reports locks that were *lost*, not locks that were never *taken*.
- **The in-memory locker keeps its two key namespaces in separate maps**, so the collision is impossible by construction rather than by convention. Re-acquiring under a live transaction is now a proper refresh instead of overwriting the record and leaking the locks it dropped.
- **One layer does the waiting.** The Postgres locker waits out contention itself with jittered exponential backoff bounded by a single deadline, and the auditor stops retrying as soon as it sees `ErrLockAcquireTimeout` — the locker saying "I already spent the whole budget". Worst case for one audit is now about one `acquireDeadline`, not `maxRetries × acquireDeadline`, and a contended wait costs a few dozen round trips instead of hundreds.
- **The contract is written down** on the `Locker` interface, and `locker/conformance_test.go` runs it against every backend. The interface had no docs at all, which is how the two implementations drifted in the first place.

## Testing

- `make checks` and `make lint-auto-fix` clean.
- `go test -race` green on `token/services/storage/auditdb/...` and `token/services/auditor/...`, with the Postgres cases running against a real container (not skipped).
- Regression value verified by running each new test against the code it targets: the concurrency test reproduces `panic: semaphore: released more than held`; the four new Postgres tests fail with the exact misclassifications and deletions described above; the conformance suite fails `EmptyEnrollmentIDs` on Postgres only and `ContentionIsReported` on both once the masking deadline is corrected.

### DIFF
--- a/token/services/storage/auditdb/locker/memory/memory.go
+++ b/token/services/storage/auditdb/locker/memory/memory.go
@@ -9,8 +9,10 @@ package memory
 import (
 	"context"
 	"sync"
+	"time"
 
 	"github.com/LFDT-Panurus/panurus/token/services/storage/auditdb/locker/dedup"
+	"github.com/LFDT-Panurus/panurus/token/services/storage/auditdb/locker/errs"
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"golang.org/x/sync/semaphore"
 )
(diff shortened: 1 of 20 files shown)


token/services/storage/auditdb/locker/memory/memory.go
```diff
@@ -33,57 +74,228 @@ func New() *Locker {
...
+// of that record: performed lock-free, two concurrent re-acquisitions that both
+// narrow the set read the same stale record, both conclude the same ID is now
+// stale, and both release its permit — which panics the process, since releasing
+// a semaphore more than it was acquired is a programming error in
+// golang.org/x/sync/semaphore rather than a no-op.
 func (m *Locker) AcquireLocks(ctx context.Context, anchor string, eIDs ...string) error {
 	deduped := dedup.AndSort(eIDs)
 
+	st := m.lockAnchor(anchor)
+	defer m.unlockAnchor(anchor, st)
```
Effi-S, line 143: hi @AkramBitar, 
Can you please double check that you didn't introduce a potential deadlock here ?

AkramBitar: Hi @Effi-S 

Good catch,  fixed. 

Regards,
Akram

Effi-S (approved):
LGTM

Issue #2020, opened 2026-07-30
tokens.amount is unindexed and the spendable-token query has no ORDER BY or LIMIT, so every candidate row is fetched and re-parsed in Go


adecaro:
## Summary

The tokens table carries a queryable `amount NUMERIC(78, 0) NOT NULL` column (`token/services/storage/db/sql/common/tokens.go:1394-1424`), but:

1. **No index covers `amount`.** The five declared indexes are:

```sql
CREATE INDEX ... idx_spent_%s              ON %s ( is_deleted, owner );
CREATE INDEX ... idx_ski_cleanup_%s        ON %s ( is_deleted, spent_at );
CREATE INDEX ... idx_owner_wallet_id_%s    ON %s ( owner_wallet_id );
CREATE INDEX ... idx_owner_wallet_part_%s  ON %s ( owner_wallet_id, token_type ) WHERE is_deleted = false AND owner = true;
CREATE INDEX ... idx_issued_%s             ON %s ( redeemed, token_type ) WHERE issuer = true;
```

2. **The selector's query ignores `amount` entirely, and is unbounded and unordered.** `buildSpendableTokensIteratorByQuery` (`token/services/storage/db/sql/common/tokens.go:392-406`) selects the hex-encoded TEXT `quantity` column with no `ORDER BY` and no `LIMIT`:

```go
return q.Select().
    FieldsByName("tx_id", "idx", "token_type", "quantity", "owner_wallet_id").
    From(q.Table(db.table.Tokens)).
    Where(HasTokenDetails(driver.QueryTokenDetailsParams{
        WalletID:           walletID,
        TokenType:          typ,
        Spendable:          driver.SpendableOnly,
        LedgerTokenFormats: db.getSupportedTokenFormats(),
    }, nil)).
    Format(db.ci)
```

`QueryTokenDetailsParams` (`token/services/storage/db/driver/token.go`) has no amount bounds at all, so amount-range filtering cannot be expressed even by callers that want it.

## Consequences

Every selection materialises the wallet's **entire** spendable set for the token type and re-parses each row's hex `quantity` into a `token.Quantity` in Go, regardless of how few tokens the transaction actually needs. Cost is linear in wallet size on every transfer, and the work is repeated per retry. Wallet fragmentation therefore feeds back into query cost: the more the wallet fragments, the more rows each selection reads. The `amount` column that would make this an indexed range scan is present but unused by this path.

## Enabling primitives already exist

No new SQL machinery is required:

[...]

PR #2221, opened 2026-08-13
fix(storage): keep token amounts exact instead of truncating to uint64 (#2021)

### LINKED ISSUE #2021: TokenRecord.Amount is uint64 but the column is NUMERIC(78,0), silently truncating token amounts on write
## Summary

`TokenRecord.Amount` is declared `uint64` (`token/services/storage/db/driver/token.go:50-51`):

```go
// Amount is the Quantity converted to decimal
Amount uint64
```

while the backing column is `amount NUMERIC(78, 0) NOT NULL` (`token/services/storage/db/sql/common/tokens.go:1394-1424`) — 78 decimal digits, chosen to hold a 256-bit value. `uint64` tops out at 20 digits.

The narrowing happens in the production write path, `AppendToken` (`token/services/tokens/storage.go:210`):

```go
Amount: q.ToBigInt().Uint64(),
```

`big.Int.Uint64()` is documented as undefined when the value does not fit in 64 bits — it returns the low 64 bits with no error and no panic.

## Consequences

For any token whose quantity exceeds 2^64-1, the `amount` column silently receives a wrapped value while the authoritative hex `quantity` column stays correct. The two columns then disagree, with no error raised at write time.

Quantities above `uint64` are representable elsewhere in the stack: `token.ToQuantity` (`token/token/quantity.go:67-85`) returns a `BigQuantity` backed by `big.Int` whenever precision is not 64, so a driver configured with a larger precision can legitimately produce such values.

Today the damage is limited because no query path reads `amount` — the selector reads the hex `quantity` column instead (`token/services/storage/db/sql/common/tokens.go:392-406`). That also means the corruption is currently invisible.

## Why this needs fixing before amount-aware selection

Any work that filters, orders or range-scans on `amount` (see the companion issue on the missing index) makes this column load-bearing. A wrapped `amount` would then produce incorrect selections — a large token appearing as a small one — rather than merely sitting unread.

## Suggested direction

Widen the field to `*big.Int` (or a decimal string) and drop the `Uint64()` conversion, so the stored value matches the column's declared range and the hex `quantity`. Worth checking whether any consumer of `TokenRecord.Amount` relies on the `uint64` type, and adding a round-trip test with a value above 2^64-1 to lock the behaviour in.

Related: #2017.

AkramBitar:
Fixes #2021

## The problem

Every token has an amount. We store it twice: once as text (the source of truth) and once
as a number, so the database can add amounts up.

The text version can hold very large amounts. The number version was declared as a type
that stops at 20 digits. When a token's amount was bigger than that, the conversion did
not fail or warn — it just kept the leftover part and stored a completely different
number. The two copies then disagreed, and nothing said so.

Nobody notices today because no query uses the number column yet. As soon as one does —
for example to pick which tokens to spend — a huge token could look tiny and the wrong
tokens would be chosen.

## The fix

- Store the amount as a big number that has no size limit, so the value written matches
  the source of truth exactly.
- If an amount is missing, or too large even for the database column, the write is now
  rejected with a clear error instead of quietly saving something wrong.
- Added tests: one saves an amount larger than the old limit and reads it back unchanged,
  one checks the bad cases are refused.

## Note for reviewers

No shipped driver can produce such an amount today — fabtoken rejects a precision above 64
bits (`fabtoken/v1/setup/setup.go:256`), zkatdlog supports 16/32/64
(`zkatdlog/nogh/v1/setup/setup.go:42`) — so this is a prerequisite for #2020, not a live
incident.

Also: only Postgres keeps such amounts exactly. SQLite gives a `NUMERIC` column NUMERIC
affinity and converts an integer wider than int64 to REAL, so it cannot store them exactly
whatever the Go type is — after this change that errors on read instead of silently
corrupting. That needs a per-driver column type plus a migration, and it matters for #2020's
index and `ORDER BY` on `amount`; happy to open a separate issue if you'd like it tracked.

Checks: `make checks` and `make lint` clean; storage and token service tests pass,
including the PostgreSQL suite.

### DIFF
--- a/token/services/storage/db/driver/token.go
+++ b/token/services/storage/db/driver/token.go
@@ -49,8 +49,9 @@ type TokenRecord struct {
 	Quantity string
 	// Type is the type of token
 	Type token.Type
-	// Amount is the Quantity converted to decimal
-	Amount uint64
+	// Amount is the Quantity converted to decimal as a big integer to support arbitrary
+	// precision. It must be non-nil and match Quantity.
+	Amount *big.Int
 	// Owner is used to mark the token as owned by this node
 	Owner bool
 	// Auditor is used to mark this token as audited by this node
@@ -371,4 +372,13 @@ var (
 	// ErrTokenAlreadyLocked is returned by TokenLockStore.Lock when the token is
 	// already locked by another transaction (primary-key conflict on the lock row).
 	ErrTokenAlreadyLocked = errors.New("token already locked")
+	// ErrAmountMissing is returned when a record carries no amount. The amount column is NOT NULL.
+	ErrAmountMissing = errors.New("no amount specified")
+	// ErrAmountOutOfRange is returned when an amount is too wide for the amount column to hold.
+	// Storing it would leave the amount disagreeing with the authoritative quantity.
+	ErrAmountOutOfRange = errors.New("amount exceeds maximum supported size")
+	// ErrAmountNegative is returned when a token record carries a negative amount. Token
+	// quantities are unsigned, so a negative amount means the record was built outside
+	// token.ToQuantity. Movement and transaction amounts are signed and are not affected.
+	ErrAmountNegative = errors.New("amount is negative")
 )
--- a/token/services/storage/db/sql/common/transactions.go
+++ b/token/services/storage/db/sql/common/transactions.go
@@ -12,6 +12,7 @@ import (
 	"encoding/json"
 	errors2 "errors"
 	"fmt"
+	"math/big"
 	"strings"
 	"time"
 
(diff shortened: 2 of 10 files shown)


token/services/storage/db/sql/common/transactions.go
```diff
@@ -12,6 +12,7 @@ import (
 	"encoding/json"
 	errors2 "errors"
 	"fmt"
+	"math/big"
 	"strings"
 	"time"
```
adecaro, line 18: **Nit:** `validateAmount` only checks `amount.BitLen() > maxAmountBits`. `BitLen()` is defined on the absolute value, so a negative `*big.Int` with small magnitude would pass this untouched. Currently unreachable via the production path — `token2.ToQuantity`/`.ToBigInt()` only ever produce non-negative values from unsigned hex parsing — but worth a comment or an explicit `Sign() < 0` check so the invariant survives the next caller that doesn't go through `ToQuantity`.

token/services/storage/db/driver/token.go
```diff
@@ -46,8 +46,9 @@ type TokenRecord struct {
 	Quantity string
 	// Type is the type of token
 	Type token.Type
-	// Amount is the Quantity converted to decimal
-	Amount uint64
+	// Amount is the Quantity converted to decimal as a big integer to support arbitrary
+	// precision. It must be non-nil and match Quantity.
+	Amount *big.Int
```
adecaro, line 54: **Nit:** confirmed this is the only place `Amount`'s type changed and that every `TokenRecord{...}` construction site in the repo was updated to match (grepped repo-wide, not just the diff) — no stale `uint64`/`int` literal remains. No action needed, just noting the check was done rather than inferred from the diff.

adecaro (commented):
## Review: does this fully address #2021?

Yes. `TokenRecord.Amount` is widened to `*big.Int` end-to-end and the truncating `.Uint64()` call in `AppendToken` is gone, replaced by a fail-closed `validateAmount` guard that rejects (instead of silently corrupting) any amount too wide for the `NUMERIC(78,0)` column or missing entirely. Test coverage reproduces the reporter's exact `2^64+1` scenario, and a Postgres-only round-trip test proves the fix holds at the DB layer, not just in-memory.

**Verified independently** (build, targeted tests including the Postgres-backed round-trip, `go vet`, `gofmt`, base-drift check, repo-wide grep for stale `TokenRecord` construction sites) — see the full write-up for the command table. No Blocking findings.

**Should-fix:** PR title doesn't carry the issue number (`titleCarriesIssueNumber: false`) — body does say "Closes #2021" so linkage is fine, just the title convention.

**Design note:** `mergeable: CONFLICTING` against current `main` — confirmed via a scratch `git merge --no-commit --no-ff main` (aborted after inspection). Three unrelated commits landed on `main` since this PR's base and touch the same two files (`dbtest/tokens.go`, `driver/token.go`). Textual, not semantic — will need a rebase before merge.

Two nits inline below. Nice fix — good root-cause type widening rather than a point patch, and the validation closes off future silent-corruption paths, not just the one in the issue.

token/services/storage/db/driver/token.go
AkramBitar: Thanks for confirming — no action needed on our end, the grep was done as part of the fix to ensure all construction sites were updated.

token/services/storage/db/sql/common/transactions.go
AkramBitar: Good catch — fixed in 57e0c44f4.

Added `validateTokenAmount`, which wraps `validateAmount` and rejects `Sign() < 0` with a dedicated `ErrAmountNegative` sentinel; `StoreToken` now goes through it, so the invariant is enforced at the write rather than relying on every caller coming from `token.ToQuantity`.

I deliberately kept the sign check out of the shared `validateAmount`: movement and transaction amounts are signed by design, a negative one meaning tokens left the wallet (`driver.MovementRecord.Amount`), so those two call sites stay magnitude-only. Both functions now carry a doc comment spelling out that split, and `dbtest` asserts that `StoreToken` with a negative amount returns `ErrAmountNegative`.

adecaro (approved):
LGTM

PR #2229, opened 2026-08-14
feat(evm): endorsers return the delta, the initiator stops building one

atharrva01:
Stacked on #2228 (the diff here will shrink to just the endorsement commit once that merges).

The initiator built its own StateDelta before contacting any endorser, running the request through
the same DeltaFactory as the responder. That is validation work, and validating is exactly what this
flow delegates to endorsers, so every client repeated what N endorsers were already doing, plus a
`getToken` read per input.

It also cost liveness in a way that was hard to diagnose. The delta covers the public parameters hash
and version, and the digest covers the whole struct, so an initiator reading parameters on the other
side of an update from its endorsers computed a different digest and discarded every signature as an
unknown signer. Its ledger read had the same shape: a node lagging behind failed to build a delta at
all and never contacted an endorser, for a request every endorser would have signed.

It bought no security property either way. The contract only checks that a threshold of registered
endorsers signed the digest, so a quorum that wanted to apply something else would never need the
initiator's cooperation.

`EndorseResponse` now carries the delta the endorser built, next to its signature, the way a Fabric
proposal response carries the RWSet. The initiator binds each returned delta to the anchor it asked
about, checks the StateDelta invariants, recovers the signature over that delta's own digest, and
groups by delta so one endorser translating differently is outvoted instead of failing the whole
transaction. A shortfall where more than one delta was seen reports that explicitly, since that is a
determinism bug in the translator, not endorsers being unavailable.

Verified: unit tests and `-race` across the module, the Week 4 gate (drives the real initiator and
responders over in-memory sessions, pinned to the committed Solidity fixture) unchanged, all 55 forge
tests, and the fungible integration suite against real Besu, twice.

### DIFF
--- a/docs/services/network-ethereum.md
+++ b/docs/services/network-ethereum.md
@@ -310,16 +310,16 @@ sequenceDiagram
         E1->>E1: Validate token request
         E1->>E1: Compute state delta
         E1->>E1: Sign state delta
-        E1-->>Driver: Signature 1
+        E1-->>Driver: State delta + signature 1
     and
         Driver->>E2: Request endorsement
         E2->>E2: Validate token request
         E2->>E2: Compute state delta
         E2->>E2: Sign state delta
-        E2-->>Driver: Signature 2
+        E2-->>Driver: State delta + signature 2
     end
     
-    Driver->>Driver: Assemble state update + signatures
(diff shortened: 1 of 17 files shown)


adecaro (approved):
LGTM

PR #2231, opened 2026-08-15
fix(evm): ten bugs from a self-review pass over the network driver

atharrva01:
Went through the evm driver directory end to end looking for correctness bugs and fixed everything that verified against the actual code, six that could produce a wrong result and four smaller ones. One commit per fix, each with its own tests.

High:
- finality/manager.go: a run of transient read errors right before the timeout used to resolve as Invalid even though the chain was never actually reached; now it only reports Invalid if at least one poll succeeded, otherwise it reports the read failure instead.
- config.go: the default finality timeout (5m) was shorter than real PoS time-to-finality (~13m) on the finalized tag, so a transaction could never be seen as final and would time out as Invalid no matter what. Raised the default and added a floor Validate enforces.
- pp/watcher.go + driver.go: a failed public-parameters reload was marked seen anyway, so the version was never retried and a node could keep running stale parameters.
- pp/provider.go: PublicParams read the parameter bytes and their version as two separate calls with no consistency check, so a version bump between them could hand back bytes for the wrong version. Now brackets the read and retries on a mismatch.
- driver.go: SetupPublicParams checked a single network-wide endorsement field instead of resolving it per TMS, so an endorsing TMS on a node running more than one TMS never got public parameter validation wired up.
- nonce.go + submitter.go: the nonce was advanced before the submit attempt, not after it succeeded, so a rejected transaction still burned a nonce and a concurrent submit could double-allocate the gap. Reworked around a single WithNonce call that only advances after the caller reports success.

Medium/low:
- driver.go: registerEndorser used sync.Once, so a second TMS trying to register as an endorser silently discarded its config instead of refusing. Now it refuses loudly and keeps the first registration, since the view registry can only ever route to one responder per process.
- config.go + endorsement/authorize.go: an empty allowlist on an enabled endorser was documented as defaulting to the TMS's own nodes, but nothing built that default, so a node came up looking healthy and never actually registered as an endorser. Validate now rejects it at startup, and the stale comments are corrected.
- network.go: Broadcast never checked that the envelope's own anchor matched the anchor baked into its delta. The chain only trusts the delta's anchor while everything local tracks the envelope's, so a mismatch would apply and commit on chain under one anchor while the local side waited forever on another. Broadcast now refuses a mismatched envelope before spending any gas.
- config.go: the blockTag field comment only mentioned finalized and safe, leaving out that latest is also accepted (used by the local Besu test harness, no reorg protection, already documented on the constant itself). Comment corrected, no behavior change.

Also bumped the module to fabric-smart-client v0.17.0, needed to build this branch (dupes #2228, will collapse on rebase).

Tested with go build, go vet, gofmt, golangci-lint on every changed package, and go test -race across the whole module including the anvil e2e test.

### DIFF
--- a/x/token/services/network/evm/config.go
+++ b/x/token/services/network/evm/config.go
@@ -131,8 +131,10 @@ type EndorsementConfig struct {
 	// Threshold is the number of distinct endorser signatures a transaction needs. It must match the
 	// threshold the EndorsementVerifier was constructed with.
 	Threshold uint `yaml:"threshold"`
-	// Allowlist is the FSC identities permitted to request endorsement. Empty means the policy is
-	// resolved from the TMS network's nodes at wiring time.
+	// Allowlist is the FSC identities permitted to request endorsement. It is fail-closed: when this
+	// node is configured to endorse, an empty allowlist is a validation error rather than a default
+	// that resolves to anyone. There is no automatic "the TMS network's nodes" fallback; every
+	// permitted requester has to be named.
 	Allowlist []string `yaml:"allowlist"`
 	// Endorsers binds each endorser's Ethereum address to its FSC identity.
 	Endorsers []EndorserBinding `yaml:"endorsers"`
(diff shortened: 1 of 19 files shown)


adecaro (approved):
LGTM

PR #2232, opened 2026-08-15
fix(evm): a second bug-hunt pass, five subsystems plus nwo and statedelta

atharrva01:
Follow-up to #2231. That one came from an unstructured first pass. This one is five separate, targeted adversarial reviews, one per trust-critical area, plus two more once the first five turned up clean enough to ask what was still unreviewed. Same process: verify every finding against the real code before touching anything, one commit per fix.

Builds on evm-bugfix-hunt-findings (#2231), so this stacks on that branch until it merges.

Crypto and wire format (crypto/, eip712/, keys/, statedelta/):
- keys.AnchorFromTxID decodes attacker-influenced hex from a request's own anchor, a transfer input's TxId, and an envelope off the wire, with zero fuzz coverage while its siblings all had it. Added FuzzAnchorFromTxID.

JSON-RPC client and ABI:
- baseFee read a null eth_getBlockByNumber result as a legitimate zero base fee (unmarshaling JSON null into a non-pointer struct is a silent no-op), which fed straight into every transaction's fee pricing. Fixed by making the target a pointer, same pattern GetTransactionReceipt already uses.
- Call didn't classify reverts the way EstimateGas does, an asymmetry in the same client for the same kind of RPC call.
- Hex-prefix parsing disagreed between files in the same package (0x/0X tolerated in one, 0x only in three others). Consolidated into one helper.
- eth_getLogs/eth_getTransactionReceipt decoders had no fuzz coverage despite parsing node-supplied bytes. Added it.

Endorsement trust boundary (endorsement/):
- Core safety properties held up under tracing: no blind signing, full EIP-712 digest binding, signature-recovery dedup, fail-closed authorization ordering, threshold enforcement. No HIGH findings.
- Fixed a stale doc comment claiming the allowlist is per TMS. Traced it through configNetworkResolver.ConfigFor: it isn't, the whole Config (allowlist included) comes from whichever TMS declares the network first.
- Two real findings need bigger design work than a single commit and are intentionally not fixed here: Ledger.GetState reads a floating block tag per call instead of a pinned block number, which can make two honest endorsers build non-identical deltas within one validation pass; and two TMS on one network structurally cannot have different endorser allowlists or thresholds, since Config bundles genuinely network-shared fields with per-TMS policy fields.

Solidity contracts (contracts/src/):
- 55/55 (now 56/56) Foundry tests passing, no HIGH findings.
- isSpent/areTokensSpent on a graph-hiding TokenState clone always answered false, including for tokens that were actually spent, because that mode never populates the marker those functions look up. On-chain enforcement in applyStateDelta was never affected, but any external caller (wallet, explorer, monitor) got a confidently wrong answer. Now reverts instead.
- Left alone: the EIP-712 domain separator has no rebuild-on-chain-id-change fallback (the standard OZ pattern), but the scenario it guards against doesn't apply to a permissioned network the TMS operator controls.

Test suite quality:
- The suite is unusually strong overall (golden vectors cross-checked with external tooling, independently recomputed digests, real concurrency tests). Found three fuzz harnesses that discarded their result and so verified nothing beyond "did not panic," including two added in this same PR's earlier commits. Proved the gap was real by injecting the exact bug the doc comment claimed each property caught and confirming the mutation was undetected before the fix, and caught after.

nwo/ (never independently reviewed before):
- Zero unit tests anywhere in the repo drove a setup/public-parameters-update request through the real endorsement pipeline (Responder.endorse, DeltaFactory.Build, Translator.writeSetup). The integration harness's own way of bumping public parameters mid-test hand-assembles and signs a delta directly instead. Added a real unit test driving one through the actual pipeline; it passed on the first try, which closes the coverage gap rather than a live bug. The deeper part, that the integration harness itself still bypasses the real pipeline, is not fixed here.
- NewSetupUpdater didn't fail fast on a zero TokenState address.
- DeploySpec.Threshold and SetupUpdaterConfig.Threshold silently disagree on what zero means, in the same package.

statedelta/ (previously treated as an already-correct precondition, first real review here):
- Validate() didn't reject a setup delta carrying metadata, even though TokenState.sol's _applySetup does.
- IsGraphHiding() was read once per output instead of once per action, a latent divergence from the Fabric reference translator this package mirrors. Not reachable by either shipped driver today, both hardcode it false, confirmed real with a mutation test using a mock that varies its answer across calls.

Round 3: every directory had already had a dedicated pass by round 2, so this time five subagents split by bug class instead (nonce handling, JSON-RPC/fee/finality robustness, crypto and signing correctness, Solidity contract risk, driver orchestration), plus a fresh look at the endorsement trust boundary since endorsers now build their own delta (#2229) instead of the initiator.

- NonceManager.Reset could reassign a nonce a concurrent, not-yet-broadcast caller was still holding, so one of two submissions could silently lose its transaction. Already closed by this PR's own WithNonce redesign (holds the lock across the whole allocate-and-use step instead of giving nonces back individually), independently rediscovered by round 3, confirmed already fixed. Nothing more to do.
- registerEndorser swallowed every failure past the empty-allowlist and second-network checks (unusable signing key, allowlist resolution, authorizer or responder construction, view-registry registration) as a log line. installEndorsement and Driver.New always reported success regardless, so a node explicitly configured to endorse could come up looking healthy while never answering a request, discoverable only once a quorum it was needed for timed out. Now returns an error that propagates.
- An endorser's reply carries the StateDelta it built. The initiator JSON-decodes and EIP-712-hashes it before it has any reason to trust the signature over it, and StateDelta.Validate had no upper bound on entry counts or field sizes. A single dishonest registered endorser could force real CPU and memory cost per request with an oversized delta. Validate now rejects one before doing any per-element work.
- NewSignerFromBytes and LoadKey both went through secp256k1.PrivKeyFromBytes, which silently reduces an out-of-range key scalar modulo the curve order instead of erroring. A key file at or above the order would load as a different, unrelated key with no error. Both now check the overflow flag.
- Log had no field for eth_getLogs's removed flag, so a log from a block a reorg undid was indistinguishable from a canonical one. TxHashByAnchor could report a transaction hash off one, or trip its own duplicate-commit guard on one. Removed is now decoded and reorged-out logs are dropped before deciding.
- TxHashByAnchor always searched up to "latest" regardless of the operator's configured block tag, unlike every other reader in the module. Now takes and uses the same tag.
- Solidity contracts and Go/Solidity crypto parity got a fresh adversarial pass (EIP-1167 clone risk in TokenStateFactory, EIP-712 domain/type-hash parity, signature malleability, address checksums). Both came back clean, no new findings. 55/55 Foundry tests still passing.

One more finding needs the same fix Ledger.GetState from round 2 still needs (resolve a block tag to a concrete block number, not currently possible with EVMClient): a multi-token query can straddle a finalization boundary the same way, reading a mix of pre- and post-finalization state across a batch. Not fixed here, deferred alongside Ledger.GetState. (A third candidate in this class, ChainProvider.PublicParams reading bytes and version as two unsynchronized calls, was already independently fixed earlier in this same PR by the torn-read retry commit, confirmed before writing this up.)

Also, the per-TMS allowlist/threshold sharing bug from round 2 turns out to be bigger than scoped there: two TMS on one network/channel share the whole Config, including Contracts.TokenState, not just the endorsement policy, because Network is memoized per (network, channel) with the namespace left out of the key. Still not fixed here, needs splitting Config into network-shared and per-TMS parts.

Round 4: rounds 1-3 were static review. This one is adversarial and PoC-driven instead: for each trust boundary already reviewed, actually try to break it with a constructed exploit or a live reproduction, not just re-read the code. Four targets, one subagent each.

- Endorsement quorum coercion: tried a malicious endorser signing a divergent delta, a signature replayed from a different anchor, double-counting one key toward threshold (via malleability and via duplicate registration), and timing manipulation of which of two diverging honest deltas wins. All four failed. Cross-checked against the real EndorsementVerifier.sol through forge too, no Go/Solidity disagreement in either direction. The harness itself was mutation-tested first (broke the dedup check on purpose, confirmed the suite catches it) before trusting any of this.
- Nonce/submitter concurrency: 300 goroutines hammering Submit with 70% randomized failure injection across every network call, under -race, nonce read back from the actual signed transaction bytes rather than internal state. Zero races, zero duplicate or gapped nonces across around 65 runs. The WithNonce fix holds under real stress, not just under reasoning about the lock scope.
- EIP-712 signatures: real reverted transactions against real deployed TokenState clones through forge. Flipped a signature to its malleable form (mathematically impossible to pass here, not just empirically clean on this fixture, since the signer only ever emits low-s), replayed a signature across two different clones and across chain ids, and tried three near-miss encodings looking for a digest collision. All rejected.
- Multi-TMS cross-contamination: built it instead of just reading it. Two TMS on one network, each with its own TokenState address, driven through the real Driver.New/Network.Connect/QueryTokens/Broadcast path with only the transport mocked. Confirms the finding above live: TMS B's reads, its broadcast, and its actual EIP-712 signing domain (read off the running object) all target TMS A's TokenState. Kept as the regression test for whenever the Config split lands.

No new bugs. Every boundary held under an actual attempt to break it.

### DIFF
--- a/.github/workflows/nightly-fuzz.yml
+++ b/.github/workflows/nightly-fuzz.yml
@@ -173,6 +173,14 @@ jobs:
             dir: ./x/token/services/network/evm
             pkg: ./client
             func: FuzzBytesToAddress
+          - name: evm-json-log-to-log
+            dir: ./x/token/services/network/evm
+            pkg: ./client
+            func: FuzzJSONLogToLog
+          - name: evm-json-receipt-to-receipt
+            dir: ./x/token/services/network/evm
+            pkg: ./client
+            func: FuzzJSONReceiptToReceipt
 
     steps:
       - name: Checkout code
--- a/.github/workflows/nightly-fuzz.yml
+++ b/.github/workflows/nightly-fuzz.yml
@@ -141,6 +141,10 @@ jobs:
             dir: ./x/token/services/network/evm
             pkg: .
             func: FuzzComputeAnchor
+          - name: evm-anchor-from-txid
+            dir: ./x/token/services/network/evm
+            pkg: ./keys
+            func: FuzzAnchorFromTxID
           - name: evm-abi-decode-bytes
             dir: ./x/token/services/network/evm
             pkg: ./abi
--- a/.gitignore
+++ b/.gitignore
@@ -15,6 +15,7 @@ cmd/token_validation_service/out/
 coverage.out
 /.codex/
 plan.md
+x/BUG_HUNT_PROMPT.md
 
 # The EVM integration suites generate a node binary tree and a network under out/ and testdata/.
 # Both hold real Go files, so a stray "git add -A" commits them and a linter then reports on them.
--- a/docs/services/network-ethereum.md
+++ b/docs/services/network-ethereum.md
@@ -323,12 +323,12 @@ sequenceDiagram
     
     Note over Driver,State: On-Chain Execution Phase
     Driver->>Node: eth_sendRawTransaction
-    Node->>Contract: Execute applyStateUpdate()
-    Contract->>Contract: Verify endorser signatures
-    Contract->>Contract: Check signature threshold
-    Contract->>Contract: Double Spending Check
+    Node->>Contract: Execute applyStateDelta()
+    Contract->>Contract: Recompute the EIP-712 digest
+    Contract->>Contract: Verify endorser quorum
+    Contract->>Contract: Check public parameters are current
     Contract->>State: Apply state delta
-    Contract->>Contract: Emit StateUpdateEvent
+    Contract->>Contract: Emit StateCommitted
     
     Node->>Node: Include in block
     Node->>Node: Mine block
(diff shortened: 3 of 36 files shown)


adecaro (approved):
LGTM

PR #2244, opened 2026-08-17
fix(selector): expire token locks by the consuming transaction

### LINKED ISSUE #2018: SQLite lease cleanup joins token_locks on the wrong column, expiring locks by the locked token's tx instead of the consumer's
## Summary

The `token_locks` lease cleanup is implemented separately per backend, and the two implementations join on **different columns**. The SQLite version joins `token_locks` to `token_requests` on `tl.tx_id = tr.tx_id`; the Postgres version correlates on `tl.consumer_tx_id = tr.tx_id`.

`token/services/storage/db/sql/sqlite/tokenlock.go:36-46`:

```go
q.Select().
    Fields(tokenLocks.Field("tx_id")).
    From(tokenLocks.Join(tokenRequests, cond.Cmp(tokenLocks.Field("tx_id"), "=", tokenRequests.Field("tx_id")))).
    Where(common4.IsExpiredToken(tokenRequests, tokenLocks, c.leaseExpiry)).
```

`token/services/storage/db/sql/postgres/tokenlock.go:72-80`:

```go
cond.Cmp(tokenRequests.Field("tx_id"), "=", tokenLocks.Field("consumer_tx_id")),
cond.FieldIn(tokenRequests.Field("status"), driver.Deleted, driver.Orphan),
```

The two columns mean different things. In `token_locks`, `(tx_id, idx)` identifies the **locked token** — i.e. the transaction that *created* it — while `consumer_tx_id` identifies the transaction that *is trying to spend* it. A lease should expire when the **consuming** transaction is `Deleted`/`Orphan`, which is what Postgres does. SQLite instead expires the lock based on the status of the transaction that produced the token.

## Consequences on SQLite

- **Locks leak.** When a consuming transaction becomes `Deleted` or `Orphan`, its locks are not released by the status branch. They are only reclaimed once `created_at` passes `leaseExpiry` (default 3m), so tokens stay unspendable for the remainder of the lease.
- **Locks are released too early.** If the transaction that *created* the token is `Deleted`/`Orphan`, the lock is deleted even though the consuming transaction may be in flight — dropping a lock that is still live.

## Second defect: the DELETE is scoped to the wrong key

The subquery selects only `tl.tx_id`, and the outer statement is `DELETE FROM <token_locks> WHERE tx_id IN (...)` (`sqlite/tokenlock.go:39-45`, `:48-51`). Since the primary key is `(tx_id, idx)`, matching on `tx_id` alone deletes locks for **every index of that transaction**, not just the expired row. One expired lock on output 0 evicts live locks on outputs 1..n of the same transaction. The Postgres version has no such issue: it deletes by correlated predicate rather than by a partial-key `IN`.

## Why it is not caught by tests

`token/services/storage/db/sql/sqlite/tokenlock_test.go:36-51` asserts the generated SQL **string** rather than the behaviour, so it encodes the current (incorrect) join as expected output:

```go
Expect(query).To(Equal("DELETE FROM TokenLocks WHERE tx_id IN (" +
    "SELECT tl.tx_id " +
    "FROM TokenLocks AS tl " +
    "LEFT JOIN Requests AS tr " +
    "ON tl.tx_id = tr.tx_id " + ...
```

There is no test asserting that a lock held by a `Deleted` consumer is actually released, nor that sibling indices survive cleanup. Note the join also renders as `LEFT JOIN`, so rows with no matching request participate with `NULL` status.

## Expected behaviour

Both backends should expire a lock when the **consumer** transaction is `Deleted`/`Orphan` or the lease has aged out, and should delete only the matching `(tx_id, idx)` rows. The shared `IsExpiredToken` predicate (`token/services/storage/db/sql/common/tokenlock.go:111-116`) already exists and can back a single implementation for both backends.

Behavioural tests should replace the string-equality assertions: insert locks for a `Deleted` consumer and assert release; insert multiple indices under one `tx_id`, expire one, and assert the siblings survive.

## Impact

SQLite is the default local/dev and in-memory backend, so selector behaviour differs between development and Postgres production deployments. Both symptoms — leaked locks and prematurely dropped locks — manifest as spurious `SelectorSufficientButLockedFunds` or as double-spend attempts surfacing later in validation.

Related: #2017.

HayimShaul:
The SQLite lease cleanup correlated `token_locks` to `token_requests` on `tl.tx_id = tr.tx_id` — the transaction that *created* the locked token — while Postgres correlated on `tl.consumer_tx_id`, the transaction that is trying to spend it. Only the consumer says anything about whether a lock is still live, so on SQLite a lock held by a `Deleted`/`Orphan` consumer leaked until the lease aged out, and a lock held by an in-flight consumer was dropped as soon as the producing transaction died. The SQLite delete was also scoped as `tx_id IN (SELECT tl.tx_id ...)` while the primary key is `(tx_id, idx)`, so one expired lock evicted the live locks on every other output of the same transaction.

The cleanup statement now lives once in the shared store as `IsStaleLock`, correlated on `consumer_tx_id` and scoped per row: SQLite inherits it, Postgres keeps only its logging of the rows about to go. The generated Postgres SQL is unchanged. The SQLite tests that asserted the generated SQL string — and so encoded the wrong join as the expected output — are replaced by behavioural cases in the shared `dbtest` suite, which runs on SQLite, the in-memory backend and Postgres; four of them fail against the previous SQLite implementation. `docs/services/selector.md` now documents the lease-expiry semantics.

Note this changes SQLite/in-memory behaviour visibly: locks of dead consumers are released on the next tick instead of up to `leaseExpiry` later, and locks of live consumers are no longer dropped when the producing transaction dies.

`make checks` passes. `make lint` reports 17 pre-existing `revive: unhandled-error` findings that reproduce identically on a pristine `origin/main`; none are in the files touched here, and the changed packages lint clean.

Fixes #2018

### DIFF
--- a/docs/services/selector.md
+++ b/docs/services/selector.md
@@ -119,6 +119,42 @@ To prevent double-spending *before* the transaction is committed to the ledger, 
 2.  **Concurrency Control**: If another concurrent process has already locked that token, the insertion fails, and the selector moves on to the next candidate.
 3.  **Lock Release**: Locks are released either when the transaction reaches finality (success/failure) or when a timeout occurs, ensuring that tokens do not remain permanently inaccessible due to crashed or abandoned transactions.
 
+#### Lease expiry
+
+Every `leaseCleanupTickPeriod`, `sherdlock` runs a cleanup pass over the `TokenLocks`
+table that releases a lock when **either** of the following holds.
+
+> **Both `leaseExpiry` and `leaseCleanupTickPeriod` must be non-zero for the cleanup
+> goroutine to start.** If either is zero the pass never runs, so locks held by
+> `Deleted` or `Orphan` consumers are never released and those tokens remain
+> permanently unselectable. Setting `leaseExpiry: 0` to disable time-based expiry
+> while relying on consumer-status release is therefore not supported.
+
+*   the **consuming** transaction — the one that took the lock, stored in
+    `consumer_tx_id` — has reached `Deleted` or `Orphan`, so it will never spend the
+    token; or
+*   the lease is older than `leaseExpiry`, which covers the consumer that crashed or was
+    abandoned without ever reaching a terminal status.
+
+Two properties of the pass are worth spelling out:
+
+*   **The status that matters is the consumer's, not the producer's.** A lock row is keyed
+    by `(tx_id, idx)`, which identifies the *locked token* and therefore the transaction
+    that created it. That transaction's status says nothing about whether the lock is
+    still live, so it is never used to expire a lease.
+*   **Expiry is per token, not per transaction.** Only the affected `(tx_id, idx)` rows are
+    deleted; the other outputs of the same transaction keep their locks.
+
+The pass is the same statement on every SQL backend (SQLite and Postgres), so lock
+expiry behaviour is identical across those backends. `created_at` is stored as
+`TIMESTAMPTZ`, so the comparison with the database-side `NOW()` expression is always
+timezone-consistent on Postgres regardless of the session `TimeZone` setting.
+On Postgres a single replica per TMS runs the pass per tick, elected through an advisory
+lock; SQLite is non-distributed and always runs it locally.
+
+The in-memory locker described below does not use the `TokenLocks` table and never
+expires locks via `Cleanup`; its lifecycle is entirely managed in process.
+
 ### In-Memory Locker Internals
 
 The `simple` driver keeps its locks in memory (`token/services/selector/simple/inmemory`)
--- a/token/services/storage/db/dbtest/tokenlock.go
+++ b/token/services/storage/db/dbtest/tokenlock.go
@@ -14,8 +14,12 @@ import (
 	driver3 "github.com/LFDT-Panurus/panurus/token/services/storage/db/driver"
 	"github.com/LFDT-Panurus/panurus/token/services/utils"
 	"github.com/LFDT-Panurus/panurus/token/token"
+	fscerrors "github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 	"github.com/stretchr/testify/require"
 )
+
+// agedLease is the leaseExpiry used for the aged-lease test.
+const agedLease = time.Hour
 
 func TokenLocksTest(t *testing.T, cfgProvider cfgProvider) {
 	t.Helper()
@@ -55,6 +59,12 @@ var tokenLockDBCases = []struct {
 	Fn   func(*testing.T, driver3.TokenStore, driver3.TokenLockStore, driver3.TokenTransactionStore)
 }{
 	{"TestFully", TestFully},
+	{"TestReleaseOnDeletedConsumer", TestReleaseOnDeletedConsumer},
+	{"TestReleaseOnOrphanConsumer", TestReleaseOnOrphanConsumer},
+	{"TestKeepOnDeletedProducer", TestKeepOnDeletedProducer},
+	{"TestKeepSiblingIndices", TestKeepSiblingIndices},
+	{"TestReleaseOnAgedLease", TestReleaseOnAgedLease},
+	{"TestKeepFreshPendingLock", TestKeepFreshPendingLock},
 }
 
 func TestFully(t *testing.T, tokenDB driver3.TokenStore, tokenLockDB driver3.TokenLockStore, tokenTransactionDB driver3.TokenTransactionStore) {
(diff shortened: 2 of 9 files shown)


token/services/storage/db/sql/common/tokenlock.go
```diff
@@ -117,3 +117,45 @@ func IsExpiredToken(tokenRequests, tokenLocks common3.Table, leaseExpiry time.Du
...
+// IsStaleLock matches the lock rows whose lease has aged out, or whose consuming
+// transaction is Deleted or Orphan. The correlation is on consumer_tx_id, the
+// transaction that is trying to spend the token: (tx_id, idx) identifies the locked
+// token, i.e. the transaction that created it, whose status says nothing about
+// whether the lock is still live. The condition is correlated rather than a
+// partial-key IN on tx_id, so cleanup removes only the matching (tx_id, idx) rows
+// and leaves the other indices of the same transaction locked. See #2018.
+func IsStaleLock(tokenLocks, tokenRequests common3.Table, leaseExpiry time.Duration) cond.Condition {
+	return cond.Or(
+		cond.OlderThan(tokenLocks.Field("created_at"), leaseExpiry),
```
AkramBitar, line 149: **Should fix here.** Timezone mismatch between the two halves of this comparison on Postgres.

`created_at` is declared `TIMESTAMP` (line 98) — no time zone — but `cond.OlderThan` renders `NOW() - INTERVAL '…'`, which is a `timestamptz`. Postgres resolves `timestamp < timestamptz` by interpreting the bare value in the **session** `TimeZone`, not UTC, even though `Lock` writes `time.Now().UTC()`.

For a lock created at 12:00 UTC with a 5s lease:

| session `TimeZone` | stored `12:00` read as | effect |
|---|---|---|
| `UTC` | 12:00 UTC | correct |
| `Asia/Tokyo` | 03:00 UTC | already 9h "old" — deleted on the first cleanup tick, freeing a token whose spend is still in flight, so two selections can race for it |
| `America/New_York` | 16:00 UTC | 4h in the future — the lease does not expire for ~4 hours |

SQLite is unaffected because `datetime('now')` is UTC on both sides. That is also why the new claim at `docs/services/selector.md:142` that lock lifetime is backend-independent does not hold as written.

Pre-existing rather than introduced here, but this PR consolidates the query into one place and documents the behaviour as uniform, so it is the natural moment. `created_at TIMESTAMPTZ`, or comparing against `NOW() AT TIME ZONE 'UTC'`, closes it. Note `postgres/tokenlock.go:109` already selects `NOW()` next to `created_at`, so the same skew shows up in the debug output too.

token/services/storage/db/sql/common/tokenlock.go
```diff
@@ -117,3 +117,45 @@ func IsExpiredToken(tokenRequests, tokenLocks common3.Table, leaseExpiry time.Du
...
 }
+
+// IsStaleLock matches the lock rows whose lease has aged out, or whose consuming
+// transaction is Deleted or Orphan. The correlation is on consumer_tx_id, the
+// transaction that is trying to spend the token: (tx_id, idx) identifies the locked
+// token, i.e. the transaction that created it, whose status says nothing about
+// whether the lock is still live. The condition is correlated rather than a
+// partial-key IN on tx_id, so cleanup removes only the matching (tx_id, idx) rows
+// and leaves the other indices of the same transaction locked. See #2018.
+func IsStaleLock(tokenLocks, tokenRequests common3.Table, leaseExpiry time.Duration) cond.Condition {
```
AkramBitar: *Optional — hardening.* `IsStaleLock(tokenLocks, tokenRequests, …)` takes its two `common3.Table` arguments in the opposite order from its sibling `IsExpiredToken(tokenRequests, tokenLocks, …)` fourteen lines above. Both parameters are the same type, so a swapped call compiles and renders valid-but-nonsense SQL (`DELETE FROM TokenLocks WHERE Requests.created_at < … OR EXISTS (SELECT 1 FROM TokenLocks WHERE TokenLocks.tx_id = Requests.consumer_tx_id …)`) that silently deletes the wrong rows, or none.

Given this PR exists precisely because a wrong table correlation went unnoticed for a while, aligning the two signatures — or giving the two tables distinct named types so the compiler catches a swap — seems worth doing here.

docs/services/selector.md
```diff
@@ -119,6 +119,31 @@ To prevent double-spending *before* the transaction is committed to the ledger,
...
+
+*   **The status that matters is the consumer's, not the producer's.** A lock row is keyed
+    by `(tx_id, idx)`, which identifies the *locked token* and therefore the transaction
+    that created it. That transaction's status says nothing about whether the lock is
+    still live, so it is never used to expire a lease.
+*   **Expiry is per token, not per transaction.** Only the affected `(tx_id, idx)` rows are
+    deleted; the other outputs of the same transaction keep their locks.
+
+The pass is the same statement on every SQL backend, so lock lifetime does not depend on
+whether a deployment runs on SQLite, the in-memory backend, or Postgres. On Postgres a
```
AkramBitar: *Doc accuracy.* "lock lifetime does not depend on whether a deployment runs on SQLite, the in-memory backend, or Postgres" — two points against it:

1. The in-memory *locker* never expires locks at all: `token/services/selector/sherdlock/inmemory/locker.go:47` implements `Cleanup` as `return nil`, and the `simple` driver's locker does not use the `TokenLocks` table. The sentence sits two lines above a heading named "In-Memory Locker Internals", so a reader will apply it there. It holds only for the `memory` *SQL* persistence (`db/sql/memory`), which is SQLite.
2. It is also not true across SQL backends while the `created_at` timezone issue above stands.

Scoping this to "the same statement on every SQL backend" and dropping the in-memory claim would keep it accurate.

docs/services/selector.md
```diff
@@ -119,6 +119,31 @@ To prevent double-spending *before* the transaction is committed to the ledger,
 2.  **Concurrency Control**: If another concurrent process has already locked that token, the insertion fails, and the selector moves on to the next candidate.
 3.  **Lock Release**: Locks are released either when the transaction reaches finality (success/failure) or when a timeout occurs, ensuring that tokens do not remain permanently inaccessible due to crashed or abandoned transactions.
 
+#### Lease expiry
+
+Every `leaseCleanupTickPeriod`, `sherdlock` runs a cleanup pass over the `TokenLocks`
```
AkramBitar, line 124: *Doc accuracy.* "Every `leaseCleanupTickPeriod`, `sherdlock` runs a cleanup pass" omits the guard at `token/services/selector/sherdlock/manager.go:61`: the cleaner goroutine is never started when `leaseExpiry == 0` or `leaseCleanupTickPeriod == 0`.

That matters more after this PR, because consumer-status release makes `leaseExpiry: 0` an attractive setting — an operator reading it as "no time-based expiry, rely on consumer status" gets **no** cleanup pass at all, so locks of `Deleted`/`Orphan` consumers are never released and those tokens stay permanently unselectable. Worth stating that both values must be non-zero for any cleanup to run.

token/services/storage/db/dbtest/tokenlock.go
```diff
@@ -97,3 +103,174 @@ func TestFully(t *testing.T, tokenDB driver3.TokenStore, tokenLockDB driver3.Tok
...
+	require.NoError(t, tx.Commit())
+}
+
+// requireLockHeld asserts that the lock on tokenID survived cleanup. The store exposes
+// no read API, so the probe is a second Lock on the same token: the (tx_id, idx)
+// primary key rejects it for as long as the row is there.
+func requireLockHeld(t *testing.T, tokenLockDB driver3.TokenLockStore, tokenID token.ID) {
+	t.Helper()
+
+	require.Error(t, tokenLockDB.Lock(t.Context(), &tokenID, "probe-"+tokenID.String(), "owner1"),
```
AkramBitar: *Optional — test rigor.* `requireLockHeld` accepts *any* error from the probe `Lock`, not specifically a primary-key conflict, so all three "Keep" cases (`TestKeepOnDeletedProducer`, `TestKeepSiblingIndices`, `TestKeepFreshPendingLock`) pass vacuously if `Lock` starts failing for an unrelated reason.

Concretely: a plausible future hardening that rejects a lock on a token whose producing transaction is `Deleted` would keep `TestKeepOnDeletedProducer` green while telling you nothing about whether cleanup preserved the row. Asserting on the conflict specifically — or adding a read path to the store — would make the probe mean what the helper's comment says it means.

AkramBitar (commented):
On SQLite, lock cleanup checked the status of the transaction that *created* the token instead of the one *spending* it — so when a spend died, its lock was never released and the token stayed unspendable.

In general the fix is OK. I verified it independently: the generated SQLite `DELETE` now correlates on `consumer_tx_id` per row and is valid SQL; the generated Postgres SQL is genuinely unchanged; and reverting only `sqlite/tokenlock.go` to `main` makes exactly four of the new shared cases fail (`TestReleaseOnDeletedConsumer`, `TestReleaseOnOrphanConsumer`, `TestKeepOnDeletedProducer`, `TestKeepSiblingIndices`), so the regression guard really guards. `go vet` and `go test ./token/services/storage/...` are clean, no references to the deleted `sqlite.IsStale` remain, and only sqlite and postgres embed `common.TokenLockStore`, so the promoted `Cleanup` resolves unambiguously.

Some minor issues, inline below. One does not sit on a line in this diff, so it goes here:

**`token/services/storage/db/driver/token.go:342`** — *should fix here.* The authoritative interface godoc still reads "Cleanup removes the locks such that either: 1. The transaction that locked that token is valid or invalid; 2. The lock is too old." That is the exact ambiguity behind #2018 — "the transaction that locked that token" is what the old SQLite query read as the *producer*, and "valid or invalid" covers every status while the implementation only collects `Deleted`/`Orphan`. This PR corrects `docs/` and adds a good explanatory comment on `IsStaleLock`, but leaves the contract a new driver author would actually implement against unchanged.

token/services/storage/db/dbtest/tokenlock.go
```diff
@@ -97,3 +103,174 @@ func TestFully(t *testing.T, tokenDB driver3.TokenStore, tokenLockDB driver3.Tok
...
+
+	addTokenRequest(t, tokenTransactionDB, "producer")
+	addTokenRequest(t, tokenTransactionDB, "consumer")
+	storeTokens(t, tokenDB, "producer", 0)
+	require.NoError(t, tokenLockDB.Lock(ctx, &tokenID, "consumer", "owner1"))
+
+	// The margin over the lease is deliberate: the SQLite interpreter renders the
+	// threshold with datetime('now', '-N seconds'), which has one-second resolution,
+	// so a sub-second margin would make this flaky.
+	time.Sleep(2200 * time.Millisecond)
```
AkramBitar: *Optional.* This compares two different clocks: `created_at` comes from the test process (`Lock` writes `time.Now().UTC()`), while the threshold is evaluated by the database. Against a Postgres container whose clock lags the test process by more than 1.2s the row still looks fresh and the case fails, and the sleep adds ~2.2s to the shared suite on each backend that runs it.

Inserting the lock row with a backdated `created_at`, or passing a zero/negative lease, would be both deterministic and instant.

AkramBitar (changes requested):

token/services/storage/db/sql/common/tokenlock.go
```diff
@@ -67,12 +73,22 @@ func (db *TokenLockStore) CreateSchema() error {
...
+}
+
+// LockAt is like Lock but records the supplied timestamp as the lock creation
+// time instead of the current time. It is intended for testing (backdating locks
+// to exercise the lease-age expiry path without sleeping).
+func (db *TokenLockStore) LockAt(ctx context.Context, tokenID *token.ID, consumerTxID transaction.ID, _ string, createdAt time.Time) error {
 	query, args := q.InsertInto(db.Table.TokenLocks).
 		Fields("consumer_tx_id", "tx_id", "idx", "created_at").
-		Row(consumerTxID, tokenID.TxId, tokenID.Index, time.Now().UTC()).
+		Row(consumerTxID, tokenID.TxId, tokenID.Index, createdAt).
```
AkramBitar: `createdAt` is stored verbatim here, while `Lock` passes `time.Now().UTC()`. On SQLite the driver serialises the `time.Time` to a string that is compared **lexicographically** against `datetime('now', '-N seconds')`, which is always UTC — so a non-UTC value is off by the whole zone offset.

Reproducible today:

```
TZ=Asia/Tokyo go test ./token/services/storage/db/sql/sqlite/... -run TestTokenLocks
→ FAIL TestTokenLocks/TestReleaseOnAgedLease
```

This is not test-only: `LockAt` is on the public `driver.TokenLockStore` port with no documented UTC requirement, so a caller passing a local `time.Time` silently gets a lease that never expires (east of UTC) or expires immediately (west).

```suggestion
		Row(consumerTxID, tokenID.TxId, tokenID.Index, createdAt.UTC()).
```

token/services/storage/db/sql/common/tokenlock.go
```diff
@@ -95,7 +111,7 @@ func (db *TokenLockStore) GetSchema() string {
 			tx_id TEXT NOT NULL,
 			idx INT NOT NULL,
 			consumer_tx_id TEXT NOT NULL,
-			created_at TIMESTAMP NOT NULL,
+			created_at TIMESTAMPTZ NOT NULL,
```
AkramBitar, line 114: `TIMESTAMP` → `TIMESTAMPTZ` is an unmigrated schema change. `CreateSchema` uses `CREATE TABLE IF NOT EXISTS`, so every existing Postgres deployment keeps `timestamp without time zone` and the exact skew this PR closes persists after upgrade — while `Cleanup`'s godoc and `docs/services/selector.md` now state it is fixed.

Please add the required `ALTER` to `docs/upgradability.md` (it already carries two such warnings):

```sql
ALTER TABLE <TokenLocks>
  ALTER COLUMN created_at TYPE TIMESTAMPTZ
  USING created_at AT TIME ZONE 'UTC';
```

AkramBitar (commented):
Two items left before this can merge; the core fix (correlating on `consumer_tx_id` and scoping the delete per row) looks correct.

AkramBitar (changes requested):

AkramBitar (approved):
LGTM

PR #2245, opened 2026-08-17
fix(tokens): publish token events only after commit

### LINKED ISSUE #2183: tokens: AddToken/DeleteToken events are published inside the open transaction, so subscribers observe tokens that are never committed
### Summary
`AddToken`/`DeleteToken` events are published synchronously *inside the still-open database transaction*, before the caller who owns that transaction has committed or rolled it back. Subscribers can therefore observe a token that is later rolled back and never actually persisted.

### Where
- `token/services/tokens/storage.go:220-226` (`AppendToken`) — `t.Notify(ctx, AddToken, ...)` runs immediately after `t.Tx.StoreToken(...)`, in the same call.
- `token/services/tokens/storage.go:159-163` (`DeleteToken`) — same pattern for `DeleteToken`.
- `token/services/tokens/tokens.go:97-161` (`AppendValid`) — never commits; it calls `AppendToken`/`DeleteTokens` on the transaction and returns. The transaction is committed by the caller (the finality listener), *after* `AppendValid` has already returned and the events have already gone out.

### Impact
The in-repo subscriber is the interactive certifier (`token/services/ttx/certifier/interactive/client.go:122` subscribes, `:248` casts, `OnReceive` enqueues the token ID for asynchronous certification). If the surrounding transaction is rolled back after `AppendValid` returns an error further up the call stack, the certifier has already been notified about a token that will never exist in the store, producing spurious certification attempts and retries. `events.Publisher` is also a public extension point (any application-level subscriber sees the same premature event).

### Reproduction
```go
tx, _ := tokens.NewTransaction(pub, &tokendb.Transaction{TokenStoreTransaction: mockTx}, tmsID)
err := tx.AppendToken(ctx, tta)
require.NoError(t, err)
require.Equal(t, 1, pub.PublishCallCount(), "AddToken should have been published")

// caller now decides to roll back
require.NoError(t, tx.Rollback())

// BUG: publish call count is unchanged — the event already escaped and cannot be retracted
assert.Equal(t, 1, pub.PublishCallCount())
```
```
=== RUN   TestTransaction_AppendToken_NotifiesBeforeRollback
--- PASS: TestTransaction_AppendToken_NotifiesBeforeRollback (0.00s)
PASS
```

### Severity
High — an event bus observing not-yet-committed, possibly-never-committed state is a correctness hazard for every current and future subscriber.

Part of #2112.

HayimShaul:
Fixes #2183

`AddToken` / `DeleteToken` events were published from inside the still-open database transaction that produced them, so a subscriber could act on a token that was later rolled back and never persisted.

The events are now buffered on `DBTransaction` and published only once the change is durable:

- `DBTransaction.Commit(ctx)` publishes them, and only if the underlying commit succeeded — this covers the transactions the service owns.
- `DBTransaction.Rollback` discards them.
- `DBTransaction.FlushEvents(ctx)` publishes them explicitly, for transactions owned by the caller.

`Service.AppendValid` applies a request to a transaction it does not own, so it now returns a `PostCommit` function (never nil, idempotent) that the owner invokes once its own commit succeeded. The finality listener calls it after `tx.Commit()` and before `NotifyStatus`, so a woken finality waiter never sees a confirmed transaction whose token events are still pending.

Tests cover both ownership models: nothing published before commit, nothing published on rollback or on a failed commit, events published in recording order after commit, and the listener ordering commit → publish → notify-status. `docs/services/tokens.md` documents the new contract.

### DIFF
--- a/token/services/tokens/storage_test.go
+++ b/token/services/tokens/storage_test.go
@@ -61,6 +61,8 @@ func TestTransaction_AppendToken(t *testing.T) {
 	}
 	err = tx.AppendToken(ctx, tta)
 	require.NoError(t, err)
+	// commit to flush the (empty) event buffer — without this the assertion is vacuous
+	require.NoError(t, tx.Commit(ctx))
 	assert.Equal(t, 0, pub.PublishCallCount())
 }
 
@@ -79,7 +81,12 @@ func TestTransaction_Notify(t *testing.T) {
 	mockTx.GetTokenReturns(&token2.Token{Type: "TOK"}, []string{"alice"}, nil)
 	err = tx.DeleteTokens(ctx, "me", ids)
 	require.NoError(t, err)
+	// the transaction is still open, nothing may be published yet
+	assert.Equal(t, 0, pub.PublishCallCount())
+
+	require.NoError(t, tx.Commit(ctx))
 	assert.Equal(t, 1, pub.PublishCallCount())
+	assert.Equal(t, tokens.DeleteToken, pub.PublishArgsForCall(0).Topic())
 }
 
 func TestTransaction_AppendToken_Notify(t *testing.T) {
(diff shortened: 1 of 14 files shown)


token/services/tokens/tokens.go
```diff
@@ -91,44 +91,61 @@ func NewService(tmsID token.TMSID, TMSProvider TMSProvider, networkProvider Netw
...
 
 	logger.DebugfContext(ctx, "check transaction exists")
 	exists, err := t.Storage.TransactionExists(ctx, string(txID))
 	if err != nil {
 		logger.ErrorfContext(ctx, "transaction [%s], failed to check existence in db [%s]", txID, err)
 
-		return errors.WithMessagef(err, "transaction [%s], failed to check existence in db", txID)
+		return noPostCommit, errors.WithMessagef(err, "transaction [%s], failed to check existence in db", txID)
 	}
 	if exists {
```
AkramBitar, line 134: *Low — worth writing down rather than changing.* This fix converts token-event delivery from at-least-once to **at-most-once**, and the `TransactionExists` short-circuit makes the loss permanent. The PR description does not mention the trade-off.

Concretely: `finality.Commit` calls `tx.Commit()`; the server commits but the client sees an error (connection reset, driver timeout on the commit ack), or the process dies between `tx.Commit()` returning and `publishTokenEvents(ctx)` at `listener.go:239`. The tokens are durably in the store but no event was published. On the `retryRunner` retry — or on the next pass of the recovery handler (`recovery.go:125`) — `AppendValid` finds `exists == true` and returns `noPostCommit`, so the `store-token`/`delete-token` events for that transaction are never emitted.

For the one in-repo subscriber (`certifier/interactive`) this self-heals only at the next process start via `Scan()` (`driver.go:172`); a third-party subscriber loses the notification silently.

The previous behaviour was worse — events for transactions that never committed — so this is still the right direction. But it belongs in the contract section of `docs/services/tokens.md`, or the recovery path should re-emit events for an already-applied transaction.

AkramBitar (commented):
The fix is real and correct — I verified it rather than taking the description at face value. On the merge-base I wired `tokens.Service` with fakes and called `AppendValid` directly: it published 2 events with `CommitCallCount() == 0`, so events really did escape from inside a still-open transaction. The new code buffers them and flushes after `tx.Commit()`, in the right order (commit → publish → notify-status).

Also checked: `go build ./...` (root + `integration`), `go vet ./...`, `gofmt -l` on the touched packages, and `go test` plus `go test -race` on `./token/services/tokens/...` and `./token/services/ttx/finality/...` — all clean. `go generate ./...` in `token/services/ttx/finality` reproduces `mock/tokens_service.go` byte-identically. `AppendValid` has exactly one non-test caller (`finality.Commit`, shared by the listener and the recovery handler) and `AppendToken`/`DeleteToken` are the only event producers, so no path is left un-flushed.

No defects in the buffering mechanism itself: `Notify`'s `Notifier == nil` guard means `pending` can never be non-empty with a nil `Notifier`, so `FlushEvents`' unguarded `Publish` cannot nil-deref; `Rollback` clears before delegating (the safe direction if the underlying rollback fails); `err` at `listener.go:212` is reassigned rather than shadowed, so the named-return rollback defer in `AppendValid` still fires on every error path; `PostCommit` is a type *alias*, so `*tokens.Service` still satisfies the finality `tokensService` interface; and double-publish is covered by `FlushEvents` draining the buffer.

Two minor findings below, both low severity — neither blocks merge.

token/services/tokens/storage_test.go
AkramBitar, line 1: *Low — test-only, but a silent coverage regression on exactly the code this PR rewrites.*

Three pre-existing **negative** assertions became vacuous, because they assert `pub.PublishCallCount() == 0` on a transaction that is never committed — and events are now buffered until commit, so they would pass no matter what the code under test does:

- `TestTransaction_AppendToken:64` (no owners → no event)
- `TestTransaction_AppendToken_NoNotify:276` (empty owner ID → no event)
- `TestTransaction_DeleteToken_AbsentTokenIsNotAnError:355` (token absent locally → no event)

Verified by mutation: deleting the `if len(id) == 0 { continue }` guard in `AppendToken` (`storage.go:235`) makes the **base** suite fail (`TestTransaction_AppendToken_NoNotify`: got 1, want 0) but leaves this PR's suite fully green.

The PR correctly added `tx.Commit(ctx)` to the two positive tests and to `TestTransaction_Notify_NoPublisher`, just not to these three. `require.NoError(t, tx.Commit(ctx))` before each assertion restores the property each comment claims to protect.

(Anchored at file level because these lines are outside the diff hunks.)

AkramBitar (changes requested):

AkramBitar (approved):
LGTM

PR #2257, opened 2026-08-18
1610_verify_integrity_of_high_value_payloads

### LINKED ISSUE #1610: Selective Cryptographic Integrity Verification of High-Value Payloads [LOW]
**Description**
The store services currently make no integrity guarantees: they store what they receive. Architects should explicitly decide, for each asset class, whether to verify cryptographic integrity before persistence, on retrieval, or not at all -- and document that decision as a security posture choice. At minimum, the highest-value assets warrant verification: endorsement ack signatures (`TransactionEndorseAck.sigma`) should be verified against the endorser's public key before insertion; serialized token request bytes should be checked for valid deserialization and a matching `pp_hash` before storage; identity bytes should be verified against a known public key before being written to `IdentityInfo` or `Signers`. For lower-value or high-volume assets (movements, locks, metadata), the cost-benefit trade-off may favour deferred or sampled verification. The key architectural requirement is that the API surface makes its verification guarantees explicit -- "we store what you give us, unverified" is an acceptable posture only if callers are required to perform verification themselves before calling.

**Summary**
Need to add the defense-in-depth checks in the table below.

| Check | Operation | Cost |
|-------|-----------|------|
| **1. Endorsement Signature Verification** | Verify signature is valid for claimed endorser | **Medium** (1 DB read + crypto verify ~1-5ms) |
| **2. Token Request Integrity** | Verify data deserializes and hash matches | **Low** (deserialize + hash ~0.1-1ms) |
| **3. Identity Validation** | Verify identity contains valid public key | **Low** (key format check ~0.1ms) |
| **4. Signer Registration Validation** | Verify signer's public key matches identity | **Low** (key comparison ~0.1ms) 

Also, 
5. Remove NoOp Bypass: Eliminate option to skip verification

Crypto related data should be checked when inserted to the database and when retrieved. This can be time consuming. We need to find the checks that protect against systemic failures and check only those.

HayimShaul:
Fixes #1610

The store services persisted whatever they were handed. Nothing in their API said whether the
bytes had been cryptographically verified, and nothing re-checked them on the way out — so a
wrong row, a truncated blob or a cross-transaction payload was indistinguishable from authentic
evidence.

This adds `token/services/storage/integrity` (typed sentinel errors, no options, no config) and
wires it into the paths where stored bytes become a claim someone acts on:

- **Token requests** — refused at insert on an empty transaction id, empty request bytes or an
  empty public parameters hash. On retrieval, `ttxdb`/`auditdb` deserialise the payload and
  require its **anchor to equal the transaction id it is filed under**. `endorserdb` holds the
  bare actions format, which has no anchor field, so it gets a version and non-empty-actions
  check instead.
- **Public parameters** — `PublicParamsByHash` now recomputes the hash of what it is about to
  return and refuses a mismatch. Parameters are addressed *by* that hash and carry the issuer
  and auditor keys plus the setup every action is validated against, which makes them the
  highest-value payload in the store.
- **Identities** — empty identities are refused, and identity lookups compare the stored
  identity against the requested one. This is substantive rather than cosmetic: an empty
  identity's unique id is the constant `"<empty>"`, so every empty identity collides on one row
  and one cache key. Applied to both the SQL and the KVS backend.
- **Endorsement acknowledgements** — refused at insert on an empty endorser or empty signature.
- **Signer registration** — refuses an empty identity or a nil verifier.

The checks are unconditional, and `nobypass_test.go` makes that enforceable rather than
conventional: it parses the AST of the affected packages and fails if anyone adds a
bypass-shaped identifier, gives a check an options parameter or a second result, exposes mutable
state from the check package, introduces a verification config key, or discards a check's
result.

Cross-backend expectations live in the shared `dbtest` spec so SQL and KVS are held to one
contract. That spec caught a real gap: SQL `GetTokenInfo` had no empty-identity guard and never
compared the stored identity against the requested one — fixed here. Two fuzz targets cover the
request parsers and are wired into the nightly fuzz matrix.

### What this deliberately does not do

Written up per asset class in the new
[`docs/security/store_integrity_verification.md`](docs/security/store_integrity_verification.md),
including:

- **Endorsement ack signatures are not re-verified on retrieval.** The row holds only
  `(endorser, sigma)`; the per-party filtered payload that was actually signed is not persisted,
  so there is nothing to verify against without a schema change.
- **A supplied verifier is not compared against the identity it is registered for.**
  `driver.Verifier` is only `Verify(message, sigma) error` and exposes no canonical key, so the
  comparison is not expressible against the current interface.
- **`pp_hash` is not a digest of the request bytes**, so it is not what makes a retrieved request
  checkable — the anchor is.
- Movements, locks and metadata are not verified, and the doc says so rather than leaving it
  implied.

### Behavioural note

Requests are now refused at insert where they would previously have been stored, and retrieval
fails closed on a payload that does not bind to its transaction id. The audit in the issue thread
found no in-tree caller affected — every current producer already verifies before storing — so
this is defence-in-depth and an explicit contract, not a fix for an exploitable hole in a current
flow. Out-of-tree callers that store unverified data would see an explicit error naming the failed
check.

### Verification

- `make checks` — passes.
- `make unit-tests` — passes, except `token/services/identity/config` `TestTranslatePath`, which
  asserts the absolute checkout path contains the string `"panurus"` and so fails for any clone
  not named that. Pre-existing and unrelated.
- `make lint-auto-fix` — clean for every file this branch touches. It still reports **17 revive
  `unhandled-error` findings in files this branch does not touch** (`token/driver/wallet.go`,
  `token/services/benchmark/runner.go`, `token/services/network/common/rws/…`,
  `token/services/storage/db/sql/query/common/builder.go`, and others). Those reproduce on a clean
  worktree at `origin/main`, so `make lint` does not currently pass on `main` either. Left alone
  here to keep this change reviewable; happy to fix them in a separate PR.
  
  ### How each item in #1610 is resolved

The issue's table asked for cryptographic verification at the store for several
asset classes. After the audit in the issue thread, the agreed resolution is that
the store enforces **structural** integrity and the **cryptographic** verification
stays at the layer that already has the state to do it — disclosed per asset class
in `docs/security/store_integrity_verification.md`. Mapping the table explicitly so
the decision is on the record rather than inferred from the diff:

| #1610 item | Resolution in this PR |
|---|---|
| **1. Endorsement Signature Verification** | **Not performed at the store — by agreement.** `CheckEndorsementAck` verifies only that `endorser` and `sigma` are non-empty. The per-party filtered payload that was signed is not persisted alongside `(endorser, sigma)`, so the signature is not re-verifiable at the store without a schema change. Cryptographic verification remains the caller's obligation at `CollectEndorsementsView.distributeTxToParty`, and is stated as Requirement 2 in the doc. |
| **2. Token Request Integrity** | **Deserialization: done.** ttxdb/auditdb re-parse on retrieval; endorserdb checks version + non-empty actions. **Hash-matching: not applicable to the request bytes — by agreement.** `pp_hash` is the hash of the public parameters, not a digest of the request, so it cannot bind the request bytes; retrieval binds bytes to their row via the request's **anchor** instead. Hash-matching *is* implemented where a payload is genuinely hash-addressed: `PublicParamsByHash` recomputes and compares. |
| **3. Identity Validation** | Done. Empty identities refused on every path; lookups compare the stored identity against the requested one, on both SQL and KVS backends. |
| **4. Signer Registration Validation** | Refuses an empty identity or a verifier that cannot be derived from the identity. A full key-vs-identity comparison is not expressible against `driver.Verifier` (`Verify(message, sigma)` only, no canonical key) — see the doc. |
| **5. Remove NoOp Bypass** | Done and made enforceable: checks are unconditional, guarded by `nobypass_test.go`. |


docs/security/store_integrity_verification.md
```diff
@@ -0,0 +1,154 @@
...
+purpose: `grep -rn "Verification:"` enumerates the contract, and `grep -rn integrity.Check`
+enumerates the enforcement.
+
+## Scope
+
+This is a **structural** integrity boundary, not a second validator. The checks need only the bytes
+themselves plus the key those bytes are stored under; none of them performs I/O, consults a ledger,
+or evaluates a zero-knowledge proof. A token request is fully verified only by a `token.Validator`
+against a ledger, and that happens where it already happened — see
+[Endorsement responder security](../services/ttx_responder_security.md).
```
AkramBitar: This relative link does not resolve: from `docs/security/`, `../services/ttx_responder_security.md` points at `docs/services/ttx_responder_security.md`, but the file is `docs/services/ttx/ttx_responder_security.md`. Same link again at line 148. Both 404, and this is the pointer to where token requests actually get validated — the half of the contract this page explicitly defers to. Should be `../services/ttx/ttx_responder_security.md`.

token/services/storage/ttxdb/store.go
```diff
@@ -284,25 +289,99 @@ func (d *StoreService) GetStatuses(ctx context.Context, txIDs []string) (map[str
...
+// under. A single failing record fails the whole call: the caller asked for a
+// set of requests and cannot be expected to notice that one key silently went
+// missing.
 func (d *StoreService) GetTokenRequests(ctx context.Context, txIDs []string) (map[string][]byte, error) {
-	return d.db.GetTokenRequests(ctx, txIDs)
+	requests, err := d.db.GetTokenRequests(ctx, txIDs)
+	if err != nil {
+		return nil, err
+	}
+	for txID, raw := range requests {
```
AkramBitar, line 342: This loop has no `raw == nil` skip, but the singular `GetTokenRequest` just above (line 306) does: there a nil value is treated as not-found and returned as `nil, nil`. Here the same row goes straight into `integrity.CheckStoredTokenRequest` and fails the whole call with `ErrEmptyTokenRequest`.

Since the godoc defers to `driver.TransactionStore.GetTokenRequests` for missing-key semantics, the two entry points now disagree: a request column that reads back nil/empty (driver-dependent for a zero-length BYTEA) is "absent" via the singular call and "corrupt, fail the batch" via this one. Anyone converting a `GetTokenRequest` loop to the batch API silently changes failure behaviour. Suggest skipping nil entries here as well, or stating the divergence in the godoc if failing is intended.

token/services/storage/db/dbtest/tokens.go
```diff
@@ -822,6 +822,20 @@ func TPublicParams(t *testing.T, db TestTokenDB) {
...
+	// query happened to reach. Public parameters carry the issuer and auditor
+	// keys every action is validated against, and a caller asking by hash is
+	// asking for the setup a specific transaction was created under.
+	res, err = db.PublicParamsByHash(ctx, utils.Hashable([]byte("never stored")).Raw())
+	require.NoError(t, err)
+	assert.Nil(t, res)
+
+	// An empty hash matches nothing and would make the hash comparison vacuous.
+	res, err = db.PublicParamsByHash(ctx, nil)
+	if err == nil {
```
AkramBitar: This assertion never runs. `PublicParamsByHash` rejects an empty hash with `integrity.ErrEmptyPublicParamsHash`, so `err == nil` is false and the block is skipped — the shared spec ends up asserting nothing about the behaviour the comment above describes.

Since this is a conformance spec every driver runs, it should pin the contract down rather than tolerate either outcome:

```go
_, err = db.PublicParamsByHash(ctx, nil)
require.ErrorIs(t, err, integrity.ErrEmptyPublicParamsHash)
```

AkramBitar (changes requested):

AkramBitar (commented):
Re-reviewed at `f5249bf27`. My three earlier comments are addressed — both relative links in `docs/security/store_integrity_verification.md` resolve, the nil skip is in `ttxdb` **and** `auditdb` with the divergence documented on the driver interfaces, and the vacuous `PublicParamsByHash(nil)` assertion in the shared `dbtest` spec is now `require.ErrorIs(..., ErrEmptyPublicParamsHash)` plus an unknown-hash case. Thanks. The whole `./token/...` suite passes at this head (only the pre-existing `TestTranslatePath` fails).

Three things I'd like resolved before merge — the first is a gap in the nil-skip fix itself, the second and third are error-handling blast radius:

1. `raw == nil` vs `len(raw) == 0` in the four token-request read paths.
2. One bad endorsement-ack row makes a whole transaction unreadable.
3. `StorePublicParams` discards the new hash-mismatch error.

The remaining comments are minor. Note also that the branch is behind `main` and will need a rebase.

token/services/storage/ttxdb/store.go
```diff
@@ -284,25 +289,109 @@ func (d *StoreService) GetStatuses(ctx context.Context, txIDs []string) (map[str
...
+// returned. Callers treat the result as authentic evidence about txID — they
+// hash it against the ledger, re-broadcast it, or show it to an auditor — so a
+// record that cannot be bound to txID must not reach them.
 func (d *StoreService) GetTokenRequest(ctx context.Context, txID string) ([]byte, error) {
-	return d.db.GetTokenRequest(ctx, txID)
+	raw, err := d.db.GetTokenRequest(ctx, txID)
+	if err != nil {
+		return nil, err
+	}
+	if raw == nil {
```
AkramBitar: The not-found guard is `raw == nil`, but the write side treats `len(raw) == 0` as empty — `integrity_test.go` asserts `[]byte{}` -> `ErrEmptyTokenRequest`. A legacy row whose `request` column is present but zero-length (writable before this PR) scans through `database/sql` as a **non-nil empty slice** on most drivers, so `raw == nil` is false and `CheckStoredTokenRequest` returns `ErrEmptyTokenRequest`.

Effect: `GetTokenRequest` reports an integrity failure where it should report not-found, and `GetTokenRequests` (line 339) fails the entire call for every *other* tx id in the batch. The godoc justifies the skip with "some drivers scan a zero-length BYTEA as nil" — the complementary representation is exactly the one left failing.

`len(raw) == 0` here and at line 339, plus `auditdb/store.go:346` and `:377`, makes it driver-independent.

token/services/storage/auditdb/store.go
```diff
@@ -324,15 +330,63 @@ func (d *StoreService) GetStatuses(ctx context.Context, txIDs []string) (map[str
...
+// anchored to a transaction other than txID is reported as an error rather than
+// returned. This store is read by auditors, for whom a request attributed to
+// the wrong transaction is worse than no request at all.
 func (d *StoreService) GetTokenRequest(ctx context.Context, txID string) ([]byte, error) {
-	return d.db.GetTokenRequest(ctx, txID)
+	raw, err := d.db.GetTokenRequest(ctx, txID)
+	if err != nil {
+		return nil, err
+	}
+	if raw == nil {
```
AkramBitar: Same as `ttxdb/store.go:306` — `raw == nil` misses the non-nil zero-length slice that most drivers produce for an empty `BYTEA`, so a legacy empty row errors instead of reading as not-found. Applies here and at line 377; `len(raw) == 0` covers both representations.

token/services/storage/ttxdb/store.go
```diff
@@ -284,25 +289,109 @@ func (d *StoreService) GetStatuses(ctx context.Context, txIDs []string) (map[str
...
+// carries no signature is reported instead of being passed off as an
+// acknowledgement.
 func (d *StoreService) GetTransactionEndorsementAcks(ctx context.Context, txID string) (map[string][]byte, error) {
-	return d.db.GetTransactionEndorsementAcks(ctx, txID)
+	acks, err := d.db.GetTransactionEndorsementAcks(ctx, txID)
+	if err != nil {
+		return nil, err
+	}
+	for endorser, sigma := range acks {
+		if len(sigma) == 0 {
```
AkramBitar: Two things here.

**Blast radius.** A single empty-signature row fails the whole `GetTransactionEndorsementAcks` call, and `ttx.TransactionInfo` (`token/services/ttx/info.go:42`) propagates that — so the transaction's *valid* acks and its token request both become permanently unreadable, and there is no API to delete the offending row. That is the mirror image of the decision made 50 lines above for token requests (skip + `delete`), and it contradicts the driver godoc added in this PR (`db/driver/ttx.go`: "the returned signatures are returned as stored"). Skipping with an ERROR log would keep the damage to the bad row; if erroring is deliberate, the godoc should say so and name the repair path.

**Asymmetry with the insert.** The godoc claims each row is checked for "the same non-vacuity `AddTransactionEndorsementAck` enforces", but `AddTransactionEndorsementAck` calls `integrity.CheckEndorsementAck(id, sigma)` (empty *endorser* too) while this loop only checks `len(sigma) == 0`. A row with an empty endorser comes back keyed `"<empty>"`.

token/services/storage/db/sql/common/tokens.go
```diff
@@ -1161,14 +1162,32 @@ func (db *TokenStore) PublicParams(ctx context.Context) ([]byte, error) {
 	return common.QueryUniqueContext[[]byte](ctx, db.readDB, query, args...)
```
AkramBitar, line 1176: (Anchored on the new godoc; the issue is in `StorePublicParams` at line 1136.)

```go
if pps, err := db.PublicParamsByHash(ctx, rawHash); err == nil && len(pps) > 0 {
```

`PublicParamsByHash` now returns `ErrPublicParamsHashMismatch`, and `err == nil &&` throws it away. Execution falls through to the INSERT against `raw_hash BYTEA PRIMARY KEY`, so a corrupt row surfaces to the operator as an opaque unique-key violation — the one call site that can detect tampering reports it as a duplicate, and the row can never be repaired through this API. `nobypass_test.go` cannot catch it because the discard is indirect.

Suggest distinguishing the cases: propagate `ErrPublicParamsHashMismatch` (wrapped, so the operator sees what it is), treat sql.ErrNoRows / empty as "not stored", and only then INSERT.

token/services/storage/ttxdb/gettoken_requests_test.go
```diff
@@ -0,0 +1,130 @@
...
+	got, err := svc.GetTokenRequests(context.Background(), []string{"tx1"})
+	require.NoError(t, err)
+	_, hasTx1 := got["tx1"]
+	assert.False(t, hasTx1, "nil entry must not appear in returned map")
+}
+
+// TestGetTokenRequests_PresentNilSignedViaGetTokenRequest confirms the
+// behavioural equivalence: GetTokenRequest(ctx, "tx1") on a nil-returning
+// driver also produces nil, nil — no difference to the caller.
+func TestGetTokenRequests_PresentNilSignedViaGetTokenRequest(t *testing.T) {
```
AkramBitar: This test asserts nothing — it builds the stub, calls `newStoreService`, then `_ = svc` with "actual test of the singular path is in store.go". It reports as passing coverage of the singular nil path that nothing exercises, which is worse than no test.

It is also the only reason `var _ time.Time` (line 84) and `var _ driver.PPHash` (line 130) exist; both go away with it. Either give it a real `GetTokenRequest` assertion through the stub, or drop it.

token/services/storage/db/dbtest/identity.go
```diff
@@ -341,6 +342,45 @@ func TRegisterIdentityDescriptor(t *testing.T, db driver.IdentityStore) {
...
+	require.Error(t, db.RegisterIdentityDescriptor(ctx, &idriver.IdentityDescriptor{
+		Identity:   empty,
+		AuditInfo:  auditInfo,
+		Signer:     &mock.Signer{},
+		SignerInfo: []byte("signer_info"),
+		Verifier:   &mock.Verifier{},
+	}, nil))
+
+	// nothing was written under the shared key
+	exists, err := db.SignerInfoExists(ctx, empty)
```
AkramBitar, line 459:
```go
exists, err := db.SignerInfoExists(ctx, empty)
if err == nil { assert.False(...) }
```

Same conditional-assertion shape I flagged on `PublicParamsByHash`: in a shared cross-backend conformance spec this tolerates either outcome instead of pinning the contract. Here the guard is dead code — neither backend errors on this call — so it can be an unconditional `require.NoError` + `assert.False`.

token/services/storage/db/sql/common/identity.go
```diff
@@ -305,55 +306,119 @@ func (db *IdentityStore) Notifier() (idriver.IdentityConfigurationNotifier, erro
...
+// Verification: an empty id is refused, for the reason given on
+// StoreIdentityData — the empty identity does not hash, it maps to the shared
+// "<empty>" row key.
 func (db *IdentityStore) StoreSignerInfo(ctx context.Context, id tdriver.Identity, info []byte) error {
+	if err := integrity.CheckIdentity(id); err != nil {
+		return errors.WithMessage(err, "refusing to store signer info")
+	}
 	_, err := db.storeSignerInfo(ctx, db.writeDB, id.UniqueID(), id, info, true)
 
 	return err
```
AkramBitar, line 435: `GetExistingSignerInfo` (line 427 below) and `SignerInfoExists` are the only identity paths left without an empty-identity guard — they still hash the empty identity into the shared `"<empty>"` composite key. `StoreSignerInfo` immediately above refuses it, so the read side is now more permissive than the write side; `identity.Provider.AreMe(ctx, emptyIdentity)` would report the empty identity as mine if such a row ever exists (e.g. written before this PR).

That contradicts Requirement 3 in `docs/security/store_integrity_verification.md` ("every write and read path must refuse an empty identity outright"). The KVS backend has the same gap in its `GetExistingSignerInfo` (`db/kvs/identitydb.go:356`).

These are also the only two methods in `identity/driver/storage.go` with no `Verification:` clause, so the `grep -rn "Verification:"` enumeration the doc advertises under-reports the contract.

token/sig.go
```diff
@@ -55,15 +56,70 @@ func (s *SignatureService) GetSigner(ctx context.Context, id Identity) (Signer,
...
+// canonical public key to compare, so establishing agreement would require a new
+// accessor on every identity type. The in-tree callers that pass a verifier are
+// the x509 and idemix key managers, which derive it from the identity they are
+// registering, so the comparison would be a tautology there; the ttx callers
+// pass nil. See docs/security/store_integrity_verification.md.
+func (s *SignatureService) checkSignerIdentity(ctx context.Context, identity Identity) error {
+	if err := integrity.CheckIdentity(identity); err != nil {
+		return err
+	}
+	if _, err := s.deserializer.GetOwnerVerifier(ctx, identity); err == nil {
```
AkramBitar: The owner -> issuer -> auditor fallback resolves to the same object three times: both in-tree drivers build `common.NewDeserializer(des, des, des, des, des)` from a single `TypedVerifierDeserializerMultiplex`, so on a genuine rejection the identity is deserialized three times before the error surfaces, and only the auditor error is kept. The "any of the three roles is accepted, since this service is role-agnostic" rationale in the godoc (and in `docs/services/identity.md`) is therefore vacuous for every driver in the tree.

It also puts a full verifier construction — idemix credential parsing — on `RegisterSigner`/`RegisterEphemeralSigner`, which run per recipient per transaction (`ttx/recipients.go`, `interop/htlc/transaction.go`). A single `GetOwnerVerifier` call plus a comment noting the multiplex would be equivalent and cheaper.

AkramBitar (dismissed):

token/services/storage/ttxdb/store.go
HayimShaul: Good catch — the two representations were split. `GetTokenRequest` and `GetTokenRequests`
in `ttxdb/store.go` now skip on `len(raw) == 0`, so a present-but-zero-length `request`
column reads as not-found whichever way the driver scans it, and one such row no longer
fails the whole batch. The godoc on both methods and on
`db/driver/ttx.go:TokenTransactionStore` was reworded to describe the empty value rather
than only the nil one.

Fixed in 0bba89ff5c29.

token/services/storage/auditdb/store.go
HayimShaul: Same change applied in `auditdb/store.go` at both sites (`GetTokenRequest` and
`GetTokenRequests` now skip on `len(raw) == 0`), with the godoc and the
`db/driver/audit.go` interface contract updated to match.

Fixed in 0bba89ff5c29.

token/services/storage/ttxdb/store.go
HayimShaul: Both points addressed. A row that fails the check is now logged at ERROR and deleted
from the returned map instead of failing the call, so the transaction's valid acks and
its token request stay readable — the same posture as the token-request skip above. The
empty-endorser case is covered too: the map is keyed by unique id, so the row is
recognised by comparing its key against `token.Identity(nil).UniqueID()`. The godoc on
`ttxdb.StoreService.GetTransactionEndorsementAcks`, on
`db/driver/ttx.go:GetTransactionEndorsementAcks` and in
`docs/security/store_integrity_verification.md` now say so, and
`ttxdb/endorsement_acks_test.go` pins it.

Fixed in 0bba89ff5c29.

token/services/storage/db/sql/common/tokens.go
HayimShaul: `StorePublicParams` no longer swallows the error: it propagates whatever
`PublicParamsByHash` returns (wrapped with the hash), treats an empty result as "not
stored", and only then INSERTs. A corrupt row now surfaces as
`ErrPublicParamsHashMismatch` instead of as a primary-key violation from the
fall-through.

Fixed in 0bba89ff5c29.

token/services/storage/ttxdb/gettoken_requests_test.go
HayimShaul: Replaced with a test that actually drives the singular path: a stub whose
`GetTokenRequest` returns the configured value, asserting `nil, nil` for both the nil and
the zero-length representation. The `var _ time.Time` and `var _ driver.PPHash`
placeholders and their imports are gone, and the two batch tests now cover the
zero-length entry as well.

Fixed in 0bba89ff5c29.

token/services/storage/db/dbtest/identity.go
HayimShaul: The guard is gone: the spec now asserts `require.NoError` + `assert.False` on
`SignerInfoExists(ctx, empty)` unconditionally, and additionally pins
`GetExistingSignerInfo(ctx, empty)` to an empty result. Both hold on both backends after
the change made for the `GetExistingSignerInfo` thread.

Fixed in 0bba89ff5c29.

token/services/storage/db/sql/common/identity.go
HayimShaul: Closed on both backends. `GetExistingSignerInfo` (SQL and KVS) now excludes an empty
identity from the lookup, so the shared `"<empty>"` key is never consulted and
`SignerInfoExists` — and through it `Provider.AreMe` — reports the empty identity as not
mine even if such a row exists. It excludes rather than errors because this is an
existence query over a batch: `areMe` passes many identities at once and drops the whole
storage answer on error, so failing the call would lose the verdict for the caller's
other identities over one bad entry. Both methods now carry a `Verification:` clause in
`identity/driver/storage.go`, so the `grep -rn "Verification:"` enumeration is complete,
and Requirement 3 in `docs/security/store_integrity_verification.md` states the
exception explicitly.

Fixed in 0bba89ff5c29.

token/sig.go
HayimShaul: `checkSignerIdentity` now makes a single `GetOwnerVerifier` call and returns its error.
The godoc there, `docs/services/identity.md` and
`docs/security/store_integrity_verification.md` now record why one call suffices — every
in-tree driver builds `common.NewDeserializer` from one
`TypedVerifierDeserializerMultiplex`, which routes by identity type rather than by role —
instead of the vacuous role-agnostic rationale.

Fixed in 0bba89ff5c29.

Effi-S (approved):

token/services/storage/db/sql/common/transactions.go
```diff
@@ -688,11 +705,22 @@ func (w *TransactionStoreTransaction) AddTransaction(ctx context.Context, rs ...
...
+// know which of the two token request formats they hold. What is refused here
+// is a row that no later check could act on: an empty transaction id, empty
+// request bytes, or an empty public parameters hash. See
+// docs/security/store_integrity_verification.md.
 func (w *TransactionStoreTransaction) AddTokenRequest(ctx context.Context, txID string, tr []byte, applicationMetadata, publicMetadata map[string][]byte, ppHash driver2.PPHash) error {
 	logger.DebugfContext(ctx, "adding token request [%s]", txID)
 	if w.txn == nil {
 		return errors.New("no db transaction in progress")
 	}
+	if err := integrity.CheckTokenRequestForStorage(txID, tr, ppHash); err != nil {
```
adecaro, line 721: **Blocking: this new check breaks an existing test in a sibling package.**

`CheckTokenRequestForStorage` now refuses an empty `ppHash` here, which is correct in isolation — but `token/services/storage/db/guard/decorator_test.go:65` (`TestGuardDisabledWritePassesThrough`) calls `AddTokenRequest(ctx, "tx", make([]byte, 100), nil, nil, nil)` with `ppHash` as `nil`. The test now fails with `empty public parameters hash` before the write ever reaches the mocked DB, so it no longer proves what its name claims ("a zero payload limit disables the check and the write proceeds to the database"). This is one of the two failures in CI's `utest (unit-tests-regression)` job.

**Suggestion**: pass a non-empty `ppHash` in that test — the guard's size-limit behavior under test is orthogonal to this check.

token/services/storage/endorserdb/store.go
```diff
@@ -118,10 +119,32 @@ func (d *StoreService) ValidationRecords(ctx context.Context, params QueryValida
...
+//
+// Note that this format carries no anchor, so the record cannot be bound to
+// txID by a structural check — see docs/security/store_integrity_verification.md.
 func (d *StoreService) AppendValidationRecord(ctx context.Context, txID string, tokenRequest []byte, meta map[string][]byte, ppHash driver2.PPHash) error {
 	logger.DebugfContext(ctx, "appending new validation record... [%s]", txID)
 
+	if err := integrity.CheckTokenRequestForStorage(txID, tokenRequest, ppHash); err != nil {
+		return errors.WithMessagef(err, "refusing to append validation record for txid [%s]", txID)
+	}
+	if err := integrity.CheckTokenRequestActions(tokenRequest); err != nil {
```
adecaro, line 144: **Blocking: breaks the #2316 regression test in this same package.**

`token/services/storage/endorserdb/store_test.go:141` (`TestSetStatus_NotifyNotDroppedByCallerCanceledContext`) calls `store.AppendValidationRecord(t.Context(), txID, []byte("tr"), nil, driver2.PPHash("pp"))`. `[]byte("tr")` is not a deserializable token request, so `CheckTokenRequestActions` now fails it with `malformed token request` before the test ever reaches the context-cancellation behavior it's actually regression-testing. This is the second of the two failures in CI's `utest (unit-tests-regression)` job.

**Suggestion**: build a real (even minimal) `driver.TokenRequest` and use its serialized bytes as the fixture — the test's subject (notification not dropped when the caller's context is already cancelled) doesn't depend on the token request's content.

token/services/storage/integrity/integrity.go
```diff
@@ -0,0 +1,241 @@
...
+// values, so a row with an empty signature is indistinguishable from a genuine
+// one and turns "this party never signed" into "this party signed". An empty
+// endorser identity collapses to a single map key and would mask, or be masked
+// by, an unrelated party's acknowledgement.
+//
+// This does not verify the signature. Verifying it requires the payload that
+// was signed, which is not persisted alongside the acknowledgement; the
+// verification therefore happens where that payload is still in scope, before
+// the acknowledgement reaches storage.
+func CheckEndorsementAck(endorser []byte, sigma []byte) error {
```
adecaro, line 192: **Design note: this does not implement issue #1610 item 1 as literally specified.**

#1610 asks for "Endorsement Signature Verification: Verify signature is valid for claimed endorser" at the store. `CheckEndorsementAck` only checks that `endorser`/`sigma` are non-empty; actual cryptographic verification is documented (here and in `docs/security/store_integrity_verification.md`'s Requirement 2) as the caller's responsibility, already performed at `CollectEndorsementsView.distributeTxToParty`. This is disclosed clearly, so I'm not blocking on it, but since the PR author is also the issue's reporter, it'd help future readers if the PR description said explicitly that this is the agreed resolution for 1610.1 (and the hash-matching half of 1610.2) rather than leaving that gap to be noticed by diffing the issue table against the code.

adecaro (commented):
## Review: does this fully address #1610?

@HayimShaul

**The design and docs here are strong, but CI's `utest (unit-tests-regression)` is currently red, and both failures are caused directly by this PR's own new checks — please fix before merge.**

Checked out `1610_verify_integrity_of_high_value_payloads` at `f97a8d147` and verified:

```
go build ./...                                                       # clean
go test -count=1 ./token/services/storage/integrity/...              # pass, 100% coverage
go test -count=1 ./token/services/storage/db/guard/...                # FAIL: TestGuardDisabledWritePassesThrough
go test -count=1 ./token/services/storage/endorserdb/...              # FAIL: TestSetStatus_NotifyNotDroppedByCallerCanceledContext
go test -count=1 ./token/services/storage/ttxdb/... ./token/services/storage/auditdb/... ./token/services/identity/... ./token/services/ttx/...   # pass
go test ./token/services/storage/integrity/... -fuzz=FuzzCheckStoredTokenRequest -fuzztime=15s     # clean, no panics
go test ./token/services/storage/integrity/... -fuzz=FuzzCheckTokenRequestActions -fuzztime=15s    # clean, no panics
git log --oneline ede8c547..origin/main                               # 0 commits behind — no base drift
```

I also re-read all 8 of @AkramBitar's prior threads against the current code rather than trusting the "Fixed in 0bba89ff5c29" comments at face value — they hold. I'm not repeating any of them here.

I grepped every caller of `CheckTokenRequestForStorage`/`CheckTokenRequestActions`/`CheckEndorsementAck` outside this PR's own files, which is how the two test failures below turned up — they're not in the PR's diff but they break because of it.

`checks`, `itest`, `bench-check`, `cgo-check`, and the Token Validation Benchmark are all green; only `utest` is red, at the `unit-tests-regression` leg.

### What it fixes

| # | Finding | Verified |
|---|---|---|
| 1610.2 | Token request integrity: deserializes, valid version, anchor matches txID | ✅ `CheckStoredTokenRequest`/`CheckTokenRequestActions`, 100% covered, fuzzed with no panics |
| 1610.3/4 | Identity/signer validation | ✅ non-empty + matched against requested identity, verifier derivable |
| 1610.5 | Remove NoOp bypass | ✅ `nobypass_test.go` statically enforces this; ran it, passes |
| 1610.1 | Endorsement signature verified against endorser's public key | ⚠️ not implemented as a store-side crypto check — see non-blocking note below |

Rows 2–5 are solid. Row 1 is a real, disclosed design choice (structural check + documented caller responsibility) rather than a bug — flagged below for visibility, not as a blocker.

### Blocking 1 — two existing tests fail because the new checks reject their fixtures

- `token/services/storage/db/guard/decorator_test.go:65` (`TestGuardDisabledWritePassesThrough`) passes `nil` as `ppHash` to `AddTokenRequest`. `CheckTokenRequestForStorage` (`token/services/storage/db/sql/common/transactions.go:721`) now returns `empty public parameters hash` before the mocked write happens, so the test no longer proves what its name claims.
- `token/services/storage/endorserdb/store_test.go:141` (`TestSetStatus_NotifyNotDroppedByCallerCanceledContext`, the #2316 regression test) passes the literal bytes `[]byte("tr")` as a token request. `CheckTokenRequestActions` (`token/services/storage/endorserdb/store.go:144`) now fails deserialization with `malformed token request` before the test reaches the context-cancellation behavior it's actually regression-testing.

Both are reproducible locally at this SHA (not flaky) and are exactly the two failures in [`utest (unit-tests-regression)`](https://github.com/LFDT-Panurus/panurus/actions/runs/34332195938/job/102403756436). Fix: give the guard test a non-empty `ppHash`, and give the endorserdb test a real deserializable token request — neither test's actual subject (size guard; notification-under-cancellation) depends on those specific fixture values.

### Non-blocking

- **Design note:** Issue #1610 item 1 asks for endorsement-signature verification "against the claimed endorser" at the store. What's shipped checks only that the endorser/signature fields are non-empty, with actual crypto verification documented as already happening at `CollectEndorsementsView.distributeTxToParty` before the ack reaches storage. This is disclosed clearly in both the package doc and `docs/security/store_integrity_verification.md`'s Requirements section, so I'm not blocking on it — but since you're both the issue's author and this PR's author, it'd help future readers if the PR description said explicitly that this is the agreed scope for 1610.1 (and the hash-matching half of 1610.2), rather than leaving that gap to be noticed by diffing the issue table against the code.
- **Nit:** `docs/security/store_integrity_verification.md`'s Requirements section (§3, lines 59-64) and its "what is deliberately not checked" section (lines 114-118) each document a different deliberate empty-identity exception; a cross-reference between them would help a reader who only skims Requirements.

### Recommendation

Fix the two test fixtures so `utest` goes green — that's the only thing blocking this. Everything else, including the very thorough resolution of the previous review round, looks good to me.

token/services/storage/db/sql/common/transactions.go
Effi-S: Fixed

token/services/storage/endorserdb/store.go
Effi-S: Fixed

token/services/storage/integrity/integrity.go
Effi-S: Updated PR description

adecaro (approved):
LGTM

AkramBitar (approved):

PR #2262, opened 2026-08-18
Fix `time.After` in `inmemory/locker` Scan Sleep Loop [mem]

### LINKED ISSUE #2127: Fix `time.After` in `inmemory/locker` Scan Sleep Loop [mem]
**Document severity:** MEDIUM · **Verified priority:** Low · **Effort:** Low · **Type:** Task

### Old issue reference

> **Sub-Task 6 — Fix `time.After` in `inmemory/locker` Scan Sleep Loop**
>
> **Status:** `[ ] pending`

| 6 | `[RUN]` | **MEDIUM** | Timer leak | `selector/simple/inmemory/locker.go` | 268 | `time.After` in scan sleep loop; should be `Ticker` |

### Intent

The scan goroutine in the in-memory locker uses `time.After(d.sleepTimeout)`
inside a `for { select { } }` loop. A new timer is allocated on every
iteration. Although the `ctx.Done()` branch provides an exit path, each
iteration before exit leaks the timer for up to `sleepTimeout` duration.

### Expected Outcomes

- A single `time.Ticker` drives the scan sleep instead of per-iteration timers.
- `ticker.Stop()` is called via `defer` when the goroutine exits.

### Todo List

1. Create `ticker := time.NewTicker(d.sleepTimeout)` before the outer loop.
2. Add `defer ticker.Stop()` immediately after.
3. Replace `case <-time.After(d.sleepTimeout):` with `case <-ticker.C:`.

### Relevant Context

- File: `token/services/selector/simple/inmemory/locker.go`
- Lines 264–284: inner sleep loop in the scan goroutine.

---

### Status in current code

**Not a leak on Go 1.26**, but the `Ticker` suggestion has independent merit. Line numbers have shifted: the loop is now at `token/services/selector/simple/inmemory/locker.go:430`, not 264–284.

The call is now at `token/services/selector/simple/inmemory/locker.go:430`
(`case <-time.After(d.sleepTimeout):`), inside the scan goroutine's sleep loop.

On the leak claim:

`go.mod` declares `go 1.26.5`. Per the [Go 1.23 release notes](https://go.dev/doc/go1.23#timer-changes): *"Timers and Tickers that are no longer referred to by the program become eligible for garbage collection immediately, even if their `Stop` methods have not been called. Earlier versions of Go did not collect unstopped Timers until after they had fired and never collected unstopped Tickers."* The timer returned by `time.After` becomes unreachable the moment the `select` completes, so it is collectable immediately rather than surviving until it fires. The document's premise — that the timer is retained for its full duration — describes pre-1.23 runtime behaviour and does not hold on this toolchain.

So "each iteration before exit leaks the timer for up to `sleepTimeout` duration" is not true
here — the timer is collectable as soon as its iteration's `select` returns.

This is the one item in the timer group where the *suggested* change is still worth making on
its own merits. A single `time.NewTicker` created before the loop with `defer ticker.Stop()`
is genuinely clearer than re-arming a timer every pass through a hot scan loop, and it
expresses the fixed-interval intent directly. It also avoids one allocation per scan cycle,
which is not a leak but is not free either.

Recommendation: fold into unrelated work in this file rather than tracking as a standalone
change. Kept open at **Low** for that purpose.

### Severity and priority

The source document rated this **MEDIUM**. Re-checked against this repository at
`origin/main` with `go 1.26.5` on 2026-08-04, the priority is **Low** for the

### Effort

**Low**

Effi-S:
Fixes #2127

**Document severity:** MEDIUM · **Verified priority:** Low · **Effort:** Low · **Type:** Task

### Old issue reference

> **Sub-Task 6 — Fix `time.After` in `inmemory/locker` Scan Sleep Loop**
>
> **Status:** `[ ] pending`

| 6 | `[RUN]` | **MEDIUM** | Timer leak | `selector/simple/inmemory/locker.go` | 268 | `time.After` in scan sleep loop; should be `Ticker` |

### Intent

The scan goroutine in the in-memory locker uses `time.After(d.sleepTimeout)`
inside a `for { select { } }` loop. A new timer is allocated on every
iteration. Although the `ctx.Done()` branch provides an exit path, each
iteration before exit leaks the timer for up to `sleepTimeout` duration.

### Expected Outcomes

- A single `time.Ticker` drives the scan sleep instead of per-iteration timers.
- `ticker.Stop()` is called via `defer` when the goroutine exits.

### Todo List

1. Create `ticker := time.NewTicker(d.sleepTimeout)` before the outer loop.
2. Add `defer ticker.Stop()` immediately after.
3. Replace `case <-time.After(d.sleepTimeout):` with `case <-ticker.C:`.

### Relevant Context

- File: `token/services/selector/simple/inmemory/locker.go`
- Lines 264–284: inner sleep loop in the scan goroutine.

---

### Status in current code

**Not a leak on Go 1.26**, but the `Ticker` suggestion has independent merit. Line numbers have shifted: the loop is now at `token/services/selector/simple/inmemory/locker.go:430`, not 264–284.

The call is now at `token/services/selector/simple/inmemory/locker.go:430`
(`case <-time.After(d.sleepTimeout):`), inside the scan goroutine's sleep loop.

On the leak claim:

`go.mod` declares `go 1.26.5`. Per the [Go 1.23 release notes](https://go.dev/doc/go1.23#timer-changes): *"Timers and Tickers that are no longer referred to by the program become eligible for garbage collection immediately, even if their `Stop` methods have not been called. Earlier versions of Go did not collect unstopped Timers until after they had fired and never collected unstopped Tickers."* The timer returned by `time.After` becomes unreachable the moment the `select` completes, so it is collectable immediately rather than surviving until it fires. The document's premise — that the timer is retained for its full duration — describes pre-1.23 runtime behaviour and does not hold on this toolchain.

So "each iteration before exit leaks the timer for up to `sleepTimeout` duration" is not true
here — the timer is collectable as soon as its iteration's `select` returns.

This is the one item in the timer group where the *suggested* change is still worth making on
its own merits. A single `time.NewTicker` created before the loop with `defer ticker.Stop()`
is genuinely clearer than re-arming a timer every pass through a hot scan loop, and it
expresses the fixed-interval intent directly. It also avoids one allocation per scan cycle,
which is not a leak but is not free either.

Recommendation: fold into unrelated work in this file rather than tracking as a standalone
change. Kept open at **Low** for that purpose.

### Severity and priority

The source document rated this **MEDIUM**. Re-checked against this repository at
`origin/main` with `go 1.26.5` on 2026-08-04, the priority is **Low** for the

### Effort

**Low**

### DIFF
--- a/token/services/selector/simple/inmemory/locker.go
+++ b/token/services/selector/simple/inmemory/locker.go
@@ -34,6 +34,10 @@ const (
 	// stopTimeout is the maximum time to wait for the scan goroutine to stop during shutdown.
 	// This prevents indefinite blocking if the goroutine fails to exit cleanly.
 	stopTimeout = 10 * time.Second
+	// defaultSleepTimeout is the fallback scan interval used when NewLocker is
+	// given a non-positive timeout, which would otherwise make the scan loop's
+	// timer fire immediately and spin without pause. It mirrors the production default.
+	defaultSleepTimeout = 2 * time.Second
 )
 
 var ErrTimeout = errors.New("timeout occurred")
@@ -81,6 +85,10 @@ type locker struct {
 
 func NewLocker(ttxdb TXStatusProvider, timeout time.Duration, validTxEvictionTimeout time.Duration) simple.Locker {
 	ctx, cancel := context.WithCancel(context.Background())
+
+	if timeout <= 0 {
+		timeout = defaultSleepTimeout
+	}
 
 	r := &locker{
 		ttxdb:                  ttxdb,
@@ -381,6 +389,16 @@ func (d *locker) lockedCount() int {
 
 func (d *locker) scan(ctx context.Context) {
 	defer close(d.scanDone)
+
+	// A single one-shot timer, re-armed after each scan, so every scan is
+	// followed by a full d.sleepTimeout pause regardless of how long the scan
+	// itself took. A time.Ticker fires on a fixed grid and would drop that pause
+	// once a scan exceeds the interval (leaving a stale tick already waiting);
+	// time.After would allocate a fresh timer on every pass. On Go 1.23+ Reset
+	// needs no channel-drain dance. NewTimer, unlike NewTicker, does not panic on
+	// a non-positive interval, but NewLocker still clamps it to avoid a spin.
+	timer := time.NewTimer(d.sleepTimeout)
+	defer timer.Stop()
 	for {
 		// Check for shutdown before starting a new scan cycle.
 		select {
@@ -461,8 +479,11 @@ func (d *locker) scan(ctx context.Context) {
 
 		for {
 			logger.DebugfContext(ctx, "token collector: sleep for some time...")
+			// Re-arm for a full interval measured from now, i.e. from the end of
+			// the scan just completed.
+			timer.Reset(d.sleepTimeout)
 			select {
-			case <-time.After(d.sleepTimeout):
+			case <-timer.C:
 			case <-ctx.Done():
 				logger.Debugf("token collector: stopping during sleep")
 
--- a/token/services/selector/simple/inmemory/locker_test.go
+++ b/token/services/selector/simple/inmemory/locker_test.go
@@ -37,6 +37,27 @@ func TestLockEntry(t *testing.T) {
 	assert.Equal(t, "b", m[id2])
 }
 
+// TestNewLockerClampsNonPositiveTimeout verifies that NewLocker substitutes
+// defaultSleepTimeout for a non-positive timeout. The scan goroutine drives
+// time.NewTicker, which panics on a <= 0 interval; clamping in NewLocker keeps
+// that panic (which would crash the whole process from the goroutine) from
+// being reachable, while a positive timeout is passed through unchanged.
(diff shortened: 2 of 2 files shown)


AkramBitar (approved):
See my comments

AkramBitar (commented):
@Effi-S,

Two items:

**1. `Ticker` drops the pause between scans (blocking).** A `Ticker` fires on a fixed grid with a 1-slot buffer. Once a scan exceeds `sleepTimeout` (one `GetStatus` per locked token, `sleepTimeout` is 2s in prod), the `select` finds a stale tick already waiting and rescans with **zero** pause — hammering the status provider. `time.After` always gave a full gap *after* the scan. Use a reusable `Timer` instead: same one allocation, original semantics.

```go
timer := time.NewTimer(d.sleepTimeout)
defer timer.Stop()
// in the inner loop, before the select:
timer.Reset(d.sleepTimeout)
select {
case <-timer.C:
case <-ctx.Done():
    return
}
```

**2. Panic on non-positive interval (minor).** `NewTicker`/`NewTimer` panic when `sleepTimeout <= 0`; `time.After` did not. Clamp it in `NewLocker`.

AkramBitar (changes requested):

PR #2263, opened 2026-08-18
fix(identity): do not hold the registry lock across WalletFactory.NewWallet

### LINKED ISSUE #2064: identity/role: WalletByID holds the write lock across NewWallet, contradicting its own comment
### Summary

`Registry.WalletByID` takes the registry's write lock and then calls `WalletFactory.NewWallet` while
still holding it — directly contradicting the method's own inline comment and the concurrency
invariant documented on the `Registry` type itself. `NewWallet` performs idemix pseudonym generation
and calls back into the registry via the `IdentitySupport` interface (`BindIdentity`), so any factory
implementation that calls back into the same registry self-deadlocks, since `sync.RWMutex` is not
reentrant.

### Where

`token/services/identity/role/registry.go:284-298`:

```go
// Register the newly created wallet but check if another goroutine already created it.
r.WalletMu.Lock()
defer r.WalletMu.Unlock()
if existing, ok := r.Wallets[wID]; ok {
	// Another goroutine created and registered the wallet in the meantime; prefer it.
	return existing, nil
}
// Create the wallet without holding the registry lock (avoid holding locks while calling external code).
r.Logger.DebugfContext(ctx, "create wallet [%s]", wID)
newWallet, err := r.WalletFactory.NewWallet(ctx, wID, role, r, idInfo)
```

The comment on line 291 says "without holding the registry lock," but the call on line 293 happens
while `r.WalletMu` is locked (`:285`, released only by the deferred `Unlock` at function return).
This also violates the type-level invariant documented at `registry.go:27-33`: *"Methods in this file
follow the pattern of ... never holding locks while calling out to external services ... to avoid
blocking and potential deadlocks."*

`r` is passed to `NewWallet` as the `IdentitySupport` argument. `role/wallets.go:587` and `:604` call
`IdentitySupport.BindIdentity` from within wallet methods — if a factory implementation calls
`BindIdentity` during construction (a plausible thing for a wallet to do while registering its
identity), it reaches back into `Registry.BindIdentity`... which does not itself take `WalletMu`, so
today it wouldn't deadlock via that specific path, but any factory that touches `r.Wallets` (e.g. via
`RegisterWallet`, which does take `r.WalletMu.Lock()` at `:172-173`) would deadlock immediately since
`sync.RWMutex` is non-reentrant.

### Impact

- Every wallet creation for a given role serializes behind one lock, including the potentially
  expensive idemix pseudonym generation, DB reads, and DB writes performed inside `NewWallet` — this
  is a concurrency bottleneck, not just a correctness smell.
- A `WalletFactory` implementation that calls back into `Registry.RegisterWallet` (or otherwise
  re-enters `WalletMu`) during construction deadlocks the calling goroutine permanently.

### Reproduction

Not yet committed. A regression test can assert the lock is not held during `NewWallet` by having a
fake `WalletFactory.NewWallet` implementation attempt `r.RegisterWallet(ctx, id, w)` from within its
own body (simulating a factory that registers itself) and confirming it does not hang; today this
would deadlock (test with a timeout/subprocess pattern, since a real deadlock has no natural
termination).

### Suggested fix

Move wallet construction outside the critical section, using `golang.org/x/sync/singleflight` (already
vendored — `go.mod:33` — and already the house pattern for this exact "coalesce concurrent creation
under one key without holding a lock across external calls" problem at
`token/services/utils/cache/ristretto.go:13,32,88`):

```go
newWallet, err, _ := r.creationGroup.Do(wID, func() (interface{}, error) {
	r.WalletMu.RLock()
	if existing, ok := r.Wallets[wID]; ok {
		r.WalletMu.RUnlock()
		return existing, nil
	}
	r.WalletMu.RUnlock()
	return r.WalletFactory.NewWallet(ctx, wID, role, r, idInfo)
})
if err != nil {
	return nil, err
}
r.WalletMu.Lock()
r.Wallets[wID] = newWallet.(driver.Wallet)
r.WalletMu.Unlock()
```

Note: singleflight alone dedupes *concurrent* creation for the same key, but does not help when
storage genuinely reports "not found" due to a transient error (see #7) — that's a separate defect
in the input to this code path, and the two should be fixed together.

### Severity

MEDIUM — real deadlock risk depends on factory implementation details, but the lock-scope violation
and its performance impact are present today regardless.

AkramBitar:
## Problem

`Registry.WalletByID` (`token/services/identity/role/registry.go`) built the wallet **while holding the registry write lock**, contradicting its own inline comment ("Create the wallet without holding the registry lock") and the invariant documented on the `Registry` type.

Flow before this PR:

1. Caller asks for a wallet → cache miss.
2. `WalletByID` takes `r.WalletMu.Lock()` (`defer Unlock`).
3. Double-checks the `Wallets` map — still missing.
4. **Still holding the write lock**, calls `WalletFactory.NewWallet` → idemix pseudonym generation, identity-store reads, identity-store writes.
5. Stores the wallet in the map and returns, releasing the lock.

Two consequences:

- **Serialization**: every wallet creation for a role queues behind step 4, and so does every other reader/writer of the map — the expensive crypto and DB work happens inside the critical section.
- **Deadlock**: the factory is handed the registry itself as `IdentitySupport`, so it may legitimately call back into it (e.g. `RegisterWallet`, which takes `WalletMu.Lock()`). `sync.RWMutex` is not reentrant, so such a factory hangs forever at step 4.

## Fix

Flow after this PR:

1. Caller asks for a wallet → cache miss.
2. Construction runs inside a `singleflight.Group` keyed by wallet id (`golang.org/x/sync/singleflight`, already a dependency and the existing house pattern in `token/services/utils/cache/ristretto.go`).
3. Short `RLock` for the double-check, released immediately.
4. `WalletFactory.NewWallet` is called with **no lock held** — a factory may safely re-enter the registry.
5. `WalletMu.Lock()` only for the map write, then unlock.

`singleflight` keeps the "build once" guarantee that the lock used to provide: concurrent callers for the same wallet id share one `NewWallet` call and get the same instance, while distinct wallet ids are now built in parallel.

## Tests

Both new tests were verified to fail against the unfixed `registry.go`:

- `TestWalletByID_FactoryReentersRegistry` — a factory that calls `RegisterWallet` from inside `NewWallet` must return; on the old code it hangs (timeout-guarded, a deadlock has no natural termination).
- `TestWalletByID_DistinctWalletsAreCreatedConcurrently` — two `NewWallet` calls for different wallet ids must be in flight simultaneously; on the old code they are serialized.
- `TestWalletByID_ConcurrentCreation` additionally asserts exactly one factory call for concurrent same-id requests.

`go test -race -count=5 ./token/services/identity/role/` green; `make checks` and `make lint-auto-fix` clean.

## Docs

New "Wallet Registry concurrency" section in `docs/services/identity.md` stating the contract: `WalletMu` guards the map only, `NewWallet` is never called under a lock, same-id creations are coalesced rather than serialized.

Fixes #2064

### DIFF
--- a/token/services/identity/role/registry.go
+++ b/token/services/identity/role/registry.go
@@ -10,11 +10,13 @@ import (
 	"context"
 	"encoding/json"
 	"sync"
+	"sync/atomic"
 
 	"github.com/LFDT-Panurus/panurus/token/driver"
 	idriver "github.com/LFDT-Panurus/panurus/token/services/identity/driver"
 	"github.com/LFDT-Panurus/panurus/token/services/logging"
 	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
+	"golang.org/x/sync/singleflight"
 )
 
 //go:generate counterfeiter -o mock/wf.go -fake-name WalletFactory . WalletFactory
@@ -31,6 +33,15 @@ type WalletFactory interface {
 //     RLocks for map reads and never holding locks while calling out to external
 //     services (identity provider, storage, wallet factory) to avoid blocking and
 //     potential deadlocks.
+//   - In particular, WalletMu MUST NOT be held while calling WalletFactory.NewWallet:
+//     wallet construction is expensive (pseudonym generation, storage access) and the
+//     factory receives the registry itself as IdentitySupport, so it may legitimately
+//     call back into it. sync.RWMutex is not reentrant, so any such callback would
+//     deadlock. Concurrent creations of the same wallet are instead coalesced with
+//     walletCreationGroup (see WalletByID).
+//   - Because construction runs outside the lock, it can overlap Done. Nothing may be
+//     written to the Wallets map once closed is set: Done has already dropped the cache
+//     and closed what it held, so a later entry would never be closed by anyone.
 type Registry struct {
 	Logger  logging.Logger
 	Role    idriver.Role
@@ -39,7 +50,22 @@ type Registry struct {
 	WalletFactory WalletFactory
 	WalletMu      sync.RWMutex
 	Wallets       map[string]driver.Wallet
+
+	// walletCreationGroup coalesces concurrent WalletFactory.NewWallet calls for the same
+	// wallet identifier, so that a wallet is built at most once without holding
+	// WalletMu across the construction.
+	walletCreationGroup singleflight.Group
+
+	// closed is set by Done. It is guarded by WalletMu and exists because construction
+	// no longer runs under that lock: a creation still in flight when Done runs must not
+	// repopulate the cache Done has just dropped.
+	closed bool
 }
+
+// errRegistryClosed is returned by wallet creation once Done has run. The cache has been
+// dropped and its wallets closed, so a wallet created afterwards would be released by
+// nobody.
+var errRegistryClosed = errors.New("wallet registry is closed")
 
 // NewRegistry returns a new registry for the passed parameters.
 // A registry is bound to a given role, and it is persistent.
@@ -245,6 +271,12 @@ func (r *Registry) GetWalletID(ctx context.Context, identity driver.Identity) (s
 	return wID, nil
 }
 
+// WalletByID returns the wallet bound to the passed lookup id, creating and registering it
+// on first use.
+//
+// Locking: WalletMu is only taken for short map reads/writes. WalletFactory.NewWallet is
+// always called with no lock held, so a factory may safely call back into the registry;
+// concurrent callers for the same wallet identifier are coalesced so the wallet is built once.
 func (r *Registry) WalletByID(ctx context.Context, role idriver.IdentityRoleType, id driver.WalletLookupID) (driver.Wallet, error) {
 	r.Logger.DebugfContext(ctx, "role [%d] lookup wallet by [%T]", role, id)
 	defer r.Logger.DebugfContext(ctx, "role [%d] lookup wallet by [%T] done", role, id)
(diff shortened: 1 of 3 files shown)


token/services/identity/role/registry.go
```diff
@@ -39,8 +50,23 @@ type Registry struct {
 	WalletFactory WalletFactory
 	WalletMu      sync.RWMutex
 	Wallets       map[string]driver.Wallet
+
+	// walletCreation coalesces concurrent WalletFactory.NewWallet calls for the same
+	// wallet identifier, so that a wallet is built at most once without holding
+	// WalletMu across the construction.
+	walletCreation singleflight.Group
```
Effi-S: Super nit: I think `walletCreationGroup` or `creationGroup` are better names

token/services/identity/role/registry.go
```diff
@@ -281,21 +313,126 @@ func (r *Registry) WalletByID(ctx context.Context, role idriver.IdentityRoleType
...
+			// Anything else - including a context error from the creation this caller ran
+			// itself - is its own error to report.
+			inherited := !ranCreation.Load() && ctx.Err() == nil &&
+				(errors.Is(res.Err, context.Canceled) || errors.Is(res.Err, context.DeadlineExceeded))
+			if !inherited || attempt >= walletCreationJoinRetries {
+				return nil, res.Err
+			}
+			r.Logger.DebugfContext(ctx,
+				"wallet [%s] creation failed with another caller's cancellation, retrying", wID)
+			r.walletCreation.Forget(wID)
```
Effi-S: Pretty sure `r.walletCreation.Forget(wID)` is unnecessary and can leak wallet.

Effi-S (changes requested):

adecaro (commented):
## Review: does this fully address #2064?

@AkramBitar

**This genuinely fixes the deadlock/serialization bug in #2064 — verified by reverting `registry.go` and watching the new tests hang exactly as described — with one still-open design concern from @Effi-S's existing review that I'd like to see addressed before merge.**

Checked out `fix/2064-registry-lock-scope` at `84fa870` and verified:

```
go build ./services/identity/role/...            # clean
go vet ./services/identity/role/...               # clean
gofmt -l -s registry.go registry_test.go           # clean
go test -count=1 ./services/identity/role/...      # ok
go test -race -count=1 ./services/identity/role/...# ok
go test -cover ./services/identity/role/...        # 87.1% of statements
git checkout <base commit> -- registry.go && go test ./services/identity/role/...
  # hangs — goroutine dump shows Done()/RegisterWallet blocked on WalletMu
  # held by WalletByID inside WalletFactory.NewWallet — restored afterwards
CI: checks, utest (race+regression), itest (full matrix), bench-check, cgo-check, DCO — all green
```

I also read `token/services/identity/wallet/service.go` (the only caller of `WalletByID`) and the `golang.org/x/sync/singleflight` source to check the `Forget` race mechanism below.

Reverting `registry.go` to the pre-fix version and re-running the new tests fails with a real deadlock (`Done` blocked at `registry.go:317`, `RegisterWallet` blocked at `registry.go:172`, both waiting on `WalletMu` held by `WalletByID` inside `NewWallet`) — the tests pin the bug, not just the diff.

### What it fixes

| # | Finding | Verified |
|---|---|---|
| 2064.1 | `WalletByID` built the wallet while holding `WalletMu.Lock()` | ✅ confirmed on reverted code — held across `WalletFactory.NewWallet` |
| 2064.2 | Re-entrant factory (calls back into `RegisterWallet`) deadlocks | ✅ confirmed — reproduces exactly via revert + `TestWalletByID_FactoryReentersRegistry` |
| 2064.3 | Wallet creation for distinct ids serialized behind one write lock | ✅ confirmed — `TestWalletByID_DistinctWalletsAreCreatedConcurrently` fails on reverted code, passes here |
| 2064.4 | Build-once guarantee preserved despite dropping the lock | ✅ `singleflight` coalescing verified by reading `createWallet`/`newWallet` and `TestWalletByID_ConcurrentCreation` |

All four rows are solid.

### Non-blocking

- **Existing thread, reinforced:** @Effi-S's comment on line 385 ("Pretty sure `r.walletCreation.Forget(wID)` is unnecessary and can leak wallet") is correct and still open. I traced the exact race: `singleflight`'s `doCall` deletes the flight from its map and delivers results to all waiters inside the same critical section, so a caller retrying via `Forget(wID)` after an "inherited cancellation" failure can end up deleting a *different*, freshly-started flight for the same `wID` if a third caller began one in the interim — this can produce two concurrent `WalletFactory.NewWallet` calls for the same id, one of which is silently dropped when `r.Wallets[wID]` is overwritten, leaking any resources (e.g. an anonymous owner wallet's background goroutine) held by the dropped instance. It needs 3+ concurrent callers to hit, which is why the current tests (max 2 callers on that path) don't catch it. Simplest fix: drop the `Forget` call — ordinary `DoChan` already removes the completed entry from its internal map on its own, so the next loop iteration starts a fresh flight regardless.
- **Existing thread, no addition:** the `walletCreation` → `walletCreationGroup` naming nit at line 57 is still open; agree it's reasonable and non-blocking.

### Recommendation

Good to merge once @Effi-S's `Forget`/leak point is resolved one way or another (fix or an explicit rationale for keeping it) — everything else, including the deadlock fix itself, checks out cleanly.

token/services/identity/role/registry.go
AkramBitar: Good point, fixed.

token/services/identity/role/registry.go
AkramBitar: Good catch. 
Fixed. 
Thanks.

Effi-S (approved):
LGTM

PR #2264, opened 2026-08-18
fix(network-driver): bound tcc query request size and item count (#2050)

### LINKED ISSUE #2050: network-driver: tcc queryStates/queryTokens/areTokensSpent have no request-size limit (resource exhaustion)
### Summary

The token chaincode's `invoke` entry point enforces a request-size ceiling before doing any per-element work, but the read-only query functions — `queryStates`, `queryTokens`, `areTokensSpent` — have no analogous guard: each JSON array element the (unauthenticated) caller supplies translates 1:1 into a `stub.GetState` (or equivalent) call, with no cap on array length or payload size.

### Where

`token/services/network/fabric/tcc/tcc.go:281-375`

The `invoke` path checks request size via `CheckRawRequestSize`/`CheckRequestLimits` (`driver.DefaultResourceLimits().MaxRequestBytes`) before processing. `queryStates`/`queryTokens`/`areTokensSpent` skip this check entirely.

### Impact

Any client able to reach the chaincode's query path (which, unlike `invoke`, does not go through the size-limited transaction-submission flow) can submit a single request with an arbitrarily large key/id array, driving an unbounded number of ledger reads inside one chaincode invocation — a straightforward resource-exhaustion / peer-slowdown vector requiring no elevated privileges.

### Reproduction

Reproduced locally with a unit test (not yet committed) that builds a 10,000-key payload already larger than `MaxRequestBytes` (the limit that *would* reject an equivalently-sized `invoke` request) and shows `queryStates` processes it in full:

```go
payload, _ := json.Marshal(keys)  // 10,000 keys
require.Greater(t, len(payload), driver.DefaultResourceLimits().MaxRequestBytes)

resp := cc.Invoke(fakestub)
require.Equal(t, int32(200), resp.Status)
require.Equal(t, n, fakestub.GetStateCallCount())  // one GetState call per key, no cap
```

### Suggested fix

Apply the same size/count limits enforced on `invoke` (`CheckRawRequestSize`/`CheckRequestLimits` or an equivalent per-element cap) to `queryStates`, `queryTokens`, and `areTokensSpent` before processing their input arrays.

### Severity

HIGH — unauthenticated resource-exhaustion vector against the chaincode/peer.

AkramBitar:
## Problem

The chaincode's read-only queries — `queryStates`, `queryTokens`, `areTokensSpent` — perform **one
ledger read per element** of a caller-supplied JSON array, with no cap on array length or payload
size. `invoke` is bounded by `driver.ResourceLimits`; the query path is not, and it doesn't go
through the size-limited transaction-submission flow either.

**Today:** attacker sends `queryStates` an array of 10M keys (no signature, wallet or privileges
needed) → the whole array is unmarshalled → 10M `GetState` calls in **one** invocation → the peer is
tied up. Repeat concurrently for resource exhaustion. Same for the other two.

## Fix

Two limits, both applied before any work proportional to the input:

| Limit | Default | Checked |
| --- | --- | --- |
| `MaxQueryRequestBytes` | 1 MiB | **before** `json.Unmarshal` |
| `MaxQueryItems` | 4096 | after decode, **before the first ledger read** |

**After:** oversized payload → rejected before decoding, **zero** `GetState` calls. Retry with 5,000
short keys → item cap rejects it before the read loop, **zero** reads. A legitimate 4,096-key request
is served exactly as before.

`WithDefaults()` replaces any field left unset *or negative*, so an unconfigured chaincode is still
bounded and a config typo can't disable a limit. The standalone process overrides via
`TOKEN_QUERY_MAX_REQUEST_BYTES` / `TOKEN_QUERY_MAX_ITEMS`. Not consensus-relevant — the query path
writes nothing and isn't an endorsement boundary — so no lockstep rollout needed.

## Bonus: a panic the new fuzzing found

`["queryTokens", "[null]"]` — 8 bytes, inside every limit — panicked: a JSON `null` decodes to a nil
`*token.ID` that `translator.QueryTokens` dereferenced. `Invoke`'s `recover()` meant no peer crash,
but every such request logged a full stack trace, and direct translator callers have no recovery at
all. Fixed at the root; both triggering payloads are kept in the fuzz corpus.

## Tests

- Exact boundaries for both limits and the env provider.
- Per query function: over-count and oversize each rejected with **zero** `GetState` calls; at exactly
  `MaxQueryItems` served with one read per element; unconfigured chaincode still bounded. Verified
  these fail with the guards removed.
- One fuzz target per query function — the surface behind the shared check differs (`queryStates` uses
  the string as a key verbatim; `areTokensSpent` adds validator init and, with graph hiding, the
  composite-key builder; `queryTokens` decodes structs). Nightly matrix entries for all three.
  `f.Add` carries the seed shapes; `testdata/fuzz/` holds only the 8 files that benefit from being on
  disk — the `MaxQueryItems` boundary pair per target and the two crash reproducers, which report as
  named subtests and outlive the build cache the generated corpus lives in.
- `make checks` and `make lint-auto-fix` clean; `./token/services/network/...` green, `tcc` also under
  `-race`.

## Docs

New `docs/security/tcc_query_limits.md`, linked from `docs/services/network-fabric.md` and
`docs/configuration.md`.

> `docs/configuration.md` is also touched by #1925 (which bounds the **storage** service and doesn't
> touch `tcc/`) — trivial rebase on that one file for whichever merges second.

Fixes #2050


Effi-S (approved):
LGTM

PR #2268, opened 2026-08-19
fix(identity): accept a boolpolicy rest recipient on long-term owner wallets

### LINKED ISSUE #2267: LongTermOwnerWallet.RegisterRecipient rejects a boolpolicy rest recipient, so change cannot return to a policy identity
## Context

#1918 implemented the recipient guard on `LongTermOwnerWallet.RegisterRecipient` (previously a no-op with a TODO): the recipient identity must equal the long-term identity the wallet is bound to.

## The gap

A transfer that spends boolpolicy-owned inputs naturally sends the rest back to the policy identity: the caller passes it via `WithRestRecipientIdentity`, and `prepareTransfer` registers it through `wallet.RegisterRecipient` on the sending long-term wallet. A boolpolicy composite identity can never equal the wallet's own long-term identity, so the guard now rejects it:

```
failed to register recipient identity [...] for the rest, wallet [...]:
identity [...] does not belong to this wallet [...]
```

`token/services/ttx/boolpolicy` avoids change by spending whole tokens (`Spend` pins exactly one token for its full amount), but `Redeem` has no whole-token variant: redeeming an arbitrary amount from policy-owned holdings necessarily produces a rest, and the only owner that keeps custody semantics intact is the policy identity itself. We hit this in our test environments on every redeem prepared from policy-owned holdings.

## Proposal

Accept a rest recipient whose identity is a well-formed boolpolicy composite that lists the wallet's own long-term identity among its components — typed-identity tag is `PolicyIdentityType`, the inner `PolicyIdentity` ASN.1 parses with no trailing bytes, the component list is non-empty, bounded, and free of empty or duplicate entries, and the audit info carries one entry per component. Everything else keeps failing exactly as before.

Accepting such a recipient grants no spending power: spending the rest output still requires the policy members' signatures on-chain; a malformed policy owner would in any case be rejected by validation before funds move.

PR follows.

EvanYan1024:
Fixes #2267

## What

`LongTermOwnerWallet.RegisterRecipient` rejects any recipient identity other than the wallet's own long-term identity. That breaks transfers that spend boolpolicy-owned inputs and send the rest back to the policy identity (the only rest owner that keeps custody semantics intact); `Redeem` in particular has no whole-token variant, so redeeming an arbitrary amount from policy-owned holdings cannot avoid a rest. Hit in our test environments on every redeem prepared from policy-owned holdings.

## How

Keep the guard as-is for everything else and accept exactly one new shape: a well-formed boolpolicy composite recipient that lists the wallet's own long-term identity among its components —

- the typed-identity tag is `PolicyIdentityType`,
- the inner `PolicyIdentity` ASN.1 parses with no trailing bytes,
- the component list is non-empty, within `MaxIdentityComponentsFrom`, and free of empty or duplicate entries (mirroring boolpolicy's `validateComponentIdentities`),
- the wallet's long-term identity is one of the components,
- the audit info decodes strictly (the `token/core/common/encoding/json` decoder) with one entry per component identity.

Accepting such a recipient grants no spending power: spending the rest output still requires the policy members' signatures on-chain.

One implementation note for review: `role` cannot import `identity/boolpolicy` (import cycle through `membership`'s test chain: `boolpolicy` tests → `x509` → `membership` → `role`), so the two stable wire formats are mirrored locally as `policyIdentityWire`/`policyAuditInfoWire` with a comment pointing at the authoritative definitions. Happy to restructure (e.g. move the wire types into a leaf package) if you prefer.

## Testing

- New cases in `TestLongTermOwnerWallet/RegisterRecipient`: a policy recipient listing the wallet's identity is accepted, also at the component fan-out limit and under context-carried limits; a policy over unrelated owners, an empty component list, empty or duplicate components, one component above the fan-out limit, audit info that does not cover every component, audit info with unknown JSON fields, and garbage under the policy type tag are all rejected. The existing four cases are unchanged.
- New `FuzzLongTermOwnerWalletRegisterRecipientNoPanic` covers the recipient parsing path, with an independent oracle re-checking every acceptance invariant; registered in the nightly fuzz matrix.
- `docs/services/ttx.md` updated: it still described `LongTermOwnerWallet.RegisterRecipient` as a no-op.
- Red-green verified: the accepted-policy-recipient case fails on current main, and the rejected-foreign-policy case fails on the previous head of this branch.
- `gofmt`, `go build ./token/...`, `go vet`, `go test -race` on the identity tree, golangci-lint v2.12.2 with the repo config (0 issues), `make gofix` — all clean.

### DIFF
--- a/token/services/identity/role/wallets.go
+++ b/token/services/identity/role/wallets.go
@@ -9,10 +9,13 @@ package role
 import (
 	"bytes"
 	"context"
+	"encoding/asn1"
 	"math/big"
 
+	"github.com/LFDT-Panurus/panurus/token/core/common/encoding/json"
 	"github.com/LFDT-Panurus/panurus/token/core/common/metrics"
 	"github.com/LFDT-Panurus/panurus/token/driver"
+	"github.com/LFDT-Panurus/panurus/token/services/identity"
 	idriver "github.com/LFDT-Panurus/panurus/token/services/identity/driver"
 	"github.com/LFDT-Panurus/panurus/token/services/identity/wallet"
 	"github.com/LFDT-Panurus/panurus/token/services/logging"
@@ -473,11 +476,23 @@ func (w *LongTermOwnerWallet) EnrollmentID() string {
 // because the wallet's identity is already resolved and bound at
 // construction time (see NewLongTermOwnerWallet); this is purely a guard
 // against mismatched recipient data.
+// A boolpolicy composite identity listing this wallet's identity among its
+// components is also accepted and registered: the rest of a transfer that
+// spends policy-owned inputs goes back to the policy identity, which by
+// construction differs from the wallet's own long-term identity.
 func (w *LongTermOwnerWallet) RegisterRecipient(ctx context.Context, data *driver.RecipientData) error {
 	if data == nil {
 		return errors.Wrapf(wallet.ErrNilRecipientData, "invalid recipient data")
 	}
 	if !w.Contains(ctx, data.Identity) {
+		if w.isMemberPolicyRecipient(ctx, data) {
+			if err := w.IdentityProvider.RegisterRecipientData(ctx, data); err != nil {
+				return errors.Wrapf(err, "failed registering audit info for owner [%s]", data.Identity)
+			}
+
+			return nil
+		}
+
 		return errors.Errorf("identity [%s] does not belong to this wallet [%s]", data.Identity, w.ID())
 	}
 	if !bytes.Equal(data.AuditInfo, w.OwnerAuditInfo) {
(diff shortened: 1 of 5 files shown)


token/services/identity/role/wallets.go
```diff
@@ -487,6 +498,63 @@ func (w *LongTermOwnerWallet) RegisterRecipient(ctx context.Context, data *drive
...
+func (w *LongTermOwnerWallet) isMemberPolicyRecipient(ctx context.Context, data *driver.RecipientData) bool {
+	ti, err := identity.UnmarshalTypedIdentity(data.Identity)
+	if err != nil || ti.Type != driver.PolicyIdentityType {
+		return false
+	}
+	pi := &policyIdentityWire{}
+	if rest, err := asn1.Unmarshal(ti.Identity, pi); err != nil || len(rest) > 0 {
+		return false
+	}
+	if n := len(pi.Identities); n == 0 || n > driver.MaxIdentityComponentsFrom(ctx) {
```
AkramBitar, line 541: **The policy expression itself is never validated.** The components are checked thoroughly, but `pi.Policy` is only unmarshalled and then ignored, so both of these are accepted:

- `Policy: ""` — `boolpolicy`'s `Parse("")` errors, so `DeserializeVerifier` can never build a verifier for this identity.
- `Policy: "$7 AND $9"` with 2 components — `evalRef` bounds-checks out-of-range refs to `false`, so the policy is unsatisfiable.

Either way the change output of the redeem is permanently unspendable, which is exactly the class of caller mistake this guard exists to catch. Note `boolpolicy.WrapPolicyIdentity` already rejects an empty policy at construction time, so accepting one here is inconsistent with how these identities are built.

A `len(pi.Policy) > 0` check costs nothing and closes the empty case. The full `$N`-in-range check needs the parser, which is blocked by the import cycle you documented above — worth a `TODO` if you don't want to duplicate it.

AkramBitar (commented):
@EvanYan1024  Thanks a lot

Three gaps in the new `isMemberPolicyRecipient` itself, verified by running the code at `3ed273b5`. I would like to consult with you if we can resolved at least first two before merge; the third is a symmetry issue with the non-policy branch.

token/services/identity/role/wallets.go
```diff
@@ -473,11 +476,19 @@ func (w *LongTermOwnerWallet) EnrollmentID() string {
...
+// A boolpolicy composite identity listing this wallet's identity among its
+// components is also accepted: the rest of a transfer that spends
+// policy-owned inputs goes back to the policy identity, which by construction
+// differs from the wallet's own long-term identity.
 func (w *LongTermOwnerWallet) RegisterRecipient(ctx context.Context, data *driver.RecipientData) error {
 	if data == nil {
 		return errors.Wrapf(wallet.ErrNilRecipientData, "invalid recipient data")
 	}
 	if !w.Contains(ctx, data.Identity) {
+		if w.isMemberPolicyRecipient(ctx, data) {
```
AkramBitar, line 488: **The accepted recipient is validated and then discarded — never registered.** This path `return nil`s without calling `w.IdentityProvider.RegisterRecipientData(ctx, data)`, which is already on the interface and is what `AnonymousOwnerWallet.RegisterRecipient` does. The caller comment at `request.go:1646` says "register it and use it".

Downstream consequence: `boolpolicy.TypedIdentityDeserializer.GetAuditInfo` prefers stored audit info for the composite, and with nothing stored it rebuilds it from a per-component store lookup. If any component's audit info isn't locally present, the transfer fails with `failed getting audit info for component identity [k]` — even though the caller supplied a complete audit-info blob that this guard had just finished verifying.

So the method currently succeeds while leaving the wallet in the state the caller was trying to avoid.

token/services/identity/role/wallets.go
```diff
@@ -487,6 +498,63 @@ func (w *LongTermOwnerWallet) RegisterRecipient(ctx context.Context, data *drive
...
+	}
+	if !member {
+		return false
+	}
+	ai := &policyAuditInfoWire{}
+	if err := json.Unmarshal(data.AuditInfo, ai); err != nil {
+		return false
+	}
+
+	return len(ai.IdentityAuditInfos) == len(pi.Identities)
```
AkramBitar: **Audit info is counted, not matched.** The only constraint is that the number of entries equals the number of components. I verified that a correct policy identity paired with audit info belonging to an unrelated pair of owners (same count) is accepted.

This is asymmetric with the non-policy branch a few lines up, which does an exact `bytes.Equal(data.AuditInfo, w.OwnerAuditInfo)`, and with `docs/services/ttx.md`, which states that a mismatched Identity/AuditInfo pair is rejected. Comparing the entry at the wallet's own component index against `w.OwnerAuditInfo` would restore that symmetry — you already know the index from the `member` loop. As written, the mismatch surfaces much later, in the auditor's `InfoMatcher`.

AkramBitar (approved):
LGTM

PR #2276, opened 2026-08-19
tokens: Parse returns TokenToAppend entries aliasing the caller's OutputStream, and those pointers are cached

### LINKED ISSUE #2186: tokens: Parse returns TokenToAppend entries aliasing the caller's OutputStream, and those pointers are cached
### Summary
`Parse` builds each `TokenToAppend.Tok` as `&output.Token` — a pointer into the caller-owned `token.OutputStream`, not a copy. Those `TokenToAppend` values are then stored verbatim in the requests cache by `CacheRequest` and read back by `getActions` on the commit path, so a cached, about-to-be-persisted entry aliases memory the original caller still owns and may reuse or mutate before the commit happens.

### Where
`token/services/tokens/tokens.go:445` and `:495`:
```go
tta := TokenToAppend{
	...
	Tok: &output.Token,
	...
}
```
Cached at `tokens.go:175-180` (`CacheRequest` → `t.RequestsCache.Add(...)`), read back at `tokens.go:354-358` (`getActions`'s cache-hit path), consumed by `AppendValid` at `tokens.go:145-150`.

### Impact
Any code path that reuses or mutates a `token.Output` after `Parse` returns — including in tests that build outputs from a pool, or a future producer that recycles output structs — will observe the mutation reflected through the cached `TokenToAppend.Tok`, potentially at commit time, well after the value was supposedly captured.

### Reproduction
```go
store, _, err := ts.Parse(ctx, auth, "tx1", md, is, os, false, 64, false)
require.NoError(t, err)

output1.Token.Quantity = "0xdeadbeef" // mutate the caller's own stream after Parse returns

// BUG: the change is visible through the value Parse already returned
assert.Equal(t, "0xdeadbeef", store[0].Tok.Quantity)
```
```
=== RUN   TestParse_AliasesCallersOutputStream
--- PASS: TestParse_AliasesCallersOutputStream (0.00s)
PASS
```

### Severity
Medium — no reproduction via the current production call graph mutates outputs post-`Parse`, but the aliasing is a latent hazard for any future caller and for the cache's implicit assumption that cached entries are immutable snapshots.

Part of #2112.

Effi-S:
Fixes #2186

### Summary
`Parse` builds each `TokenToAppend.Tok` as `&output.Token` — a pointer into the caller-owned `token.OutputStream`, not a copy. Those `TokenToAppend` values are then stored verbatim in the requests cache by `CacheRequest` and read back by `getActions` on the commit path, so a cached, about-to-be-persisted entry aliases memory the original caller still owns and may reuse or mutate before the commit happens.

### Where
`token/services/tokens/tokens.go:445` and `:495`:
```go
tta := TokenToAppend{
	...
	Tok: &output.Token,
	...
}
```
Cached at `tokens.go:175-180` (`CacheRequest` → `t.RequestsCache.Add(...)`), read back at `tokens.go:354-358` (`getActions`'s cache-hit path), consumed by `AppendValid` at `tokens.go:145-150`.

### Impact
Any code path that reuses or mutates a `token.Output` after `Parse` returns — including in tests that build outputs from a pool, or a future producer that recycles output structs — will observe the mutation reflected through the cached `TokenToAppend.Tok`, potentially at commit time, well after the value was supposedly captured.

### Reproduction
```go
store, _, err := ts.Parse(ctx, auth, "tx1", md, is, os, false, 64, false)
require.NoError(t, err)

output1.Token.Quantity = "0xdeadbeef" // mutate the caller's own stream after Parse returns

// BUG: the change is visible through the value Parse already returned
assert.Equal(t, "0xdeadbeef", store[0].Tok.Quantity)
```
```
=== RUN   TestParse_AliasesCallersOutputStream
--- PASS: TestParse_AliasesCallersOutputStream (0.00s)
PASS
```

### Severity
Medium — no reproduction via the current production call graph mutates outputs post-`Parse`, but the aliasing is a latent hazard for any future caller and for the cache's implicit assumption that cached entries are immutable snapshots.

Part of #2112.

### DIFF
--- a/token/services/tokens/tokens.go
+++ b/token/services/tokens/tokens.go
@@ -9,6 +9,7 @@ package tokens
 import (
 	"context"
 	"runtime/debug"
+	"slices"
 
 	"github.com/LFDT-Panurus/panurus/token"
 	"github.com/LFDT-Panurus/panurus/token/driver"
@@ -439,18 +440,20 @@ func (t *Service) Parse(
 				continue
 			}
 
+			// clone the token and the caller-owned byte slices so the cached entry does not
+			// alias the output stream, which may be reused or mutated after Parse returns
 			tta := TokenToAppend{
 				TxID:                  string(requestAnchor),
 				Index:                 output.Index,
-				Tok:                   &output.Token,
-				TokenOnLedger:         output.LedgerOutput,
+				Tok:                   output.Clone(),
+				TokenOnLedger:         slices.Clone(output.LedgerOutput),
 				TokenOnLedgerFormat:   output.LedgerOutputFormat,
-				TokenOnLedgerMetadata: output.LedgerOutputMetadata,
+				TokenOnLedgerMetadata: slices.Clone(output.LedgerOutputMetadata),
 				OwnerType:             "",
 				OwnerIdentity:         []byte{},
 				OwnerWalletID:         "",
 				Owners:                nil,
-				Issuer:                issuer,
+				Issuer:                slices.Clone(issuer),
 				Precision:             precision,
 				Flags: Flags{
 					Mine:     false,
@@ -489,18 +492,20 @@ func (t *Service) Parse(
 			return nil, nil, errors.Wrapf(err, "failed to extract owner type for token [%s:%d]", requestAnchor, output.Index)
 		}
 
+		// clone the token and the caller-owned byte slices so the cached entry does not
+		// alias the output stream, which may be reused or mutated after Parse returns
 		tta := TokenToAppend{
 			TxID:                  string(requestAnchor),
 			Index:                 output.Index,
-			Tok:                   &output.Token,
-			TokenOnLedger:         output.LedgerOutput,
+			Tok:                   output.Clone(),
+			TokenOnLedger:         slices.Clone(output.LedgerOutput),
 			TokenOnLedgerFormat:   output.LedgerOutputFormat,
-			TokenOnLedgerMetadata: output.LedgerOutputMetadata,
+			TokenOnLedgerMetadata: slices.Clone(output.LedgerOutputMetadata),
 			OwnerType:             identity.TypeToString(ownerType),
 			OwnerIdentity:         ownerIdentity,
 			OwnerWalletID:         ownerWalletID,
 			Owners:                ids,
-			Issuer:                output.Issuer,
+			Issuer:                slices.Clone(output.Issuer),
 			Precision:             precision,
 			Flags: Flags{
 				Mine:    mine,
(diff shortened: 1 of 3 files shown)


token/services/tokens/tokens.go
```diff
@@ -439,10 +439,13 @@ func (t *Service) Parse(
...
 			}
 
+			// copy the token so the cached entry does not alias the caller-owned output stream,
+			// which may be reused or mutated after Parse returns
+			tok := output.Token
 			tta := TokenToAppend{
 				TxID:                  string(requestAnchor),
 				Index:                 output.Index,
-				Tok:                   &output.Token,
+				Tok:                   &tok,
```
AkramBitar: The PR fixes the pointer aliasing but misses the slice aliasing on Owner []byte. A struct copy in Go does not copy slice backing arrays.

I think missing at both sites, please recheck that:

tok.Owner = bytes.Clone(output.Token.Owner)

AkramBitar (changes requested):

token/services/tokens/tokens.go
Effi-S: `Token`'s `Quantity` and `Type` are strings so they are immutable, 
`tok := output.Token` is a shallow copy.  
The copy gets its own header for cache.

bytes.Clone is more defensive (and probably best practice) 
So it's a nice addition

AkramBitar (dismissed):
See my comment

Effi-S (commented):
Done

token/services/tokens/tokens.go
```diff
@@ -439,18 +440,20 @@ func (t *Service) Parse(
...
+			// clone the token and the caller-owned byte slices so the cached entry does not
+			// alias the output stream, which may be reused or mutated after Parse returns
 			tta := TokenToAppend{
 				TxID:                  string(requestAnchor),
 				Index:                 output.Index,
-				Tok:                   &output.Token,
-				TokenOnLedger:         output.LedgerOutput,
+				Tok:                   output.Clone(),
+				TokenOnLedger:         slices.Clone(output.LedgerOutput),
 				TokenOnLedgerFormat:   output.LedgerOutputFormat,
```
AkramBitar, line 450: can we check if we have test or add new one to check this change?

AkramBitar (changes requested):

token/services/tokens/tokens.go
Effi-S: @AkramBitar , 
`token/token/token_test.go:198` (`TestToken_Clone`)  - tests the cloning is correct and tests mutating the original does not affect the cloned token. 

This should be enough for cache alias issue in my opinion.

AkramBitar (approved):
LGTM

AkramBitar (approved):

PR #2278, opened 2026-08-20
feat(nwo): add fabricx+evm gateway backend for token integration tests

### LINKED ISSUE #1989: nwo: bootstrap a fabricx+EVM backend for token integration tests

arunabha003:
### Addressed issue
Closes #1989 — add fabricX + EVM support to NWO, so the fungible integration
suite can run against a fabric-x-evm gateway node (alongside the existing Fabric
and Besu backends).

### Summary
- New self-contained node-boot layer under `integration/nwo/token/evm/gateway/`
  (reachability probe, docker-CLI container launcher, `testnode` helper) that
  boots a fabric-x-evm gateway in its gasless, in-memory `testnode` mode.
- Made the EVM node pluggable behind a small `Node` interface; Besu stays the
  default and is untouched. A `gatewayNode` boots fabric-x-evm and reports its
  chain id (fixing a funder rejection from a zero chain id).
- Gasless-aware provisioning: submitters get keys only, no ETH funding — the
  gateway is gasless, so Besu-style funding txs are dropped there.
- Registered `fabricx-evm` as a selectable backend + gateway topology, wired
  into the fungible topology and the NWO factory.
- New `integration/token/fungible/evmgw` suite runs the same shared fungible
  bodies against the gateway. Image-gated so it skips cleanly in CI (which does
  not pre-pull the image).

Uses the published `ghcr.io/hyperledger/fabric-x-evm:0.1.3` (first release with
the self-contained `--listen`/`--chain-id` testnode mode) — no build-from-source.

### Validation done
- `make checks`, `go build ./...`, `go vet`, `golangci-lint` — all green.
- Full suite passes end-to-end against `0.1.3` (issue → transfer → redeem,
  ~15 min, `1 Passed | 0 Failed | 0 Skipped`, no dropped commits).

### Reproduce locally (from the repo root):

    docker pull ghcr.io/hyperledger/fabric-x-evm:0.1.3
    export FAB_BINS=$PWD/../fabric/bin
    cd integration
    go test -timeout 40m -v ./token/fungible/evmgw/...
    
  ### Screenshots
  
  
Screenshot 2026-08-21 at 1 20 35 PM


integration/token/fungible/evmgw/evmgw_test.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/hyperledger-labs/fabric-smart-client/integration/nwo/fsc"
+	"github.com/hyperledger-labs/fabric-smart-client/pkg/node"
+	. "github.com/onsi/ginkgo/v2"
+)
+
+// gatewayImage is the published fabric-x-evm image the suite boots; the suite skips when it is absent locally.
+const gatewayImage = "ghcr.io/hyperledger/fabric-x-evm:0.1.3"
+
+// The EVM gateway suite runs the shared fungible test bodies against the fabric-x-evm gateway node.
+var _ = Describe("EndToEnd", func() {
```
adecaro, line 25: please, extend itest-evm in @.github/workflows/tests.yml to execute also these tests so we can get a confirmation from the CI. You will need to add entries in makefile as well. Just follow the existing pattern.

integration/token/fungible/evmgw/evmgw_test.go
```diff
@@ -0,0 +1,81 @@
...
+	Describe("Fungible with Auditor ne Issuer", func() {
+		ts, selector := newTestSuite(fsc.LibP2P, 0, "alice", "bob", "charlie")
+		// Runs before ts.Setup so the image check gates the whole spec.
+		BeforeEach(func() {
+			if err := exec.Command("docker", "image", "inspect", gatewayImage).Run(); err != nil {
```
adecaro: I would move this to the NWO extension. Actually, using the moby api you can check if exists. Even better, you can follow what we do in @integration/nwo/token/evm/besu.go to start the Besu image.

adecaro (changes requested):
Please, check the comments.

integration/token/fungible/evmgw/evmgw_test.go
arunabha003: addressed

integration/token/fungible/evmgw/evmgw_test.go
arunabha003: addressed

integration/token/fungible/evmgw/evmgw_suite_test.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+func TestEndToEnd(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "EndToEnd EVM Gateway Suite")
+}
+
+// StartPortEVMGateway returns the port range this suite allocates from, shared with the Besu suite.
+func StartPortEVMGateway() int {
+	return integration.EVMFungible.StartPortForNode()
```
atharrva01: nit: this reuses EVMFungible's port range instead of getting its own entry in integration/ports.go. That file documents the convention explicitly (evm suites are appended, not inserted, so each gets its own block) and evm-fabtoken followed it with EVMFungibleFabToken. Sharing Besu's range here is fine in CI since matrix entries run on isolated runners, but running evm and evmgw locally at the same time would collide on the same ports. Might be worth a dedicated EVMFungibleGateway entry to stay consistent.

atharrva01 (commented):
Both of Angelo's comments look properly addressed, checked the moby-based node boot against besu.go and it's a faithful mirror (same CheckImagesExist/container-remove/waitReady pattern), and the CI/Makefile wiring matches the existing evm/evm-fabtoken targets. One nit inline, non-blocking.

integration/token/fungible/evmgw/evmgw_suite_test.go
arunabha003: addressed

PR #2279, opened 2026-08-20
fix(tcc): reject null token ids in queryTokens

### LINKED ISSUE #2051: network-driver: tcc QueryTokens panics on a null id element ([null])
### Summary

`QueryTokens` unmarshals attacker-controlled JSON directly into `[]*token.ID` with no per-element validation. A JSON array element of `null` decodes to a nil `*token.ID` pointer (a valid, no-error outcome for `encoding/json`), which the translator then dereferences, panicking on a nil-pointer dereference.

### Where

`token/services/network/fabric/tcc/tcc.go:281-296` (chaincode entry point) and `token/services/network/common/rws/translator/translator.go` (`QueryTokens`, dereferences `id.TxId`/`id.Index`)

### Impact

Any unauthenticated caller of the chaincode's `queryTokens` function can trigger a panic on demand by including a `null` element in the ids array. `Invoke`'s top-level `recover()` downgrades this to an HTTP-500-equivalent response instead of crashing the peer process, but every such request still pays the panic/recover cost and generates log noise — a cheap, repeatable "poke it and see it panic" primitive for an attacker probing the chaincode.

### Reproduction

```go
require.Panics(t, func() { cc.QueryTokens([]byte(`[null]`), fakestub) })

fakestub.GetArgsReturns([][]byte{[]byte(chaincode2.QueryTokensFunctions), []byte(`[null]`)})
resp := cc.Invoke(fakestub)
require.Equal(t, int32(500), resp.Status)  // recovered, not crashed — but still panics internally
```

### Suggested fix

Validate each unmarshaled `*token.ID` element is non-nil before passing the slice to the translator, returning a normal validation error for `null` elements instead of letting the dereference panic (even though it's currently caught by the top-level `recover()`).

### Severity

MEDIUM (dampened by the top-level `recover()` in `Invoke`, but still an unauthenticated panic-on-demand primitive)

AkramBitar:
Fixes #2051

## The problem, in one line

Anyone who can query the token chaincode could make it crash on demand by asking for a token id that is `null`.

## What happened before (the bug)

```
1. Caller asks the chaincode: queryTokens  [null]
2. Chaincode turns that text into a list of token ids
   -> "null" becomes an empty slot (no id at all), and no error is raised
3. Chaincode passes the list on, empty slot included
4. The next layer reads "the transaction id of this token" from the empty slot
   -> nothing there -> CRASH (nil pointer panic)
5. A safety net higher up catches the crash and returns an error 500
```

So nothing was stolen and the peer stayed up — but every such request cost a
crash-and-recover plus a stack trace in the logs, and it could be repeated as
often as the caller liked.

## What happens now (the fix)

```
1. Caller asks the chaincode: queryTokens  [null]
2. Chaincode turns that text into a list of token ids
3. NEW: chaincode checks every entry in the list
   -> finds the empty slot and stops right there
4. Caller gets a plain error: "invalid token id at position [0]: null"
   -> no crash, no stack trace, safety net never needed
```

A second check was added one layer deeper (in the translator), so the crash
cannot happen even if an empty slot reaches it from somewhere else in the code.

## What about an id that is *present but empty*?

Worth asking, so it was checked: `[{}]` (an id with no transaction id in it) never
crashed. It went all the way to the ledger, looked up a key that cannot exist, and
came back as an error. Harmless — but it wasted one ledger lookup per bogus id and
the error it produced was confusing.

An empty transaction id is never a real token id, so it is now rejected in the same
place as `null`, before any ledger lookup:

```
Caller asks for [{}]  ->  "invalid token id at position [0]: empty tx id"
```

## Changes

- `token/services/network/fabric/tcc/tcc.go` — reject `null` ids and ids with an empty `tx_id` in `queryTokens`.
- `token/services/network/common/rws/translator/translator.go` — treat a nil id as an error instead of reading from it.
- Tests for both layers, reproducing the original crash before the fix.
- A fuzz test that throws random/broken input at `queryTokens` and requires it never to crash, added to the nightly fuzz job.
- `docs/services/network-fabric.md` — short section on what the query functions accept.

## Verification

`make checks` clean, `make lint` 0 issues, `go test -race` green on both packages, 25s fuzz run with no findings.

### DIFF
--- a/.github/workflows/nightly-fuzz.yml
+++ b/.github/workflows/nightly-fuzz.yml
@@ -123,6 +123,9 @@ jobs:
           - name: driver-config-key-field-round-trip
             pkg: ./token/driver
             func: FuzzConfigKeyFieldRoundTrip
+          - name: tcc-query-tokens
+            pkg: ./token/services/network/fabric/tcc
+            func: FuzzQueryTokensNoPanic
 
     steps:
       - name: Checkout code
--- a/docs/services/network-fabric.md
+++ b/docs/services/network-fabric.md
@@ -48,6 +48,27 @@ The chaincode exposes the following functions:
 | `areTokensSpent` | Check if tokens are spent | Token IDs, metadata      | Boolean array |
 | `queryStates` | Query arbitrary state keys | State keys               | State values |
 
+### Input Validation
+
+The query functions are reachable by any client that can query the channel, so their
+arguments are treated as untrusted input:
+
+- `queryTokens` expects a JSON array of token ids (`{"tx_id": ..., "index": ...}`). Each
+  element is validated before any state is read:
+  - a `null` element decodes to a nil token id and is rejected with
+    `invalid token id at position [i]: null`, rather than being passed on to the translator,
+    which dereferences each id. `translator.QueryTokens` rejects nil ids as well, so an id
+    list assembled elsewhere in the codebase cannot panic it either;
+  - an element with an empty `tx_id` is rejected with
+    `invalid token id at position [i]: empty tx id`. It is not a valid token id, and letting
+    it through would spend a state read on a key that cannot exist.
+- `areTokensSpent` and `queryStates` expect a JSON array of strings; anything else fails to
+  unmarshal and returns an error response.
+
+Malformed arguments always produce an ordinary error response (status 500). `Invoke`'s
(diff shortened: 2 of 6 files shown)


Effi-S (approved):
LGTM

PR #2283, opened 2026-08-20
fix(core): resolve the first usable public parameters, not the first retrievable

### LINKED ISSUE #2281: newTMS commits to the first retrievable public parameters instead of the first usable one
### Original report (by Angelo)

> fix for `func (m *TMSProvider) newTMS(opts *driver.ServiceOptions) (driver.TokenManagerService, error) {`. Instead of fetching the pp bytes first and then trying to instantiate the TMS, we should do the following: for each option, retrieve pp raw, try to instantiate TMS, if you can, go ahead with it, otherwise try next opt.

### Description

`newTMS` (`token/core/tms.go:219`) resolves the public parameters and instantiates the token service as two separate steps:

```go
ppRaw, err := m.loadPublicParams(opts)  // first source that returns bytes wins
ts, err := m.tokenDriverService.NewTokenService(tmsID, opts.PublicParams)  // only one attempt
```

`loadPublicParams` tries four PP sources in priority order:

1. `opts.PublicParams` (caller-provided)
2. local PP storage
3. local config file (`publicParameters.path`)
4. `PublicParamsFetcher` — the authoritative one, reads the PP from the chaincode

The loop moves on to the next source only when a source **returns no bytes**. It does not move on when a source returns bytes that turn out to be **unusable**: the loop commits to the first blob it can retrieve, and the single `NewTokenService` call then decides pass/fail for the whole TMS. Every remaining source — including the authoritative fetcher — is skipped.

The rejection happens inside the driver's PP deserialization (`token/core/zkatdlog/nogh/v1/setup/setup.go:474` and the fabtoken equivalent), which is where PP **format** problems surface:

- `invalid identifier, expecting [dlog.v1], got [dlog.v2]` — driver name/version skew
- `failed to deserialize public parameters` / `failed unmarshalling public parameters` — container or proto shape changed
- missing or out-of-range curve id, and the other field-level checks

### Impact

A PP blob that is retrievable but unusable permanently shadows every lower-priority source. When a node's local PP copy is stale or malformed — the typical situation during an upgrade, where the on-chain PP is updated before the nodes are — the TMS cannot be instantiated at all, even though the authoritative on-chain PP one step further down the list is valid. The node cannot recover on its own; it needs manual config or database intervention.

Diagnosis is also harder than it needs to be. Only one source ever reaches the deserialization code, so the reported error carries a single format error from a single source and says nothing about the other three. Retrieval errors from the earlier sources are only `logger.Warnf`'d (`token/core/tms.go:243`) and never surfaced, so the logs do not identify which of the four sources supplied the offending bytes.

AkramBitar:
Fixes #2281

## Problem

**Old behaviour — stops at the first source that returns bytes, not at the first usable one:**

```text
1. Try options       -> got bytes -> pass to driver -> driver rejects (wrong version)
                                                      |
                                                      v
                                                 STOP. Return error.
                                                 Storage / config / ledger never tried.
```

`newTMS` (`token/core/tms.go`) resolved the public parameters and instantiated the token service as two independent steps: `loadPublicParams` walked the four sources (caller options -> local storage -> local configuration -> network fetcher) and returned the bytes of the first one that had *any*, and a single `NewTokenService` call then decided pass or fail for the whole TMS.

So the walk moved on only when a source returned *no bytes* — never when a source returned bytes the driver **rejects** (driver name/version skew such as `invalid identifier, expecting [dlog.v1], got [dlog.v2]`, an unparsable container, an out-of-range curve id, ...). A retrievable but unusable copy permanently shadowed every lower-priority source, including the authoritative one on the ledger.

Result: a stale or corrupted local copy permanently blocks the authoritative copy on the ledger. During a public parameters upgrade — where the on-chain parameters are updated before the nodes are — the affected TMS cannot be instantiated at all, and cannot recover without manual configuration or database intervention.

Diagnosis was equally hard: only one source ever reached the driver, and the retrieval failures of the earlier ones were `Warnf`'d and dropped, so nothing identified which source supplied the offending bytes.

## Fix

**New behaviour — retrieve and try each source in order, first accepted wins:**

```text
1. Try options       -> got bytes -> driver rejects  -> skip, try next
2. Try storage       -> got bytes -> driver rejects  -> skip, try next
3. Try config file   -> got bytes -> driver rejects  -> skip, try next
4. Try ledger        -> got bytes -> driver ACCEPTS  -> use these, done

If all 4 produced bytes, none usable  -> error "failed to instantiate token service"
                                         (deliberately NOT ErrTMSNotFound)
If no source produced any bytes       -> ErrTMSNotFound ("TMS not set up yet")
```

For each source, `newTMS` retrieves the public parameters **and immediately tries to instantiate the token service with them**, inside the loop. The first source the driver accepts wins. A source that yields nothing, fails to yield, or yields public parameters the driver rejects falls through to the next one. `loadPublicParams` is gone.

- **Named sources** (`ppSource{name, retrieve}`: `options`, `storage`, `configuration`, `fetcher`) so every failure is attributable; the returned error aggregates each source's failure, tagged with the source name.
- **`ErrTMSNotFound` semantics preserved** — returned if and only if *no* source produced any bytes ("TMS not set up yet"), never when a source produced unusable bytes. It is now decided by an explicit count of the sources that produced bytes. The callers that branch on it (`token/services/network/fabric/endorsement/fsc`) are unaffected.
- **No duplicate work**: public parameters byte-identical to a blob a higher-priority source already tried are not deserialized twice, and the fetcher is still never called when a higher-priority source succeeds.
- **`newTokenService`** converts a panic raised while deserializing stale or attacker-controlled public parameters into an error, and logs the stack at the recover site. Without it, one panicking blob would still shadow the authoritative public parameters — the same failure mode.
- Walking every source reaches two paths a short-circuiting source used to hide: `ppFromConfig` dereferenced a `nil` `driver.Configuration` returned with a `nil` error (the pre-existing test even carried a "To avoid panic in ppFromConfig" comment), and `ppFromStorage` a nil storage. Both are now reported as errors.

## Bonus fix (commit 2) — the `Update` path

> **This is not `main`'s behaviour — the two are not the same.** On `main`, `Update` always won on source 1: `options` is first in the priority order and an update always carries public parameters, so `main` never consulted storage on this path, never fell back, and was never destructive (`service.Done()` sits behind `if err == nil`). The fallback described below was **introduced by commit 1 of this PR**, which routed `Update` through the new walk. Commit 2 removes it. It is a regression caught in review, not a pre-existing bug.

**After commit 1:** `Update(newPP)` -> driver rejects newPP -> falls through to local storage -> builds a service from an older copy -> **reports success**, having already unloaded the service that was running. Reached from the ledger setup listener (`token/services/network/fabric/network.go`), that turned a node which merely needed a binary upgrade into a node silently running on superseded public parameters, with `ManagementServiceProvider.Update` also shutting down the selector manager and clearing its cache on the way. On the update path the fallback cannot even reach the ledger, because the service options an update is built from carry no `PublicParamsFetcher` — so it could only ever land on a stale local copy, the opposite of what the walk is for.

**After commit 2:** `Update(newPP)` -> driver rejects newPP -> **fails, and leaves the running service untouched.** `Update` resolves from source 1 alone (`updatePublicParamsSources`) and accepts only what it was given. No fallback.

Also in commit 2:

- **the panic's stack is kept.** `newTokenService` turns a panic into an error and the walk moves on, so the stack was lost and nothing downstream could say where a driver panicked. It is now logged at the recover site.
- **`ErrTMSNotFound` is decided by a count**, not inferred from `len(instantiationErrs)`. The previous form relied on "a duplicate implies an earlier failed attempt", an invariant nothing enforces.
- **the write to `opts.PublicParams` is gone.** Every caller passes a pointer to its own by-value copy, so the documented post-condition was unobservable.

## Tests

`TestTMSProviderPublicParamsFallback` in `token/core/tms_test.go`, 9 sub-cases:

- fallback to a usable source when `options` / `storage` / `configuration` hold unusable public parameters;
- a driver panic on one source does not abort the walk;
- identical public parameters held by two sources are tried once;
- all sources unusable -> error names every source and is **not** an `ErrTMSNotFound`;
- no source holds any public parameters -> `ErrTMSNotFound`, and every source's reason is reported;
- a `nil` configuration returned with a `nil` error does not panic;
- a failed resolution is not cached, so a later call succeeds once the network serves usable public parameters.

These sub-cases were verified to fail against the pre-fix `newTMS` (four fail outright, the panic case aborts the run) and to pass after it.

For the `Update` path: `TestTMSProviderUpdateRejectsUnusablePublicParams` (the update fails, and never reads storage or queries the network) and `TestTMSProviderUpdateKeepsTheCachedServiceOnFailure` (the running service is neither unloaded nor evicted). These replace `TestTMSProviderUpdateFallback`, which asserted the behaviour commit 1 had introduced.

`TestTMSProviderConcurrentGet` covers 32 concurrent resolvers, asserting the TMS is instantiated once and the network queried once.

## Verification

- `go test ./token/core/` passes, also under `-race`.
- `golangci-lint run token/core/...` -> 0 issues; `make checks` -> exit 0.
- `go test ./token/...` shows one unrelated pre-existing failure, `TestTranslatePath` in `token/services/identity/config`: it asserts the checkout path contains "panurus" and the local worktree is named differently. It fails identically on unmodified `main`.

## Docs

`docs/public_parameters.md` gains a "Resolution Order" section under *Discovery and Fetching*, documenting the four sources, the retrieve-and-try-in-order semantics, and the error/`ErrTMSNotFound` contract — plus a "Resolution vs. update" subsection stating that the walk does not apply to `Update`, and a note that source 4 requires a configured `PublicParamsFetcher`.

## Known limitation (not addressed here)

`GetTokenManagerService` holds the provider-wide write lock across the whole walk, which may now include the ledger fetch, so a hung fetch blocks TMS lookups for every namespace on the node. This is pre-existing — the fetcher was always reachable under that lock — but this change makes reaching it more likely. Fixing it means per-key creation locking, which changes when the post-init callback can run concurrently, so it is tracked separately in #2306 rather than folded in here.

### DIFF
--- a/docs/public_parameters.md
+++ b/docs/public_parameters.md
@@ -49,6 +49,35 @@ When an FSC node starts or connects to a TMS, it must fetch the current public p
 - **Fetching**: The node uses the `PublicParamsFetcher` (provided by the **Network Service**) to query the ledger (via TCC or the Query Service).
 - **Local Cache**: Fetched parameters are cached in the node's local `PublicParametersStorage` to ensure they are available even if the network is temporarily unreachable.
 
+### Resolution Order
+
+`TMSProvider` (`token/core/tms.go`) resolves the public parameters of a TMS from four sources, in priority order:
+
+| # | Source | Description |
+| :--- | :--- | :--- |
+| 1 | `options` | The public parameters passed by the caller in `driver.ServiceOptions.PublicParams` (for example, the new parameters delivered by an update event). |
+| 2 | `storage` | The node's local `PublicParametersStorage`. |
+| 3 | `configuration` | The file referenced by `publicParameters.path` in the TMS configuration. |
+| 4 | `fetcher` | The `PublicParamsFetcher`, which reads the parameters from the ledger. This is the authoritative source. |
+
+For each source, the provider retrieves the parameters **and immediately tries to instantiate the token service with them**. The first source whose parameters the driver accepts wins, and its parameters become the ones the TMS runs with. A source is skipped, and the next one is given its chance, when it:
+
+- holds no public parameters, or fails to return them (storage error, missing file, …); or
+- holds public parameters the driver rejects — driver name/version skew (`invalid identifier, expecting [dlog.v1], got [dlog.v2]`), an unparsable container, an invalid curve id, or any other field-level validation failure; or
+- holds public parameters byte-identical to those a higher-priority source already tried (they are not deserialized twice).
+
+This is what lets a node recover on its own during an upgrade: when the local copy of the parameters is stale or malformed, it does not shadow the authoritative parameters on the ledger — the provider simply falls through to the fetcher. Note that source 4 is only available when the caller supplied a `PublicParamsFetcher` in the service options.
+
+### Resolution vs. update
+
+The walk above applies when a TMS is being **resolved** (`GetTokenManagerService`, `NewTokenManagerService`). It does **not** apply to `Update`, which resolves from source 1 only.
+
+`Update` carries the public parameters the TMS is to adopt — typically the new parameters just committed to the ledger. If the driver rejects them, `Update` fails and the currently cached service is left running and cached, untouched. Falling back to another source would report success while the node kept running on an older copy, after having already unloaded the service it had; and since the service options an update is built from carry no `PublicParamsFetcher`, such a fallback could only ever land on a stale local copy — the opposite of what the walk exists for.
+
+If no source produces usable public parameters, the returned error aggregates the failure of every source, each tagged with the source name, so the offending copy can be identified. The error wraps `ErrTMSNotFound` if, and only if, no source produced any parameters at all, which means the TMS has not been set up yet — as opposed to being set up with parameters this node cannot use.
+
+Note that both resolution and update hold a single provider-wide lock for the whole walk, not one lock per TMS, so a slow source — a driver deserialization or, on the resolution path, the ledger fetch — blocks lookups for every other network/channel/namespace on the node. Tracked in [#2306](https://github.com/LFDT-Panurus/panurus/issues/2306).
+
 ---
 
 ## Dynamic Updates and Synchronization
--- a/token/core/tms.go
+++ b/token/core/tms.go
@@ -110,7 +110,7 @@ func (m *TMSProvider) GetTokenManagerService(opts driver.ServiceOptions) (servic
 	}
 
 	logger.Debugf("creating new token manager service for [%s] with key [%s]", opts, key)
-	service, err = m.getTokenManagerService(opts)
+	service, err = m.getTokenManagerService(opts, m.publicParamsSources())
 	if err != nil {
 		return nil, err
 	}
@@ -130,7 +130,7 @@ func (m *TMSProvider) NewTokenManagerService(opts driver.ServiceOptions) (driver
 	}
 	logger.Debugf("creating new token manager service for [%s]", opts)
 
-	service, err := m.newTMS(&opts)
+	service, err := m.newTMS(&opts, m.publicParamsSources())
 	if err != nil {
 		return nil, err
 	}
@@ -140,6 +140,8 @@ func (m *TMSProvider) NewTokenManagerService(opts driver.ServiceOptions) (driver
 
 // Update updates the public parameters for the specified TMS.
 // If the service is already cached and the public parameters are different, it unloads the old service and reloads it.
+// The TMS is rebuilt from the public parameters carried by the update and from nothing else: if the
+// driver rejects them, Update fails and the currently cached service is left untouched.
 func (m *TMSProvider) Update(opts driver.ServiceOptions) (err error) {
 	if len(opts.Network) == 0 {
 		return errors.Errorf("network not specified")
@@ -171,8 +173,9 @@ func (m *TMSProvider) Update(opts driver.ServiceOptions) (err error) {
 		logger.Debugf("service found, unload token management system for [%s:%s:%s] for key [%s] and reload it", opts.Network, opts.Channel, opts.Namespace, key)
 	}
 
-	// create the service for the new public params
-	newService, err := m.getTokenManagerService(opts)
+	// create the service for the new public params. Only the public params carried by the update
+	// are eligible: see updatePublicParamsSources.
+	newService, err := m.getTokenManagerService(opts, m.updatePublicParamsSources())
 	if err == nil {
 		// unload the old service, if set
 		if service != nil {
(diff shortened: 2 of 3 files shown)


adecaro (commented):
## Review: does this fully address #2281?

@adecaro

**Approve — the fix does exactly what #2281 asks: `newTMS` now walks every public-parameters source and tries to *instantiate* with each one instead of stopping at the first source that merely *retrieved* bytes.**

Checked out `2281_newtms_first_usable_pp` at `fab0226e` and verified:

```
$ go build ./...                                          # clean
$ go vet ./token/core/                                     # clean
$ gofmt -l -s token/core/tms.go token/core/tms_test.go     # clean
$ go test -count=1 ./token/core/                           # ok, coverage: 96.8% of statements
$ go test -race -count=1 ./token/core/...                  # ok (all subpackages)
$ golangci-lint run ./token/core/...                        # 0 issues in changed files
```

I also grepped every caller of `TMSProvider.Update`/`newTMS`/`ErrTMSNotFound` across the repo (`token/services/network/fabric/endorsement/fsc/service.go`, `responder.go`) to confirm none of them pattern-match on the old error text or rely on `newTMS` mutating `opts.PublicParams` — none do, and `opts` is passed by value into the exported methods anyway, so that mutation was never externally visible even pre-PR.

Reverting `token/core/tms.go` to `bdcd99d` (pre-PR) and re-running `TestTMSProviderPublicParamsFallback` fails four of its subtests, and one panics uncaught (`panic: invalid curve id`, since base has no `recover()`), confirming the tests pin the bug rather than the diff. I also checked out the intermediate commit `2c78ccf2f` (commit 1 only, without commit 2's `Update`-specific restriction) and confirmed `TestTMSProviderUpdateRejectsUnusablePublicParams` fails there — commit 1 alone would let `Update` silently fall back to stale storage; commit 2 exists specifically to close that. Both commits are load-bearing and correctly ordered.

`checks`, `utest`, `itest` (base is `main`, so the full matrix ran), `bench-check`, `cgo-check`, CodeQL, and DCO are all green at `fab0226e`.

### What it fixes

| # | Finding | Verified |
|---|---|---|
| 2281.1 | `newTMS` committed to the first source that returned bytes, even if the driver then rejected them, instead of trying the next source | ✅ traced in `newTMS` (`tms.go:279-346`): retrieval and instantiation now happen per-source inside the same loop iteration, and a rejection continues the walk |
| 2281.2 | A stale/malformed local copy (storage or config) could shadow the authoritative network fetcher | ✅ `TestTMSProviderPublicParamsFallback/unusable_storage_public_params_fall_back_to_the_fetcher` + my own revert test above |
| 2281.3 | Failures from skipped sources were not surfaced, making the eventual error unhelpful | ✅ `retrievalErrs`/`instantiationErrs` are aggregated via `errors.Join` and every source's failure is present in the final error (`TestTMSProviderPublicParamsSourceErrors`) |

All three rows check out; no gaps.

### Non-blocking

- **Design note:** `GetTokenManagerService`/`Update` hold the single provider-wide `m.lock` (not one per TMS key) for the entire `newTMS` walk. Before this PR that walk normally did one retrieval; now, in the recovery scenario this fix targets, it can do several driver deserializations plus a network fetch, all while every other network/channel/namespace key is blocked. Not a regression from this PR (the coarse lock predates it) and the trade-off is reasonable for a rare path — flagging in case a future PR wants per-key lock granularity if this walk gets exercised more than occasionally.
- **Nice catch, not a finding:** the PR incidentally fixes a pre-existing typo in `ppFromStorage`'s error string (`"no public parames found"` → `"no public params found"`), and the test suite (priority matrix, concurrency, dedup-by-digest, `Update` keeping the cached service on a rejected update) goes well beyond what #2281 asked for.

### Recommendation

Nothing to change before merge. The Design note above is worth a one-line mention in `docs/public_parameters.md`'s existing "Resolution vs. update" section if you want to preempt the question later, but it's not blocking.

adecaro (approved):
LGTM

PR #2284, opened 2026-08-20
fix(zkatdlog): upgrade dlog tokens whose format changed

### LINKED ISSUE #2282: zkatdlog cannot upgrade dlog tokens whose format changed
### Original description (by Angelo)

> zkatdlog does not support token upgrade via issuer for dlog tokens with a different format (because for example the pedersen commitments changed). The issuer should be able to open these commitments with the old public parameters. Alice could send: commitment, opening, and pp hash. The issuer checks if it still has pp hash, if yes, it tries to open the commitment with the pedersen bases in the referenced public params.

### Symptom

An upgrade regenerated the public parameters, changing the format identifier of older tokens. Transfers skip them, the balance still counts them; the tokens are safe and unspent, just unusable.

### Cause

`SupportedTokenFormat` (`zkatdlog/nogh/v1/token/service.go:283`) hashes `pp.PedersenGenerators` into the format, so new generators change every token's format. Both upgrade paths are fabtoken-only: local reinterpretation rejects the unknown format (`service.go:181`), and the issuer path admits only fabtoken formats (`crypto/upgrade/service.go:79-88`) and expects cleartext, not a commitment.

### What needs to happen

1. Upgrade dlog tokens via the issuer protocol: commitment + opening + pp hash, opened with the referenced pp's bases.
2. Trace a token back to the pp that created it — the request store already records a pp hash per txID.
3. Investigate the pp upgrade scenarios end to end, including the restart failure on the affected database.

AkramBitar:
Fixes #2282

## Problem

When public parameters are regenerated, every old token gets a format the driver no longer recognises. It can't be transferred, redeemed, or upgraded — permanently stuck on the ledger.

A zkatdlog token's `Format` is a digest **over the Pedersen generators**. So:

1. Public parameters are regenerated with different generators (or a new curve).
2. Every token created before gets a format the driver no longer lists in `SupportedTokenFormats()`.
3. Transfers skip those tokens; the balance still counts them.
4. Neither upgrade path can rescue them — local reinterpretation rejects the unknown format, and the issuer path accepted only fabtoken formats and expected cleartext, not a commitment.

Result: safe, unspent, permanently unusable tokens.

## Fix

The owner tells the issuer which generation of public parameters created the token. The issuer verifies this, opens the commitment with the old Pedersen bases to recover the type and value, then atomically burns the old token and issues a new one under the current format. After that the token is spendable again.

Extend the existing issuer-mediated "burn and re-issue" flow to commitment tokens:

1. **Owner** lists its unsupported tokens — it already holds the commitment and its opening.
2. **Owner** resolves which generation of public parameters produced that format and puts its hash in the proof (`Proof.PublicParamsHashes`).
3. **Issuer** loads those parameters by hash from its own store.
4. **Issuer** re-checks they actually generate the format the token carries on the ledger — the hash is a lookup hint, never a trusted claim, so a peer cannot pick the bases used.
5. **Issuer** opens the commitment with those bases, recovers type + value, and re-issues; the ledger burns the old inputs atomically.

Supporting changes:

- The public parameters table was already a full history (`StorePublicParams` never overwrites) — exposed via `TokenStore.PublicParamsHashes` and `driver.QueryEngine`.
- New `upgrade.PublicParamsHistory` resolves format → generation (`ByFormat`) and verifies hash + format (`ByHashAndFormat`), with caching.
- Formats the driver already supports directly are now always rejected for upgrade, whichever family they belong to.

No ledger change and no storage schema change.

## Notes

- Issuers **must retain historical public parameters**; an issuer that dropped them cannot open the commitment and refuses the upgrade. Documented in `docs/upgradability.md`.
- Two items from the issue are deliberately **not** in scope: the restart failure on the affected database (not reproducible without that database) and the balance counting unspendable tokens (a separate, user-visible decision).

## Testing

- New unit tests: round trip, issuer dropped the parameters, proof tampering (wrong hash / no hash / bad opening), already-supported format, mixed fabtoken + dlog batch, resolver + caching + error paths.
- Fuzz seeds extended with a proof carrying pp hashes.
- `make checks` and `make lint-auto-fix` clean.


adecaro (commented):
## Review: fix(zkatdlog) — upgrade dlog tokens whose format changed

Went through this end-to-end: read `pphistory.go`, `service.go`, `proof.go`, the `driver.go` wiring, the doc changes, and the new tests in full; ran `go build ./...`, the touched packages' unit tests, and cross-checked the fuzz-CI wiring, mock freshness, call-site consistency, and the error-wrap chain the fallback logic depends on. No Blocking, Should-fix, or Design-note items.

### Acceptance criteria (#2282)

| # | Requirement | Verdict |
|---|---|---|
| 1 | Upgrade dlog tokens via issuer protocol (commitment + opening + pp hash) | ✅ `upgrade.Service.processCommToken`/`processTokensWith`. The pp hash is only ever a lookup hint — the format is re-verified against the ledger-recorded value before the commitment is ever opened, so a peer can't steer which Pedersen bases get used. |
| 2 | Trace a token back to the pp generation that created it | ✅ `PublicParamsHistory.ByFormat`/`ByHashAndFormat`, backed by the pre-existing full-history `PublicParamsHashes`/`PublicParamsByHash` (no schema change needed). |
| 3 | Investigate pp upgrade scenarios end-to-end, incl. the restart failure on the affected DB | ⚠️ Deliberately, transparently partial — the PR body discloses the original restart failure isn't reproducible without that specific database. In its place, `docs/upgradability.md` gets a full recovery runbook for the general class of restart/rebuild-after-regeneration failures, with the exact error strings and both owner- and issuer-side remediation steps. Reasonable scope call, clearly stated, not hidden. |

### Verification

- `go build ./...` — clean
- `go test ./token/core/zkatdlog/nogh/v1/crypto/upgrade/... ./token/core/zkatdlog/nogh/v1/token/... ./token/services/storage/db/sql/sqlite/...` — all PASS
- Broader `./token/...` suite — PASS (only failures are `sql/postgres`, which needs a local Docker daemon I don't have — unrelated to this diff)
- `FuzzProcessCommTokenNoPanic` correctly wired into `nightly-fuzz.yml`'s matrix, seeded with an adversarial cross-curve case tied to the historical `processCommToken` panic
- `errors.Is`/`Wrapf` (fabric-smart-client) delegate to `cockroachdb/errors` and preserve the `Unwrap()` chain, so the `ErrPublicParamsNotFound`-triggered fallback in `processTokensWith` behaves as documented
- Repo-wide grep confirms every caller of `upgrade.NewService` was updated for the new 6-arg signature; mocks (`mock/pps.go`, `driver/mock/qe.go`, `services/tokens/mock/*`) match their interfaces
- No new `fmt`/`fmt.Errorf` usage introduced (AGENTS.md ban)
- `MaxUpgradeRequestTokens` (256) matches `driver.DefaultResourceLimits().MaxInputs`, as the doc claims

### Overall

Solid, well-scoped fix. The commitment-upgrade path has unusually thorough coverage: round trip, pruned-issuer-history, proof tampering (bad hash / no hash / bad opening), already-supported formats, mixed fabtoken+commitment batches, spurious hashes on fabtoken entries, owner-side pruned history, and a concurrency test on the history cache. No ledger or storage-schema changes required. Nothing here blocks merge from my side.

adecaro (approved):
LGTM

PR #2286, opened 2026-08-21
Close the write channel and synchronise localSession state

### LINKED ISSUE #2133: Fix Missing Channel Closure in `localSession.Close()` [mem]
**Document severity:** MEDIUM · **Verified priority:** Medium · **Effort:** Medium · **Type:** Bug

### Old issue reference

> **Sub-Task 12 — Fix Missing Channel Closure in `localSession.Close()`**
>
> **Status:** `[ ] pending`

| 12 | `[RUN]` | **MEDIUM** | Channel leak | `ttx/session.go` | 30–31, 133–135 | `Close()` only sets flag; channels never closed |

### Intent

`localSession.Close()` only sets `s.info.Closed = true` but leaves the
`writeChannel` and `readChannel` open. Any code that sends to a closed session
(e.g., via `SendWithContext`) will block on `s.writeChannel <- msg` forever
since `send()` checks the `Closed` flag _before_ the send but the check and
the send are not atomic.

In practice this can cause a sender goroutine to be permanently blocked if the
receiver side has called `Close()` concurrently or just before the send.

### Expected Outcomes

- After `Close()` is called, send operations return immediately with an error
  instead of blocking.
- Buffered messages already in the channels are drained so goroutines waiting
  on receive also unblock.

### Todo List

1. Add a `closeOnce sync.Once` field to `localSession`.
2. In `Close()`, use `closeOnce.Do(...)` to close `writeChannel` (which
   will cause any goroutine blocked on send to receive a panic, or — preferred
   — use the `Closed` flag to gate sends and just close `readChannel` to
   unblock receivers).
3. Update `send()` to hold a read lock or use an atomic check so the
   closed-flag check and channel send are safe from races.
4. Consider using a `context.Context` tied to the session lifetime instead
   of a flag, so all `send/receive` calls respect the same cancellation signal.

### Relevant Context

- File: `token/services/ttx/session.go`
- Lines 106–135: `send()` and `Close()` methods.
- Lines 30–31: channel creation.

---

### Status in current code

**Confirmed — still present.** Two defects here, not one: the missing channel closure the document describes, plus an unsynchronised `info.Closed` that is a data race in its own right.

All references confirmed at HEAD in `token/services/ttx/session.go`:

- `:30-31` — both channels created buffered with capacity 10:
  `lr := make(chan *view.Message, 10)` and `rl := make(chan *view.Message, 10)`, wired into
  the paired sessions at `:43` and `:52`.
- `:133-134` — `func (s *localSession) Close()` whose entire body is `s.info.Closed = true`.
  Neither channel is closed.
- `:107` and `:126` — `send()` and `Receive()` each test `if s.info.Closed` before touching
  the channel.

**Defect 1 — non-atomic check-then-send (as documented).** The `Closed` test at `:107` and
the subsequent `s.writeChannel <- msg` are separate operations. A concurrent `Close()` landing
between them leaves the sender committed to a send on a channel nobody will drain. The
capacity-10 buffer masks this for the first ten messages, which makes it load-dependent and
intermittent rather than harmless — it will surface under exactly the traffic where it is
hardest to debug.

**Defect 2 — data race on `info.Closed`, which the document does not mention.** `Closed` is a
plain `bool` written at `:134` and read at `:107` and `:126` from different goroutines with no
mutex, no atomic, and no channel handoff. That is a data race by the Go memory model
independent of whether a send ever blocks: `go test -race` should flag it, and the compiler is
free to keep the read hoisted so a closed session may never be observed as closed. Todo item 3
gestures at this ("hold a read lock or use an atomic check") but the document frames it as a
consequence of the leak rather than as its own bug.

**On the proposed fixes.** Todo item 2's first option — closing `writeChannel` — would make a
blocked or subsequent sender panic on send to a closed channel, converting a hang into a crash.
The document flags this itself and marks the second option preferred. Agreed: gate sends on
state and close only the receive side, or adopt Todo item 4 and give the session a
`context.Context` so both directions observe one cancellation signal. Item 4 is the cleanest
end state and is what the effort estimate below assumes.

Effort **Medium** rather than Low: unlike the other confirmed items this is not a local patch.
Getting it right means reworking how session state is represented across `send()`, `Receive()`
and `Close()`, and the fix needs a race-detector test to demonstrate it, since the current
behaviour is timing-dependent.

### Severity and priority

The source document rated this **MEDIUM**. Re-checked against this repository at
`origin/main` with `go 1.26.5` on 2026-08-04, the priority is **Medium** for the

### Effort

**Medium**

srikary12:
## Summary

`localSession.Close()` only set `info.Closed` and left both channels open, and `info.Closed`
itself was an unsynchronised `bool` — written by `Close()` while `send()` and `Receive()` read
it from other goroutines. That is a data race under the Go memory model regardless of whether a
send ever blocks, and the non-atomic check-then-send meant a `Close()` landing between the
`Closed` test and `s.writeChannel <- msg` left the sender committed to a send nobody would drain
(masked for the first ten messages by the capacity-10 buffer).

## Changes

- **Guard `info.Closed` with a `sync.RWMutex`.** `Close()` takes the write lock; `send()`,
  `Receive()` and `Info()` take the read lock.
- **`Close()` now closes `writeChannel`,** which unblocks a peer parked on `Receive()` while
  leaving already-buffered messages deliverable. The issue flagged that closing `writeChannel`
  could panic a sender — that cannot happen here, because `send()` holds the read lock across
  the channel send, so `Close()` cannot interleave between the closed check and the send.
- **`Close()` is idempotent** — a second call returns early instead of double-closing.
- **`send()` no longer blocks forever on a full buffer.** It tries a non-blocking send first,
  then blocks until either the buffer drains or the caller's context is done.

Each side closes only its own `writeChannel`, so closing one end leaves the peer able to keep
sending on its own direction.

## Tests

Eight tests in `session_test.go` covering: idempotent close, closing both sides, peer sends
after a one-sided close, `Close()` unblocking a peer's `Receive()`, buffered messages surviving
close, a full-buffer send returning on context cancellation, and two race-detector tests for
concurrent send/close and concurrent `Info()`/`Close()`.

`go test ./token/services/ttx/... -race` passes clean.

> **Note:** the tests in this PR were generated by Claude; the `session.go` change is hand-written.

Fixes #2133


token/services/ttx/session.go
```diff
@@ -105,11 +110,14 @@ func (s *localSession) SendErrorWithContext(ctx context.Context, payload []byte)
 }
 
 func (s *localSession) send(ctx context.Context, payload []byte, status int32) error {
+	s.mu.RLock()
```
AkramBitar: **Blocker: `Close()` can now deadlock — this is a regression.**

`send()` holds `s.mu.RLock()` (line 113, `defer`red to function exit) across the blocking `select` on `s.writeChannel` at line 138. So whenever a sender is parked on a full buffer (cap 10), `Close()` blocks forever on `s.mu.Lock()`. Worse, Go's `RWMutex` queues new readers behind a pending writer, so once `Close()` is waiting, `Info()` and `Receive()` block too.

Pre-PR `Close()` never blocked, so this is new behaviour, not a pre-existing wart. The blocking `select` only escapes via `ctx.Done()`, and the `Send`/`SendError` convenience wrappers pass `s.ctx` — the context handed to `NewLocalBidirectionalChannel` — which in the self-audit path has no deadline of its own. Concretely, the deferred `cleanupSessions` in `collectendorsements.go` hangs while still holding `c.sessionsMu`, and `CollectEndorsementsView` never returns.

The mutex can't simply be dropped either: `close(s.writeChannel)` at line 164 means a sender selecting on that channel concurrently with the close would panic. Closing the channel and keeping `Close()` non-blocking needs the two to be separated — a `done` channel to release parked senders, and the actual `close()` deferred until in-flight senders have provably left:

```go
type localSession struct {
    // ...
    mu      sync.RWMutex   // guards info only, never held across a send
    done    chan struct{}  // closed by Close()
    senders sync.WaitGroup // in-flight send() calls
}

func (s *localSession) send(ctx context.Context, payload []byte, status int32) error {
    s.mu.RLock()
    if s.info.Closed {
        s.mu.RUnlock()
        return errors.New("session is closed")
    }
    s.senders.Add(1)
    s.mu.RUnlock()          // released before the send
    defer s.senders.Done()

    msg := &view.Message{ /* ... */ }
    select {
    case s.writeChannel <- msg:
        return nil
    case <-s.done:
        return errors.New("session is closed")
    case <-ctx.Done():
        return errors.Wrapf(ctx.Err(), "failed sending message on session [%s]", s.info.ID)
    }
}

func (s *localSession) Close() {
    s.mu.Lock()
    if s.info.Closed {
        s.mu.Unlock()
        return
    }
    s.info.Closed = true
    close(s.done)           // parked senders bail out immediately
    s.mu.Unlock()

    s.senders.Wait()        // no lock held; returns as soon as they observe done
    close(s.writeChannel)   // provably no sender in the select now
}
```

`done` needs initialising in `NewLocalBidirectionalChannel` (one per side, not shared). This keeps `Close()` bounded, keeps the channel close that fixes #2133, and removes the panic window without a lock around the send.

token/services/ttx/session.go
```diff
@@ -132,5 +154,12 @@ func (s *localSession) Receive() <-chan *view.Message {
...
 
 func (s *localSession) Close() {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
+	if s.info.Closed {
+		return
+	}
 	s.info.Closed = true
+	close(s.writeChannel)
```
AkramBitar: **One-sided close leaves the peer writing into a channel nobody reads.**

`left` and `right` each hold their own *copy* of `info` (it's a `view.SessionInfo` value, assigned to both in the constructor), and `left.writeChannel` is `lr` while `right.writeChannel` is `rl`. So after `left.Close()`:

- `right.Info().Closed` is still `false`;
- `right.Send()` returns `nil` for 10 more messages, which land in `rl` — but `left` is closed, so `left.Receive()` returns `nil` and nothing ever drains `rl`;
- the 11th `right.Send()` parks on the full buffer, and with finding 1 above that wedges `right.Close()` permanently.

`TestLocalBidirectionalChannel_ClosePeerSendStillSafe` (line 486) currently asserts that silently-dropped send as intended behaviour. Two coherent options — worth picking deliberately:

1. **Close means the conversation is over** (my preference, and closer to what a real session close does): hoist the closed state to the shared `LocalBidirectionalChannel`, or have `Close()` mark both sides, so `right.Send()` returns `"session is closed"` instead of dropping the message on the floor.
2. **Keep it one-sided**, but then the test should assert the send is lossy and say so in a comment, because a caller reading `Send() == nil` as "delivered" is going to be wrong.

token/services/ttx/session.go
```diff
@@ -132,5 +154,12 @@ func (s *localSession) Receive() <-chan *view.Message {
 }
 
 func (s *localSession) Close() {
```
AkramBitar, line 230: Minor, and arguably out of scope — but while you're here: `Close()` closes `s.writeChannel`, which is the *peer's* `readChannel`, so it only wakes a reader blocked on the other side. A goroutine already parked on **this** session's `readChannel` isn't released, and `ReceiveRawWithTimeout` (`token/services/utils/session/session.go:82`) captures `j.s.Receive()` before it selects, so it doesn't re-check `Closed` either. It waits out its full timeout — up to `time.Minute * 4` via `ReceiveTransactionView` (`receivetx.go:42`) — after a local close.

Fixing both sides falls out naturally from option 1 in my comment on line 164.

token/services/ttx/session_test.go
```diff
@@ -442,3 +443,214 @@ func TestLocalBidirectionalChannel_UniqueSessionIDs(t *testing.T) {
...
+// TestLocalBidirectionalChannel_ConcurrentSendAndClose verifies that Close racing with concurrent sends neither panics nor races.
+func TestLocalBidirectionalChannel_ConcurrentSendAndClose(t *testing.T) {
+	ctx := t.Context()
+	channel, err := ttx.NewLocalBidirectionalChannel(ctx, "caller", "ctx-id", "endpoint", []byte("pkid"))
+	require.NoError(t, err)
+
+	leftSession := channel.LeftSession()
+	rightSession := channel.RightSession()
+
+	// Drain the peer so senders are never blocked on a full buffer.
```
AkramBitar, line 600: The suite can't catch the `Close()` deadlock, because the two tests that come closest each avoid the other half of it:

- here, `ConcurrentSendAndClose` deliberately drains the peer (`// Drain the peer so senders are never blocked on a full buffer`), so no sender ever parks and `Close()` always gets the lock;
- `SendFullBufferContextDone` (line 542) does park a sender on a full buffer, but never calls `Close()`.

The missing case is both at once. Something like:

```go
func TestLocalBidirectionalChannel_CloseWithSenderBlockedOnFullBuffer(t *testing.T) {
    channel, err := ttx.NewLocalBidirectionalChannel(t.Context(), "caller", "ctx-id", "endpoint", []byte("pkid"))
    require.NoError(t, err)
    left := channel.LeftSession()

    for range 10 { // fill the buffer; nothing reads the right side
        require.NoError(t, left.Send([]byte("fill")))
    }
    blocked := make(chan struct{})
    go func() { defer close(blocked); _ = left.Send([]byte("parks here")) }()

    closed := make(chan struct{})
    go func() { defer close(closed); left.Close() }()

    select {
    case <-closed:
    case <-time.After(2 * time.Second):
        t.Fatal("Close blocked behind a sender parked on a full buffer")
    }
    <-blocked // the parked sender must be released too
}
```

On the current branch this hangs on both selects; with the `done`-channel shape it passes. Would be good to keep it in the suite as a regression guard.

AkramBitar (commented):
@srikary12 , 

Thanks for looking into this.

The diagnosis is right: the old code had both a data race on `info.Closed` and a check-then-send hole that could panic with *send on closed channel*, and the mutex does close the panic window. But holding the read lock across the blocking channel send trades that benign race for a hard deadlock, so this needs one more round before it can go in.

I verified each point below against `d15f816` directly (channel wiring in `NewLocalBidirectionalChannel`, `ReceiveRawWithTimeout` in `token/services/utils/session/session.go`, and the 4-minute default in `receivetx.go:42`).

Worth flagging up front: the branch's own tests pass and `go test ./token/services/ttx/ -race` is green, so CI being clean isn't evidence against finding 1 — the deadlock sits just outside what the new suite exercises (see the comment on `session_test.go`).

For reference, FSC's own copy of this type (`platform/view/services/comm/session/bidi.go` in v0.17.0) uses `atomic.Bool` and never takes a lock around the send — but it also never closes `writeChannel`, which is precisely the gap you're trying to fix, so it isn't a drop-in model for the whole design. The sketch in finding 1 is what it takes to close the channel *and* keep `Close()` non-blocking.

AkramBitar (changes requested):

token/services/ttx/session.go
srikary12: Fixed, with one deviation from the sketch.

mu now guards only the closed flag and is never held across a send: beginSend registers the in-flight send under the read lock and releases it before the select; Close marks closed, closes done, drops the lock, waits for the in-flight senders, and only then closes the message channels. So Close is bounded, and no sender can still be selecting on a channel being closed.

The deviation: done is shared by both ends rather than one per side. Once the closed state is shared (option 1 on your other thread), a per-side done would leave the peer's parked sender stuck. Same reasoning for the senders WaitGroup — the close waits for in-flight sends in both directions before closing either channel.

token/services/ttx/session.go
srikary12: Close closes both message channels, so a goroutine parked on this session's own readChannel is released too. One thing that didn't fall out on its own — Receive() no longer returns nil when closed. It returns the closed channel, so a receive started after the close gets a nil message immediately rather than blocking forever on a nil channel.

token/services/ttx/session.go
srikary12: Went with option 1. The closed flag, the done channel, the in-flight sender count and both message channels now live in one channelState that both ends point at. After left.Close(), right.Info().Closed is true and right.Send returns session is closed instead of dropping the message into rl. Info() derives Closed from the shared state, so each session's info copy is immutable now.

token/services/ttx/session_test.go
srikary12: Added tests for the same

Effi-S (approved):
Hi @srikary12, 
Thank you for your work here.
LGTM.
Can you please squash and rebase?

AkramBitar:
Hello @srikary12 

Thanks a lot for all the effort you put into this PR. I think it’s ready to be merged. Could you please resolve the conflict and, once that’s done, make sure all the commits are squashed into a single commit?

Regards,
Akram

Effi-S:
@srikary12, Please squash to a single commit

srikary12:
Done

AkramBitar (approved):

AkramBitar:
## Review

The synchronisation design is sound — shared `channelState` with `RWMutex` + `WaitGroup` + `done` channel correctly prevents both "send to closed channel" panics and "close blocking behind full buffer". Idempotent close via `finished` channel is correct. Tests are comprehensive and cover all the interesting race scenarios.

Two things to confirm before merge:

**1. `wg.Go` requires Go 1.22**
`sync.WaitGroup.Go` was added in Go 1.22. The tests use it in `TestLocalBidirectionalChannel_ConcurrentSendAndClose` and `TestLocalBidirectionalChannel_ConcurrentCloseBothSides`. Please verify the module's `go` directive in `go.mod` is at least 1.22.

**2. New test imports — verify no new module deps**
`TestLocalBidirectionalChannel_CloseEndsReceiveWithTimeout` imports `token/services/utils/json/session` and `token/services/utils/session`. Please confirm these are already transitive dependencies of the `ttx` package and do not introduce new module-level deps.

Otherwise ready to merge.

AkramBitar (changes requested):
@srikary12, 

Thanks a lot. I did additional review round. Please see my comments.

Regards,
Akram

PR #2287, opened 2026-08-21
fix(recovery): bound per-transaction recovery with a timeout

### LINKED ISSUE #2192: Recovery sweep has no per-transaction timeout, so one hung status query stalls every replica
## Problem

The recovery manager builds its context once at startup, with no deadline
(`token/services/storage/services/recovery/manager.go:114`):

```go
m.ctx, m.cancel = context.WithCancel(context.Background())
```

That context is passed straight through `runSweep(m.ctx)` into leadership acquisition,
`ClaimPendingTransactions`, and every per-transaction `Recover` call. Nothing on that path
narrows it, so the only thing that can ever cancel it is `Stop()`.

`Recover` is a network call. `TTXRecoveryHandler.Recover` reaches
`h.network.GetTransactionStatus(ctx, ...)` at `token/services/ttx/finality/recovery.go:76`,
which queries the ledger for the transaction's status. If that query hangs rather than
failing, an unresponsive peer, a connection that dies without surfacing an error, an orderer
in catch-up, then the worker blocks for as long as the query does.

## Why it matters

The blast radius is not one transaction, because leadership is held across the whole sweep.
`runSweep` (`manager.go:216`) acquires the advisory lock, defers its release, and only then
calls `recoverTransactions`, which both claims and processes the batch. So while the leader
is parked on a hung query:

- its own sweep makes no progress
- every other replica's `AcquireRecoveryLeadership` returns `acquired=false` and skips its tick
- no pending transaction on that TMS is recovered by anyone, for the duration of the hang

Recovery is the mechanism that repairs transactions whose finality listener was lost. One
unresponsive peer can therefore disable it across every replica of a TMS, quietly, with
nothing timing out to break it.

There is also nothing an operator can configure to bound this. `validateConfig`
(`manager.go:199`) only checks that `TTL`, `ScanInterval`, `BatchSize`, `WorkerCount` and
`LeaseDuration` are greater than zero. There is no per-transaction or per-sweep deadline.

## Related, worth confirming at the same time

`LeaseDuration` (default 30s) can expire while the leader is still processing, since nothing
renews it. Today that is masked, because leadership serialises the sweep so no other replica
is claiming while one is at work. It stops being masked if sweeps ever run concurrently on a
path where leadership is not enforced. If a renewal mechanism is wanted, there is already a
precedent in `auditdb/locker/postgres`, whose `heartbeatLoop`
(`token/services/storage/auditdb/locker/postgres/postgres.go:304`) renews at roughly TTL/3.

atharrva01:
The recovery sweep's context had no deadline, so a hung GetTransactionStatus call could block a worker forever. Since leadership is held for the whole sweep, that stalls recovery on every replica of the TMS, not just the one that hit the slow peer.

This adds a `TransactionTimeout` config field (default 10s, validated like the other durations) and wraps the handler's `Recover` call in `context.WithTimeout` at the per-transaction call site. Claim release and orphan promotion still use the sweep's own context, so a timed-out attempt is cleaned up normally instead of being cut off itself.

Added a test where the mock handler actually honors ctx.Done() (a real hang simulation) and asserts leadership is released and reacquired rather than held for the duration of the hang.

The LeaseDuration renewal question raised in the same issue is separate and not addressed here.

Fixes #2192

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -240,23 +240,59 @@ token:
               # Performance impact: Each scan queries the transaction database for pending transactions.
               scanInterval: 5s
               
-              # batchSize is the maximum number of pending transactions claimed per scan. Default: 100.
+              # batchSize is the maximum number of pending transactions claimed per scan. Default: 16.
               # Limits the number of transactions processed in a single recovery sweep to prevent
               # overwhelming the system. Increase for high-throughput environments with many pending transactions.
               # Performance impact: Larger batches reduce scan overhead but increase memory usage and processing time per sweep.
-              batchSize: 100
+              batchSize: 16
               
-              # workerCount is the number of local workers that process claimed transactions in parallel. Default: 4.
+              # workerCount is the number of local workers that process claimed transactions in parallel. Default: 8.
               # Increase to improve recovery throughput in high-volume scenarios.
               # Decrease to reduce resource consumption on constrained systems.
               # Performance impact: More workers increase CPU and network utilization but improve recovery speed.
-              workerCount: 4
+              workerCount: 8
               
-              # leaseDuration is how long a claimed transaction remains leased to this instance before it can be reclaimed. Default: 30s.
+              # leaseDuration is how long a claimed transaction remains leased to this instance before it can be reclaimed. Default: 5m.
               # This prevents stuck transactions from blocking recovery indefinitely if a worker crashes.
               # Should be longer than the typical time to query and process a single transaction.
               # Relationship: Should be greater than the expected network latency + processing time for transaction status queries.
-              leaseDuration: 30s
+              leaseDuration: 5m
+
+              # transactionTimeout bounds a single transaction's recovery attempt: the
+              # handler's Recover call, which queries the ledger for status and applies
+              # finality logic. Default: 60s. If set explicitly (and non-zero), rejected
+              # below a 10s floor: a shorter deadline risks abandoning recoveries that were
+              # merely slow, not genuinely stuck.
+              # Without this, a hung status query blocks the worker indefinitely, and this
+              # instance holds recovery leadership for as long as anything is abandoned in
+              # the background (see the leadership note further down), so a permanently
+              # hung status query stalls recovery on every replica.
+              # Relationship: workerCount workers can each be stuck for the full
+              # transactionTimeout at once, so size leaseDuration against
+              # leaseDuration > (batchSize / workerCount) x transactionTimeout. Also keep
+              # scanInterval comfortably below leaseDuration: a stuck transaction's claim
+              # is kept alive only by this instance's own next sweep reclaiming it, so a
+              # scanInterval close to or above leaseDuration risks the claim lapsing
+              # between sweeps. The manager Warns at startup (does not fail to start) when
+              # either does not hold: an attempt still in flight when this happens keeps
+              # its claim rather than releasing it, so this is a throughput concern, not a
+              # correctness one.
+              # Set to 0 to disable the deadline entirely. Do this only if you accept the
+              # tradeoff: a Recover call that ignores its context can then block Stop()
+              # indefinitely, since a deadline is the only thing that can interrupt a call
+              # the callee itself never reads.
+              transactionTimeout: 60s
+
+              # stuckTransactionAlertThreshold escalates the per-attempt failure log from
+              # Warn to Error once the same transaction has hit transactionTimeout this many
+              # consecutive times, so a persistently-unresponsive peer is visible to alerting
+              # instead of blending into routine sweep failures. Default: 5.
+              # Log-severity only: it never changes the transaction's status, and it never
+              # changes when the manager skips a re-attempt (that happens independently,
+              # whenever an earlier attempt for the same tx hasn't returned yet). Past the
+              # threshold, the log backs off by doubling rather than firing every sweep.
+              # Set to 0 to disable the escalation.
+              stuckTransactionAlertThreshold: 5
               
               # instanceID identifies this replica as the owner of recovery claims.
               # If empty, a process-local identifier is generated automatically at startup using a UUID.
@@ -643,11 +679,13 @@ token:
               enabled: true
               ttl: 30s
               scanInterval: 5s
-              batchSize: 100
-              workerCount: 4
-              leaseDuration: 30s
+              batchSize: 16
+              workerCount: 8
+              leaseDuration: 5m
+              transactionTimeout: 60s
               instanceID:
               notFoundGracePeriod: 30m
+              stuckTransactionAlertThreshold: 5
 ```
 
 Default values:
@@ -655,27 +693,32 @@ Default values:
 - enabled: true
 - ttl: 30s
 - scanInterval: 5s
-- batchSize: 100
-- workerCount: 4
-- leaseDuration: 30s
+- batchSize: 16
+- workerCount: 8
+- leaseDuration: 5m
+- transactionTimeout: 60s (set to 0 to disable the per-transaction deadline; an explicit non-zero value below 10s is rejected)
 - instanceID: empty, auto-generated when the recovery manager starts
 - notFoundGracePeriod: 30m (set to 0 to disable promotion to Orphan)
+- stuckTransactionAlertThreshold: 5 (set to 0 to disable the log escalation)
 
 **Parameter Relationships and Tuning:**
 
 - **Recovery is enabled by default** to ensure automatic recovery of lost finality listeners.
 - **Only pending transactions older than `ttl` are considered for recovery** to avoid interfering with active transaction processing.
-- **The manager validates** that `ttl`, `scanInterval`, `batchSize`, `workerCount`, and `leaseDuration` are all greater than zero.
-- **Recovery leadership** uses a PostgreSQL advisory lock so that only one replica performs a recovery sweep at a time, on the owner transaction path. The lock identifier is no longer configurable: it is derived from the TMS's own `requests` table name, which already carries the network, channel and namespace. That makes it unique per TMS automatically, so several TMSes sharing one PostgreSQL database no longer compete for a single lock, and there is nothing left for an operator to keep unique by hand. The audit transaction store does not yet participate in this coordination — its recovery leader factory is nil, so every replica is granted leadership and the audit sweep runs redundantly on each of them; see [#2143](https://github.com/LFDT-Panurus/panurus/issues/2143).
+- **The manager validates** that `ttl`, `scanInterval`, `batchSize`, and `workerCount` are all greater than zero, that `leaseDuration` is greater than zero, and that `transactionTimeout` is not negative (`0` means unbounded; any positive value is accepted without an upper bound, though loading the configuration separately rejects an explicit non-zero value below a 10s floor, see the next bullet). It additionally **Warns at startup, without failing to start,** when `leaseDuration > (batchSize / workerCount) x transactionTimeout` or `scanInterval < leaseDuration` do not hold, see the next bullet.
+- **Recovery leadership** uses a PostgreSQL advisory lock so that only one replica performs a recovery sweep at a time, on the owner transaction path. The lock identifier is no longer configurable: it is derived from the TMS's own `requests` table name, which already carries the network, channel and namespace. That makes it unique per TMS automatically, so several TMSes sharing one PostgreSQL database no longer compete for a single lock, and there is nothing left for an operator to keep unique by hand. Leadership is normally released at the end of each sweep, but held across sweeps for as long as any transaction is still abandoned in the background from a previous sweep (see `transactionTimeout` below), so this instance keeps winning it — and thereby keeps its own reclaim renewing that transaction's lease — instead of leaving that to chance; `Stop()` always releases it regardless. The audit transaction store does not yet participate in this coordination — its recovery leader factory is nil, so every replica is granted leadership and the audit sweep runs redundantly on each of them; see [#2143](https://github.com/LFDT-Panurus/panurus/issues/2143).
+- **`transactionTimeout`** bounds a single `Handler.Recover()` call, not the whole sweep. The call runs in its own goroutine so the worker can give up on the deadline even if the call itself never honours the context. A per-transaction bound does not bound the sweep: a batch of `batchSize` claims split across `workerCount` workers can take up to `(batchSize / workerCount) × transactionTimeout` worst case, so the invariant that actually matters is `leaseDuration > (batchSize / workerCount) × transactionTimeout`. With the shipped defaults that worst case is `2 × 60s = 120s`, comfortably inside the default 5m `leaseDuration`. Separately, `scanInterval` should stay comfortably below `leaseDuration`: a stuck transaction's claim is kept alive only by this instance's own next sweep reclaiming the still-`Pending` row (which refreshes the claim as a side effect), and this instance holds recovery leadership across sweeps for as long as any transaction stays abandoned specifically so that reclaim keeps happening on this instance rather than depending on it winning leadership again by chance, but a `scanInterval` close to or above `leaseDuration` still risks that renewal arriving too late between this instance's own sweeps. If you only set `leaseDuration` and leave `transactionTimeout` unset, the manager automatically lowers the 60s default below whatever `leaseDuration` you configured (floored at 10s) rather than leaving it at a disproportionate default. An explicit, non-zero `transactionTimeout` below 10s is rejected when the configuration is loaded: a deadline that tight is likely to abandon recoveries that were merely slow on a loaded network, not genuinely stuck. `transactionTimeout: 0` disables the deadline entirely, but `Stop()` does **not** bound a ctx-blind call the way the deadline does; disabling it can leave `Stop()` blocked indefinitely on a hung `Recover` call (the manager logs a Warn at startup when this is set). Note the bound stops at `Handler.Recover()`: `ClaimPendingTransactions`, `ReleaseRecoveryClaim`, and `SetStatus` still run on the sweep's own context, which carries no deadline, so a wedged database connection in any of those still blocks the worker and, since leadership is held for as long as anything is abandoned, every replica of the TMS in the meantime, the same stall `transactionTimeout` was added to fix for the ledger side. `transactionTimeout` is a bound on the handler call only, not a general no-stall guarantee for the sweep. A `recoverCtx.Done()` firing because `Stop()` cancelled the manager, rather than because `transactionTimeout` actually elapsed, is not counted or alerted on as a timeout.
 - **`instanceID`** is used as the lease owner identifier for claimed transactions; if omitted, the manager generates a unique UUID automatically at startup.
 - **`notFoundGracePeriod`** controls how long a transaction whose status query keeps returning `NotFound` is left in `Pending` before being promoted to the terminal `Orphan` status. This protects the recovery sweep from being permanently blocked by transactions whose audit log was persisted but whose broadcast never reached the ledger. The promoted row is marked `Orphan` rather than `Deleted` so operators can distinguish broadcast failures from ledger-rejected transactions. Raise the default if your network has long catch-up windows after committer/orderer restarts; set it to `0` to disable the promotion entirely.
+- **`stuckTransactionAlertThreshold`** escalates the per-attempt failure log from Warn to Error once the same transaction has hit `transactionTimeout` this many consecutive times, so a persistently-unresponsive peer surfaces to alerting instead of blending into routine sweep-failure logs. Past the threshold, the log backs off by doubling (at the threshold, then `2×`, `4×`, `8×`, ...) rather than firing on every sweep, so a transaction stuck for hours does not flood alerting. It is log-severity only: a timeout means the status query didn't answer in time, not that the transaction failed. Unlike `notFoundGracePeriod`, it never changes the transaction's status. The row is still reclaimed every sweep, but while an earlier attempt is still running past its own `transactionTimeout`, the manager skips starting a second one for the same transaction, without releasing its claim, rather than running duplicate concurrent `Recover` calls; the claim is released once that earlier attempt actually returns, whenever that is. Set to `0` to disable the escalation.
 
 **Tuning Recommendations:**
 
 1. **For High-Throughput Environments:**
-   - Increase `batchSize` to 200-500 to process more transactions per sweep
-   - Increase `workerCount` to 8-16 to improve parallel processing
+   - Increase `batchSize` to 100-500 to process more transactions per sweep
+   - Increase `workerCount` to 16-32 to improve parallel processing
    - Decrease `scanInterval` to 2-3s for faster recovery detection
+   - Raise `leaseDuration` alongside `batchSize`/`workerCount` to keep `leaseDuration > (batchSize / workerCount) x transactionTimeout`
 
 
 ### Optional: token.storage.tableNames
@@ -884,7 +927,7 @@ multiply `acquireDeadline` there.
    - Keep default backoff settings to balance retry attempts with resource usage
 
 2. **For Resource-Constrained Environments:**
-   - Decrease `batchSize` to 50 to reduce memory usage
+   - Decrease `batchSize` to 8 to reduce memory usage
    - Decrease `workerCount` to 2 to reduce CPU load
    - Increase `scanInterval` to 10-15s to reduce database queries
 
--- a/docs/services/storage/recovery.md
+++ b/docs/services/storage/recovery.md
@@ -65,11 +65,13 @@ recovery:
   enabled: true              # Enable/disable recovery
   ttl: 30s                   # Minimum age before recovery
   scanInterval: 5s           # How often to scan
-  batchSize: 100             # Max transactions per scan
-  workerCount: 4             # Parallel workers
-  leaseDuration: 30s         # Claim lease duration
+  batchSize: 16              # Max transactions per scan
+  workerCount: 8             # Parallel workers
+  leaseDuration: 5m          # Claim lease duration
+  transactionTimeout: 60s    # Deadline for a single transaction's recovery attempt (0 = unbounded, or >= 10s)
   instanceID: ""             # Instance identifier
   notFoundGracePeriod: 30m   # Promote NotFound rows to Orphan after this age (0 disables)
+  stuckTransactionAlertThreshold: 5  # Escalate to Error log after this many consecutive timeouts (0 disables)
 ```
 
 ## Usage Example
@@ -78,13 +80,15 @@ Creating a recovery manager:
 
 ```go
 config := recovery.Config{
-    Enabled:             true,
-    TTL:                 30 * time.Second,
-    ScanInterval:        5 * time.Second,
-    BatchSize:           100,
-    WorkerCount:         4,
-    LeaseDuration:       30 * time.Second,
-    NotFoundGracePeriod: 30 * time.Minute,
+    Enabled:                        true,
+    TTL:                            30 * time.Second,
+    ScanInterval:                   5 * time.Second,
+    BatchSize:                      16,
+    WorkerCount:                    8,
+    LeaseDuration:                  5 * time.Minute,
+    TransactionTimeout:             60 * time.Second,
+    NotFoundGracePeriod:            30 * time.Minute,
+    StuckTransactionAlertThreshold: 5,
 }
 
 manager := recovery.NewManager(
@@ -121,14 +125,14 @@ func (h *MyHandler) Recover(ctx context.Context, txID string) error {
 
 ## Recovery Process Flow
 
-1. Manager acquires leadership (PostgreSQL advisory lock)
+1. Manager acquires leadership (PostgreSQL advisory lock). Normally released at the end of each sweep, but held across sweeps for as long as any transaction is still abandoned in the background from a previous sweep (see step 5): this instance's own next `ClaimPendingTransactions` call is what keeps that transaction's claim lease alive (step 5), and that only reliably happens if this instance keeps winning leadership, rather than leaving it to the next non-blocking lock attempt succeeding by chance. `Stop()` always releases leadership when called, regardless of whether anything is still abandoned
 2. Manager queries for pending transactions older than TTL
 3. Manager atomically claims a batch of transactions, each returned as a `RecoveryClaim` (`TxID` + `StoredAt`)
 4. Manager distributes claimed transactions to worker pool
-5. Each worker calls `Handler.Recover()` for its transactions
+5. Each worker calls `Handler.Recover()` for its transactions, bounded by `transactionTimeout`: the call runs in its own goroutine, and the worker moves on as soon as the deadline fires rather than waiting for the call to return, so a peer that hangs without honouring the context (as some ledger calls do) still cannot block the sweep indefinitely, and since leadership is held for as long as anything is abandoned (step 1), neither can every other replica in the meantime. The abandoned goroutine keeps running in the background until the underlying call eventually completes; its claim stays held (not released) in the meantime, and its eventual result is used, not discarded (see step 8). If the same transaction is still claimed and re-dispatched to a worker before that abandoned goroutine returns, the manager skips it rather than starting a second concurrent `Recover` call for it
 6. Handler queries network and applies finality logic
-7. If the handler reports `NotFound` and the row was stored more than `notFoundGracePeriod` ago, the manager promotes the row to `Orphan` via `SetStatus` so it exits the eligible scan range
-8. Manager releases claims with success/failure message
+7. If the handler reports `NotFound` and the row was stored more than `notFoundGracePeriod` ago, the manager promotes the row to `Orphan` via `SetStatus` so it exits the eligible scan range, unless this is a step-5 background finalization arriving after `Stop()` has already run, in which case the promotion is skipped: leadership was released by `Stop()` (step 1), so another replica may have since legitimately resolved the same transaction, and an unconditional `SetStatus` could overwrite that outcome
+8. Manager releases each claim with a success/failure message, as soon as a final result for it is known: synchronously, for a `Recover()` call that returns within `transactionTimeout`, or later in the background, for one that did not (step 5), whichever comes first, and only once. The in-flight guard for a transaction is cleared only after this release actually completes, not before, so a sweep that reclaims the same transaction in between cannot start a second attempt out from under the release still in progress
 9. Process repeats on next scan interval
 
 ## Transaction Status Lifecycle
@@ -166,10 +170,24 @@ All three terminal statuses (`Confirmed`, `Deleted`, `Orphan`) are excluded from
 - Increase `ttl` (60s or more)
 - Ensure `leaseDuration` > expected processing time
 
+### Transaction Timeout
+- `transactionTimeout` bounds a single `Handler.Recover()` call (ledger status query plus finality logic), not the whole sweep. It is not a general no-stall guarantee for the sweep: `ClaimPendingTransactions`, `ReleaseRecoveryClaim`, and `SetStatus` still run on the sweep's own context, which has no deadline, so a wedged database connection in any of those can still stall recovery on every replica of the TMS, the same failure mode `transactionTimeout` was added to fix on the ledger side
+- That per-transaction bound does not by itself bound the sweep: a batch of `batchSize` claims split across `workerCount` workers, each taking the full `transactionTimeout`, takes up to `(batchSize / workerCount) × transactionTimeout` worst case. The invariant to size `leaseDuration` against is `leaseDuration > (batchSize / workerCount) × transactionTimeout`. With the shipped defaults (`batchSize: 16`, `workerCount: 8`, `transactionTimeout: 60s`) that worst case is `2 × 60s = 120s`, comfortably inside the default `leaseDuration: 5m`. The manager Warns at startup, rather than refusing to start, when this does not hold: an attempt still in flight when this happens keeps its claim rather than releasing it (see the "stuck transaction" bullet below), so this is a throughput concern, not a correctness one
+- `leaseDuration` should also comfortably exceed `scanInterval`. A transaction still recovering past its `transactionTimeout` keeps its claim alive only because this instance's own next sweep reclaims the same still-`Pending` row, which refreshes the claim's lease as a side effect, and this instance holds recovery leadership across sweeps for as long as that transaction stays abandoned specifically so that reclaim keeps happening on schedule (see step 1 of the process flow above) rather than depending on this instance winning leadership again by chance. If `scanInterval` is not shorter than `leaseDuration`, that renewal can still arrive too late and the claim can lapse between this instance's own sweeps. The manager Warns at startup when this does not hold
+- Raise it if legitimate recoveries (e.g. large token requests re-verified from the database) routinely take longer than the default 60s; lower it if a slow or flaky peer should be given up on sooner, down to a 10s floor: an explicit `transactionTimeout` below 10s is rejected when the configuration is loaded, since a deadline that tight is likely to abandon recoveries that were merely slow rather than genuinely stuck. Setting it to `0` disables the per-transaction deadline entirely, but that also removes the only thing that can interrupt a `Recover` call that ignores its context: `Stop()` itself can then block indefinitely on a hung call, and every later `Start`/`Stop` wedges behind it. The manager logs a Warn at startup when the timeout is disabled; it does not stop you from doing it
+- A transaction that keeps timing out stays `Pending` and is re-claimed on every sweep, but the manager will not start a second concurrent `Recover` call for it while an earlier attempt is still running past its own `transactionTimeout`: it skips the attempt instead. Unlike earlier versions of this manager, the claim is **not** released when this happens: releasing it while the earlier attempt might still be running is what let a second replica claim and run a concurrent `Recover` (and `Commit`) against the same transaction. The claim stays held, kept alive by the leadership-holding and reclaim-renewal described above, until the original stuck attempt actually returns, at which point the manager finalizes it in the background, releasing the claim with the real outcome and only then clearing the in-flight guard, whenever that turns out to be, including well after the sweep (or even the process's `Stop()` call) that first hit the timeout has already moved on. A `recoverCtx.Done()` firing because `Stop()` cancelled the manager, rather than because `transactionTimeout` actually elapsed, is not counted or alerted on as a timeout: it is an ordinary shutdown catching an attempt mid-flight, not evidence the transaction itself is stuck
+- `stuckTransactionAlertThreshold` escalates the per-attempt failure log from Warn to Error once the same transaction has timed out this many times in a row, so a persistently-unresponsive peer is visible to alerting instead of blending into routine sweep failures. Past the threshold the log backs off by doubling (threshold, `2×`, `4×`, `8×`, ...) rather than firing every sweep, so a transaction stuck for hours does not flood alerting. It is purely a log-severity signal: a timeout means the status query didn't answer in time, not that the transaction failed. Unlike a persistent `NotFound`, the manager does not promote it to `Orphan`. Set to `0` to disable the escalation
+
 ## Thread Safety
 
 The Manager is thread-safe and can be safely started/stopped from multiple goroutines. The Handler implementation must also be thread-safe as it will be called concurrently by multiple workers.
 
 ## Shutdown Behaviour
 
-`Stop()` cancels the manager context and waits for the recovery loop to return. If a sweep is mid-batch, the fan-out to the worker pool aborts on cancellation: workers stop after their in-flight `Handler.Recover()` call and any claims not yet dispatched are simply left undispatched. Those rows keep their `Pending` status, so they become eligible again once their lease (`leaseDuration`) expires and are picked up by the next sweep — on this or another replica. The aborted sweep reports a `recovery fan-out cancelled` error. Because this is an ordinary shutdown rather than a failure, the loop logs it at debug level and only warns for genuine sweep errors. The sweep summary counts successes against the claims actually dispatched, so a partial sweep logs `claimed=N, dispatched=M, ...` at warn level instead of crediting the undispatched tail as succeeded.
+`Stop()` cancels the manager context and waits for the recovery loop to return. If a sweep is mid-batch, the fan-out to the worker pool aborts on cancellation: workers exit as soon as they observe the cancellation, and any claims not yet dispatched are simply left undispatched. Those rows keep their `Pending` status, so they become eligible again once their lease (`leaseDuration`) expires and are picked up by the next sweep — on this or another replica. The aborted sweep reports a `recovery fan-out cancelled` error. Because this is an ordinary shutdown rather than a failure, the loop logs it at debug level and only warns for genuine sweep errors. The sweep summary counts successes against the claims actually dispatched, so a partial sweep logs `claimed=N, dispatched=M, ...` at warn level instead of crediting the undispatched tail as succeeded.
+
+With a `transactionTimeout` configured, a worker mid-`Handler.Recover()` does not wait out that call on shutdown: it abandons the call once the deadline fires and moves on (see [Transaction Timeout](#transaction-timeout)), so `Stop()` returning does **not** guarantee every in-flight recovery attempt has actually stopped. An abandoned call can keep running after `Stop()` returns and eventually reach `Commit`, writing into a store the SDK may already be tearing down as part of node shutdown. This is the same accepted tradeoff the timeout mechanism makes everywhere else: bounding the worker takes priority over waiting for a ctx-blind call to finish. With `transactionTimeout: 0` there is no deadline to abandon the call at, so `Stop()` instead blocks on `Handler.Recover()` directly, potentially indefinitely if the handler ignores its context.
+
+The claim on a transaction abandoned this way is not released by `Stop()` either: it is released only once the abandoned call actually returns and the manager finalizes it in the background, which can be well after `Stop()` has already returned. That finalization uses its own independent context rather than the (by then long since cancelled) sweep or manager context, specifically so the release can still succeed after shutdown.
+
+Recovery leadership, unlike the claim above, is always released by `Stop()`, even if a transaction is still abandoned in the background: this instance can no longer productively use it once its own sweep loop has exited, and holding onto it would only block a live peer that could otherwise take over. That means a transaction still abandoned when `Stop()` runs is no longer guaranteed to have its lease renewed by this instance, and a live peer can legitimately reclaim and finish it while this instance's own attempt is still finishing up in the background. For that reason, if the background finalization's handler eventually reports the transaction as `NotFound`, the manager does not promote it to `Orphan` once it observes that `Stop()` already ran: another replica may have since resolved it to something else, and unlike the owner-scoped claim release, `SetStatus` has no way to tell that has happened.
--- a/token/services/storage/services/recovery/config.go
+++ b/token/services/storage/services/recovery/config.go
@@ -11,11 +11,14 @@ import (
 
 	"github.com/LFDT-Panurus/panurus/token/services/config"
 	"github.com/LFDT-Panurus/panurus/token/services/logging"
+	"github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors"
 )
 
 const (
 	// ConfigKeyRecovery is the configuration key for recovery settings
 	ConfigKeyRecovery = "services.network.fabric.recovery"
+
+	defaultStuckTransactionAlertThreshold = 5
 )
 
 var logger = logging.MustGetLogger()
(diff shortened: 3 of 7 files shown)


token/services/storage/services/recovery/manager_test.go
```diff
@@ -465,6 +473,77 @@ func TestManager_PromoteOrphanOnNoSuchTransactionID(t *testing.T) {
...
 }
 
+// TestManager_TransactionTimeoutBoundsHungRecover walks the scenario from
+// issue #2192: a Recover call that never returns on its own (a hung status
+// query). Without a per-transaction deadline, the worker would block forever
+// and, since leadership is held for the whole sweep, no replica would ever
+// recover another pending transaction. The stub here honours ctx.Done() the
+// way a real network call would, so this only passes if the manager actually
+// derives a bounded context for the call.
+func TestManager_TransactionTimeoutBoundsHungRecover(t *testing.T) {
```
AkramBitar, line 505: The test `TestManager_TransactionTimeoutBoundsHungRecover` uses a fixed `time.Sleep(2s)` to wait for two sweep cycles. With up to 1s initial jitter, this can be flaky on a loaded CI machine. Replace the sleep with `require.Eventually` polling on `AcquireRecoveryLeadershipCallCount() >= 2` — same assertion, no timing fragility.

token/services/storage/services/recovery/manager.go
```diff
@@ -371,8 +374,14 @@ func (m *Manager) worker(ctx context.Context, wg *sync.WaitGroup, work <-chan *t
 func (m *Manager) recoverTransaction(ctx context.Context, txID string, storedAt time.Time) error {
 	m.logger.Debugf("recovering transaction [%s]", txID)
 
+	// Bound the handler's ledger query and finality logic: without a deadline, a
+	// hung status query blocks this worker forever and, since leadership is held
+	// for the whole sweep, stalls recovery on every replica of the TMS.
+	recoverCtx, cancel := context.WithTimeout(ctx, m.config.TransactionTimeout)
```
AkramBitar: **This deadline cannot bound the hang the PR is targeting.**

`Recover` is a synchronous call, and `context.WithTimeout` only *signals* cancellation — it cannot interrupt a callee that never reads the context. The hang described in the PR body and #2192 is the ledger status query, and that call is ctx-blind:

- `TTXRecoveryHandler.Recover` → `network.GetTransactionStatus(ctx, ...)` → `ledger.GetTransactionStatus(ctx, ...)` → **`l.l.GetTransactionByID(txID)`** — FSC's `Ledger.GetTransactionByID` takes no `context` (checked v0.17.0 and every other version in the module cache).
- On the fabric adapter (`token/services/network/fabric/network.go:101-143`) there is no ctx-aware call anywhere on that path. On fabricx the context is only reached later, at `executor.QueryState`.

So when a peer hangs: `recoverCtx` expires after 10s, the worker stays parked inside `GetTransactionByID`, `runSweep` never returns, and leadership is still held for the full duration on every replica — the exact behaviour #2192 reports. What *is* bounded today is only the DB/finality tail of `Recover` (`GetTokenRequest`, `ProcessTokenRequest`, `Commit`).

To actually bound it, the worker has to be able to walk away from the call: run `Recover` in a goroutine and select on `recoverCtx.Done()`, abandoning the goroutine on timeout (with the usual caveat that it leaks until the peer call returns), or get a ctx-aware ledger API into FSC.

token/services/storage/services/recovery/manager_test.go
```diff
@@ -465,6 +473,77 @@ func TestManager_PromoteOrphanOnNoSuchTransactionID(t *testing.T) {
...
+		return ctx.Err()
+	}
+
+	manager := recovery2.NewManager(logger, mockDB, mockHandler, config)
+	require.NoError(t, manager.Start())
+
+	// Long enough for the initial sweep's jitter (up to 1s), the transaction
+	// timeout to fire, and at least one further scan tick to prove leadership
+	// was released and reacquired rather than held forever.
+	time.Sleep(2 * time.Second)
```
AkramBitar: Two notes on this test.

**It's a genuine guard on the manager** — I removed the `context.WithTimeout` locally and both the `AcquireRecoveryLeadershipCallCount() >= 2` assertion (got 1) and the `"context deadline exceeded"` message assertion fail, so it is not a rubber stamp. Nice.

**But it cannot catch the problem in my comment on `manager.go`:** the stub here honours `ctx.Done()`, while the production ledger call does not, so this proves the manager derives a bounded context rather than that a real hang is bounded.

(The `require.Eventually` suggestion in my other comment on this file still stands — for reference, `go test -race -count=3` on the package now takes ~31s.)

token/services/storage/services/recovery/manager.go
```diff
@@ -371,8 +374,14 @@ func (m *Manager) worker(ctx context.Context, wg *sync.WaitGroup, work <-chan *t
...
+	defer cancel()
+
 	// Attempt recovery using the injected handler
-	err := m.handler.Recover(ctx, txID)
+	err := m.handler.Recover(recoverCtx, txID)
```
AkramBitar: **A permanently-slow transaction now blocks the queue head.**

`DeadlineExceeded` is not matched by `isNotFoundError`, so the Orphan promotion below never fires for a timed-out attempt, and there is no attempt counter or give-up path. The row is released still `Pending` with its claim cleared (`token/services/storage/db/sql/postgres/transactions.go:192`), and the claim query is `status = Pending AND stored_at < olderThan ORDER BY stored_at ASC LIMIT $5` — so being the oldest, it sits at the head of *every* subsequent batch and burns a full `TransactionTimeout` per sweep, forever.

That is the same "queue head permanently blocked" condition the Orphan promotion was added to prevent, now reachable for slow transactions instead of missing ones — and `docs/services/storage/recovery.md:175` explicitly anticipates recoveries that legitimately exceed the default ("large token requests re-verified from the database").

Worth a consecutive-timeout counter, or promoting to `Orphan` (or a distinct status) after N timed-out attempts.

docs/configuration.md
```diff
@@ -640,7 +651,8 @@ Default values:
 
 - **Recovery is enabled by default** to ensure automatic recovery of lost finality listeners.
 - **Only pending transactions older than `ttl` are considered for recovery** to avoid interfering with active transaction processing.
-- **The manager validates** that `ttl`, `scanInterval`, `batchSize`, `workerCount`, and `leaseDuration` are all greater than zero.
+- **The manager validates** that `ttl`, `scanInterval`, `batchSize`, `workerCount`, `leaseDuration`, and `transactionTimeout` are all greater than zero.
+- **`transactionTimeout`** bounds a single `Handler.Recover()` call, not the whole sweep. Keep it under `leaseDuration` for the same reason `leaseDuration` itself should exceed expected processing time — a claim whose lease expires mid-attempt can be reclaimed by another replica.
```
AkramBitar: **This relationship can't deliver the guarantee as stated.** A per-transaction bound does not bound the sweep, and it is the *sweep* that holds the claims.

With the shipped defaults — `batchSize: 100`, `workerCount: 4`, `transactionTimeout: 10s`, `leaseDuration: 30s` — a batch of slow transactions takes 25 × 10s = 250s. Claims 5..100 wait in the fan-out channel while their 30s leases expire, so another replica can reclaim them mid-sweep: precisely the outcome this line says keeping the timeout under `leaseDuration` prevents.

Either soften the wording (the honest invariant is `leaseDuration > (batchSize / workerCount) × transactionTimeout`), or say plainly that lease expiry mid-sweep is still possible for the tail of a large batch. Same text at `docs/services/storage/recovery.md:174` and `config.go:34`.

token/services/storage/services/recovery/manager.go
```diff
@@ -208,6 +209,8 @@ func (m *Manager) validateConfig() error {
 		return errors.Errorf("invalid recovery worker count [%d]", m.config.WorkerCount)
 	case m.config.LeaseDuration <= 0:
 		return errors.Errorf("invalid recovery lease duration [%s]", m.config.LeaseDuration)
+	case m.config.TransactionTimeout <= 0:
```
AkramBitar: The docs state the `transactionTimeout < leaseDuration` relationship twice as a requirement, but nothing enforces it — this only rejects non-positive values.

`transactionTimeout: 60s` against the default `leaseDuration: 30s` starts happily, and then every single attempt can outlive its own claim lease, letting a second replica claim and recover the same transaction concurrently. A `case m.config.TransactionTimeout >= m.config.LeaseDuration:` arm would catch it at startup, where the other invariants are already caught.

token/services/storage/services/recovery/config.go
```diff
@@ -90,6 +97,9 @@ func LoadConfig(cfg *config.Configuration) (Config, error) {
 	if config.LeaseDuration > 0 {
 		result.LeaseDuration = config.LeaseDuration
 	}
+	if config.TransactionTimeout > 0 {
```
AkramBitar: `transactionTimeout: 0` is treated as "unset" and falls back to 10s, so there is no way to opt out of the new deadline.

`NotFoundGracePeriod` below uses a `cfg.IsSet` gate for exactly this reason — so an explicit zero can mean "disabled" rather than silently reverting to the default. That reasoning applies at least as strongly here: this PR newly introduces the ability to *abort* an operation that was previously unbounded, so an escape hatch is worth having. An operator on a network with long catch-up windows who writes `transactionTimeout: 0` intending "behave as before" currently gets 10s and no log line explaining why their recoveries start failing.

(Note `validateConfig` rejects 0, so allowing it as "unbounded" needs the IsSet gate *and* a validation carve-out.)

docs/services/storage/recovery.md
```diff
@@ -168,10 +170,14 @@ All three terminal statuses (`Confirmed`, `Deleted`, `Orphan`) are excluded from
...
 - Ensure `leaseDuration` > expected processing time
 
+### Transaction Timeout
+- `transactionTimeout` bounds a single `Handler.Recover()` call (ledger status query plus finality logic), not the whole sweep
+- Keep it comfortably under `leaseDuration` — a claim whose lease expires while still being processed can be reclaimed by another replica
+- Raise it if legitimate recoveries (e.g. large token requests re-verified from the database) routinely take longer than the default 10s; lower it if a slow or flaky peer should be given up on sooner
+
 ## Thread Safety
 
 The Manager is thread-safe and can be safely started/stopped from multiple goroutines. The Handler implementation must also be thread-safe as it will be called concurrently by multiple workers.
```
AkramBitar, line 183: Unrelated whitespace change: the blank line before `## Shutdown Behaviour` was removed here, so the heading now abuts this paragraph (markdownlint MD022, and some renderers fold it into the paragraph). Worth restoring since the rest of this file is append-only in the PR.

token/services/storage/services/recovery/manager.go
atharrva01: Good catch, you're right that context.WithTimeout alone can't interrupt GetTransactionByID since it takes no context. Fixed by running Recover in its own goroutine and giving up on recoverCtx.Done() instead of waiting for the call to return. The goroutine is abandoned on a real hang and leaks until the underlying call eventually completes, but the worker (and the leadership it holds) is no longer stuck.

Added a test where the mock handler never reads ctx at all (select{}) to prove this actually works and isn't just relying on the handler being cooperative: TestManager_TransactionTimeoutBoundsRecoverIgnoringContext.

token/services/storage/services/recovery/manager_test.go
atharrva01: Swapped for require.Eventually polling AcquireRecoveryLeadershipCallCount() >= 2, same as your suggestion.

token/services/storage/services/recovery/manager_test.go
atharrva01: Right, this test alone couldn't catch the manager.go bug you flagged since the stub was cooperative. Fixed the underlying issue (goroutine + select instead of relying on the handler honouring ctx) and added TestManager_TransactionTimeoutBoundsRecoverIgnoringContext, whose handler does `select{}` and never touches ctx at all, so it actually exercises the real failure mode instead of a well-behaved one. Also swapped this test's sleep for require.Eventually.

token/services/storage/services/recovery/manager.go
atharrva01: You're right this is real. I thought about auto-promoting to Orphan here too, but a timeout doesn't tell us the tx never reached the ledger, it just means the status query didn't answer in time. The tx could easily be sitting Confirmed on the ledger and we just can't see it because of a slow/flaky peer. Marking that Orphan would be actively wrong, not just imprecise, since Orphan specifically means "never reached the ledger" elsewhere in this codebase (and downstream tooling may read it that way).

A real fix (backing off retries on a specific stuck tx, or a distinct non-misleading status) needs either a claim-renewal API or a schema change, both bigger than this PR. For now I added stuckTransactionAlertThreshold: after N consecutive timeouts on the same tx it escalates the log from Warn to Error so it's visible to alerting instead of blending into routine failures, but the row stays Pending and keeps being retried. Happy to file a follow-up issue for the real backoff/status fix if you think it's worth tracking separately.

docs/configuration.md
atharrva01: Reworded to the actual invariant (leaseDuration > (batchSize/workerCount) x transactionTimeout) with the worked example (25 x 10s = 250s vs the default 30s lease), same wording in both docs/configuration.md and recovery.md now.

token/services/storage/services/recovery/manager.go
atharrva01: Added a validateConfig case for transactionTimeout >= leaseDuration (rejected at startup, exempting the 0/unbounded case since there's no bound to compare against there).

token/services/storage/services/recovery/config.go
atharrva01: Switched to the same cfg.IsSet gate NotFoundGracePeriod uses. transactionTimeout: 0 now means unbounded (validateConfig has a carve-out too), and it's called out in the config docs.

docs/services/storage/recovery.md
atharrva01: Restored the blank line.

AkramBitar (commented):
@atharrva01 

Thanks a lot for all the effort and the great work that you put on that PR. I appreciate it a lot. 

Reviewed the fix against issue #2192 and against my earlier review comments.

**The fix works.** `callHandler` running `Recover` in its own goroutine and returning on `recoverCtx.Done()` is the right shape — a ctx-blind ledger call can no longer pin the worker or the leadership it holds for the whole sweep. I verified this empirically with a `select{}` handler: leadership is released and reacquired, and `Stop()` returns.

**All 8 of my earlier comments were addressed**, several of them well — the goroutine-based timeout, the new `...BoundsRecoverIgnoringContext` test that genuinely exercises the ctx-blind case, and the reworded docs invariant with the `25 x 10s = 250s` worked example. The "permanently-slow tx blocks the queue head" one is only half-addressed: the counter is log-only, and the queue-head block is still there (and now compounded, see the `callHandler` comment).

`go test`, `go test -race`, and `go vet` on `token/services/storage/services/recovery/` all pass clean at this head (race 10.7s).

Two of the fixes introduced new problems, though, and I think three things should block merge:

1. **`manager.go:228`** — the new `timeout >= lease` rejection hard-fails startup on configs that were valid before, because the 10s default isn't clamped to a smaller configured `leaseDuration`. Any deployment with `leaseDuration <= 10s` refuses to start after upgrade.
2. **`manager.go:396`** — timed-out transactions are re-claimed every sweep, so abandoned goroutines accumulate without bound and run duplicate *concurrent* `Recover` calls for the same txID (measured: 101 concurrent invocations for one txID). Since `Recover` reaches `Commit`, that defeats the mutual exclusion the claim lease provides.
3. **`manager.go:397`** — `transactionTimeout: 0` deadlocks `Stop()` while holding `m.mu`, and both doc sites tell operators the opposite.

Details and suggested fixes inline, plus six lower-severity items. Findings 1-3 were each reproduced with throwaway probe tests.

token/services/storage/services/recovery/manager.go
```diff
@@ -208,6 +218,15 @@ func (m *Manager) validateConfig() error {
...
 	case m.config.LeaseDuration <= 0:
 		return errors.Errorf("invalid recovery lease duration [%s]", m.config.LeaseDuration)
+	case m.config.TransactionTimeout < 0:
+		return errors.Errorf("invalid recovery transaction timeout [%s]", m.config.TransactionTimeout)
+	// A timeout that can outlive its own claim lease lets a second replica
+	// claim and recover the same transaction concurrently, so reject it at
+	// startup alongside the other invariants. TransactionTimeout == 0 means
+	// "unbounded" (opted out, see LoadConfig) and is exempt: there is no
+	// bound to compare against.
+	case m.config.TransactionTimeout > 0 && m.config.TransactionTimeout >= m.config.LeaseDuration:
```
AkramBitar: **HIGH — this rejection hard-fails node startup on configs that were valid before this PR.**

`LoadConfig` starts from `DefaultConfig()`, which now sets `TransactionTimeout: 10s` (`config.go:77`), and only overrides it when the operator explicitly set `transactionTimeout` (`config.go:122`). So an existing deployment that configured only `leaseDuration: 5s` ends up with `TransactionTimeout=10s`, `LeaseDuration=5s` — and this case fires:

```
invalid recovery transaction timeout [10s]: must be less than lease duration [5s]
```

That error propagates out of `Start()` → `createRecoveryManager` (`token/services/network/fabric/network.go:530`) → network setup, so **the node refuses to start after upgrading**. Reproduced with a probe building `DefaultConfig()` and setting only `LeaseDuration = 5s`.

Any `leaseDuration <= 10s` is affected, and this isn't exotic: `docs/configuration.md` recommends sizing `leaseDuration` at roughly "2x expected processing time", which is well under 10s on a healthy network.

Suggested fix — clamp the *default* rather than rejecting it, e.g. in `LoadConfig`:

```go
if !cfg.IsSet(ConfigKeyRecovery + ".transactionTimeout") && result.TransactionTimeout >= result.LeaseDuration {
    result.TransactionTimeout = result.LeaseDuration / 2
}
```

or only enforce the invariant when the value was set explicitly. Either way the hard error should be reserved for a config the operator actually wrote.

token/services/storage/services/recovery/manager.go
```diff
@@ -356,6 +375,62 @@ func (m *Manager) worker(ctx context.Context, wg *sync.WaitGroup, work <-chan *t
...
+// (or the process exits), and its result is discarded into the buffered
+// channel. That is the accepted cost of bounding a call this package does
+// not control the cancellation semantics of; the alternative is leaving the
+// worker (and the leadership it holds for the whole sweep) blocked
+// indefinitely, which is the bug this exists to fix.
+//
+// TransactionTimeout == 0 means "unbounded" (see LoadConfig / validateConfig):
+// Recover is then called directly on ctx, with no derived deadline and no
+// extra goroutine.
+func (m *Manager) callHandler(ctx context.Context, txID string) error {
```
AkramBitar: **HIGH — abandoned goroutines accumulate without bound, and run duplicate _concurrent_ `Recover` calls for the same txID.**

A timed-out transaction keeps its `Pending` status, and being the oldest it is re-claimed on **every** subsequent sweep. Each sweep starts a fresh goroutine here and abandons it, so they stack up for the same txID.

Measured with a `select{}` handler at `scanInterval: 20ms`: **101 `Recover` invocations for one txID, all 101 concurrently in flight, +103 leaked goroutines in 2.5s.** At the default 5s `scanInterval` that's ~12/min, ~17k/day per stuck transaction, each one pinning a ledger call and its connection.

The bigger problem is that `Recover` is not read-only. `TTXRecoveryHandler.Recover` reaches `Commit` — `token/services/ttx/finality/listener.go:188` does `AppendValid` + `SetStatus(Confirmed)` + `NotifyStatus`. If the peer eventually answers, N abandoned attempts can commit concurrently for the same txID, which is precisely the mutual exclusion the claim lease exists to provide. The doc comment above accounts for the leak but not for this duplicate-writer case.

Suggested fix: keep an in-flight set keyed by txID and skip a transaction whose previous attempt hasn't returned (the existing `timeoutMu`/`timeoutCounts` pair is a natural place), and/or apply per-tx backoff so a stuck row isn't re-attempted every sweep.

token/services/storage/services/recovery/manager.go
```diff
@@ -356,6 +375,62 @@ func (m *Manager) worker(ctx context.Context, wg *sync.WaitGroup, work <-chan *t
...
+	if m.config.TransactionTimeout <= 0 {
```
AkramBitar, line 636: **MEDIUM — `transactionTimeout: 0` reintroduces a hard `Stop()` deadlock, and the docs promise the opposite.**

On this path `Recover` is called directly, so a handler that ignores ctx (the exact case the goroutine below exists for) blocks the worker → `workerWG.Wait()` → `recoveryLoop` → `m.wg.Wait()` inside `Stop()`. And `Stop()` holds `m.mu` while waiting, so every later `Start`/`Stop` wedges too. Verified by probe: with `TransactionTimeout: 0` and a `select{}` handler, `Stop()` never returned within 3s.

That's defensible as a documented trade-off, but the docs currently say the opposite:

- `docs/configuration.md:657` — "the call is then bounded only by the sweep's own cancellation, i.e. `Stop()`"
- `docs/services/storage/recovery.md:178` — "set it to `0` to disable the per-transaction deadline entirely"

`Stop()` does *not* bound a ctx-blind call — that's what this function's own comment explains. So an operator who sets `0` to "behave as before" gets an unshutdownable manager with no warning. Either document the deadlock risk explicitly at both doc sites, or log a Warn at `Start()` when the timeout is disabled.

docs/services/storage/recovery.md
```diff
@@ -127,7 +131,7 @@ func (h *MyHandler) Recover(ctx context.Context, txID string) error {
 2. Manager queries for pending transactions older than TTL
 3. Manager atomically claims a batch of transactions, each returned as a `RecoveryClaim` (`TxID` + `StoredAt`)
 4. Manager distributes claimed transactions to worker pool
-5. Each worker calls `Handler.Recover()` for its transactions
+5. Each worker calls `Handler.Recover()` for its transactions, bounded by `transactionTimeout`: the call runs in its own goroutine, and the worker moves on as soon as the deadline fires rather than waiting for the call to return, so a peer that hangs without honouring the context (as some ledger calls do) still cannot block the sweep indefinitely, and since leadership is held for the whole sweep, neither can every other replica. The abandoned goroutine keeps running in the background until the underlying call eventually completes; its result is discarded
```
AkramBitar: **MEDIUM — the "Shutdown Behaviour" section (lines 185–187) is now stale, and it hides a real consequence.**

It still says workers "stop after their in-flight `Handler.Recover()` call". With a `transactionTimeout` configured, that's no longer true — as this step correctly notes, the worker *abandons* the call.

Worth spelling out what that means at shutdown, because it isn't obvious: once `Stop()` returns, recovery has **not** necessarily stopped. An abandoned `Recover` can still complete afterwards and run `Commit` against a store the SDK is already tearing down — a write into a closing DB pool, or a spurious `Confirmed` status write after shutdown. Given the SDK shuts these managers down as part of node teardown, that deserves a sentence in the shutdown section rather than only being implied here.

token/services/storage/services/recovery/manager.go
```diff
@@ -371,8 +446,27 @@ func (m *Manager) worker(ctx context.Context, wg *sync.WaitGroup, work <-chan *t
...
+	// callHandler). Track consecutive timeouts per txID purely so the log can
+	// escalate if the same transaction keeps timing out sweep after sweep;
+	// this deliberately does not change the transaction's stored status. A
+	// timeout only means the status query didn't answer in time, not that the
+	// transaction never reached the ledger, so it cannot be promoted to
+	// Orphan the way a persistent NotFound can: unlike NotFound, we have no
+	// signal that would let us conclude one way or the other.
+	if errors.Is(err, context.DeadlineExceeded) {
+		count := m.recordTimeout(txID)
+		if m.config.StuckTransactionAlertThreshold > 0 && count >= m.config.StuckTransactionAlertThreshold {
```
AkramBitar: **LOW/MEDIUM — this escalated Error log fires on every sweep, forever, with no rate limit.**

The condition is `count >= threshold`, and a stuck transaction is re-claimed every `scanInterval`, so once the threshold is crossed this emits an Error line every 5s indefinitely, per stuck transaction. The stated goal is to make the transaction "visible to alerting"; at this rate it floods alerting instead, which tends to get the whole pattern muted.

Suggested fix: log once on the crossing (`count == threshold`) and then back off — e.g. re-log only at powers of two, or at a fixed multiple of the threshold.

token/services/storage/services/recovery/manager.go
```diff
@@ -68,6 +69,14 @@ type Manager struct {
 	wg      sync.WaitGroup
 	started bool
 	mu      sync.Mutex
+
+	// timeoutCounts tracks, per txID, how many consecutive attempts have hit
+	// the TransactionTimeout deadline. It only exists for the log escalation
+	// in recoverTransaction; it is best-effort in-memory state (reset on
+	// restart, or if leadership moves to another replica) and never changes a
+	// transaction's stored status.
+	timeoutCounts map[string]int
```
AkramBitar: **LOW — `timeoutCounts` is never bounded or expired, so it can grow for the manager's lifetime.**

`clearTimeoutCount` only runs on a *later attempt of the same txID*. But a transaction that times out once and is then confirmed by an independent finality listener leaves the `Pending` scan range and is never claimed again — so its entry is never cleared. On a long-lived node with a flaky peer this accumulates, keyed by 64-char txIDs.

Suggested fix: delete the entry when the claim is released for a non-timeout reason regardless of path, or cap the map size.

token/services/storage/services/recovery/manager.go
```diff
@@ -356,6 +375,62 @@ func (m *Manager) worker(ctx context.Context, wg *sync.WaitGroup, work <-chan *t
...
+	recoverCtx, cancel := context.WithTimeout(ctx, m.config.TransactionTimeout)
+	defer cancel()
+
+	resultCh := make(chan error, 1)
+	go func() {
+		resultCh <- m.handler.Recover(recoverCtx, txID)
+	}()
+
+	select {
+	case err := <-resultCh:
```
AkramBitar: **LOW — this `select` can report `DeadlineExceeded` for a `Recover` that actually succeeded.**

If `Recover` returns just as the deadline fires, both cases are ready and Go picks pseudo-randomly. The successful result is then silently discarded: the claim is released with `recovery failed: context deadline exceeded`, the sweep counts a failure, and `recordTimeout` increments toward the stuck-transaction escalation — for a recovery that worked.

Suggested fix: prefer the real result when both are ready.

```go
case <-recoverCtx.Done():
    select {
    case err := <-resultCh:
        return err
    default:
        return recoverCtx.Err()
    }
```

token/services/storage/services/recovery/manager_test.go
```diff
@@ -474,4 +630,5 @@ func TestDefaultConfig(t *testing.T) {
 	assert.Equal(t, 100, config.BatchSize)
 	assert.Equal(t, 4, config.WorkerCount)
 	assert.Equal(t, 30*time.Second, config.LeaseDuration)
+	assert.Equal(t, 10*time.Second, config.TransactionTimeout)
```
AkramBitar: **LOW — neither new `validateConfig` arm nor the `LoadConfig` zero-means-unbounded gate has a test.**

All three were added specifically in response to review, and there is no `config_test.go` and nothing here asserting that:

- `TransactionTimeout >= LeaseDuration` is rejected
- a negative `TransactionTimeout` is rejected
- `transactionTimeout: 0` survives `LoadConfig` as `0` rather than falling back to the 10s default

A test that constructs `DefaultConfig()` with a small `LeaseDuration` and calls `Start()` would have caught the startup regression I flagged on `manager.go:228`.

`StuckTransactionAlertThreshold` is also entirely untested — no test asserts the escalation fires at the threshold, or that `0` disables it.

docs/configuration.md
```diff
@@ -250,6 +250,15 @@ token:
...
               leaseDuration: 30s
+
+              # transactionTimeout bounds a single transaction's recovery attempt: the
+              # handler's Recover call, which queries the ledger for status and applies
+              # finality logic. Default: 10s.
+              # Without this, a hung status query blocks the worker indefinitely, and since
+              # leadership is held for the whole sweep, stalls recovery on every replica.
+              # Relationship: keep comfortably under leaseDuration, so a claim whose lease
+              # expires while still processing isn't reclaimed by another replica mid-attempt.
+              transactionTimeout: 10s
```
AkramBitar: **LOW — `stuckTransactionAlertThreshold` is missing from this annotated reference block.**

`transactionTimeout` got a proper annotated entry here, but the new threshold option appears only in the compact yaml (line 635), the defaults list (line 650), and the relationships prose (line 661) — not in the primary annotated section, which is where operators actually read option semantics. Worth adding a sibling entry right after this one.

AkramBitar (changes requested):

token/services/storage/services/recovery/manager.go
atharrva01: Fixed: LoadConfig now clamps the transactionTimeout default below whatever leaseDuration you configured, when you didn't set transactionTimeout explicitly, instead of letting the stale 10s default hit validateConfig. An operator setting only leaseDuration now gets a working default rather than a refusal to start. Added TestLoadConfig_ClampsDefaultTransactionTimeoutBelowConfiguredLeaseDuration, which reproduces your exact scenario and asserts Start() succeeds against the LoadConfig output.

token/services/storage/services/recovery/manager.go
atharrva01: Fixed: added an in-flight guard (tryMarkInFlight/clearInFlight) keyed by txID. While a previous attempt for a txID is still running past its transactionTimeout, callHandler returns immediately without starting a second goroutine or a second Recover call; recoverTransaction releases the claim with a skip message instead. The abandoned goroutine still leaks until the underlying call eventually returns, same as before, but only one goroutine and one Recover call per stuck txID exist at a time now, so no more duplicate concurrent Commits racing each other. Added TestManager_SkipsReclaimedTransactionStillInFlight, which uses a handler that never returns and asserts RecoverCallCount stays at 1 across 5+ reclaim sweeps.

token/services/storage/services/recovery/manager.go
atharrva01: Documented rather than fixed further, since there's no way to bound a call that both ignores its context and has no deadline: Start() now logs a Warn when transactionTimeout is disabled explaining that it can block Stop() indefinitely, and both docs.configuration.md and recovery.md say this explicitly instead of the old (wrong) claim that Stop() bounds it.

docs/services/storage/recovery.md
atharrva01: Fixed: added a paragraph to Shutdown Behaviour explaining that with transactionTimeout configured, Stop() returning does not mean every in-flight attempt has stopped, an abandoned call can keep running and reach Commit after Stop() returns, against a store that may already be tearing down.

token/services/storage/services/recovery/manager.go
atharrva01: Fixed: added maybeAlertStuckTransaction, which logs at Error on the threshold crossing and then backs off by doubling (threshold, 2x, 4x, 8x, ...) instead of every sweep. Covered by TestManager_StuckTransactionAlertThreshold_EscalatesAndBacksOff, which asserts the number of Error logs stays well below the number of timeouts observed.

token/services/storage/services/recovery/manager.go
atharrva01: Fixed: timeoutCounts entries now carry a lastSeen timestamp and pruneTimeoutCounts drops anything untouched for over an hour, called once per sweep. Went with a TTL rather than clearing on claim-release for a non-timeout reason, since the case you flagged is exactly the one where this manager never sees the row again to clear it (an independent listener resolves it), so TTL is the only thing that actually catches that path.

token/services/storage/services/recovery/manager.go
atharrva01: Fixed: added preferResult, which checks resultCh non-blockingly before falling back to the deadline error, so a Recover that completed at the same instant the deadline fired is no longer discarded as a timeout. Added a direct unit test (TestPreferResult in manager_internal_test.go) that constructs both channel states directly rather than relying on real timing to force the race.

token/services/storage/services/recovery/manager_test.go
atharrva01: Added: TestManager_StartValidatesConfig (table-driven, covers negative timeout, timeout==leaseDuration, timeout>leaseDuration, and the other validateConfig arms), TestManager_StartAcceptsZeroTransactionTimeoutAsUnbounded, TestLoadConfig_TransactionTimeoutZeroSurvivesAsUnbounded, TestLoadConfig_ClampsDefaultTransactionTimeoutBelowConfiguredLeaseDuration (the one that would've caught the startup regression), and two StuckTransactionAlertThreshold tests (escalates+backs off, and 0 disables it) using a small recordingLogger since the package's MockLogger is a no-op and doesn't record calls.

docs/configuration.md
atharrva01: Fixed: added a stuckTransactionAlertThreshold entry to the annotated reference block right after transactionTimeout.

token/services/storage/services/recovery/manager.go
```diff
@@ -289,10 +360,17 @@ func (m *Manager) recoverTransactions(ctx context.Context) error {
...
-	if failures > 0 || dispatched < len(records) {
-		m.logger.Warnf("completed recovery sweep: claimed=%d, dispatched=%d, succeeded=%d, failed=%d",
-			len(records), dispatched, dispatched-failures, failures)
-	} else {
+	succeeded := dispatched - failures - skipped
+	switch {
+	case failures > 0 || dispatched < len(records):
+		m.logger.Warnf("completed recovery sweep: claimed=%d, dispatched=%d, succeeded=%d, failed=%d, skipped=%d",
+			len(records), dispatched, succeeded, failures, skipped)
+	case skipped > 0:
```
AkramBitar: **MEDIUM - sweep summary misreports partial progress as a full stall.**

The `case skipped > 0:` arm logs `"claimed=%d, all skipped (previous attempts still in flight), no progress made"`, but its condition only establishes `skipped >= 1`, not that *all* claims were skipped - and it discards the `succeeded`/`failed` counts computed on line 363.

Reproduced with a probe test using a 2-claim batch (one `select{}` handler, one healthy handler): the sweep logged

```
claimed=2, all skipped (previous attempts still in flight), no progress made
```

in the very same sweep where `txOK` was successfully recovered.

Operational impact: with one stuck transaction in a 100-claim batch, **every** sweep tells the operator recovery is completely wedged while 99 rows are being recovered normally.

Suggested fix: guard the arm with `skipped == dispatched`, and keep the `succeeded`/`failed` counts in the message for the partial case.

token/services/storage/services/recovery/manager.go
```diff
@@ -356,6 +434,153 @@ func (m *Manager) worker(ctx context.Context, wg *sync.WaitGroup, work <-chan *t
...
+// previous attempt is still in flight, callHandler returns errRecoveryInFlight
+// immediately instead of starting a second one.
+//
+// TransactionTimeout == 0 means "unbounded" (see LoadConfig / validateConfig):
+// Recover is then called directly on ctx, with no derived deadline, no
+// in-flight guard, and no extra goroutine, so a ctx-blind call can block
+// Stop() indefinitely (Start logs a Warn about this when the timeout is
+// disabled).
+func (m *Manager) callHandler(ctx context.Context, txID string) error {
+	if m.config.TransactionTimeout <= 0 {
```
AkramBitar, line 636: **MEDIUM - residual design gap (previously acknowledged); suggest a tracked follow-up rather than blocking this PR.**

For a genuinely ctx-blind hung `Recover` call, the txID stays in `m.inFlight` until that call returns. The row therefore makes no forward progress and - staying `Pending` and oldest - permanently occupies one slot of `ORDER BY stored_at ASC LIMIT batchSize` on every subsequent sweep.

Worst case: if a peer outage hangs `batchSize` status queries, every sweep is all-skips and **no other pending transaction can be recovered** until those calls return or the process restarts. The claim-release plus skip guard correctly prevents duplicate `Commit`s, but it does not stop the batch window from being starved.

Neither `docs/services/storage/recovery.md` nor `docs/configuration.md` mentions this batch-slot consequence. Real backoff or a distinct terminal status for abandoned attempts would fix it; happy to see that as a follow-up issue rather than growing this PR.

token/services/storage/services/recovery/config.go
```diff
@@ -90,6 +114,33 @@ func LoadConfig(cfg *config.Configuration) (Config, error) {
...
+		// reserved for a transactionTimeout the operator set explicitly.
+		//
+		// Integer division truncates result.LeaseDuration/2 to 0 for a
+		// leaseDuration below 2ns, and 0 here means "unbounded" (see the
+		// IsSet gate above and callHandler): exactly the opt-out this clamp
+		// exists to avoid landing an operator in by accident. Floor at 1ns,
+		// still bounded, if not a practically useful deadline at that scale.
+		result.TransactionTimeout = result.LeaseDuration / 2
+		if result.TransactionTimeout <= 0 {
+			result.TransactionTimeout = 1
```
AkramBitar: **LOW - the clamp's `1`ns floor still violates the invariant the clamp exists to satisfy.**

For `leaseDuration: 1ns`: `LeaseDuration/2` truncates to `0`, the floor raises it to `1ns`, and `validateConfig` then rejects `TransactionTimeout >= LeaseDuration`. Reproduced:

```
invalid recovery transaction timeout [1ns]: must be less than lease duration [1ns]
```

That is exactly the startup refusal this clamp was added to prevent.

`TestLoadConfig_ClampFloorsAboveZeroForTinyLeaseDuration` only asserts `Positive`, unlike its sibling test which asserts `Start()` succeeds, so it does not catch this.

Practically unreachable (needs a sub-2ns lease), but the fix is trivial - floor at `result.LeaseDuration - 1`.

AkramBitar (commented):
### Review: all prior comments addressed; three small items left

Verified against head `e08ede548` (rebased on `origin/main`, MERGEABLE).

**All 17 prior review comments are addressed.**

- **Round 1 (8, in `ae051b0c3`)** — goroutine + `select` on `recoverCtx.Done()` so a ctx-blind `Recover` no longer pins the worker; `require.Eventually` replacing the fixed sleep; the `...BoundsRecoverIgnoringContext` test with a `select{}` handler; `validateConfig` arm for `timeout >= lease`; `IsSet` gate so `transactionTimeout: 0` means unbounded; corrected `leaseDuration > (batchSize/workerCount) x transactionTimeout` invariant; restored blank line; consecutive-timeout counter + `stuckTransactionAlertThreshold`.
- **Rounds 2-3 (9, in `79bb061f4` + `e08ede548`)** — `LoadConfig` clamp of the unset default below a configured `leaseDuration` (with the regression test that would have caught the startup failure); `tryMarkInFlight`/`clearInFlight` guard (confirmed `RecoverCallCount` stays at 1 for a stuck txID across many reclaim sweeps); Warn at `Start()` for a disabled timeout plus both doc sites; Shutdown-Behaviour paragraph; doubling alert backoff in `maybeAlertStuckTransaction`; `timeoutCounts` TTL prune; `preferResult` tie-break; new `config_test.go` + `manager_internal_test.go`; annotated `stuckTransactionAlertThreshold` entry.

**Verification on `token/services/storage/services/recovery`:** `go test` clean; `go test -race -count=4` clean (46s); `go vet` clean; `gofmt -l` clean; `golangci-lint run` -> 0 issues. Production wiring reaches `Config` only via `LoadConfig`/`DefaultConfig` (`token/services/network/fabric/network.go:497-505`), so the clamp does cover the startup regression path.

**Verdict: merge-ready once three items are fixed** — none require design changes:

1. `manager.go:368` — sweep-summary reports a healthy sweep as fully wedged (reproduced).
2. `recovery.md:179` — states the row stays claimed; the code releases it.
3. `configuration.md:261` — annotated block still carries the naive `transactionTimeout < leaseDuration` claim that round 1 asked to correct.

The `manager.go:466` batch-slot exhaustion gap is worth a **follow-up issue**, not a blocker on this PR. None of the earlier blocking issues are reopened: the startup regression, the unbounded-goroutine / duplicate-`Commit` problem, and the `Stop()` deadlock documentation are all genuinely resolved and covered by the new tests.

docs/services/storage/recovery.md
```diff
@@ -168,10 +172,19 @@ All three terminal statuses (`Confirmed`, `Deleted`, `Orphan`) are excluded from
...
+- Raise it if legitimate recoveries (e.g. large token requests re-verified from the database) routinely take longer than the default 10s; lower it if a slow or flaky peer should be given up on sooner. Setting it to `0` disables the per-transaction deadline entirely, but that also removes the only thing that can interrupt a `Recover` call that ignores its context: `Stop()` itself can then block indefinitely on a hung call, and every later `Start`/`Stop` wedges behind it. The manager logs a Warn at startup when the timeout is disabled; it does not stop you from doing it
+- A transaction that keeps timing out stays `Pending` and is re-claimed on every sweep, but the manager will not start a second concurrent `Recover` call for it while an earlier attempt is still running past its own `transactionTimeout`: it skips the attempt and releases the claim instead. This avoids both unbounded goroutine growth and duplicate concurrent writes racing to `Commit` the same transaction, at the cost of the row staying claimed rather than making forward progress until the stuck attempt eventually returns (or the process restarts)
```
AkramBitar: **LOW - contradicts the code and the sibling doc.**

This says the skip's cost is "the row staying claimed rather than making forward progress". The code does the opposite: `manager.go:605-611` explicitly releases the claim ("so the row does not stay locked to this owner"), and `docs/configuration.md` describes it correctly ("just releases the claim").

An operator debugging a stuck row would go looking for a claim that isn't there.

AkramBitar: **LOW - the duplicate-`Commit` guarantee is process-local and should be qualified.**

"This avoids ... duplicate concurrent writes racing to `Commit` the same transaction" reads as an absolute guarantee, but `m.inFlight` is per-process. Because the claim is released while the abandoned goroutine is still running, once leadership rotates another replica can claim the same txID and run `Recover` concurrently with replica A's abandoned attempt.

In practice A's attempt is largely neutered by the cancelled `recoverCtx` (see the `defer cancel()` comment), so the real risk is small - but the sentence should say "within this process".

token/services/storage/services/recovery/manager.go
```diff
@@ -356,6 +434,153 @@ func (m *Manager) worker(ctx context.Context, wg *sync.WaitGroup, work <-chan *t
...
+		return m.handler.Recover(ctx, txID)
+	}
+
+	if !m.tryMarkInFlight(txID) {
+		return errRecoveryInFlight
+	}
+
+	recoverCtx, cancel := context.WithTimeout(ctx, m.config.TransactionTimeout)
```
AkramBitar, line 644: **LOW - `defer cancel()` can leave a confirmed row labelled as a recovery failure.**

`defer cancel()` fires as soon as `callHandler` gives up, cancelling the abandoned goroutine's remaining ctx-aware work.

In the narrow window where `Commit`'s `tx.Commit()` (`token/services/ttx/finality/listener.go:226`, no ctx) has already succeeded but `NotifyStatus(ctx, ...)` (line 245) has not, the row ends up `Confirmed` in the DB while the manager records a `DeadlineExceeded`:

- `ReleaseRecoveryClaim` overwrites `status_message` with `recovery failed: context deadline exceeded` on a row that is actually confirmed, and
- `recordTimeout` counts it toward the stuck-transaction escalation.

`preferResult` cannot cover this case because `Recover` has not returned yet.

docs/configuration.md
```diff
@@ -252,6 +252,30 @@ token:
               # Should be longer than the typical time to query and process a single transaction.
               # Relationship: Should be greater than the expected network latency + processing time for transaction status queries.
               leaseDuration: 30s
+
+              # transactionTimeout bounds a single transaction's recovery attempt: the
+              # handler's Recover call, which queries the ledger for status and applies
+              # finality logic. Default: 10s.
+              # Without this, a hung status query blocks the worker indefinitely, and since
+              # leadership is held for the whole sweep, stalls recovery on every replica.
+              # Relationship: keep comfortably under leaseDuration, so a claim whose lease
```
AkramBitar: **LOW - the round-1 correction did not reach the annotated block.**

This `transactionTimeout` annotation still carries the naive relationship ("keep comfortably under `leaseDuration`, so a claim whose lease expires while still processing isn't reclaimed by another replica mid-attempt") - the exact claim round 1 asked to correct.

The corrected invariant (`leaseDuration > (batchSize/workerCount) x transactionTimeout`) landed only in the "Parameter Relationships" prose and in `recovery.md`, not in the annotated block operators read first.

token/services/storage/services/recovery/manager.go
```diff
@@ -371,8 +605,44 @@ func (m *Manager) worker(ctx context.Context, wg *sync.WaitGroup, work <-chan *t
...
+		count := m.recordTimeout(txID)
+		m.logger.Debugf("recovery: tx [%s] skipped, a previous attempt is still running past transactionTimeout (%d consecutive timeout(s))", txID, count)
+		m.maybeAlertStuckTransaction(txID, count)
+
+		if releaseErr := m.releaseClaim(ctx, txID, "skipped: previous recovery attempt still in flight"); releaseErr != nil {
+			m.logger.Warnf("failed to release recovery claim for transaction [%s]: %v", txID, releaseErr)
+		}
+
+		return errSkippedInFlight
+	}
```
AkramBitar, line 866: **MEDIUM/HIGH — the timeout releases the claim while the abandoned `Recover` is still running, so another replica can run a concurrent `Recover` -> `Commit`.**

(Anchored here on the `errRecoveryInFlight` release; the same applies to the `DeadlineExceeded` release at `manager.go:679`, which is outside the diff.)

`releaseClaim` -> `ReleaseRecoveryClaim` NULLs `recovery_claimed_by`/`recovery_claim_expires_at` (`token/services/storage/db/sql/postgres/transactions.go:195`), and `ClaimPendingTransactions` claims any `Pending` row where `recovery_claimed_by IS NULL`.

Sequence:
1. Replica A claims tx T; `Recover` hangs.
2. At `transactionTimeout`, A releases the claim. The abandoned goroutine is still inside `Recover` and may still reach `Commit`.
3. On its next sweep replica B claims T and runs its own `Recover` -> `Commit`, concurrently with A's abandoned attempt.

The new `inFlight` guard cannot help — it is process-local. And this is precisely the failure mode `validateConfig`'s `transactionTimeout < leaseDuration` check says it prevents ("an attempt that outlives its own claim lease can be reclaimed by another replica mid-flight"), but the code drops the claim at `transactionTimeout` rather than holding it to `leaseDuration`, so the lease provides no protection at all.

Consequence is not just a duplicate log line: `AppendValid`'s dedup is a non-transactional `TransactionExists` read (`token/services/tokens/tokens.go:128`), so two concurrent commits can both observe `exists == false` and both append the same tokens.

Suggested fix: keep the claim held (or re-assert/renew it) for the lease remainder while a local attempt is still in flight, and release only when the goroutine actually returns.

token/services/storage/services/recovery/manager.go
```diff
@@ -356,6 +443,153 @@ func (m *Manager) worker(ctx context.Context, wg *sync.WaitGroup, work <-chan *t
...
+	go func() {
+		defer m.clearInFlight(txID)
+		resultCh <- m.handler.Recover(recoverCtx, txID)
+	}()
+
+	select {
+	case err := <-resultCh:
+		return err
+	case <-recoverCtx.Done():
+		return preferResult(resultCh, recoverCtx.Err())
```
AkramBitar: **MEDIUM — the deadline can fire inside the handler's commit tail, leaving a durably-`Confirmed` transaction whose status event is dropped.**

In `Commit` (`token/services/ttx/finality/listener.go:219-243`):
1. `tx.SetStatus(ctx, ...)` succeeds just before the deadline.
2. `tx.Commit()` takes no context and still succeeds — the row is now durably `Confirmed`.
3. `ttxDB.NotifyStatus(ctx, ...)` then runs on the already-expired `recoverCtx`. `StatusSupport.safeSend` selects on `<-event.Ctx.Done()` (`token/services/storage/db/common/status.go:130`) and drops the event — deterministically when the listener channel buffer is full.

Meanwhile `recoverTransaction` classifies the attempt as `DeadlineExceeded`, counts a sweep failure, and writes `status_message = "recovery failed: context deadline exceeded"` over a row that is now `Confirmed`. Finality waiters only wake via fallback polling.

`preferResult` cannot cover this — `Recover` has not returned yet.

Suggested fix: use a non-cancellable context (`context.WithoutCancel`) for the post-commit publication path. That lives in `listener.go`, so a follow-up PR is probably the right home.

token/services/storage/services/recovery/manager.go
```diff
@@ -22,12 +22,29 @@ import (
...
 const (
-	defaultLockID        int64 = 0x74746b7265636f76
-	defaultBatchSize           = 100
-	defaultWorkers             = 4
-	defaultLeaseDuration       = 30 * time.Second
+	defaultLockID             int64 = 0x74746b7265636f76
+	defaultBatchSize                = 100
+	defaultWorkers                  = 4
+	defaultLeaseDuration            = 30 * time.Second
+	defaultTransactionTimeout       = 10 * time.Second
```
AkramBitar: **MEDIUM — a `10s` default can livelock recoveries that legitimately exceed it, where pre-PR they completed.**

`TTXRecoveryHandler.Recover` does a ledger status query, then `GetTokenRequest` + `ProcessTokenRequest` (full token-request deserialize/hash) + `AppendValid` + `SetStatus` + commit — all on the passed context.

Scenario: a large ZK token request whose recovery genuinely takes 12s on a loaded node. Every sweep starts it, the deadline cancels it at 10s, `Commit` rolls back, the claim is released with a failure, and the next sweep repeats identically. The transaction never confirms. Before this PR it would have.

There is no escalating timeout, no backoff, and no give-up path — the only signal is the log escalation, which reads as "stuck peer" rather than "timeout too tight".

Suggested fix: raise the default well above realistic slow-path latency, and/or let a repeatedly-timing-out tx escape via one bounded no-deadline retry.

token/services/storage/services/recovery/manager.go
```diff
@@ -208,12 +260,23 @@ func (m *Manager) validateConfig() error {
...
 	case m.config.LeaseDuration <= 0:
 		return errors.Errorf("invalid recovery lease duration [%s]", m.config.LeaseDuration)
+	case m.config.TransactionTimeout < 0:
+		return errors.Errorf("invalid recovery transaction timeout [%s]", m.config.TransactionTimeout)
+	// A timeout that can outlive its own claim lease lets a second replica
+	// claim and recover the same transaction concurrently, so reject it at
+	// startup alongside the other invariants. TransactionTimeout == 0 means
+	// "unbounded" (opted out, see LoadConfig) and is exempt: there is no
+	// bound to compare against.
+	case m.config.TransactionTimeout > 0 && m.config.TransactionTimeout >= m.config.LeaseDuration:
```
AkramBitar: **LOW — the enforced invariant is not the one the docs say matters, and the check adds a new fatal-startup path.**

Two separate problems:

1. **Wrong invariant.** This enforces `transactionTimeout < leaseDuration`, but the PR's own docs state the relationship that actually matters is `leaseDuration > (batchSize / workerCount) x transactionTimeout` — and the shipped defaults violate *that* one by roughly 8x (`100/4 x 10s = 250s` against a `30s` lease). So the check rejects valid configurations while giving false assurance for the default one.

2. **New hard startup failure.** An operator who sets `transactionTimeout: 30s` against the default `30s` lease now fails `manager.Start()`, and `createRecoveryManager` (`token/services/network/fabric/network.go:528`) propagates that as a fatal error for the whole TMS. That is a behaviour regression for existing configs.

Suggested fix: check the documented invariant, or drop the check — and either way make it a `Warn` rather than a fatal. An advisory config relationship should not prevent a node from starting.

token/services/storage/services/recovery/manager.go
```diff
@@ -289,10 +360,26 @@ func (m *Manager) recoverTransactions(ctx context.Context) error {
 	// against len(records): a cancelled fan-out leaves the tail undispatched,
 	// and those claims never reach errCh. Attributing them to len(records)-failures
 	// would report never-attempted work as succeeded.
-	if failures > 0 || dispatched < len(records) {
-		m.logger.Warnf("completed recovery sweep: claimed=%d, dispatched=%d, succeeded=%d, failed=%d",
-			len(records), dispatched, dispatched-failures, failures)
-	} else {
+	succeeded := dispatched - failures - skipped
+	switch {
+	case skipped > 0 && skipped == dispatched && dispatched == len(records):
```
AkramBitar: **LOW — a single nil entry in `records` silently disables this new "all skipped, no progress made" warning.**

`fanOut` (`manager.go:410-412`, outside the diff) `continue`s past nil claims *without* incrementing `dispatched`. So with any nil entry, `dispatched < len(records)` always holds, `dispatched == len(records)` in this guard can never be true, and a fully-wedged sweep degrades to the ordinary breakdown line in the `case` below — exactly the misreport this guard was added to prevent, in the other direction.

Defensive-only in practice (`ClaimPendingTransactions` shouldn't return nils), but it does mean the guard has a silent off-switch.

Suggested fix: count nil-skips into `dispatched`, or compare against a non-nil count rather than `len(records)`.

token/services/storage/services/recovery/manager.go
```diff
@@ -371,8 +605,44 @@ func (m *Manager) worker(ctx context.Context, wg *sync.WaitGroup, work <-chan *t
...
+	// Recover call, and release the claim so the row does not stay locked to
+	// this owner. Counted toward the same consecutive-timeout tally as an
+	// actual DeadlineExceeded below, since a skip still means the transaction
+	// is stuck.
+	if errors.Is(err, errRecoveryInFlight) {
+		count := m.recordTimeout(txID)
+		m.logger.Debugf("recovery: tx [%s] skipped, a previous attempt is still running past transactionTimeout (%d consecutive timeout(s))", txID, count)
+		m.maybeAlertStuckTransaction(txID, count)
+
+		if releaseErr := m.releaseClaim(ctx, txID, "skipped: previous recovery attempt still in flight"); releaseErr != nil {
```
AkramBitar: **LOW — a permanently-hung transaction is skipped on every subsequent sweep forever, with no status escape hatch.**

Contrast the `NotFound` -> `Orphan` promotion added for exactly this shape of problem: there, a row that can never make progress eventually gets a terminal status. Here it never does.

Two consequences beyond the row itself:

- `errSkippedInFlight` never reaches `firstErr`, so `recoverTransactions` returns `nil` and the sweep reports **success** to any health check driven off its error. Only the log distinguishes a fully-wedged sweep from a clean one.
- The row permanently occupies a slot in the `ORDER BY stored_at ASC LIMIT batchSize` claim window. If a peer outage hangs `batchSize` status queries, no *other* pending transaction can be recovered until those calls return or the process restarts.

Neither doc page mentions the batch-slot consequence. A distinct terminal status for abandoned attempts would address both — reasonable as a follow-up rather than in this PR.

AkramBitar (commented):
@atharrva01 

The three items from my previous review are all fixed in `086205690` — sweep-summary guard, the `recovery.md` claim-release contradiction, and the `configuration.md` annotated block. Thanks, and the new tests are good.

This pass reviewed the full PR against the new head. Tests pass clean under `-race`, including 4x repeats of the new timing-sensitive tests. Six findings below; one is more serious than anything in the previous rounds.

Findings 1, 2 and 6 are three symptoms of a single unresolved decision: *what happens to a row whose recovery attempt cannot be cancelled?* Today the timeout abandons a goroutine it cannot stop, and then releases the DB claim that was the only cross-process mutual exclusion. Each round has patched a symptom of that — the process-local `inFlight` guard, then the sweep-summary wording, then the log escalation — while release-while-still-running stayed in place.

I'd suggest deciding the semantics once rather than converging incrementally. Concretely: **do not release the claim while a local attempt is still in flight** — hold it, or renew the lease, and release only when the goroutine actually returns. That single change:

- removes the cross-replica concurrent `Commit` (finding 1),
- makes the `transactionTimeout < leaseDuration` check mean something,
- demotes `inFlight` to belt-and-braces instead of the sole protection,
- gives a natural place to mark the row with a distinct status so it stops occupying a batch slot forever (finding 6).

**Also worth changing alongside it (findings 3 and 4):** raise the `10s` default well above realistic slow-path latency; replace `validateConfig`'s check with the invariant the docs actually document, or drop it; and make that check a **Warn, not a fatal** — failing whole-TMS startup on an advisory config relationship is the wrong severity and is a live regression for anyone running `transactionTimeout: 30s` against the default `30s` lease.

**Suggested follow-ups rather than this PR:** the `NotifyStatus`/`safeSend` commit-tail drop (finding 2 — the real fix is `context.WithoutCancel` in `token/services/ttx/finality/listener.go`, a different package), and an `Orphan`-style escape for permanently-hung rows.

Two findings anchor slightly off their exact site because the line in question is not in the diff; each comment says where the real code is.

AkramBitar (changes requested):

AkramBitar (approved):
LGTM

token/services/storage/services/recovery/manager.go
Effi-S: In recoverTransaction, when the handler doesn't finish (!finished), the claim is no longer released. See as well manager.go:828-835

PR #2288, opened 2026-08-21
feat(evm): endorsers return the delta, the initiator stops building one (#2229)

atharrva01:
This restores #2229. It merged into `feature/evm-network-driver` on 08-20, then about an hour later a rebase from a stale local branch overwrote the branch history and dropped it (the containedctx lint fix from the same period got dropped too, but that one already made it back in independently via #2231/#2232). Confirmed the tip was missing it, `NewInitiator` still had the old `builder DeltaBuilder` param before this PR.

Cherry-picked the original squash-merge commit onto the current tip, plus the one follow-up fix it needs (`driver_test.go`'s stale `TMS` field on `FactoryConfig`, since #2231/#2232 touched that file independently of #2229). Same content that already got reviewed and merged once.

`go build`/`go vet`/`gofmt`/`go test -race`/`golangci-lint`/`go mod tidy` all clean.

### DIFF
--- a/docs/services/network-ethereum.md
+++ b/docs/services/network-ethereum.md
@@ -310,16 +310,16 @@ sequenceDiagram
         E1->>E1: Validate token request
         E1->>E1: Compute state delta
         E1->>E1: Sign state delta
-        E1-->>Driver: Signature 1
+        E1-->>Driver: State delta + signature 1
     and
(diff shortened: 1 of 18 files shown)


adecaro (approved):
LGTM

PR #2289, opened 2026-08-21
docs(evm): drop the implementation plan, document the real endorsement flow

atharrva01:
keeping the design doc, drop the plan.

The implementation plan was a week-by-week build tracker. All 8 weeks are done (the one unchecked item, fabric-x-evm's superseded-tx handling, is explicitly out of scope for v1), so it's not carrying its weight anymore, same category of file that's been flagged for removal elsewhere in this repo. Dropped it and fixed the two places that pointed at it (design.md's header note and doc.go).

Kept the design doc as is, it's still the only place the finalized design decisions live.

Didn't link either file into docs/README.md or docs/services.md directly, neither Fabric nor FabricX cross-link their package internals like that. Instead, updated network-ethereum.md itself, its contract-side walkthrough was accurate but the driver side still had the original pre-implementation sketch. Worst part: the old sketch had the initiator picking `endorsements[0].Delta`, which is the exact thing the real initiator was changed to stop doing (building its own delta instead of relying on what endorsers actually signed, see design.md §6.5). Swapped that for the real interfaces and the actual responder/initiator flow, and added a short finality section that was just missing before.

Also fixed network.md's "See Also" list, which linked Fabric and FabricX but not Ethereum even though it's referenced earlier in the same file.

### Not touched
Root `plan.md` is a separate file outside this PR's scope, and it's already sitting at 0 bytes on this branch from an earlier rebase, so nothing there is actually dangling.


adecaro (approved):
LGTM

PR #2294, opened 2026-08-23
fix(fabricx): dynamically fetch namespace version for PP deployment

### LINKED ISSUE #2256: fabricx: PP deployment tx hardcodes NsVersion 0 — redeploy rejected once the namespace policy was ever updated
## Summary

`token/services/network/fabricx/tms/deployer.go` (`createPublicParametersTx`) builds the public-parameters deployment transaction with a hardcoded `NsVersion: 0`. Verified present on `main` as of 2026-08-18.

## Effect

On a chain whose token namespace endorsement policy has been updated at least once (namespace version ≥ 1, e.g. after adding a second endorsing organization via `fxconfig namespace update`), the deployment transaction is invalidated by the committer. This blocks the natural governance sequence "add an org to the endorsement policy, then rotate auditor keys in the public parameters" — the PP redeploy triggered by the endorser's `Init` can never commit.

## Reproduction (live, Fabric-X 4-party network)

1. `fxconfig namespace update` the token namespace policy (version 0 → 1).
2. Change the local PP file (e.g. `tokengen update` adding an auditor) and trigger the endorser's Init → the deployment tx fails finality as invalid.
3. Hand-crafting the byte-identical transaction with `ns_version` set to the actual value (1) and submitting it through fxconfig endorse/submit **commits fine**, confirming the hardcoded field is the only problem.

## Suggested fix

Query the namespace's current version before building the deployment transaction and set `NsVersion` accordingly.

Related: `pp/versionkeeper.go` has a sibling issue (endorser restart after a PP update — filed separately).

SurbhiAgarwal1:
## Summary
This PR fixes a bug in `token/services/network/fabricx/tms/deployer.go` where the public-parameters deployment transaction was hardcoding `NsVersion: 0`. This caused deployment transactions to be invalidated by the committer on chains where the token namespace endorsement policy had been updated at least once (e.g., after adding a new endorsing organization).

## Fix Details
- Modified `PublicParametersService` to implement a new `FetchNamespaceVersion` function which dynamically queries the current namespace version.
- Updated `deployPublicParametersRaw` to query and fetch the namespace's current version before building the deployment transaction.
- Passed the dynamically fetched `nsVersion` into `createPublicParametersTx` instead of hardcoding `0`. 
- Added a fallback to default to `0` if the version cannot be fetched, ensuring backward compatibility.
- Added comprehensive unit tests in `deployer_test.go` to ensure `NsVersion` is properly populated dynamically.

Fixes #2256


AkramBitar (commented):
@SurbhiAgarwal1,

Thanks for taking this on — fetching the real namespace version is the right fix for #2256, and the code reads cleanly.

My main concern is that the version can still silently end up as `0`, which is exactly the bug being fixed:

1. **Two silent fallbacks to 0** — a failed type assertion and a swallowed fetch error (`deployer.go:126`, `:130`). Either one reproduces #2256 with no diagnostics.
2. **The test doesn't cover the fix** — it never calls `deployPublicParametersRaw`, so reverting the change leaves it green.
3. **A second source of truth for `NsVersion`** — `GetNamespacePolicies().Version` here vs `GetState("_meta", ns).Version` everywhere else in the stack.

Details inline. 1 and 2 feel like blockers; 3 is worth a decision.

token/services/network/fabricx/tms/deployer.go
```diff
@@ -122,7 +122,19 @@ func (s *deployerService) deployPublicParameters(tmsID token.TMSID) error {
 // deployPublicParametersRaw constructs a public parameters transaction and
 // submits it to the network.
 func (s *deployerService) deployPublicParametersRaw(tmsID token.TMSID, ppRaw []byte) error {
-	tx, err := s.createPublicParametersTx(ppRaw, tmsID.Namespace)
+	var nsVersion uint64
+	if fetcher, ok := s.ppFetcher.(interface {
```
AkramBitar: The fix only fires if `s.ppFetcher` happens to satisfy this anonymous interface — there's no `else` and no log. `ppFetcher` is typed `fabric.NetworkPublicParamsFetcher`, and that interface exists precisely so other implementations can be plugged in (e.g. `fabric.Driver.defaultPublicParamsFetcher`, `fabricx/driver.go:145`, `fabricx/network.go:44`). The moment anything other than `*pp.PublicParametersService` is wired here, `nsVersion` silently stays `0` and #2256 is back with zero diagnostics.

Since `NewTMSDeployerService` already takes the concrete `*pp.PublicParametersService`, could we either type the field concretely, or add `FetchNamespaceVersion` to `NetworkPublicParamsFetcher` and drop the assertion? Then the compiler guarantees it.

token/services/network/fabricx/tms/deployer.go
```diff
@@ -122,7 +122,19 @@ func (s *deployerService) deployPublicParameters(tmsID token.TMSID) error {
...
+		FetchNamespaceVersion(network cdriver.Network, channel cdriver.Channel, namespace cdriver.Namespace) (uint64, error)
+	}); ok {
+		ver, err := fetcher.FetchNamespaceVersion(tmsID.Network, tmsID.Channel, tmsID.Namespace)
+		if err != nil {
```
AkramBitar, line 133: This downgrades a real failure to a warning and then submits with `NsVersion: 0`. On a transient query-service failure (gRPC timeout, committer restart, query service not yet reachable) against a chain whose policy version is >= 1, the committer invalidates the tx and the operator sees an opaque submit/finality error instead of the root cause.

I don't think the backward-compat fallback is needed: a namespace at its initial policy version already reports `0`, so there's nothing to preserve. Suggest wrapping and returning the error instead.

token/services/network/fabricx/tms/deployer.go
```diff
@@ -122,7 +122,19 @@ func (s *deployerService) deployPublicParameters(tmsID token.TMSID) error {
...
+			logger.Warnf("Failed fetching namespace version for [%s], defaulting to 0: %v", tmsID, err)
+		} else {
+			nsVersion = ver
+		}
+	}
+
+	tx, err := s.createPublicParametersTx(ppRaw, tmsID.Namespace, nsVersion)
```
AkramBitar: Minor / optional: small TOCTOU window here — the version is read, then the tx is built and submitted with no retry. If a namespace policy update commits in between (adding an endorsing org during rollout, i.e. the very scenario this PR targets), the tx is invalidated and `DeployTMSs` fails the whole loop at line 92. A single retry on a version-mismatch invalidation would make this robust.

token/services/network/fabricx/tms/deployer_test.go
```diff
@@ -0,0 +1,39 @@
...
+	return 2, nil
+}
+
+func TestCreatePublicParametersTxDynamicNsVersion(t *testing.T) {
+	s := &deployerService{
+		ppFetcher:     &mockFetcherWithVersion{},
+		keyTranslator: &keys.Translator{},
+	}
+
+	tx, err := s.createPublicParametersTx([]byte("raw-pp"), "test-ns", 2)
```
AkramBitar: This test passes the literal `2` and then asserts `NsVersion == 2`, so it only proves that a parameter is copied into a struct field. `mockFetcherWithVersion` is assigned to `ppFetcher` but never consulted, because `deployPublicParametersRaw` — where the type assertion, the version plumbing and the error fallback all live — is never invoked.

Concretely: reverting `deployPublicParametersRaw` to pass `0` leaves this test green, so it can't catch a regression of #2256.

Could we drive `deployPublicParametersRaw` through a stub `Submitter` and assert `NsVersion` on the captured tx, plus cases for (a) `FetchNamespaceVersion` returning an error and (b) a fetcher that doesn't implement the method?

token/services/network/fabricx/pp/service.go
```diff
@@ -64,6 +64,28 @@ func (f *PublicParametersService) Fetch(network driver.Network, channel driver.C
...
+	}
+	if policies != nil {
+		for _, p := range policies.GetPolicies() {
+			if p.GetNamespace() == namespace {
+				return p.GetVersion(), nil
+			}
+		}
+	}
+
+	return 0, nil
```
AkramBitar: Falling out of the loop returns `(0, nil)`, which makes "this namespace has no registered policy" indistinguishable from "version 0". If the TMS namespace was never registered on the chain (config typo, wrong channel, policy registration not yet committed), the deployer treats it as a legitimate version 0, builds and submits the tx, and it's rejected — with nothing anywhere pointing at the missing policy.

Suggest returning a distinguishable result (an error, or a `found bool`) and logging it, so a misconfigured namespace surfaces at deploy time.

token/services/network/fabricx/pp/service.go
```diff
@@ -64,6 +64,28 @@ func (f *PublicParametersService) Fetch(network driver.Network, channel driver.C
...
 }
 
+// FetchNamespaceVersion retrieves the current namespace version from the query service
+// for the specified network, channel, and namespace.
+func (f *PublicParametersService) FetchNamespaceVersion(network driver.Network, channel driver.Channel, namespace driver.Namespace) (uint64, error) {
+	qs, err := f.qsProvider.Get(network, channel)
+	if err != nil {
+		return 0, errors.Wrapf(err, "failed getting query service")
+	}
+	policies, err := qs.GetNamespacePolicies()
```
AkramBitar: This introduces a second source of truth for the same field. Everywhere else in this stack a fabricx `TxNamespace.NsVersion` is derived from `queryService.GetState("_meta", ns).Version` (FSC `platform/fabricx/core/vault/vault.go`, `rwSetWrapper.Bytes()` -> `marshal.go:148`); here it comes from `GetNamespacePolicies()` -> `PolicyItem.Version`. If those two counters ever diverge, PP deployment txs get invalidated again while ordinary token txs keep committing, which would be very hard to trace.

`qs.GetState("_meta", namespace)` is also a targeted single-key lookup rather than pulling every namespace's policy over gRPC, and it reuses the pattern already used in `Fetch` just above. Worth considering?

Nit while you're here: the `if policies != nil` guard below is dead — generated protobuf getters are nil-receiver-safe, so `policies.GetPolicies()` is already safe to range over.

AkramBitar (changes requested):

token/services/network/fabricx/tms/deployer.go
SurbhiAgarwal1: Hi @AkramBitar, added a single retry in deployPublicParametersRaw to handle the TOCTOU window — on submit failure, the version is re-fetched and the tx is resubmitted once. Added a test that asserts the retry uses the refreshed version. Thanks for the suggestion!

token/services/network/fabricx/tms/deployer_test.go
```diff
@@ -0,0 +1,204 @@
...
+// directly, allowing tests to drive deployPublicParametersRaw without needing
+// a real *pp.PublicParametersService (which requires a live qsProvider).
+type deployerServiceShim struct {
+	mock          ppVersionFetcher
+	sub           Submitter
+	keyTranslator *keys.Translator
+}
+
+func (s *deployerServiceShim) deployPublicParametersRaw(tmsID token.TMSID, ppRaw []byte) error {
+	const maxAttempts = 2
```
Effi-S: @SurbhiAgarwal1,
reimplementing the production method means we don't actually have any regression protection. 
This test should use the actual code

token/services/network/fabricx/pp/service.go
```diff
@@ -64,6 +65,37 @@ func (f *PublicParametersService) Fetch(network driver.Network, channel driver.C
 	return value.Raw, nil
 }
 
+// FetchNamespaceVersion retrieves the current namespace policy version from the
+// ledger using qs.GetState("_meta", namespace). This is the same key and encoding
+// used by the fabricx vault marshaller (platform/fabricx/core/vault/marshal.go) to
+// derive NsVersion for ordinary token transactions — a single source of truth.
+//
+// Returns an error when the namespace has no "_meta" entry so a misconfigured or
+// unregistered namespace is surfaced at deploy time rather than silently submitting
```
Effi-S: nit: godoc was not updated to reflect latest changes

token/services/network/fabricx/tms/deployer.go
```diff
@@ -120,19 +121,45 @@ func (s *deployerService) deployPublicParameters(tmsID token.TMSID) error {
...
+		if err != nil {
+			return err
+		}
+
+		tx, err := s.createPublicParametersTx(ppRaw, tmsID.Namespace, nsVersion)
+		if err != nil {
+			return err
+		}
+
+		err = s.nsSubmitter.Submit(tmsID.Network, tmsID.Channel, tx)
```
Effi-S, line 147: `Submit` blocks on finality and can return a transient error (e.g. "max retries reached") even when the tx eventually commits. The loop then re-fetches the now-bumped version and resubmits the same PP deployment against an already-initialized namespace — a conflicting/duplicate submission. Scope the retry to version-mismatch/validation failures rather than every error.

token/services/network/fabricx/pp/service.go
```diff
@@ -64,6 +65,37 @@ func (f *PublicParametersService) Fetch(network driver.Network, channel driver.C
...
+	if value == nil {
+		logger.Debugf("namespace [%s] has no _meta entry on [%s/%s] — returning initial version 0", namespace, network, channel)
+
+		return 0, nil
+	}
+
+	// Version bytes are encoded as a protobuf varint — same as UnmarshalVersion in
+	// platform/fabricx/core/vault/marshal.go.
+	version, _ := protowire.ConsumeVarint(value.Version)
+
```
Effi-S, line 97: Should Check the returned length < 0 and surface an error.

Effi-S (changes requested):

AkramBitar:
Hello @SurbhiAgarwal1 

Thanks a lot for all the efforts that you put in this PR.
Could you please have a look why these tests fails. 
Also, did you have the chance to resolve all the above comments? Is that PR ready for additional review round? 

Regards,
Akram

PR #2295, opened 2026-08-23
fabricx: PrepareNamespace panics on unknown backend topology type

### LINKED ISSUE #2039: fabricx: PrepareNamespace panics on unknown backend topology type
### Summary

`Backend.PrepareNamespace`'s type switch over `tms.BackendTopology` only special-cases `*fabrictopology.Topology` and `*fabricx.Topology`; the `default` branch panics instead of returning an error.

### Where

`integration/nwo/token/fabricx/factory.go:45`:

```go
switch n := tms.BackendTopology.(type) {
case *fabrictopology.Topology:
    ...
case *fabricx.Topology:
    ...
default:
    panic(fmt.Sprintf("unknown backend network type %T", n))   // factory.go:45
}
```

### Impact

Any new or misconfigured `BackendTopology` implementation reaching this code — for example a future backend added to the test topology framework, or a caller wiring up a TMS with the wrong topology type — crashes the entire test-network setup process (a shared, expensive operation building the whole Fabric network topology), rather than failing just that one TMS with a diagnosable error.

### Reproduction

```go
func TestPrepareNamespace_UnknownBackendTypePanics(t *testing.T) {
    b := &Backend{}
    tms := &tokentopology.TMS{..., BackendTopology: unknownBackendTopology{}}
    require.Panics(t, func() { b.PrepareNamespace(tms) })
}
```

### Suggested fix

Replace the `panic` in the `default` branch with a returned error (`PrepareNamespace`'s signature would need to support returning one, or the caller needs a recoverable path), so an unsupported topology type produces a diagnosable failure instead of crashing network setup.

### Severity

MEDIUM

Effi-S:
Fixes #2039

### Summary

`Backend.PrepareNamespace`'s type switch over `tms.BackendTopology` only special-cases `*fabrictopology.Topology` and `*fabricx.Topology`; the `default` branch panics instead of returning an error.

### Where

`integration/nwo/token/fabricx/factory.go:45`:

```go
switch n := tms.BackendTopology.(type) {
case *fabrictopology.Topology:
    ...
case *fabricx.Topology:
    ...
default:
    panic(fmt.Sprintf("unknown backend network type %T", n))   // factory.go:45
}
```

### Impact

Any new or misconfigured `BackendTopology` implementation reaching this code — for example a future backend added to the test topology framework, or a caller wiring up a TMS with the wrong topology type — crashes the entire test-network setup process (a shared, expensive operation building the whole Fabric network topology), rather than failing just that one TMS with a diagnosable error.

### Reproduction

```go
func TestPrepareNamespace_UnknownBackendTypePanics(t *testing.T) {
    b := &Backend{}
    tms := &tokentopology.TMS{..., BackendTopology: unknownBackendTopology{}}
    require.Panics(t, func() { b.PrepareNamespace(tms) })
}
```

### Suggested fix

Replace the `panic` in the `default` branch with a returned error (`PrepareNamespace`'s signature would need to support returning one, or the caller needs a recoverable path), so an unsupported topology type produces a diagnosable failure instead of crashing network setup.

### Severity

MEDIUM

### DIFF
--- a/integration/nwo/token/fabricx/factory.go
+++ b/integration/nwo/token/fabricx/factory.go
@@ -7,7 +7,6 @@ SPDX-License-Identifier: Apache-2.0
 package fabricx
 
 import (
-	"fmt"
 	"time"
 
 	"github.com/LFDT-Panurus/panurus/integration/nwo/token/fabric"
@@ -42,7 +41,7 @@ func (b *Backend) PrepareNamespace(tms *tokentopology.TMS) {
 
 		addNamespace(n.Topology, tms, orgs...)
 	default:
-		panic(fmt.Sprintf("unknown backend network type %T", n))
+		gomega.Expect(false).To(gomega.BeTrue(), "unknown backend network type %T", n)
 	}
 }
 


integration/nwo/token/fabricx/factory.go
```diff
@@ -42,7 +41,7 @@ func (b *Backend) PrepareNamespace(tms *tokentopology.TMS) {
 
 		addNamespace(n.Topology, tms, orgs...)
 	default:
-		panic(fmt.Sprintf("unknown backend network type %T", n))
+		gomega.Expect(false).To(gomega.BeTrue(), "unknown backend network type %T", n)
```
AkramBitar, line 44: @Effi-S 

The swap from `panic(fmt.Sprintf(...))` to `gomega.Expect(false).To(gomega.BeTrue(), ...)` is a lateral move — gomega's default fail handler calls `panic` internally, so the process still crashes on an unknown topology type. The defect described in #2039 is not fixed, only the error message format changes.

The correct fix is to return an error from `PrepareNamespace`. The `Backend` interface already has `InstallPublicParams` and `UpdatePublicParams` returning errors; `PrepareNamespace` is the odd one out. Changing its signature to `PrepareNamespace(tms *TMS) error` and propagating the error up to the caller gives the diagnosable, recoverable failure the issue asks for.

Thanks, 
Akram

AkramBitar (changes requested):

integration/nwo/token/fabricx/factory.go
Effi-S: gomega framework can manage what happens and not straight up panic 

This is not production code and already uses gomega

AkramBitar (approved):

PR #2296, opened 2026-08-23
fix(evm): misclassified failures and configs that pass startup but cannot work

atharrva01:
Six bugs in the EVM network driver. Each one was reproduced against a real node before it was fixed.

### Failures reported as the wrong kind

- A broadcast the node refuses outright was classified as "node unreachable", so the caller retried a transaction the chain will never take. The ordinary way to hit this is the account that pays for gas running out of funds.
- A retry of an anchor that already landed was reported as invalid. Acting on that means discarding a transfer that is on chain and final. This is the follow-on from the lost broadcast reply fix: the submitter recovers its nonce and retries correctly now, and the retry was getting the wrong answer.
- A chain head that is momentarily absent aborted the whole token transaction. Any node may answer null for "latest" while its head is moving, and it clears within a block time.

### Configurations that pass startup but cannot work

- `contracts.tokenState` pointing at an address with no contract. An eth_call there returns empty rather than failing, so nothing on the startup path noticed and the node came up healthy, then failed one transaction at a time.
- The same endorser identity bound twice. One node signs with the one key it holds, so the quorum is smaller than it looks, but validation and the on-chain policy check both agreed it was fine.

### Diagnosis

- A failed endorsement said only "collected 0 of 3 required". Unreachable, declined, wrong anchor, bad signature and duplicate signer all produced that sentence, and the reasons only went to debug logs. They travel with the error now, which is what made the first bug above findable at all.

### Tests

The four EVM integration suites ran one spec each, against fifteen for the Fabric dlog suite. They now run seven or nine, covering public parameter updates, malicious transactions, redeem and the token selector.

New anvil coverage for shapes nothing exercised: a real 2 of 3 quorum over multiple outputs, concurrent submissions through one account, a mined but reverted apply, and deltas carrying transfer metadata checked against the contract's own EIP-712 digest.

### Left alone on purpose

A metadata entry written with an empty value cannot be read back: `getTransferMetadata` returns the value alone, and the contract's `metadataExists` mapping has no getter. No shipped token driver writes one, and closing it properly means a contract change. There is a note in the test.

The anvil tests need `anvil` and `forge` on PATH.

### DIFF
--- a/integration/token/fungible/evm/evm_test.go
+++ b/integration/token/fungible/evm/evm_test.go
@@ -7,6 +7,7 @@ SPDX-License-Identifier: Apache-2.0
 package evm
 
 import (
+	integration2 "github.com/LFDT-Panurus/panurus/integration"
 	"github.com/LFDT-Panurus/panurus/integration/nwo/token"
 	tevm "github.com/LFDT-Panurus/panurus/integration/nwo/token/evm"
 	"github.com/LFDT-Panurus/panurus/integration/nwo/token/generators/crypto/zkatdlognoghv1"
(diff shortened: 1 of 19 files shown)


adecaro (approved):
LGTM

PR #2310, opened 2026-08-25
fix(metrics): make the Grafana overview dashboard work by default

### LINKED ISSUE #2309: Grafana overview dashboard renders No data by default: template variables have no allValue and are all sourced from the transfer service
## Problem

The Grafana overview dashboard shipped at `docs/monitoring/grafana/panurus.json` renders **No data** on a healthy node in its default state. Two defects in the template variables cause it.

**1. No `allValue` on any query variable.** All four variables — `network`, `channel`, `namespace`, `method` — are defined with `includeAll: true` but no `allValue`. When a variable resolves to an empty option list, the shipped "All" selection interpolates to the empty string, so `network=~""` matches only series with an absent or empty label — that is, nothing.

**2. All four variables are sourced from one service.** Every variable resolves against `panurus_core_common_metrics_transfer_service_operations_total`, which exports no series until the first transfer on that node. Combined with defect 1, every panel carrying the TMS filters renders No data on a node that has only issued, or only run ttx — a node that does have data.

Defect 2 also breaks `$method` even on a node that has transferred. `$method` is populated from the transfer service alone, so "All" expands to `(DeserializeTransferAction|Transfer|VerifyTransfer)`. That regex is then applied as `method=~"$method"` to the issue, tokens, auditor and upgrade series, whose label values are disjoint (`Issue|VerifyIssue|DeserializeIssueAction`, `AuditorCheck`, `Deobfuscate|Recipients|SupportedTokenFormats`, `CheckUpgradeProof|GenUpgradeProof|NewUpgradeChallenge`). Four of the five series in three panels are therefore empty by construction.

## Impact

An operator who imports the dashboard as shipped sees empty panels and has no way to tell that apart from a node that is genuinely not emitting metrics — the exact failure the metrics documentation and its guard tests exist to prevent. The guard in `token/services/metricsdoc` cannot catch it: `dashboard_test.go` validates label *names* against the registered metrics, not label *values*, so a query filtering on values that can never match still passes.

## Other defects in the same artifact

- **Unit mismatches.** Five panels mix seconds, rates and bytes on a single axis with one `unit` and an empty `overrides` array, so values render with the wrong unit — panel 501 (`"unit": "s"`) displays an append-error rate of 0.5/s as "500 ms". Same class: panels 402, 601, 602, 701.
- **Hardcoded `[5m]` windows.** 18 sites use a fixed `[5m]` while every counter panel uses `$__rate_interval`. On a deployment whose scrape interval exceeds 2.5 minutes the fixed window yields no samples and the p95 panels go blank.
- **Duplicate tag.** `"tags": ["panurus", "panurus", "lfdt"]` renders as two identical chips in the dashboard list.
- **Raw metric names in legends.** Panel 702 uses `"legendFormat": "idemix panurus_core_common_metrics_cache_level"` and the `recipient_data_cache_level` equivalent, while the third series in the same panel uses a clean label.

## Related gaps in the guard

- `token/services/metricsdoc/doc.go` states that the provider a constructor receives "is checked rather than assumed". It is not: `TestExportedNamesCarryPackagePrefix` derives the provider from the `scope` field it is meant to be verifying, so the assertion is circular. Routing a plain metrics provider past the TMS-scoped wrapper in either token driver invalidates six documented metric names and blanks six panels with the package still green.
- `TestDashboardQueriesReferenceRegisteredMetrics` scans every string in the dashboard with the metric-name regex but, unlike `extractMetricNames`, does not skip names ending in `_`. A panel description mentioning the namespace prefix in prose fails the build.
- `capturingRegisterer.Register` drives `Describe` on a spawned goroutine writing into a capacity-1 channel while asserting on the test goroutine. A failing assertion calls `runtime.Goexit()`, leaving the describer blocked on send forever and `close(descs)` never reached.
- `.golangci.yml` adds FSC's `metrics.Gauge` and `metrics.Histogram` to the repo-wide `ireturn` allow-list to accommodate two methods in one test file, so production code returning those interfaces instead of concrete types now passes lint silently.

## Version

Present in `main` as of `bdcd99db9`. The dashboard and its guard tests were added in #2111.

AkramBitar:
Fixes #2309

## Problem

The shipped Grafana overview dashboard renders **No data** on a healthy node in its default state:

```text
$network / $channel / $namespace / $method
  |
  +- all four sourced from panurus_core_common_metrics_transfer_service_operations_total
  |     -> no series until this node's first transfer
  +- includeAll: true, but NO allValue
        -> "All" (the shipped default) interpolates to ""
        -> network=~"" matches only series with an absent label -> nothing

$method, even on a node that has transferred:
  populated from the transfer service only  -> "All" = (DeserializeTransferAction|Transfer|VerifyTransfer)
  applied as method=~"$method" to issue / tokens / auditor / upgrade series
  whose label values are disjoint            -> 4 of 5 series empty by construction, in 3 panels
```

The guard in `token/services/metricsdoc` cannot see this: `dashboard_test.go` validates label *names* against the registered metrics, not label *values*, so a query filtering on values that can never match still passes.

## Change

**Dashboard (`docs/monitoring/grafana/panurus.json`)**

- `allValue: ".*"` on all four query variables, so the default "All" no longer interpolates to the empty string.
- `$method` now unions all five driver services via a `__name__` matcher instead of reading only the transfer service.
- Per-series `fieldConfig.overrides` on panels 402, 501, 601, 602 and 701, which previously mixed seconds, rates and bytes on one axis under a single `unit` (panel 501 rendered an append-error rate of 0.5/s as "500 ms").
- Hardcoded `[5m]` → `[$__rate_interval]` at 18 sites, matching every counter panel and keeping the p95 panels populated on deployments whose scrape interval exceeds 2.5 minutes.
- Duplicate `"panurus"` tag removed; panel 702 legends no longer print raw metric names.

**Guard (`token/services/metricsdoc`)**

- New `TestPlainProviderDoesNotBypassTheWrapper` closes the circular assertion `doc.go` claimed to make. It requires `d.metricsProvider` to appear exactly once per token driver, so it can reach nothing but `NewTMSProvider`. Verified to catch the regression it exists for: routing the plain provider into `NewWalletService` fails with `expected: 1, actual: 2`, where previously that silently invalidated six documented metric names with every test green.
- `dashboard_test.go` now skips bare namespace prefixes (names ending in `_`) as `extractMetricNames` already does. Also required by the new `__name__` matcher.
- `capturingRegisterer.Register` drives `Describe` on the test goroutine, so a failing assertion can no longer leave a sender blocked on a capacity-1 channel forever.
- The `newWalletService` / `NewWalletService` hop is added to the identity group's wiring list.

**Lint (`.golangci.yml`)**

- Reverts the repo-wide `ireturn` allow-list widening, which had added FSC's `metrics.Gauge` and `metrics.Histogram` to accommodate two methods in one test file. Replaced with two scoped `//nolint:ireturn`, the pattern already used for `assertFileContains` in the same file.

**Docs**

- `docs/monitoring/grafana/README.md` and `docs/development/metrics.md` describe the new variable sourcing; `README.md`'s previous claim that "the unfiltered panels still work" understated the breakage, since 10 of the 19 panels carry the TMS filters. `doc.go` documents the third leg of the wiring guard.
- Two links in `docs/development/metrics.md` read "Coverage gaps" while pointing at the `#metric-hygiene` anchor; the link text now matches the section it targets.

## Verification

- `go test ./token/services/metricsdoc/...` passes.
- `go vet` clean; `golangci-lint` 0 issues on all four touched packages; `gofmt` clean.
- The dashboard round-trips as byte-identical JSON formatting.
- Rebuilt on current `main` rather than replaying an earlier patch: `main` had drifted, including two new metrics (`cache_provision_failures_total`, `recipient_data_provision_failures_total`) and 41 new dashboard lines, so a cherry-pick would have been wrong.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -119,8 +119,6 @@ linters:
         - go.opentelemetry.io/otel/trace.Span
         - github.com/hyperledger/fabric-lib-go/common/metrics.Gauge
         - github.com/hyperledger/fabric-lib-go/common/metrics.Histogram
-        - github.com/hyperledger-labs/fabric-smart-client/platform/view/services/metrics.Gauge
-        - github.com/hyperledger-labs/fabric-smart-client/platform/view/services/metrics.Histogram
         - github.com/hyperledger-labs/fabric-smart-client/platform/common/driver.ConfigService
         - github.com/hyperledger-labs/fabric-smart-client/integration/nwo/api.ViewClient
         - github.com/hyperledger-labs/fabric-smart-client/integration/nwo/api.Platform
--- a/docs/development/metrics.md
+++ b/docs/development/metrics.md
@@ -8,8 +8,10 @@ in the SDK the way production wiring does, reads the resulting names back out of
 and compares them against `token/services/metricsdoc/testdata/metrics.golden` and against this page.
 Adding, renaming or moving a metric fails that test until this page is updated. Because a name also
 depends on *which* provider production hands the constructor, the same test pins that wiring: it
-asserts that the token drivers are the only place a TMS-scoped provider is built, so dropping or
-relocating the wrapper fails too rather than silently invalidating the names below.
+asserts that the token drivers are the only place a TMS-scoped provider is built, and that the provider
+they get from the dependency-injection container reaches nothing but that wrapper. Dropping the wrapper,
+or routing the plain provider past it, therefore fails too rather than silently invalidating the names
+below.
 
 See [Monitoring](./monitoring.md) for the wider monitoring setup and [Driver Metrics](../drivers/metrics.md)
 for how the driver instrumentation is wired. Every metric below already has a panel in the importable
@@ -52,7 +54,7 @@ Two consequences are worth internalising before writing a query or a dashboard:
 - **Some names stutter**, because the metric name repeats its package: hence
   `panurus_services_auditor_auditor_audit_duration_seconds` and
   `panurus_services_network_fabricx_finality_queue_finality_queue_pending_events`. They are correct as
-  listed; see [Coverage gaps](#metric-hygiene) for why they are not renamed here.
+  listed; see [Metric hygiene](#metric-hygiene) for why they are not renamed here.
 
 ### The TMS-scoped provider changes the prefix
 
@@ -160,7 +162,7 @@ Source: `token/services/utils/json/session/metrics.go`. Wired in `token/sdk/dig/
 
 Recorded by the auditor around `Audit()`, `Append()` and `Release()`. These metrics carry **no TMS
 labels**, so a node auditing several TMSs reports them aggregated; see
-[Coverage gaps](#metric-hygiene).
+[Metric hygiene](#metric-hygiene).
 
 | Metric | Type | Labels | Description |
 |---|---|---|---|
--- a/docs/monitoring/grafana/README.md
+++ b/docs/monitoring/grafana/README.md
@@ -14,10 +14,15 @@ queue).
    `DS_PROMETHEUS`.
 
 The dashboard declares four template variables — `network`, `channel`, `namespace` and `method` — whose
-values are discovered with `label_values` against
-`panurus_core_common_metrics_transfer_service_operations_total`. A node that has never issued or
-transferred a token exports no series for that metric, so the pickers stay empty until the first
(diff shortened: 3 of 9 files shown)


Effi-S (approved):
LGTM

PR #2311, opened 2026-08-26
feat(zkatsnark): implement zkatsnark token services and integration tests

neetance:
## Description

This PR implements the core token services and comprehensive integration testing for the `zkatsnark` driver.

### Changes
* **Token Services**: Implemented `token.Service` in `x/token/core/zkatsnark/token/service.go` and integrated it across `IssueService`, `TransferService`, and `AuditorService`.
* **Driver Updates**: Refactored `driver.go`, `transfer.go`, and `issue.go` to utilize the new token service logic effectively.
* **Network Orchestrator (NWO)**: Added support for `zkatsnarkv1` crypto and public parameters generators, updating topology and platform integration.
* **Integration Tests**: Added `zkatsnark_test.go` and `zkatsnark_suite_test.go` under `integration/token/fungible/zkatsnark` to provide end-to-end testing of the fungible token flows using `zkatsnark`.
* **Make/Dependencies**: Updated `fungible.mk`, `go.mod`, and `go.sum` to handle the new test suite and dependencies appropriately.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -142,5 +142,3 @@ require (
 	modernc.org/mathutil v1.7.1 // indirect
 	modernc.org/memory v1.11.0 // indirect
 )
-
-replace github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark => ./x/token/core/zkatsnark


token/request.go
```diff
@@ -663,7 +663,7 @@ func (r *Request) extractIssueOutputs(ctx context.Context, i int, counter uint64
 		if err != nil {
 			return nil, 0, errors.Wrapf(err, "failed getting issue action output in the clear [%d,%d]", i, j)
 		}
-		if !issuer.Equal(issueAction.GetIssuer()) {
+		if len(issuer) != 0 && !issuer.Equal(issueAction.GetIssuer()) {
```
adecaro: why do we need this?

token/services/network/fabric/tcc/main/main.go
```diff
@@ -16,6 +16,7 @@ import (
 	dlog "github.com/LFDT-Panurus/panurus/token/core/zkatdlog/nogh/v1/driver"
 	"github.com/LFDT-Panurus/panurus/token/services/logging"
 	"github.com/LFDT-Panurus/panurus/token/services/network/fabric/tcc"
+	zkatsnark "github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark/driver"
```
adecaro: I would prefer that we don't modify this part. Please, test the new driver using fabricx so we don't have to modify the chaincode. Later, we will see what to do to have also the test with fabric.

adecaro (changes requested):
please, have a look at the comments.

token/request.go
neetance: I added this as a workaround for an issue during integration tests where Deobfuscate returned an empty issuer for some token metadata. But, the issuer validation should be enforced for issue actions. I have removed this change and will instead work on the issuer being properly handled within the driver/note serialization.

token/services/network/fabric/tcc/main/main.go
neetance: Understood. I have removed the zkatsnark import and validator registration. We will test using fabricx as suggested without modifying the chaincode.

go.mod
```diff
@@ -142,3 +142,5 @@ require (
 	modernc.org/mathutil v1.7.1 // indirect
 	modernc.org/memory v1.11.0 // indirect
 )
+
+replace github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark => ./x/token/core/zkatsnark
```
adecaro: we don't need this

neetance: Removed 🙏

adecaro (approved):
LGTM

Issue #1266, opened 2025-11-17
ReceiveTransactionView timeout issue


adecaro:
When setting the default timeout of ReceiveTransactionView to 1 minute, then some of tests fail.

adecaro:
Hi @sid200727 , yes. I noticed in the past that when I set the default timeout to 1 minute in `ReceiveTransactionView` (token/services/ttx/receivetx.go), the integration tests fail. Not sure if this is still the case.

sid200727:
Hi @adecaro, thank you for the context. I'll investigate token/services/ttx/receivetx.go to check if the 1 minute timeout issue still exists. I'll report back with findings.

sid200727:
Hi @adecaro, here are my findings after investigating token/services/ttx/receivetx.go:
The bug still exists, it was masked with a workaround, not fixed.

What I found:
The current default timeout is 4 * time.Minute (line 46), but the comment above it still says "1 minutes is used as default" , a clear sign the workaround was applied without updating the comment
The root cause is never addressed: multi-party flows (issuer → recipient → auditor → ordering → finality) can exceed 1 minute under load, and 4 minutes is equally arbitrary
multisig/spend.go independently uses time.Minute * 4 ,confirming this is a known lower bound for CI, not a principled solution
WithTimeout exists on TxOptions but no call site actually uses it ,all call sites pass no opts, so the timeout is always the hardcoded default

Proposed fix:
Expose the default as a named constant (e.g. DefaultReceiveTimeout = 4 * time.Minute)
Fix the stale comment to reflect the actual value
Make the timeout configurable via the FSC config provider so operators can tune it for slow environments.

Does this approach sound right before I proceed? Happy to start with a PR for the constant and comment fix first as a safe incremental step.

sid200727:
Hi @adecaro, following up on my analysis mentioned, does the proposed approach (expose DefaultReceiveTimeout as a named constant, fix the stale comment, and make it configurable via FSC config) sound right? 
Happy to open a PR once you confirm the direction.

sid200727:
Hi @AkramBitar @adecaro! if the proposal aligns with your thoughts, can I start working on this issue ?

AkramBitar:
Hi @sid200727 — answering your original question and your latest one together.

[...]

PR #2312, opened 2026-08-26
test(ttx): assert responder timeout budget invariant

sid200727:
## Summary

Implements #1266: ReceiveTransactionView's responder timeout doesn't reliably exceed the sum of the waits it depends on, and a prior fix to the stale 4-minute comment (#1716) papered over the fact that the underlying invariant was never checked. Per @AkramBitar's breakdown on the issue, this adds the check as code instead of a comment.

## Changes

-> Documented the timeout budget in `docs/services/ttx.md`: a table of every relevant wait with file:line and value, plus the invariant `responder receive > sig fan-out + audit + approval + slack`
-> Added `receivetx_timeout_test.go`, asserting that invariant arithmetically against the real constants (`receivetx.go:42`, `endorse.go:111`, `auditor.go:189`, `network/fabric/endorsement/fsc/initiator.go:99`)
-> Did not touch the responder timeout constant itself — the right value/slack is a product decision I'd like input on rather than picking one unilaterally

## Testing

Test currently fails intentionally: the 4 min responder budget equals but does not exceed the 4 min sum of its FSC-endorsement-mode components (1 min sig fan-out + 1 min audit + 2 min approval), leaving zero slack. Per Akram: "that failure is the bug, please don't paper over it." Fixing the underlying constant will make it pass with no test changes needed.

Full test suite otherwise green (go build ./..., gofmt -l clean).

## Notes for the Reviewer

Happy to open the follow-up fixing `receivetx.go:42` once we agree on a value/slack margin. Also happy to file the ErrTimeout diagnostics improvement (naming step + counterparty) Akram flagged as a separate optional issue rather than bundling it here, let me know if you'd rather see either included in this PR instead.


AkramBitar (commented):
A few targeted issues with the test and doc table — all within scope of this PR.

token/services/ttx/receivetx_timeout_test.go
```diff
@@ -0,0 +1,35 @@
...
+// KNOWN FAILURE (issue #1266): as of this test's authorship, the responder budget
+// (4 min) equals — but does not exceed — the sum of its component legs in
+// FSC-endorsement mode (1 min sig fan-out + 1 min audit + 2 min approval = 4 min),
+// leaving zero slack margin. This is a real bug, not a test bug — do not "fix"
+// this test by loosening the invariant. Fix it by correcting the responder
+// timeout (or a component timeout) in the source files below, then this test
+// will pass on its own.
+func TestReceiveTransactionTimeoutBudget(t *testing.T) {
+	const (
+		responderReceiveTimeout = 4 * time.Minute // receivetx.go:42
```
AkramBitar: These constants are copies of the source magic numbers, not references to them. If someone fixes the bug by bumping `receivetx.go:42` to `5 * time.Minute`, this test stays red because it reads its own local `4 * time.Minute`. Export the constant from `receivetx.go` — e.g. `const DefaultReceiveTransactionTimeout = 4 * time.Minute` — and import it here. A source fix then flips this test green with no test edits needed.

token/services/ttx/receivetx_timeout_test.go
```diff
@@ -0,0 +1,35 @@
...
+		sigFanOutTimeout        = 1 * time.Minute // endorse.go:111
+		auditTimeout            = 1 * time.Minute // auditor.go:189
+		approvalTimeout         = 2 * time.Minute // network/fabric/endorsement/fsc/initiator.go:99
+		slack                   = 0 * time.Minute // no slack currently budgeted
+	)
+
+	requiredMinimum := sigFanOutTimeout + auditTimeout + approvalTimeout + slack
+
+	require.Greater(t, responderReceiveTimeout, requiredMinimum,
```
AkramBitar: `slack = 0` is a no-op in the arithmetic. Remove it, or give it a real non-zero value with a comment explaining the required margin.

token/services/ttx/receivetx_timeout_test.go
```diff
@@ -0,0 +1,35 @@
...
+		"responder receive timeout (%s) must exceed sig fan-out + audit + approval + slack (%s) — see issue #1266",
```
AkramBitar: With `slack = 0` this is `4min > 4min` — permanently false regardless of any source change. Use `require.GreaterOrEqual` to match the actual invariant ("budget must not be less than the sum"), and switch back to `Greater` only once `slack` has a non-zero value.

docs/services/ttx.md
```diff
@@ -571,3 +571,31 @@ The recovery service is part of the **Storage Service** and is instantiated by t
...
+every wait it depends on, or the responder times out on a transaction that was still legitimately
+in progress.
+
+**Invariant:** `responder receive > sig fan-out + audit + approval + slack`
+
+| Leg | File:Line | Value |
+|---|---|---|
+| Responder receive (the budget) | `token/services/ttx/receivetx.go:42` | 4 min |
+| Signature fan-out (per endorser) | `token/services/ttx/endorse.go:111` | 1 min |
+| Signature collection (path 1) | `token/services/ttx/collectendorsements.go:373` | 1 min |
```
AkramBitar: Wrong line. `collectendorsements.go:373` is a `logger.DebugfContext` call. The `ReceiveTypedWithTimeout(... time.Minute)` is at line **398**.

docs/services/ttx.md
```diff
@@ -571,3 +571,31 @@ The recovery service is part of the **Storage Service** and is instantiated by t
...
+| Signature collection (path 2) | `token/services/ttx/collectendorsements.go:566` | 1 min |
```
AkramBitar: Wrong line. `collectendorsements.go:566` is `entry := &remote[i]`. The timeout is at line **612**.

docs/services/ttx.md
AkramBitar: These two rows are initiator-side waits inside `collectEndorsements`, not responder-side. The responder never calls into that path — its budget only needs to cover the initiator end-to-end, already represented by the `FSC-endorsement approval` row. Including them here will mislead anyone re-deriving the arithmetic. Please remove or annotate as "initiator-internal, subsumed by approval row".

AkramBitar (commented):
One blocker: **the test passes at head**, so it does not currently catch the inversion it exists to catch. Details inline, plus a godoc displacement in `receivetx.go` and a doc/test disagreement about which invariant is the real one.

There's also a sequencing question worth settling here: once the assertion is strict, the test is legitimately red and can't land on `main` under required checks. So either this PR also bumps `receivetx.go`'s constant, or the constant fix goes first and the test follows. Happy to give a value/slack figure — just say which shape you'd prefer.

token/services/ttx/receivetx_timeout_test.go
```diff
@@ -0,0 +1,29 @@
...
+func TestReceiveTransactionTimeoutBudget(t *testing.T) {
+	const (
+		sigFanOutTimeout = 1 * time.Minute // endorse.go:111
+		auditTimeout     = 1 * time.Minute // auditor.go:189
+		approvalTimeout  = 2 * time.Minute // network/fabric/endorsement/fsc/initiator.go:99
+	)
+
+	requiredMinimum := sigFanOutTimeout + auditTimeout + approvalTimeout
+
+	require.GreaterOrEqual(t, DefaultReceiveTransactionTimeout, requiredMinimum,
```
AkramBitar: **Blocking:** this assertion is satisfied at equality, so the test is green while the bug is live.

Both sides are exactly `4m`, and `require.GreaterOrEqual` permits equality, so at head `1217da61`:

```
=== RUN   TestReceiveTransactionTimeoutBudget
--- PASS: TestReceiveTransactionTimeoutBudget (0.00s)
```

That contradicts the PR description ("Test currently fails intentionally... that failure is the bug, please don't paper over it"). It also means there are three different statements of the invariant in this PR that don't agree:

- `docs/services/ttx.md` — `responder receive > sig fan-out + audit + approval + slack` (strict, with slack)
- the docstring above — "must not be less than" (non-strict)
- this line — `GreaterOrEqual` (non-strict, no slack term)

As written, CI stays green through exactly the inversion the test is meant to guard, which is the outcome the PR sets out to prevent. `require.Greater` plus an explicit slack constant would make the assertion match the documented invariant and fail honestly until the source constant is raised.

token/services/ttx/receivetx_timeout_test.go
```diff
@@ -0,0 +1,29 @@
...
+// on: sig fan-out + audit + approval.
+//
+// KNOWN FAILURE (issue #1266): as of this test's authorship, the responder budget
+// equals the sum of its FSC-endorsement-mode components exactly, with zero slack.
+// Do not "fix" this test by loosening the invariant — fix the source constants.
+func TestReceiveTransactionTimeoutBudget(t *testing.T) {
+	const (
+		sigFanOutTimeout = 1 * time.Minute // endorse.go:111
+		auditTimeout     = 1 * time.Minute // auditor.go:189
+		approvalTimeout  = 2 * time.Minute // network/fabric/endorsement/fsc/initiator.go:99
```
AkramBitar: These are local literals rather than the constants they name. All three sites are still inline expressions on `main` (`time.Minute` at `endorse.go:111` and `auditor.go:189`, `2*time.Minute` at `initiator.go:99`), so nothing ties this arithmetic to them — raise any one of them and this test stays green while the invariant breaks, which is the drift the test exists to detect. The values are correct today; I checked all three.

The PR description says the test asserts "against the real constants (`receivetx.go:42`, `endorse.go:111`, `auditor.go:189`, ...)", but only `DefaultReceiveTransactionTimeout` is a real reference. Making that true means extracting named constants at the three sites, and the `fsc/initiator.go` one is in another package so it would need exporting — reasonable to scope out of this PR, but worth not describing as already done.

token/services/ttx/receivetx.go
```diff
@@ -33,13 +33,15 @@ func NewReceiveTransactionView(opts ...TxOption) *ReceiveTransactionView {
 // Call listens to a versioned envelope from the context's session and returns a transaction.
 // Supported message types are TypeTransaction, TypeTransactionResponse, and TypeSignatureRequest.
 // If no timeout is specified via the opts, 4 minutes is used as default.
+const DefaultReceiveTransactionTimeout = 4 * time.Minute
```
AkramBitar: The constant landed between `Call`'s doc comment and `Call` itself, so this comment now documents the constant instead: godoc will render "Call listens to a versioned envelope from the context's session and returns a transaction. Supported message types are..." on `DefaultReceiveTransactionTimeout`, and `Call` is left with no doc comment at all — AGENTS.md requires godoc on exported functions.

Moving the const above the comment block with its own one-line doc fixes both, and the "4 minutes is used as default" sentence can then refer to the constant rather than restating the value.

Minor, while here: does this need to be exported? `receivetx_timeout_test.go` is `package ttx`, so an unexported `defaultReceiveTransactionTimeout` would serve the test just as well, and public API is harder to walk back later.

docs/services/ttx.md
```diff
@@ -571,3 +571,30 @@ The recovery service is part of the **Storage Service** and is instantiated by t
...
 - [Configuration Guide - Recovery Parameters](../configuration.md), Section `Optional: token.tms.<name>.services.network.fabric.recovery`
+
+### Interactive Protocol Timeout Budget
+
+`ReceiveTransactionView` (and the `boolpolicy`/`multisig` spend responders) wait up to a fixed
+timeout for the full endorsement round-trip to complete. That budget must exceed the sum of
+every wait it depends on, or the responder times out on a transaction that was still legitimately
+in progress.
+
+**Invariant:** `responder receive > sig fan-out + audit + approval + slack`
```
AkramBitar: Two things on this section:

1. This states the invariant in strict-plus-slack form, but the test asserts `GreaterOrEqual` with no slack term (see the comment on `receivetx_timeout_test.go`). Whichever form is intended, the doc, the test docstring and the assertion should all state the same one — right now a reader gets a different answer from each.

2. The table below cites `token/services/ttx/receivetx.go:42` for the budget, but after this PR line 42 is the closing `}` of the `CompileOpts` error check — the constant is on line 36 and its use on line 44. These `file:line` references go stale as soon as anything above them shifts, and this one went stale inside its own PR. Symbol names (`DefaultReceiveTransactionTimeout`, `CollectEndorsements`) would survive edits, here and in the test's trailing comments.

AkramBitar (changes requested):

token/services/ttx/receivetx_timeout_test.go
```diff
@@ -0,0 +1,25 @@
+package ttx_test
```
AkramBitar, line 1: **1. BLOCKING (CI break): missing Apache license header.**

Every other `.go` file in the repo -- including `receivetx_test.go` right next to it -- starts with the `Copyright IBM Corp.` / `SPDX-License-Identifier: Apache-2.0` block, and `checks.mk:22-27` runs `addlicense -check` over all non-`.pb.go` `*.go` files as part of `checks-fast`.

Verified locally:

```
$ addlicense -check token/services/ttx/receivetx_timeout_test.go
token/services/ttx/receivetx_timeout_test.go
exit=1
```

The CI `checks` job will fail on this as written.

token/services/ttx/receivetx_timeout_test.go
```diff
@@ -0,0 +1,25 @@
...
+)
+
+// requiredSlack is the margin the responder budget must keep above the sum of
+// its component legs, to absorb network jitter and scheduling delay.
+const requiredSlack = 1 * time.Minute
+
+// TestReceiveTransactionTimeoutBudget asserts the invariant that the responder's
+// overall receive timeout must exceed the sum of every wait it depends on:
+// sig fan-out + audit + approval + slack. See issue #1266.
+func TestReceiveTransactionTimeoutBudget(t *testing.T) {
```
AkramBitar, line 19: **2. MEDIUM: the invariant test guards the wrong constants, so it cannot catch the regression class it advertises.**

The waits that actually elapse *inside* the responder's 6-minute window are initiator-side:

- `signRemote`'s per-endorser receive -- `collectendorsements.go:398`, still a bare `time.Minute`
- the fan-out backstop `answerCollectionTimeout` -- `collectendorsements.go:809`, 2 min

`ttx.SigFanOutTimeout` (`endorse.go:86`) is the *responder's* wait to receive a signature request. It elapses and returns *before* `EndorseView.receiveTransaction` opens the 6-min window, so it is not a summand of that budget at all.

Concrete miss: bump `answerCollectionTimeout` to 4 min, or `collectendorsements.go:398` to 3 min. This test stays green, but a responder that already signed times out at 6 min on a transaction the initiator is still legitimately collecting -- which is exactly #1266.

docs/services/ttx.md
```diff
@@ -571,3 +571,33 @@ The recovery service is part of the **Storage Service** and is instantiated by t
...
+timeout for the full endorsement round-trip to complete. That budget must exceed the sum of
+every wait it depends on, or the responder times out on a transaction that was still legitimately
+in progress.
+
+**Invariant:** `ReceiveTransactionView` (and the `boolpolicy`/`multisig` spend responders) wait up to a fixed
+timeout for the full endorsement round-trip to complete. That budget must exceed the sum of
+every wait it depends on, or the responder times out on a transaction that was still legitimately
+in progress.
+
+**Invariant:** `responder receive > sig fan-out + audit + approval + slack`
```
AkramBitar: **3. MEDIUM: the documented invariant is already violated for the multisig/boolpolicy case the doc claims to cover.**

`requestSignatures` (`collectendorsements.go:204`) recurses per multisig identity (line 233) and per boolpolicy identity (line 263), and *each* recursion level runs its own sequential `fanOut` bounded by `answerCollectionTimeout` (2 min). On top of that, `requestSignaturesOnIssues` and `requestSignaturesOnTransfers` are two sequential phases (`collectendorsements.go:114,119`).

So a transfer with two multisig owners can burn ~4 min in the signature phase alone, plus 1 min audit plus 2 min approval = 7 min, against the 6-min budget.

Separately, `ReceiveSpendTxView` (`token/services/ttx/multisig/spend.go:252`, `token/services/ttx/boolpolicy/spend.go:244`) opens its 6-min window right after ACKing the spend request -- i.e. *before* the initiator has assembled the transaction. That window therefore also has to cover token selection and zkatdlog proof generation, a leg the table omits entirely, even though the paragraph above explicitly names these responders as in scope.

docs/services/ttx.md
```diff
@@ -571,3 +571,33 @@ The recovery service is part of the **Storage Service** and is instantiated by t
...
 - [Configuration Guide - Recovery Parameters](../configuration.md), Section `Optional: token.tms.<name>.services.network.fabric.recovery`
+
+### Interactive Protocol Timeout Budget
+
+`ReceiveTransactionView` (and the `boolpolicy`/`multisig` spend responders) wait up to a fixed
+timeout for the full endorsement round-trip to complete. That budget must exceed the sum of
+every wait it depends on, or the responder times out on a transaction that was still legitimately
+in progress.
+
+**Invariant:** `ReceiveTransactionView` (and the `boolpolicy`/`multisig` spend responders) wait up to a fixed
```
AkramBitar: **4. LOW: duplicated paragraph from a bad edit.**

Lines 577-580 are repeated verbatim here at 582-585, the second copy prefixed `**Invariant:**` as though the prose *were* the invariant -- immediately followed by the real `**Invariant:**` at line 587. A reader sees two contradictory "Invariant:" labels.

docs/services/ttx.md
```diff
@@ -571,3 +571,33 @@ The recovery service is part of the **Storage Service** and is instantiated by t
...
+**Invariant:** `responder receive > sig fan-out + audit + approval + slack`
+
+| Leg | Constant | Value |
+|---|---|---|
+| Responder receive (the budget) | `DefaultReceiveTransactionTimeout` (`receivetx.go`) | 6 min |
+| Signature fan-out (per endorser) | `SigFanOutTimeout` (`endorse.go`) | 1 min |
+| Auditor signature | `AuditTimeout` (`auditor.go`) | 1 min |
+| FSC-endorsement approval | `fsc.ApprovalTimeout` (`network/fabric/endorsement/fsc/initiator.go`) | 2 min |
+
+**Known issue:** The budget carries an explicit 1-minute slack margin above the sum of its legs (1 min sig
```
AkramBitar: **5. LOW: "Known issue:" mislabels the resolved state, and the slack figure contradicts its own arithmetic.**

The line says "an explicit 1-minute slack margin" while the numbers on the same line give 6 - 4 = 2 min. The test's `requiredSlack` is a third value (1 min, used as a floor rather than as the actual margin). Three inconsistent numbers for one quantity.

token/services/ttx/endorse.go
```diff
@@ -82,6 +82,9 @@ func (s *EndorseView) Call(context view.Context) (any, error) {
 	return s.tx, nil
 }
 
+// SigFanOutTimeout bounds how long the endorser waits for a signature request.
+const SigFanOutTimeout = 1 * time.Minute
```
AkramBitar: **7. LOW: `SigFanOutTimeout` is misnamed for what it bounds.**

It bounds the responder's `ReceiveTypedWithTimeout(TypeSignatureRequest, ...)` at line 114 -- a single receive, not a fan-out -- yet the docs table labels it "Signature fan-out (per endorser)".

A maintainer who needs to lengthen the initiator's tolerance for a slow endorser will raise this constant and observe no effect, because the initiator's wait is the unnamed `time.Minute` at `collectendorsements.go:398`.

token/services/ttx/receivetx.go
```diff
@@ -30,16 +30,19 @@ func NewReceiveTransactionView(opts ...TxOption) *ReceiveTransactionView {
 	}
 }
 
+// DefaultReceiveTransactionTimeout is used when no timeout is specified via the opts.
+const DefaultReceiveTransactionTimeout = 6 * time.Minute
```
AkramBitar, line 34: **8. LOW: the 4 -> 6 min bump applies to every `ReceiveTransaction` caller, not just the endorsement round-trip that needs it.**

First-contact receives -- `integration/.../accept.go`, `AuditView`, `collectactions.go:175`, `nfttx/transaction.go:59` -- have no budget requirement of this kind, and now pin an FSC view plus session for 6 minutes per abandoned or never-sent transaction. That is a 50% increase in resource hold time on precisely the paths where the invariant is irrelevant.

Scoping the longer budget to `EndorseView.receiveTransaction` (and the two `ReceiveSpendTxView`s) via `WithTimeout` would fix #1266 without widening the blast radius.

AkramBitar (commented):
## Review: correctness pass over the responder timeout budget

One blocking CI issue, two substantive correctness findings about the invariant this PR asserts, and four lower-severity ones. Verified locally against head `02a3ae0b3`: `go test ./token/services/ttx/... -count=1` passes and the new test compiles and passes, so the issues below are the CI failure, the soundness of the asserted invariant, and the doc text -- not a build break.

Inline comments cover findings 1-5, 7 and 8. Finding 6 is below because its anchor line is not part of this diff.

### 6. `token/services/network/fabric/endorsement/fsc/setup_initiator.go:107` -- LOW: the other `CollectEndorsements` call still hard-codes `2*time.Minute`

This PR makes `ApprovalTimeout` the named source of truth that both the docs table and the new test key off, but the setup path keeps its literal:

```go
// setup_initiator.go:107
err = s.EndorserService.CollectEndorsements(ctx, tx, 2*time.Minute, s.Endorsers...)
```

Lowering or raising `ApprovalTimeout` will silently leave the setup approval at 2 min while `TestReceiveTransactionTimeoutBudget` still reports the invariant as satisfied.

AkramBitar:
## Review summary — head `a64c0ea`

Three items required before merge:

### 🔴 BLOCKING: missing Apache license header

`receivetx_timeout_test.go` still starts with `package ttx_test` at line 1 — no `Copyright IBM Corp.` / `SPDX-License-Identifier: Apache-2.0` block. `checks-fast` runs `addlicense -check` over all non-`.pb.go` `*.go` files; this will fail CI. Same blocker as finding #1 from the last review — still not applied.

### 🟡 MEDIUM: doc scope claim is broader than the invariant the test can assert

The opening sentence of the new section says the invariant covers `boolpolicy`/`multisig` spend responders. The table only accounts for a single `AnswerCollectionTimeout` (2 min). `requestSignatures` recurses per multisig/boolpolicy identity and runs sequential `fanOut` phases (`requestSignaturesOnIssues` then `requestSignaturesOnTransfers`, `collectendorsements.go:114,119`), so a transfer with two multisig owners can burn ~4 min in the signature phase alone — immediately violating the invariant the doc claims holds. Either narrow the scope statement to single-owner FSC-endorsement mode, or add an explicit note acknowledging the shortfall with a follow-up issue reference.

### 🟠 HOUSEKEEPING: squash + PR metadata

Still 9 commits (squash was explicitly requested and acknowledged). Milestone, labels, and assignee are all unset — required by AGENTS.md before merge.

---

Everything else looks good at current head: all constants are exported and referenced in the test (drift-proof), godoc placement is correct on all three sites, `setup_initiator.go` hardcoded literal is replaced, initiator-internal rows are out of the doc table, `require.GreaterOrEqual` with a non-zero `requiredSlack` is the right assertion form. Fix the three items above and this is ready to go.

sid200727:
Rebuilt the branch onto current main, the old one had unrelated commits tangled in from squash-merged PRs, so this is a fresh single commit. All three findings addressed: license header added, doc scope narrowed to single-owner FSC-endorsement mode, squashed to one.

One change beyond the review: the invariant couldn't be asserted as written, since the budget equalled the sum of its legs exactly  neither Greater nor GreaterOrEqual with non-zero slack holds at 4 min. I've raised DefaultReceiveTransactionTimeout to 5 min so it holds with 1 min slack, and updated the doc accordingly. Happy to drop that and weaken the assertion instead if you'd rather keep this test-only.

AkramBitar:
@sid200727 
Could you please rebase and resolve the conflicts?
Regards,
Akram

sid200727:
Hi @AkramBitar !
Rebased onto current main. Upstream had extracted the per-signer loop into handleSignatureRequestFor, so the fan-out timeout applies there now and the doc's line reference is updated to match.
Thank You 🙏

token/services/ttx/auditor.go
```diff
@@ -169,6 +169,9 @@ func newAuditingViewInitiator(tx *Transaction, local, skipAuditorSignatureVerifi
 // Call initiates an auditing session (local or remote), sends the transaction to the auditor,
 // receives the auditor's signature, verifies it, and adds it to the transaction.
 // Returns the session used for communication.
+// AuditTimeout bounds how long the initiator waits for the auditor's signature.
+const AuditTimeout = 1 * time.Minute
```
AkramBitar, line 173: The const is spliced into the *middle* of `Call`'s Godoc block, between the existing three-line comment and the function it documents:

```go
// Call initiates an auditing session (local or remote), sends the transaction to the auditor,
// receives the auditor's signature, verifies it, and adds it to the transaction.
// Returns the session used for communication.
// AuditTimeout bounds how long the initiator waits for the auditor's signature.
const AuditTimeout = 1 * time.Minute

func (a *AuditingViewInitiator) Call(context view.Context) (any, error) {
```

Concretely, `go doc ttx.AuditTimeout` now prints "Call initiates an auditing session (local or remote), sends the transaction to the auditor…" as the documentation of a `time.Duration`, and `go doc ttx.AuditingViewInitiator.Call` prints nothing at all — which also trips the repo rule that all exported functions must have Godoc.

Moving the const *above* the `// Call …` block fixes both, and matches what the sibling hunks in `endorse.go:85` and `receivetx.go:33` already do correctly.

token/services/ttx/receivetx.go
```diff
@@ -30,6 +30,10 @@ func NewReceiveTransactionView(opts ...TxOption) *ReceiveTransactionView {
 	}
 }
 
+// DefaultReceiveTransactionTimeout bounds the full endorsement round-trip a
+// responder waits for when no timeout is given via opts.
+const DefaultReceiveTransactionTimeout = 5 * time.Minute
+
 // Call listens to a versioned envelope from the context's session and returns a transaction.
 // Supported message types are TypeTransaction, TypeTransactionResponse, and TypeSignatureRequest.
 // If no timeout is specified via the opts, 4 minutes is used as default.
```
AkramBitar: Stale doc: the default is now 5 minutes (line 35 / line 46), but this line still says 4. This is the single line an operator reads when diagnosing a responder `ErrTimeout`, so it now contradicts the very code the PR changed — exactly the drift the PR set out to eliminate.

Suggest referencing the constant so it can't go stale again:

```go
// If no timeout is specified via the opts, DefaultReceiveTransactionTimeout is used.
```

token/services/ttx/receivetx_timeout_test.go
```diff
@@ -0,0 +1,26 @@
...
+	"github.com/LFDT-Panurus/panurus/token/services/network/fabric/endorsement/fsc"
+	"github.com/LFDT-Panurus/panurus/token/services/ttx"
+	"github.com/stretchr/testify/require"
+)
+
+// TestReceiveTransactionTimeoutBudget asserts the invariant that the responder's
+// overall receive timeout must exceed the sum of every wait it depends on:
+// sig fan-out + audit + approval + slack. See issue #1266.
+func TestReceiveTransactionTimeoutBudget(t *testing.T) {
+	requiredMinimum := ttx.SigFanOutTimeout + ttx.AuditTimeout + fsc.ApprovalTimeout
```
AkramBitar, line 21: This asserts against the wrong signature leg, so the test doesn't guard the regression it claims to.

`ttx.SigFanOutTimeout` (`endorse.go:126`) is the **responder's** wait for an *incoming* signature request. That wait completes *before* `EndorseView.receiveTransaction` starts, so it is not inside the receive budget at all.

The leg that actually consumes the responder's budget is the **initiator's** per-signature wait in `CollectEndorsementsView.signRemote` — still a bare literal at `collectendorsements.go:459`:

```go
if err := ts.ReceiveTypedWithTimeout(TypeSignature, &signaturePayload, time.Minute); err != nil {
```

The two are numerically equal today, which is the only reason this test is green. Two ways to see that it isn't measuring what it says:

- Raise `collectendorsements.go:459` to `3 * time.Minute` and the real invariant is violated (3 + 1 + 2 = 6 > 5) while `TestReceiveTransactionTimeoutBudget` still passes.
- Lower `SigFanOutTimeout` to 10s and the test reports 1m50s of slack that doesn't exist.

Fix either by extracting `collectendorsements.go:459` into the constant this test already uses, or by pointing the test at that value instead.

docs/services/ttx.md
```diff
@@ -573,3 +573,38 @@ The recovery service is part of the **Storage Service** and is instantiated by t
...
+signature wait rather than one per identity, but the phases themselves are serial and the
+table below accounts for a single fan-out leg. Multi-owner spends therefore add signature
+waits the invariant does not account for; it is not asserted for those paths.
+
+**Invariant:** `responder receive > sig fan-out + audit + approval`
+
+| Leg | File:Line | Value |
+|---|---|---|
+| Responder receive (the budget) | `token/services/ttx/receivetx.go:35` | 5 min |
+| Signature fan-out (per endorser) | `token/services/ttx/endorse.go:126` | 1 min |
```
AkramBitar, line 596: This row understates the leg, and with the real number the new 5-minute budget still has zero slack — the same inversion the PR is fixing.

The initiator's signature fan-out phase is bounded not by the endorser's 1-minute receive but by `answerCollectionTimeout = 2 * time.Minute` (`collectendorsements.go:905`), and `fanOut` arms a **fresh** 2-minute timer on every `collector.Collect(ctx, 1)` iteration (`collectendorsements.go:911`). A worker can legitimately exceed its own 1-minute receive window in the code *around* the receive, which that window does not cover: `context.GetSession(context.Initiator(), party)` at `collectendorsements.go:449` (endpoint resolution + dial) and `verifierGetter` / `verifier.Verify` at 463–469. So the phase's true upper bound is 2 min, and the worst legitimate case for a single remote signer is 2 (sig) + 1 (audit) + 2 (approval) = **5 min — exactly the new budget**.

Separately, the caveat paragraph above (lines 583–589) scopes the serial-phases gap to "multi-owner spends" and the `boolpolicy`/`multisig` responders, but the seriality at `collectendorsements.go:113-122` is unconditional: `requestSignaturesOnIssues` and `requestSignaturesOnTransfers` run one after the other for *any* transaction carrying both. A plain single-owner issue+transfer with remote signers on both sides therefore needs 1 + 1 + 1 + 2 = 5 min while not being a multi-owner spend at all, so the stated exclusion doesn't actually cover it.

AkramBitar (commented):
Reviewed at `023a3b45`. The extraction of the four waits into named constants is the right shape, and pinning the invariant in a test so CI catches a regression is the right instinct. Four inline comments below; two of them (the test's chosen leg, and the fan-out row in the table) mean the invariant as written is not quite the invariant that actually binds.

One more that falls outside the diff so it can't be an inline comment:

**`token/services/network/fabric/endorsement/fsc/setup_initiator.go:107` (low)** — `CollectEndorsements(ctx, tx, 2*time.Minute, s.Endorsers...)` is the same value in the same package as the newly extracted `ApprovalTimeout`, still a literal. The commit's stated rationale is that the values "cannot drift from them"; these two now can, and a future bump of `ApprovalTimeout` will silently leave the setup path behind. Worth folding into the constant while you're here.

Verified and *not* findings, for the record: the test compiles and passes; `go vet` and `gofmt` are clean on all touched files; `require.Greater` on `time.Duration` is fine with testify v1.12.1; `ttx_test` importing `.../endorsement/fsc` creates no import cycle (external test package, and the directory already mixes `ttx` and `ttx_test`); every `file:line` in the new doc table resolves correctly at this commit; `RequestApprovalView.Call` having no Godoc is pre-existing, not introduced here; and the `time.Minute*4` waits in `multisig/spend.go` and `boolpolicy/spend.go` are pre-flight first-message waits, genuinely outside this invariant.

sid200727:
Thank you @AkramBitar, this is a really thorough review. You're right on both substantive points: SigFanOutTimeout is the responder's incoming wait and completes before the receive budget starts, so the test is asserting the wrong leg, and the initiator's per-signature wait is bounded by answerCollectionTimeout at 2 min rather than 1, which means 5 min still leaves zero slack. I also see the serial-phases caveat doesn't hold as I scoped it.

I'd rather rework this properly than patch it: extract collectendorsements.go:459, point the test at the real leg, and recompute the budget from there. Will push once I've worked through it.

sid200727:
Thank you @AkramBitar !
This was a genuinely useful review, both catches were right and the rework is pushed.
On the wrong leg. You're right that SigFanOutTimeout is the responder's wait for an incoming request and completes before receiveTransaction starts, so it was never inside the budget. The test was green by coincidence. I've extracted the per-signer wait at collectendorsements.go:459 into SigResponseTimeout, and pointed the test at the leg that actually binds.

On the fan-out row and a number I think your comment understates. Taking your two points together changes the arithmetic more than the review suggests. If the phase bound is AnswerCollectionTimeout (2m) rather than the per-signer receive, and the seriality at collectendorsements.go:113–122 is unconditional, then a transaction carrying both issues and transfers pays two 2-minute phases, not one: 2 + 2 + 1 + 2 = 7 min. Your comment computes the single-owner issue+transfer case as 1 + 1 + 1 + 2 = 5, using the 1-minute per-signer receive for each phase but that's the same bound your other comment shows is too tight. I've gone with 2m per phase for consistency.

So 5m wasn't merely slack-free, it was below the floor. The budget is now 8 min (7 floor + 1 slack). I verified the test actually bites this time: at 7m it fails with "7m0s is not greater than 7m0s", at 8m it passes.

Also done: AuditTimeout moved above Call's Godoc block, the receivetx.go doc line now references DefaultReceiveTransactionTimeout instead of a literal, setup_initiator.go:107 folded into ApprovalTimeout, and the docs caveat rewritten, I've dropped the "multi-owner spends" scoping entirely since you showed it doesn't hold. The doc table now names constants rather than file:line so it can't drift again.

One question I'd rather ask than guess at. distributeTxToParty has its own time.Minute ack receive and also runs through fanOut, so distribution looks like a third serial phase inside the same Call, after approval. Is the responder still blocked in ReceiveTransactionView at that point? If so the floor is higher than 7 and the budget needs to go up again. I've left it out of this PR rather than guess the way I guessed at the fan-out leg

PR #2313, opened 2026-08-26
fix(identity): isolate idempotent identity registration inserts in savepoints

### LINKED ISSUE #2036: db: registerIdentityDescriptor breaks idempotent retry after partial-conflict transaction abort
### Summary

`registerIdentityDescriptor`'s duplicate-key handling on a shared Postgres transaction does not account for Postgres's transaction-abort semantics: once one statement in a transaction hits a duplicate-key error, *every subsequent statement* on that transaction fails with a generic "current transaction is aborted" error, not the specific `UniqueKeyViolation` the code is designed to detect and treat as idempotent success. This breaks idempotent re-registration in partial-conflict scenarios (e.g. retry after a crash where only some rows were already written).

### Where

`token/services/storage/db/sql/common/identity.go:492-511` (`registerIdentityDescriptor`). `storeSignerInfo` and `storeIdentityData` each individually swallow `UniqueKeyViolation` so that re-registering an identical identity is treated as a no-op success. But when they run on the *same* `*sql.Tx` (as `registerIdentityDescriptor` does) and the audit-info insert (`storeIdentityData`, line 493) hits a duplicate key, Postgres aborts the whole transaction. The next statement on that same `tx` — the alias-branch `storeSignerInfo` call at line 501 — then fails with the generic aborted-transaction error, which does not match `UniqueKeyViolation`, so it propagates as a hard failure instead of being treated as idempotent success.

### Failure scenario

Register an identity+alias where the identity's signer-info row is new but a leftover audit-info row already exists for the same key (e.g. from a partial write during a prior crash/retry). The second registration attempt fails outright instead of completing idempotently, even though from the caller's perspective this should be a safe, retryable operation.

### Test coverage gap

`TRegisterIdentityDescriptor` in `token/services/storage/db/dbtest/identity.go:264-265` calls `registerIdentityDescriptor` twice, but with **identical** arguments both times — it never exercises the partial-conflict scenario (new signer-info + pre-existing audit-info, or vice versa) where the abort cascades. The bug is real but currently untested.

### Suggested fix

Either (a) check for `UniqueKeyViolation` from the *first* failing statement and short-circuit the rest of the function as idempotent success without issuing further statements on the aborted transaction, or (b) use savepoints (`SAVEPOINT`/`ROLLBACK TO SAVEPOINT`) around each individually-idempotent insert so a duplicate-key error on one insert doesn't poison the rest of the transaction.

### Severity

HIGH — breaks a documented idempotency guarantee for identity registration under retry-after-crash conditions; discovered via a systematic bug-hunting review of `token/services/storage/db/sql/common`.

AkramBitar:
## Problem

When registering an identity, up to 4 rows are written in one database transaction. If any row already existed (e.g. from a previous crashed attempt), PostgreSQL treats that as a hard error and cancels the entire transaction — so the missing rows never get written. The registration fails instead of completing. SQLite had a quieter version of the same bug: it silently stopped after the first existing row and left the others unwritten, with no error reported.

`IdentityStore.registerIdentityDescriptor` writes up to four rows — signer info and audit info, for the identity and for its alias — on a **single** transaction. Each insert individually swallows a duplicate key so that re-registering an identity is an idempotent no-op.

On PostgreSQL that assumption breaks down as soon as more than one insert is involved: the *first* duplicate key aborts the whole transaction, and every following statement then fails with a generic `current transaction is aborted` error (`SQLSTATE 25P02`) that no longer matches `UniqueKeyViolation`. A retry after a partial write — e.g. a registration interrupted by a crash, where some rows already exist — therefore failed hard instead of completing.

There was a second, silent half to the bug: the `if exists { return nil }` short-circuit after the first insert. When only the identity's signer info was already present, the function returned success without ever writing the audit info or anything for the alias, and the in-memory caches were then populated as if it had. Nothing failed, but the rows were missing after a restart.

## Fix

Wrap each of the 4 inserts in its own mini "checkpoint" (SAVEPOINT). If a row already exists, roll back only that one checkpoint — not the whole transaction — and move on. The remaining missing rows are then written normally. A crashed partial registration is now completed on retry instead of failing.

Suggested option (b) from the issue — savepoints:

- New `insertIdempotently` helper wraps each insert in its own `SAVEPOINT`, rolling back to it when the row is already there (or when the insert failed) and releasing it otherwise. Rolling back to the savepoint is what clears PostgreSQL's aborted-transaction state, so the still-missing rows can be written and only genuine errors abort the registration. Savepoint names are package-level constants, never derived from input.
- `storeIdentityData` now reports whether the row already existed (`(bool, error)`), mirroring `storeSignerInfo`, so the wrapper can tell a swallowed duplicate key from a clean insert.
- The `exists` short-circuit is removed, so a half-written registration is completed by the retry.
- The idempotency guarantee is now documented on the `IdentityStoreService` interface method, on the implementation, and in `docs/services/storage.md`.

## Test coverage

`TRegisterIdentityDescriptor` in the shared dbtest suite only ever called `registerIdentityDescriptor` twice with identical arguments, so it never reached the abort cascade. Added:

- `TRegisterIdentityDescriptorPartialConflict` (case name `RegisterPartialConflict`), covering four partial-conflict combinations: pre-existing audit info only, pre-existing signer info only, alias rows only, and a mix of both sides.
- `TRegisterIdentityDescriptor` now also asserts on the stored rows, not just on the absence of an error.

Assertions go through `GetSignerInfo`, which bypasses the store's in-memory caches — the caches were otherwise reporting rows as present even when the registration never wrote them.

Verified the new case actually catches the bug by reverting the fix:

- **PostgreSQL**: `failed to store signer info for alias: ERROR: current transaction is aborted, commands ignored until end of transaction block (SQLSTATE 25P02)` — exactly the reported failure.
- **SQLite**: no error, but `signer info mismatch for [cherry]` / `[elderberry]` — the silent-incompleteness half.

Green with the fix: sqlite, in-memory, PostgreSQL (container) and Hashicorp-Vault KVS identity suites, plus `make lint-auto-fix` (0 issues) and `make checks`.

Fixes #2036

### DIFF
--- a/docs/services/storage.md
+++ b/docs/services/storage.md
@@ -60,6 +60,21 @@ Manages the cryptographic identities and logical wallet groupings used by the no
 *   **IdentityInfo**: Stores sensitive identity-related data, including audit information and PII required for transaction processing.
 *   **IdentitySigners**: Tracks which identities have locally available signing keys and their associated metadata.
 
+#### Idempotent identity registration
+
+`RegisterIdentityDescriptor` writes up to four rows in a single transaction: the signer info and
+the audit info of the identity, plus the same two for its alias. The operation is **idempotent and
+safe to retry**. Registering the same descriptor twice succeeds without error, and a registration
+that was only partially persisted — for example because a previous attempt was interrupted by a
+crash — is completed by the retry rather than rejected.
+
+Each of those inserts is individually idempotent (a duplicate key means "already registered"), but
+they share one transaction, so each one is isolated by its own `SAVEPOINT`. Without it, the first
+duplicate key would abort the whole transaction on PostgreSQL and every following statement would
+fail with a generic `current transaction is aborted` error (`SQLSTATE 25P02`) that no longer
+identifies itself as a unique-key violation. Rolling back to the savepoint clears that state, so
+the rows that are still missing are written and only genuine errors abort the registration.
+
 ### Generic Store
 *   **KeyStore**: A secure, generic key-value store used for persisting various cryptographic materials and small sensitive states. The keystore uses identifiers that are expected to be the hexadecimal representation of the key's Subject Key Identifier (SKI). This convention is relied upon by other packages, particularly for operations like keystore cleanup where SKIs are derived from owner identities to locate and manage keys.
 
--- a/token/services/identity/driver/storage.go
+++ b/token/services/identity/driver/storage.go
@@ -194,7 +194,9 @@ type IdentityStoreService interface {
 	SignerInfoExists(ctx context.Context, id []byte) (bool, error)
 	// GetSignerInfo returns the signer info bound to the given identity
 	GetSignerInfo(ctx context.Context, id []byte) ([]byte, error)
-	// RegisterIdentityDescriptor registers a descriptor for an identity and associates it with an alias
+	// RegisterIdentityDescriptor registers a descriptor for an identity and associates it with an alias.
+	// The operation is idempotent and safe to retry: registering the same descriptor again succeeds
+	// without error, and a registration that was only partially persisted is completed by the retry.
 	RegisterIdentityDescriptor(ctx context.Context, descriptor *IdentityDescriptor, alias driver.Identity) error
 	// IterateSigners returns a page of SignerEntry values from the Signers table ordered by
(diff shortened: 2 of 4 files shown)


Effi-S (approved):
LGTM

PR #2314, opened 2026-08-26
identity: areMe converts storage errors into false-negative ownership answers

### LINKED ISSUE #2319: identity: areMe converts storage errors into false-negative ownership answers
Part 1: https://github.com/LFDT-Panurus/panurus/issues/2066 (PR-https://github.com/LFDT-Panurus/panurus/pull/2172)

Here we implements this suggestion in review [comment](https://github.com/LFDT-Panurus/panurus/pull/2172#issuecomment-5422289741_ ):  

Two more findings, on code that isn't in this diff but is directly in scope for the bug being fixed:

**1. `token/services/storage/db/kvs/identitydb.go:312` — the fix doesn't hold for the KVS-backed identity store, so #2066 is only closed for the SQL backend.**

`return s.kvs.GetExisting(ctx, keys...), nil` can never report an error, and FSC's `KVS.GetExisting` (fabric-smart-client v0.18.0, `platform/view/services/storage/kvs/kvs.go:109-111`) does:

```go
it, err := o.store.GetStateSetIterator(...)
if err != nil {
    return result
}
```

i.e. it swallows the store failure and returns the partial/empty list. So: node restarts (in-memory `p.signers` cache cold), backing store is briefly unavailable, `Provider.areMe` receives a short slice with `err == nil` and still answers "not mine" for an owned identity — exactly the false negative this PR is meant to eliminate.

The error-less signature is baked into the `KVS` interface at `token/services/storage/db/kvs/kvs.go:17`, so fixing it needs an interface change or a `Get`-based existence check.

**2. `token/services/ttx/collectendorsements.go:462` — the auditor-locality decision in a file this PR edits is still fail-open.**

`local := sigService.IsMe(context.Context(), c.tx.Opts.Auditor)` uses FSC's `sig.Service`, whose `AreMe` logs and discards a `FilterExistingSigners` error (`platform/view/services/sig/service.go:182-184`).

Scenario: a self-auditing issuer restarts (cold signer cache) and the signer store errors → `local=false` → `AuditingViewInitiator.Call` takes `startRemote` → `context.GetSession(a, <own identity>)`, which FSC explicitly does not support for self (see the comment on `startLocal` in `token/services/ttx/auditor.go`) → the auditing session fails or blocks until the one-minute `ReceiveTypedWithTimeout`.

Worth resolving locality through `c.tx.TokenService().SigService().IsMe(...)` (now error-returning) or FSC's `SignerInfoExists`.

Effi-S:
Fixes  #2319

### Part 2 
(Part 1: https://github.com/LFDT-Panurus/panurus/pull/2172 (Issue https://github.com/LFDT-Panurus/panurus/issues/2066) ) 

Here we implements this suggestion in review [comment](https://github.com/LFDT-Panurus/panurus/pull/2172#issuecomment-5422289741_ ):  

Two more findings, on code that isn't in this diff but is directly in scope for the bug being fixed:

**1. `token/services/storage/db/kvs/identitydb.go:312` — the fix doesn't hold for the KVS-backed identity store, so #2066 is only closed for the SQL backend.**

`return s.kvs.GetExisting(ctx, keys...), nil` can never report an error, and FSC's `KVS.GetExisting` (fabric-smart-client v0.18.0, `platform/view/services/storage/kvs/kvs.go:109-111`) does:

```go
it, err := o.store.GetStateSetIterator(...)
if err != nil {
    return result
}
```

i.e. it swallows the store failure and returns the partial/empty list. So: node restarts (in-memory `p.signers` cache cold), backing store is briefly unavailable, `Provider.areMe` receives a short slice with `err == nil` and still answers "not mine" for an owned identity — exactly the false negative this PR is meant to eliminate.

The error-less signature is baked into the `KVS` interface at `token/services/storage/db/kvs/kvs.go:17`, so fixing it needs an interface change or a `Get`-based existence check.

**2. `token/services/ttx/collectendorsements.go:462` — the auditor-locality decision in a file this PR edits is still fail-open.**

`local := sigService.IsMe(context.Context(), c.tx.Opts.Auditor)` uses FSC's `sig.Service`, whose `AreMe` logs and discards a `FilterExistingSigners` error (`platform/view/services/sig/service.go:182-184`).

Scenario: a self-auditing issuer restarts (cold signer cache) and the signer store errors → `local=false` → `AuditingViewInitiator.Call` takes `startRemote` → `context.GetSession(a, <own identity>)`, which FSC explicitly does not support for self (see the comment on `startLocal` in `token/services/ttx/auditor.go`) → the auditing session fails or blocks until the one-minute `ReceiveTypedWithTimeout`.

Worth resolving locality through `c.tx.TokenService().SigService().IsMe(...)` (now error-returning) or FSC's `SignerInfoExists`.

### DIFF
--- a/token/services/storage/db/kvs/hashicorp/identitydb_test.go
+++ b/token/services/storage/db/kvs/hashicorp/identitydb_test.go
@@ -32,13 +32,9 @@ func TestIdentityDBWithHashicorpVault(t *testing.T) {
 			Namespace: "strawberries",
 		})
 		t.Run(c.Name, func(xt *testing.T) {
-			// The KVS-backed identity store returns composite storage keys from
-			// GetExistingSignerInfo instead of the identity hashes the interface
-			// contract requires. Tracked in #1892; skip until the KVS backend
-			// is brought in line with the SQL implementation.
-			if c.Name == "GetExistingSignerInfo" {
-				xt.Skip("KVS GetExistingSignerInfo returns composite keys, not identity hashes")
-			}
+			// GetExistingSignerInfo now translates the internal composite storage keys
+			// back to identity hashes, in line with the SQL backend and the driver
+			// contract (#1892), so the shared case runs against the KVS backend too.
 			c.Fn(xt, db)
 		})
 	}
--- a/token/services/storage/db/kvs/hashicorp/vault_kvs.go
+++ b/token/services/storage/db/kvs/hashicorp/vault_kvs.go
@@ -68,42 +68,57 @@ func (v *KVS) deNormalizeID(id string) string {
 	return normilzedId
 }
 
-func (v *KVS) GetExisting(ctx context.Context, ids ...string) []string {
+func (v *KVS) GetExisting(ctx context.Context, ids ...string) ([]string, error) {
 	results := make([]string, 0)
 
 	for _, id := range ids {
-		if v.Exists(ctx, id) {
+		ok, err := v.existsE(ctx, id)
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed checking existence of id [%s]", id)
+		}
+		if ok {
 			results = append(results, id)
 		}
 	}
 
-	return results
+	return results, nil
 }
 
-func (v *KVS) Exists(_ context.Context, id string) bool {
+// existsE reports whether id holds a value, distinguishing a genuine vault read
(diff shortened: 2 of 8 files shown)


AkramBitar:
@Effi-S 
Thanks a lot for all the effort that you did in this PR. 
Is that PR ready for additional review round? 
Could you please resolve the conflicts? 

Regards,
Akram

PR #2315, opened 2026-08-26
fix(identity): close assorted robustness defects across the identity stack

### LINKED ISSUE #2073: identity: assorted robustness defects (stale reload map, boolpolicy bounds-check fragility, JSON strictness asymmetries, misc)
### Summary

A bundle of smaller robustness defects across the identity stack, filed together in the style of
#2044. The most significant is a genuine stale-read bug in `LocalMembership.Load`; the rest are
hygiene/fragility items worth tracking but individually too small for their own issue.

### Where

- **Stale map after reload (the MEDIUM item):** `token/services/identity/membership/lm.go:363-366`
  `Load` resets three `localIdentitiesBy*` maps but not `localIdentitiesByIdentity` (written at
  `:830`). A second `Load` call therefore leaves stale entries in that fourth map, and both `lookup`
  (`:903`) and `getLocalIdentity` (`:922`) will return those stale entries as if they were current.

- `token/services/identity/boolpolicy/sig.go:96` bounds-checks index `i` against `len(sigs)`, but
  then indexes `v.Verifiers[i]` at `:100` — safe today only because `:80` already enforces
  `len(sig.Signatures) == len(v.Verifiers)` before `evalNode` runs; a future refactor that removes or
  loosens that equality check would turn this into an index-out-of-range panic with no local defense.

- `token/services/identity/idemixnym/nym/audit.go:24` `FromBytes` uses stdlib `encoding/json`, while
  the embedded `crypto.AuditInfo` type is elsewhere always parsed via the project's strict
  `DisallowUnknownFields` JSON wrapper — two different strictness levels for the same logical type
  depending on entry point. The exported `FromBytes` also lacks the nil-embedded-pointer guard present
  in `DeserializeAuditInfo` (`crypto/audit.go:63`), so `{"AuditInfo":null}` produces a nil embedded
  pointer that later dereferences panic.

- `token/services/identity/interop/htlc/validator.go:11` and `info.go:11` import stdlib
  `encoding/json`, while `deserializer.go:12` uses the project's strict wrapper for the same wire
  type — the validator is laxer than the deserializer on data that should be parsed identically.

- `token/services/identity/interop/htlc/validator.go:111` — `bytes.Equal(value, claim.Preimage)` is a
  non-constant-time comparison. Noted for hygiene; not currently a practical timing oracle since the
  preimage is revealed in the same transaction that reveals `claim.Preimage`.

- `token/services/identity/role/cache.go:58-61, 73, 80` — `start` is assigned only when debug logging
  is enabled, but `time.Since(start)` is evaluated eagerly at `:73` and `:80` regardless of whether
  the result is used; the same lines also mix `c.Logger` with the package-level `logger`
  (`cache.go:20`) — two loggers with independently configurable levels for what should be one
  logical component.

- `token/services/identity/idemix/crypto/audit.go:121-133` `Validate` checks `len(Attributes) >= 4`
  but never validates `Schema` against a known set. Defense-in-depth note only — per the correction in
  this campaign's plan, `Schema` is currently always overwritten with a trusted value before matching,
  so this isn't reachable with attacker-controlled input today, but there's no guard if that
  overwrite-before-use invariant is ever broken by a future change.

- Unbounded caches with no eviction or teardown hook: `token/services/identity/signer_router.go:36`
  `byConfID` (each entry pins a `KeyManager`, and for idemix specifically a BCCSP instance plus an
  `IdentityCache` backed by a live goroutine — see the goroutine-leak issue filed separately for that
  cache) and `token/services/identity/role/registry.go:41` `Wallets`.

### Impact

The `Load` stale-map item can cause `lookup`/`getLocalIdentity` to return identity data from before a
reload — this is the one item here with a real correctness impact rather than pure hygiene. The rest
are either latent fragility (one refactor away from a panic) or defense-in-depth gaps that aren't
independently exploitable given today's call graph, but are worth closing opportunistically.

### Reproduction

For the `Load` item (the only one with clear reproduction value): call `Load` twice with different
identity sets on the same `LocalMembership`, then call `lookup`/`getLocalIdentity` for an identity
present only in the *first* load's `localIdentitiesByIdentity` map and confirm it's still returned
after the second `Load`.

### Suggested fix

- `Load`: reset `localIdentitiesByIdentity` alongside the other three maps.
- `boolpolicy/sig.go`: add an explicit `i >= len(v.Verifiers)` guard at the point of use, independent
  of the earlier length-equality check, so the two don't have to be kept in sync by convention.
- `idemixnym/nym/audit.go`: switch `FromBytes` to the project's strict JSON wrapper and add the same
  nil-embedded-pointer guard as `crypto/audit.go:63`.
- `interop/htlc/validator.go`/`info.go`: switch to the strict JSON wrapper used by `deserializer.go`.
- `role/cache.go`: guard the `time.Since(start)` calls behind the same debug-level check that guards
  assigning `start`, and consolidate on one logger.
- Consider a bounded/evicting cache or an explicit `Close`/teardown hook for `byConfID` and `Wallets`.

### Severity

LOW–MEDIUM (the `Load` stale-map item is the MEDIUM; everything else is LOW hygiene/defense-in-depth).

AkramBitar:
Fixes #2073

## The one real bug

`LocalMembership.Load` (`token/services/identity/membership/lm.go`) resets three of its four
identity indices — `localIdentities`, `localIdentitiesByName`, `localIdentitiesByConfig` — but not
`localIdentitiesByIdentity`. That fourth map is what both `lookup` and `getLocalIdentity` consult
after a miss on the by-name map, so a second `Load` left entries from the first one being served as
current: `GetIdentifier` and `GetIdentityInfo` returned pre-reload identity data. It is now reset
with the others, so a reload is the full replacement it reads as.

## The rest of the bundle

Latent fragility and defense-in-depth, none independently reachable today:

- **`boolpolicy/sig.go`** — `evalRef` bounds-checked its index against the signature slots and then
  indexed `Verifiers`, safe only because `Verify` enforces the two lengths equal before the
  traversal starts: an invariant held by convention in a different function. `sigs`, `memo` and
  `Verifiers` are now each checked where they are indexed, and a nil `Verifiers` entry is an
  unsatisfied reference rather than a nil dereference.

- **`idemixnym/nym/audit.go`** — `FromBytes` parsed with stdlib `encoding/json` while the embedded
  `crypto.AuditInfo` goes through the project's `DisallowUnknownFields` wrapper on every other entry
  point, and it lacked the nil-embedded-pointer guard `crypto.DeserializeAuditInfo` has, so a
  payload carrying none of the promoted fields returned success with a nil embedded pointer for the
  caller to dereference. Now strict, and the nil embedded pointer is rejected in `FromBytes` itself.

- **`interop/htlc/{validator,info}.go`** — parsed `htlc.Script`, `ScriptInfo` and `ClaimSignature`
  with stdlib `encoding/json` while `deserializer.go` parses the same wire types strictly; the
  validator must not accept what the deserializer rejects. The claim preimage is also now compared
  against the metadata entry with `subtle.ConstantTimeCompare`.

- **`role/cache.go`** — `start` was assigned only under a debug-level check but `time.Since(start)`
  was evaluated unconditionally, and a package-level logger was mixed with the instance one for the
  same component. Both reads are behind the same check now, and everything is on `c.Logger`; the
  package-level logger had no other user (every other reference in the package is a shadowing
  parameter) and was removed rather than left dead.

- **`idemix/schema/schema.go`** — `EidNymAuditOpts`/`RhNymAuditOpts` index `attrs` at positions fixed
  by the schema name (up to `attrs[27]` for `w3c-v0.0.1`) while `crypto.AuditInfo.Validate` only
  guarantees the four entries the default schema needs, and the schema string travels in the same
  payload with nothing making the two consistent. Bounds-checked, so a mismatched pair is a rejected
  audit info rather than an index-out-of-range panic.

- **`signer_router.go`** — `byConfID` had no eviction or teardown hook, so each entry pinned its
  `KeyManager` (and for idemix a BCCSP instance plus an `IdentityCache` backed by a live goroutine)
  for the router's lifetime. Adds `Unregister`, `Close` and `Len`, and `LocalMembership.Close` now
  unregisters its own conf_ids right after releasing their `KeyManager`s — which also stops the
  router handing out signers from a closed `KeyManager` with the cryptographic probe skipped.
  `conf_id` is derived from `(ID, Type, URL)` with `Type` being the membership's identity type, so
  one role's teardown cannot unpin another's.

## Deviations from the issue's suggested fix

- The issue suggests validating `Schema` against a known set inside `crypto.AuditInfo.Validate`. A
  whitelist there would hardcode `DefaultManager`'s two schemas into a type whose `SchemaManager` is
  an interface, breaking any custom manager supporting other schemas. The reachable consequence of an
  unvalidated `Schema` is exactly the unchecked `attrs[...]` indexing in `DefaultManager`, so the
  guard went there; an unsupported schema already errors from those methods.
- `role.Registry.Wallets`, the second unbounded cache named in the issue, already has the teardown
  hook it asks for in `Registry.Done` (closes wallets implementing `Close`, drops the map), so only
  `byConfID` needed one.

## Testing

- Unit tests for each behavioural change. The `Load` regression test and the boolpolicy bounds tests
  were both confirmed to fail against the unfixed code — the latter with the actual
  `index out of range [1] with length 1` panic.
- New `FuzzVerifyOwnerNoPanic` and `FuzzMetadataClaimKeyCheckNoPanic` over the HTLC owner and
  claim-signature parsers (both touched here), each clean at `-fuzztime=20s`, and wired into the
  `nightly-fuzz.yml` matrix. The pre-existing audit-info fuzz targets were re-run against the strict
  decode.
- `make checks`, `make lint-auto-fix`, `make unit-tests` and `go test -race` on the identity and
  interop trees are clean. One unrelated pre-existing failure: `TestTranslatePath` asserts the
  absolute checkout path contains `"panurus"`, which does not hold in a worktree named otherwise.

## Docs

`docs/services/identity.md`: reload-replaces-not-merges and the `Close` contract under
`LocalMembership`, the `SignerRouter` teardown API, the audit-info JSON strictness plus the
schema/attribute-count coupling, the boolpolicy verification bounds guarantees, and the HTLC parsing
and preimage-comparison notes.

### DIFF
--- a/.github/workflows/nightly-fuzz.yml
+++ b/.github/workflows/nightly-fuzz.yml
@@ -141,6 +141,12 @@ jobs:
           - name: identity-role-register-recipient
             pkg: ./token/services/identity/role
             func: FuzzLongTermOwnerWalletRegisterRecipientNoPanic
+          - name: identity-htlc-verify-owner
+            pkg: ./token/services/identity/interop/htlc
+            func: FuzzVerifyOwnerNoPanic
+          - name: identity-htlc-metadata-claim-key-check
+            pkg: ./token/services/identity/interop/htlc
+            func: FuzzMetadataClaimKeyCheckNoPanic
 
     steps:
       - name: Checkout code
--- a/docs/services/identity.md
+++ b/docs/services/identity.md
@@ -98,6 +98,15 @@ The `LocalMembership` component (`token/services/identity/membership`) plays a p
     each returned `KeyManager` must either be independently owned by the caller or safe for concurrent `EnrollmentID` calls — and the results
     are committed to the in-memory indices sequentially in the original store order, so identity ordering (e.g. fallback default selection,
     same-name tie-breaks) is deterministic.
+*   **Reloading replaces, it does not merge**: `Load` may be called more than once on the same `LocalMembership`, and each call is a full
+    replacement of the in-memory view. It resets **all four** indices — `localIdentities`, `localIdentitiesByName`,
+    `localIdentitiesByConfig` and `localIdentitiesByIdentity` — plus the cached default identifier, so an identity that was present only in an
+    earlier `Load` no longer resolves. `localIdentitiesByIdentity` is the map consulted after a miss on `localIdentitiesByName` (by both
+    `lookup` and `getLocalIdentity`, reached from `GetIdentifier` and `GetIdentityInfo`); leaving it unreset was the stale-read fixed for
+    [#2073](https://github.com/LFDT-Panurus/panurus/issues/2073). Anything added to `LocalMembership` that indexes identities must be reset
+    there too.
+*   **Teardown**: `Close` releases every `KeyManager` this membership loaded (those implementing `Close`), unregisters its own `conf_id`s from
+    the `SignerRouter` (see below), and unsubscribes from the identity-store notifier. It is idempotent.
 
 ### Example: Wiring Services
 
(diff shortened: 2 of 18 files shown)


Effi-S (approved):
LGTM

PR #2318, opened 2026-08-26
fix(storage): repair keyset pagination and guard the query DSL's landmines

### LINKED ISSUE #2044: db/sql/query: misc medium/low findings (dead KeysetWithId panics, unguarded InTuple, perf)
### Summary

A batch of medium/low-severity findings from a systematic bug-hunting review of `token/services/storage/db/sql/query/**` (the SQL query-builder DSL used by the Postgres/SQLite drivers), not severe enough to warrant individual issues but worth tracking and fixing. Note: a full trace of every value-writing code path confirmed **no SQL injection vector** exists in this DSL today — table/column names are always Go compile-time literals, and all values go through bound parameters.

### 1. `pagination.KeysetWithId` is dead code that panics whenever actually called (MEDIUM — landmine)

`pagination/keyset.go:54` — `KeysetWithId[I comparable, V id[I]]` returns `*keyset[I, V]` for the caller's concrete `V`. But `pagination/interpreter.go:45-49` and `pagination/page.go:18-23` only type-switch on `*keyset[string, any]` / `*keyset[int, any]`. Go generic instantiations are distinct concrete types, so e.g. `*keyset[int, MyRecord]` matches neither `case`, and execution falls through to `panic(...)` in both `interpreter.go:55` and `page.go:28`. Verified: zero call sites or tests use `KeysetWithId` anywhere in the repo today — it is shipped dead and broken, only `KeysetWithField` (which deliberately returns `*keyset[I, any]`) actually works with the interpreter.

**Suggested fix:** either fix `KeysetWithId` to route through the same `any`-erased path `KeysetWithField` uses, or remove it entirely and document that only `KeysetWithField` is supported, so a future caller doesn't reach for the broken function.

### 2. `cond.InTuple` has no empty-list guard (MEDIUM — unguarded public API)

`cond/in.go:34-40` — `InTuple(fields, vals)` has no check on `len(vals)`. Both dialect interpreters short-circuit to a no-op write when `len(vals)==0` (`postgres/conditions.go:37-39`, `sqlite/conditions.go:38-40`) — i.e. they write *nothing* to the builder. If such a condition is composed inside `And`/`Or`, `andOr.WriteString` (`cond/andor.go:37-46`) does not filter it out (it only filters conditions that are `==` the `AlwaysTrue`/`AlwaysFalse` sentinel), so it can emit invalid SQL like `( AND status = $1)` — a leading operator with no left operand. The safer helpers `In`/`FieldIn` (`in.go:18-21,30-32`) explicitly return `AlwaysTrue` on empty input, and `hasTokens` in `db/sql/common/tcondition.go:23-25` also explicitly guards. **No current call site is vulnerable** — this is a latent landmine in the public API surface, and `condition_test.go` never exercises the empty-tuple case.

**Suggested fix:** make `InTuple` return the `AlwaysTrue`/`AlwaysFalse` sentinel on empty input, matching `In`/`FieldIn`'s convention, and add a regression test.

### 3. Keyset pagination's "no cursor" sentinel can collide with legitimate data (LOW/MEDIUM)

`pagination/keyset.go:105-117` (`nilElement[I]()`) uses `-1` (int) / `""` (string) as the "no cursor" sentinel; `handleKeysetPreProcess` (`interpreter.go:24-33`) branches on `FirstID != nilElement()`. If the real ID column legitimately contains `-1` or `""`, the logic silently mistakes a valid cursor for "no cursor" and falls back to `OFFSET`, reintroducing the skip/duplicate-under-concurrent-write problem keyset pagination exists to prevent. No enforcement anywhere that the ID domain excludes these values.

**Suggested fix:** use an explicit `(value, hasCursor bool)` pair or a pointer/`Option` type instead of an in-band sentinel value.

### 4. Keyset pagination has no tiebreaker/composite-key support and hardcodes ASC (LOW)

`handleKeysetPreProcess` (`interpreter.go:24-33`) always does `AddOrderBy(Asc(pagination.SQLIDName))` and compares with `>` only — single column, ascending only, no DESC option, no secondary tiebreaker for non-unique sort columns. Not presently exploitable: production stores (`ttxdb`/`auditdb` `QueryTransactions`) use `offset`/`none`/`empty` pagination with a `stored_at`-based order, never keyset — keyset is effectively unused/untested in production code paths today, only unit-tested in isolation.

**Suggested fix:** if/when keyset pagination is adopted for a production query path, add DESC support and a unique tiebreaker column before relying on it.

### 5. Validate-after-write in `WriteValueTuples`/`WriteTuples` (LOW)

`common/builder.go:56-79` — the `len(tuple) != cols` check (line 73) runs *after* the tuple's cells are already written to the builder (harmless today since the panic discards the buffer, but the pattern should validate before writing). `WriteTuples` (`builder.go:116-145`) trusts `len(tuples[0])` as the column count for every row with no explicit re-check per row — currently safe only because `insert/query.go:45-48`'s `Row()` enforces `len(tuple)==len(q.fields)` upstream; `WriteTuples` itself has no defense if called directly by a future caller with mismatched rows.

**Suggested fix:** validate row-length consistency before writing any output, not after.

### 6. Error-wrapping convention violations (LOW)

`pagination/offset.go:24,27` and `pagination/keyset.go:89,92,125` use `fmt.Errorf` instead of the mandated `fabric-smart-client/pkg/utils/errors` package.

### Perf suggestions

- **Postgres `InTuple` decomposes into OR-of-ANDs** (`postgres/conditions.go:36-49`) instead of native row-value `(a,b) IN ((..),(..))`, which Postgres supports directly. For large ID lists (e.g. `HasTokens` with many token IDs, `db/sql/common/tcondition.go:18-33`) this produces much larger SQL text and prevents the planner from doing an efficient composite comparison the way SQLite's interpreter already does (`sqlite/conditions.go:37-51`, native tuple IN). Switching Postgres to native row-value `IN` would reduce query size and likely improve planner behavior for large ID lists.
- **No query-template/prepared-statement caching**: every store call rebuilds the full field list/condition tree/SQL text from scratch even though the query "shape" is constant per call site and only parameter values change. Caching the built template and only re-binding params on hot paths would cut allocations.
- OFFSET pagination is correctly implemented but note (not a bug): `OFFSET N` forces the DB to scan/discard N rows every page — worth a doc comment for callers expecting deep pagination.

### Severity

MEDIUM (items 1-2), LOW/MEDIUM (item 3), LOW (items 4-6) — mostly landmines in unused/underused code paths rather than currently-triggered bugs, but worth fixing before a future caller reaches for the broken paths.

AkramBitar:
Fixes #2044

Closes out the batch of medium/low findings from the review of the SQL
query-builder DSL under `token/services/storage/db/sql/query`. Almost all of them
are **landmines rather than live bugs**: public entry points that panic or emit
invalid SQL, which no caller reaches today, so a future caller would be the one
to find out.

## Correctness

**1. `pagination.KeysetWithId` was shipped dead and panicking.** It returned
`*keyset[I, V]` for the caller's row type `V`, but `interpreter.PreProcess` and
`NewPage` only type-switch on the `any`-erased instantiation. Generic
instantiations are distinct concrete types, so every call fell through to
`panic(...)` in both places.

Collapsed the internal type to `keyset[I]` with an `idGetter func(any) I`, so all
four constructors (`Keyset`, `KeysetWithField`, `KeysetWithId`, `KeysetFromRaw`)
produce the one type the dispatch actually matches.

**2. `cond.InTuple` had no empty-list guard.** Both dialect interpreters
short-circuit to writing *nothing* for an empty list, and `And`/`Or` can only
filter out the `AlwaysTrue`/`AlwaysFalse` sentinels — so a condition that
rendered to the empty string stayed in the list and emitted a leading operator
with no left operand, e.g. `( AND status = $1)`. It now returns `AlwaysTrue` on
an empty field or value list, matching `In`/`FieldIn`.

**3. The keyset "no cursor" sentinel could collide with real data.** The marker
was the in-band value `-1` (int) / `""` (string), so a page whose last row
legitimately carried one of those ids was mistaken for "no cursor" and fell back
to `OFFSET`, reintroducing the skip/duplicate-under-concurrent-write window that
keyset pagination exists to close. `FirstID`/`LastID` are now `*I`; `nilElement`
is gone.

**4. Keyset ordering limitations** are documented rather than implemented — the
issue scopes DESC/tiebreaker support to "if/when keyset pagination is adopted for
a production query path", and it has no caller today. `Keyset` now spells out
that it is ascending, single-column, requires a unique id, and only walks forward
one page at a time.

**5. Validate-before-write in the builder.** `WriteValueTuples` checked a row's
length only *after* writing its cells, and `WriteTuples` derived the column count
from row 0 and never checked the rest. Both now validate the shape before
emitting any output.

**6. Error-wrapping convention.** `fmt.Errorf` replaced with
`fabric-smart-client/pkg/utils/errors` in `pagination/{keyset,offset}.go`; `fmt`
is no longer imported by the package at all. The four "must be greater than zero"
messages were also wrong — the guards are `< 0`, so they now say "must not be
negative".

Three more panic/silent-failure paths found along the way: an unsupported keyset
id type, a nil id getter, and an empty `PropertyName` (which indexed
`idFieldName[0]`) now return errors; `newKeysetTypedPage` returns an error instead
of a silent `nil, nil`.

## Performance

**Postgres now uses native row-value `IN`.** It decomposed tuple membership into
an OR-of-ANDs, which grew the SQL text with the id list and hid the composite
comparison from the planner. It now emits `(tx_id, idx) IN (($1, $2), ($3, $4))`
through a shared `common.WriteInTuple` helper, which SQLite uses too — and which
also replaces the two copies of this logic in the test doubles.

Because both dialects now render identically, the
`QueryConstructorTraits{SupportsIN, MultipleParenthesis}` flags in the shared test
utilities are obsolete and removed (one of the two branches they guarded was
already byte-identical).

`OFFSET`'s deep-pagination cost is documented on the `Offset` constructor, as the
issue suggests.

**Left for its own change:** the query-template / prepared-statement caching
suggestion — a much larger change on hot paths, with partial work already tracked
under #1183 / #1919.

## Testing

- Regression tests for every item, including the two paths that previously
  panicked (`TestKeysetWithId`, `TestKeysetCursorSentinelCollision`) and the
  empty-tuple composition the issue notes was never exercised.
- `condition_test.go`'s table is now named subtests, so a failure identifies its
  case, and asserts parameters with `HaveExactElements` instead of `ConsistOf` —
  bind order matters for `$n` placeholders.
- New `query/common/builder_test.go` covering the shape validation and
  `WriteInTuple`.
- **The Postgres row-value `IN` is verified against a real database, not just
  sqlmock**: `TestTokens` runs the `GetTokenMetadata`, `DeleteMultiple` and
  `QueryTokenDetails` cases against a Postgres container, and all three issue
  multi-tuple `HasTokens` queries and assert on the rows returned.
- `FuzzKeysetFromRawNoPanic` added — `KeysetFromRaw` decodes a pagination token
  that can come from an untrusted client — and wired into
  `.github/workflows/nightly-fuzz.yml`.

`make checks` is clean and `make lint` reports nothing in any touched file.

## Docs

New `docs/development/sql-query-dsl.md`, linked from
`docs/development/storage.md`: package layout, the bound-parameter rule, why
empty lists must return a sentinel rather than an empty fragment, tuple
membership, the four pagination strategies and their trade-offs, and a checklist
for adding a condition.

## Compatibility

One edge worth flagging: an old serialized keyset token containing
`"first_id": -1` now decodes as a real cursor rather than "no cursor". No
deployed client holds one — keyset is not wired into any production query path —
but it is the one behavioural change in the serialization.


token/services/storage/db/sql/query/common/builder.go
```diff
@@ -144,6 +142,44 @@ func (b *builder) WriteTuples(tuples []Tuple) Builder {
...
+// Callers must have already ruled out empty input (see cond.InTuple, which
+// returns a sentinel in that case); the guard here is defensive only.
+func WriteInTuple(fields []Serializable, vals []Tuple, sb Builder) {
+	if len(fields) == 0 || len(vals) == 0 {
+		return
+	}
+	if len(fields) == 1 && len(vals) == 1 {
+		sb.WriteSerializables(fields[0]).
+			WriteString(" = ").
+			WriteParam(vals[0][0])
```
Effi-S, line 172: We index `vals[0][0]` without checking the tuple is non-empty

Effi-S (approved):

PR #2320, opened 2026-08-26
network-driver: ListenerManagerConfig getters inconsistently treat a literal 0 as unset

### LINKED ISSUE #2060: network-driver: ListenerManagerConfig getters inconsistently treat a literal 0 as unset
### Summary

The five config getters on `serviceListenerManagerConfig` use inconsistent zero-handling: one correctly treats a configured `0` as "unset" and substitutes its default, while the other four return a literal `0` unchanged — indistinguishable from an operator who simply never set the key (`GetInt`/`GetDuration` on a missing key also return the zero value).

### Where

`token/services/network/fabric/config/config.go:51-89`:

```go
func (c *serviceListenerManagerConfig) DeliveryMapperParallelism() int {
    if v := c.c.GetInt(DeliveryMapperParallelism); v > 0 { return v }     // 0 falls through to default
    return DefaultDeliveryMapperParallelism
}
func (c *serviceListenerManagerConfig) DeliveryLRUSize() int {
    if v := c.c.GetInt(DeliveryLRUSize); v >= 0 { return v }   // 0 IS returned as-is
    return DefaultDeliveryLRUSize
}
```

`DeliveryMapperParallelism` treats a configured `0` as "unset" and substitutes the default (10), while `DeliveryBlockProcessParallelism`, `DeliveryLRUSize`, `DeliveryLRUBuffer`, and `DeliveryListenerTimeout` all use `>= 0` and therefore return a literal, explicit zero value unchanged when that's what's configured.

### Impact

Because `GetInt`/`GetDuration` on a missing config key return the zero value rather than a distinguishable "not set" sentinel, an operator who simply never sets `token.finality.delivery.lruSize`/`lruBuffer`/`listenerTimeout` (or `blockProcessParallelism`) gets a silent effective value of `0` instead of the documented default of 30/15/10s (or 10) — for the LRU size/buffer, a cache configured to hold zero entries; for the listener timeout, an instantly-expiring timeout. This is a "config typo/omission causes silent degradation to non-functional, not a clear error" class of bug, and the inconsistency between the getters (one clamps 0→default, the rest don't) means the safe behavior exists in the codebase already (in the sibling `DeliveryMapperParallelism` getter) but isn't applied uniformly.

### Reproduction

Reproduced locally with a table-driven unit test (not yet committed) across the three affected getters, each set to an explicit configured `0`, asserting the literal zero is returned rather than the documented default — plus a contrast test showing `DeliveryMapperParallelism`'s correct `> 0` guard does fall back as intended:

```go
fake := &fakeConfigService{ints: map[string]int{config.DeliveryLRUSize: 0}, ...}
cfg := config.NewListenerManagerConfig(fake)
got := cfg.DeliveryLRUSize()
assert.Equal(t, 0, got)                    // BUG: literal 0 returned, not DefaultDeliveryLRUSize (30)
assert.NotEqual(t, config.DefaultDeliveryLRUSize, got)
```

### Suggested fix

Make `DeliveryBlockProcessParallelism`, `DeliveryLRUSize`, `DeliveryLRUBuffer`, and `DeliveryListenerTimeout` use the same `> 0` guard as `DeliveryMapperParallelism`, so a configured/defaulted `0` consistently falls back to the documented default across all five getters.

### Severity

MEDIUM — silent degradation to a non-functional cache/timeout on a common config omission, not a clear error.

Effi-S:
Fixes #2060

### Summary

The five config getters on `serviceListenerManagerConfig` use inconsistent zero-handling: one correctly treats a configured `0` as "unset" and substitutes its default, while the other four return a literal `0` unchanged — indistinguishable from an operator who simply never set the key (`GetInt`/`GetDuration` on a missing key also return the zero value).

### Where

`token/services/network/fabric/config/config.go:51-89`:

```go
func (c *serviceListenerManagerConfig) DeliveryMapperParallelism() int {
    if v := c.c.GetInt(DeliveryMapperParallelism); v > 0 { return v }     // 0 falls through to default
    return DefaultDeliveryMapperParallelism
}
func (c *serviceListenerManagerConfig) DeliveryLRUSize() int {
    if v := c.c.GetInt(DeliveryLRUSize); v >= 0 { return v }   // 0 IS returned as-is
    return DefaultDeliveryLRUSize
}
```

`DeliveryMapperParallelism` treats a configured `0` as "unset" and substitutes the default (10), while `DeliveryBlockProcessParallelism`, `DeliveryLRUSize`, `DeliveryLRUBuffer`, and `DeliveryListenerTimeout` all use `>= 0` and therefore return a literal, explicit zero value unchanged when that's what's configured.

### Impact

Because `GetInt`/`GetDuration` on a missing config key return the zero value rather than a distinguishable "not set" sentinel, an operator who simply never sets `token.finality.delivery.lruSize`/`lruBuffer`/`listenerTimeout` (or `blockProcessParallelism`) gets a silent effective value of `0` instead of the documented default of 30/15/10s (or 10) — for the LRU size/buffer, a cache configured to hold zero entries; for the listener timeout, an instantly-expiring timeout. This is a "config typo/omission causes silent degradation to non-functional, not a clear error" class of bug, and the inconsistency between the getters (one clamps 0→default, the rest don't) means the safe behavior exists in the codebase already (in the sibling `DeliveryMapperParallelism` getter) but isn't applied uniformly.

### Reproduction

Reproduced locally with a table-driven unit test (not yet committed) across the three affected getters, each set to an explicit configured `0`, asserting the literal zero is returned rather than the documented default — plus a contrast test showing `DeliveryMapperParallelism`'s correct `> 0` guard does fall back as intended:

```go
fake := &fakeConfigService{ints: map[string]int{config.DeliveryLRUSize: 0}, ...}
cfg := config.NewListenerManagerConfig(fake)
got := cfg.DeliveryLRUSize()
assert.Equal(t, 0, got)                    // BUG: literal 0 returned, not DefaultDeliveryLRUSize (30)
assert.NotEqual(t, config.DefaultDeliveryLRUSize, got)
```

### Suggested fix

Make `DeliveryBlockProcessParallelism`, `DeliveryLRUSize`, `DeliveryLRUBuffer`, and `DeliveryListenerTimeout` use the same `> 0` guard as `DeliveryMapperParallelism`, so a configured/defaulted `0` consistently falls back to the documented default across all five getters.

### Severity

MEDIUM — silent degradation to a non-functional cache/timeout on a common config omission, not a clear error.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -71,23 +71,28 @@ token:
     #   c) The transaction reached finality long ago, so we query the whole ledger for this specific transaction. If the query returns no result, we proceed to step d.
     #   d) The transaction will reach finality at some point beyond the timeout or never, so we return Unknown. Then it is up to the client to either append another listener or accept that the transaction will never reach finality.  
     delivery:
-      # mapperParallelism is the number of goroutines that process incoming transactions in parallel. Defaults to 1.
+      # mapperParallelism is the number of goroutines that process incoming transactions in parallel.
+      # A non-positive or unset value falls back to the default (10).
       mapperParallelism: 10
       # blockProcessParallelism is the number of blocks we can process in parallel when they arrive from the delivery service.
-      # If the value is <= 1, then blocks are processed sequentially.
-      # This is the suggested configuration if we are not sure about the dependencies between blocks.
+      # Set it to 1 (and only 1) to process blocks sequentially; this is the suggested
+      # configuration if we are not sure about the dependencies between blocks.
+      # A non-positive or unset value falls back to the default (10), NOT sequential mode.
       # The total go routines processing transactions will be blockProcessParallelism * mapperParallelism.
       blockProcessParallelism: 1
       # lruSize detects how many transactions we should guarantee to keep in our recent cache.
       # If the transaction is not among these elements, we proceed to step b, as described above.
-      # If lruSize and lruBuffer are not set, then we will never evict past transactions (the cache will grow infinitely).
+      # Set it to 0 to disable the bound and never evict past transactions (the cache grows without limit).
+      # An unset or negative value falls back to the default (30).
       lruSize: 30
       # eviction will happen when the cache size exceeds lruSize + lruBuffer.
+      # Set it to 0 to disable the bound and never evict (same effect as lruSize: 0).
+      # An unset or negative value falls back to the default (15).
       lruBuffer: 15
       # listenerTimeout is the duration to listen when we can't find a transaction in the cache (most probably it is about to become final).
       # We will listen for this amount of time and then we will query the whole ledger, as described in step c.
-      # If the timeout is not set, then the listener will never be evicted and we will never proceed to step c.
-      # We will wait forever for the transaction to return (as is done for the 'committer' type).
+      # Set it to 0 to disable the timeout and wait forever for the transaction (as is done for the 'committer' type).
+      # An unset or negative value falls back to the default (10s).
       listenerTimeout: 10s
     # Only applicable for fabricx networks
     # notification: The manager is notified about finality events via a notification service (e.g. for FabricX).
--- a/token/services/network/fabric/config/config.go
+++ b/token/services/network/fabric/config/config.go
@@ -64,60 +64,57 @@ type serviceListenerManagerConfig struct {
 	c driver.ConfigService
 }
 
-func (c *serviceListenerManagerConfig) DeliveryMapperParallelism() int {
-	if v := c.c.GetInt(DeliveryMapperParallelism); v > 0 {
+// positiveOrDefault returns v when it is strictly positive, otherwise def. Use it
+// for knobs with no "disabled" state, where 0 is meaningless and indistinguishable
+// from an unset key: a non-positive or unset value is always the documented default.
+func positiveOrDefault[T int | time.Duration](v, def T) T {
+	if v > 0 {
 		return v
 	}
 
-	return DefaultDeliveryMapperParallelism
+	return def
+}
+
+// configuredOrDefault preserves an explicit, non-negative value the operator wrote
+// — including 0, which the delivery layer reads as "remove this bound" (an unbounded
+// cache, or a listener that never times out). isSet distinguishes a deliberate 0
+// from an absent key, which GetInt/GetDuration also report as 0: only an unset key
+// or a negative typo fall back to def. Contrast positiveOrDefault, which also
+// replaces 0 and so cannot express that disabled mode.
+func configuredOrDefault[T int | time.Duration](isSet bool, v, def T) T {
+	if isSet && v >= 0 {
+		return v // explicit 0 -> disabled mode, positive -> honoured as-is
+	}
+
+	return def // unset key or negative typo -> documented default
+}
+
+func (c *serviceListenerManagerConfig) DeliveryMapperParallelism() int {
+	return positiveOrDefault(c.c.GetInt(DeliveryMapperParallelism), DefaultDeliveryMapperParallelism)
 }
 
 func (c *serviceListenerManagerConfig) DeliveryBlockProcessParallelism() int {
-	if v := c.c.GetInt(DeliveryBlockProcessParallelism); v >= 0 {
-		return v
-	}
-
-	return DefaultDeliveryBlockProcessParallelism
+	return positiveOrDefault(c.c.GetInt(DeliveryBlockProcessParallelism), DefaultDeliveryBlockProcessParallelism)
 }
 
 func (c *serviceListenerManagerConfig) DeliveryLRUSize() int {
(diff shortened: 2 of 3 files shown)


docs/configuration.md
```diff
@@ -71,23 +71,24 @@ token:
     #   c) The transaction reached finality long ago, so we query the whole ledger for this specific transaction. If the query returns no result, we proceed to step d.
     #   d) The transaction will reach finality at some point beyond the timeout or never, so we return Unknown. Then it is up to the client to either append another listener or accept that the transaction will never reach finality.  
     delivery:
-      # mapperParallelism is the number of goroutines that process incoming transactions in parallel. Defaults to 1.
+      # mapperParallelism is the number of goroutines that process incoming transactions in parallel.
+      # A non-positive or unset value falls back to the default (10).
       mapperParallelism: 10
       # blockProcessParallelism is the number of blocks we can process in parallel when they arrive from the delivery service.
-      # If the value is <= 1, then blocks are processed sequentially.
```
AkramBitar, line 77: Nit / potential confusion: the comment now says "Set it to 1 to process blocks sequentially" but doesn't make clear that `0` no longer means sequential — before this fix `blockProcessParallelism: 0` (or omitted) fell through to `0`, which callers treated as sequential. After the fix it falls back to the default of `10`, which is parallel. Worth being explicit here to avoid an operator silently getting parallel mode after an upgrade:

```yaml
# Set it to 1 (and only 1) to process blocks sequentially.
# A non-positive or unset value falls back to the default (10), not sequential mode.
```

token/services/network/fabric/config/config.go
```diff
@@ -64,60 +64,44 @@ type serviceListenerManagerConfig struct {
...
+// configured or defaulted 0 (and any negative value) falls back to the
+// documented default.
+func positiveOrDefault[T int | time.Duration](v, def T) T {
+	if v > 0 {
 		return v
 	}
 
-	return DefaultDeliveryMapperParallelism
+	return def
 }
```
AkramBitar, line 76: This is the correct fix, but it is a **silent behaviour change** for existing deployments: before, an omitted or explicitly-zeroed `blockProcessParallelism` key returned `0` (which callers treated as sequential). After, it returns `10` (parallel). The other four getters are safe because `lruSize=0`/`lruBuffer=0`/`listenerTimeout=0` were already broken (zero-capacity cache, instant timeout) and nobody would rely on them intentionally.

`blockProcessParallelism=0` as a "sequential" sentinel is plausible operator intent, so this warrants a note in the PR description or a CHANGELOG/upgrade callout so operators know to add an explicit `blockProcessParallelism: 1` if they were relying on the omit-or-zero default.

AkramBitar (commented):
The fix is correct and cleanly done — `positiveOrDefault` removes the duplicated if-blocks and `DeliveryLedgerInfoAttempts`/`DeliveryLedgerInfoRetryDelay` folding is a free consistency win. Two comments below, one on the doc and one on the behaviour change worth calling out for upgraders.

AkramBitar (changes requested):

token/services/network/fabric/config/config.go
AkramBitar: Correcting myself here — my claim above that "`lruSize=0`/`lruBuffer=0`/`listenerTimeout=0` were already broken (zero-capacity cache, instant timeout)" is wrong, and @Effi-S is right in the thread below.

Checking the downstream consumer, `fabric-smart-client@v0.18.0/platform/fabric/core/generic/events/listenermanager.go`:

```go
:116  if config.LRUSize > 0 && config.LRUBuffer > 0 {
          events = cache.NewLRUCache(config.LRUSize, config.LRUBuffer, ...)
      } else {
          events = cache.NewMapCache[EventID, T]()   // unbounded — never evicts
      }

:136  if config.ListenerTimeout > 0 {
          listeners = cache.NewTimeoutCache(ctx, config.ListenerTimeout, ...)
      }                                              // else: never constructed
```

So a 0 there gives an **unbounded cache that never evicts**, not a zero-capacity one, and **no listener timeout at all**, not an instant one. FSC treats 0 as a deliberate "disabled" sentinel, and the `>= 0` guards in these getters existed to let that value through. `docs/configuration.md:83` and `:88-91` advertised exactly those two modes:

> If lruSize and lruBuffer are not set, then we will never evict past transactions (the cache will grow infinitely).

> If the timeout is not set, then the listener will never be evicted and we will never proceed to step c. We will wait forever for the transaction to return (as is done for the 'committer' type).

This PR's doc change deletes both of those lines, so code and docs agree again — but the capability is gone, and deleting the lines is what makes the removal invisible. That makes this a deliberate breaking change rather than a bug fix, so the upgrade callout I asked for above should cover `lruSize`/`lruBuffer`/`listenerTimeout` too, not just `blockProcessParallelism`.

AkramBitar (approved):
LGTM

PR #2323, opened 2026-08-27
feat(storage): run the ledger drift checks in the background

### LINKED ISSUE #2166: Ledger drift checks are never run on a live node
The SDK has a set of checkers that compare local state against the ledger, in `token/services/storage/db/common/checks.go`: `CheckTransactions`, `CheckUnspentTokens` and `CheckTokenSpendability`. They are wired through the DI container and exposed as `auditor.Check()` and `ttx.Owner.Check()`. `tokens.Service.PruneInvalidUnspentTokens` repairs one of the drift classes they detect.

None of it is reachable on a running node. Every caller of `Check()` and `PruneInvalidUnspentTokens()` in the repo is an integration test view (`integration/token/fungible/views/checks.go`, `integration/token/interop/support.go`). There is no scheduler and no coordination, so an operator has no way to find out that a node's token DB has diverged from the ledger short of writing their own view and invoking it by hand.

The divergence is silent when it happens. A node missing a token it actually owns cannot spend it and nothing reports the shortfall. A node holding tokens the ledger has already spent keeps selecting them and failing at approval.

Three gaps behind that:

**Nothing schedules the checks.** `recovery` and `cleanup` both run as background managers with advisory lock leadership, configurable intervals and graceful shutdown. The checks need the same treatment in any multi replica deployment and have none of it.

**Findings are unstructured.** A checker returns `[]string`. No severity, no token or transaction id to act on, nothing persisted. An operator cannot tell a new problem from one they already triaged last week, and nothing survives a restart.

**Only one direction is checked.** `CheckUnspentTokens` walks the local unspent tokens and asks the ledger about each one. Tokens the ledger says the node owns that never landed in the local DB are invisible to it, and that is the case that costs someone money.

Separately, `CheckUnspentTokens` will not scale as written. It collects every unspent token id into a single slice and issues one `QueryTokens` call for all of them, then for each local token linearly scans the whole ledger response looking for a match. `PruneInvalidUnspentTokens` batches at 50, the checker does not batch at all.

I would like to pick this up, but I would rather agree on where it should live and how far it should go before writing any code.

atharrva01:
Fixes #2166

The drift checks (CheckTransactions, CheckUnspentTokens, CheckTokenSpendability) already existed but nothing on a running node ever called them, only integration test views did. This adds a background sweep, modeled on recovery and cleanup: runs on an interval, leader election through a PostgreSQL advisory lock, one sweep per store a node owns (owner over ttxdb, auditor over auditdb).

What changed:

- Findings are now structured (checker, code, severity, tx/token id) instead of plain strings, persisted in a findings table keyed by a stable finding key. A repeated problem ages instead of getting re-reported, and closes once a sweep stops seeing it. A check that fails never closes its own findings, so an unreachable ledger can't look like a clean bill of health.
- CheckUnspentTokens now resolves tokens against the ledger in batches instead of one call for everything followed by a linear scan per token.
- New CheckLocalCompleteness check covers the direction the others can't: tokens the ledger says the node owns that never landed locally. That's the case that actually costs the owner money, so it runs in the background sweep.
- The on-demand Check API on auditor/owner is unchanged in shape (still plain messages) but now backed by the same finding checkers, prefixed with severity and code.

Docs at docs/services/storage/checks.md, wired into docs/services/storage.md and docs/configuration.md.

## Test plan
- [x] `go test ./token/services/storage/... ./token/sdk/...` (postgres, sqlite, race where applicable)
- [x] New unit tests for the checks manager/scheduler (leader election, sweep lifecycle, resolvable-checker semantics)
- [x] `golangci-lint run` clean on everything touched
- [x] Manually verified the findings upsert/aging against a real Postgres container


AkramBitar (changes requested):
@atharrva01 

Thanks a lot for this PR. 

See my comments. 

Regards,
Akram

token/services/storage/db/common/checks.go
```diff
@@ -128,48 +386,101 @@ func (a *DefaultCheckers) CheckTransactions(ctx context.Context) ([]string, erro
...
+
+// compareTransactionStatus resolves one local transaction record against the ledger.
+func (a *DefaultCheckers) compareTransactionStatus(l *network.Ledger, transactionRecord *driver.TransactionRecord) []Finding {
+	lVC, _, err := l.Status(transactionRecord.TxID)
+	if err != nil {
+		// nothing is known about this transaction's real ledger status. Falling back
+		// to network.Unknown and comparing it as if it were a real answer would turn
+		// a connectivity failure into a false claim that the ledger disagrees with
+		// the local record, so this reports only that the check was inconclusive.
+		return []Finding{{
```
AkramBitar, line 504: **Blocker: a ledger that cannot answer `Status` closes previously recorded critical findings.**

This branch emits `CodeTxStatusUnavailable` at `SeverityInfo` — not `CodeCheckFailed`. But `resolvableCheckers` (`services/checks/manager.go:394`) only suppresses resolution for checkers that reported `CodeCheckFailed`, so `CheckerTransactions` stays resolvable.

Concretely:
1. Sweep 1 records `Transaction Check|tx_status_mismatch|T|` as critical.
2. Sweep 2 hits a chaincode/query error for `T`, emits this Info finding.
3. `ResolveFindingsNotSeenSince` closes the critical row, because its `last_seen` predates `seenAt`.

A ledger outage is turned into a clean bill of health — the exact failure mode this comment, `resolvableCheckers`' own godoc, and `docs/services/storage/checks.md:85` all say the design prevents.

Either give the unavailable case `CodeCheckFailed`, or exclude a checker from resolution when it reported *any* inconclusive code.

token/services/storage/db/sql/common/findings.go
```diff
@@ -0,0 +1,243 @@
...
+		db.table.Findings,
+		db.table.Findings, db.table.Findings,
+		db.table.Findings, db.table.Findings,
+		db.table.Findings, db.table.Findings,
+	)
+}
+
+// UpsertFindings records the findings observed by one sweep. See the driver
+// interface for the aging semantics.
+func (db *TransactionStore) UpsertFindings(ctx context.Context, findings []dbdriver.FindingRecord, seenAt time.Time) error {
```
AkramBitar, line 84: **Blocker: one statement for every finding, so a large sweep records nothing.**

11 bind parameters per row against SQLite's 32766-variable ceiling (~2977 rows) and Postgres's 65535 (~5957 rows). Verified on SQLite: `n=1000` succeeds, `n=3000` fails with `too many SQL variables`.

The error aborts `sweep()` before `ResolveFindingsNotSeenSince`, so the sweep fails and re-fails every interval having persisted nothing.

This is reachable, not theoretical: `CheckTokenSpendability` emits one critical finding **per unspent token** when a public-parameters upgrade makes the stored format unsupported, and `CheckLocalCompleteness` one per missing token. A node with >3000 unspent tokens loses the entire report exactly when it matters most.

Suggest chunking into `BatchSize`-sized batches that all share one `seenAt`.

token/services/storage/services/checks/config.go
```diff
@@ -0,0 +1,85 @@
...
+	if !cfg.IsSet(ConfigKeyChecks) {
+		return result, nil
+	}
+
+	var loaded Config
+	if err := cfg.UnmarshalKey(ConfigKeyChecks, &loaded); err != nil {
+		return result, err
+	}
+
+	result.Enabled = loaded.Enabled
```
AkramBitar: **Blocker: any `checks:` block without an explicit `enabled` silently disables the sweep.**

`result.Enabled = loaded.Enabled` is unconditional, so a config that only tunes e.g. `scanInterval: 2h` yields `Enabled: false`.

That contradicts three things at once: `LoadConfig`'s own godoc four lines above ("falling back to DefaultConfig for anything it does not set"), the `Config.Enabled` field comment ("Default true"), and `docs/configuration.md:334` ("Default: true").

The result is a node whose operator believes drift checking is on while nothing runs — the state this service exists to end. `TestLoadConfig_EnabledDefaultsFalseWhenKeyPresentButUnset` documents the behaviour deliberately, but a `*bool` (or `cfg.IsSet(ConfigKeyChecks+".enabled")`) removes the trap. Note the same fix would be needed in `recovery`/`cleanup` for consistency.

token/services/storage/db/common/checks.go
```diff
@@ -267,25 +686,297 @@ func (a *DefaultCheckers) CheckTokenSpendability(ctx context.Context) ([]string,
...
+	ids []*token2.ID,
+) ([]Finding, error) {
+	if _, _, err := qe.WhoDeletedTokens(ctx, ids...); err == nil {
+		// every id resolved, so all of them are in the local store, spent or not
+		return nil, nil
+	}
+
+	var missing []*token2.ID
+	for _, id := range ids {
+		if _, _, err := qe.WhoDeletedTokens(ctx, id); err != nil {
```
AkramBitar: **Blocker: a local store query failure is reported as missing money.**

`checkPresenceBatch` treats any error from `qe.WhoDeletedTokens` as "token absent". The SQL implementation (`sql/common/tokens.go:1108`) returns `token not found` for a genuine miss but also propagates raw driver errors (connection reset, statement timeout) identically.

So a transient local DB failure puts every id in the batch through `reportMissingLocally`, and since the ledger still holds them unspent, each becomes a `SeverityCritical` `token_missing_locally` finding logged at error level — "the node owns money it cannot see".

This is the precise conflation that `checkUnspentBatch:557` and `checkUnspentOneByOne:593` go out of their way to avoid. Distinguish not-found from a query error (a sentinel, or a presence query returning a bool).

token/services/storage/db/common/checks.go
```diff
@@ -194,33 +520,118 @@ func (a *DefaultCheckers) CheckUnspentTokens(ctx context.Context) ([]string, err
...
-					return nil
-				}
-			}
+		// the ledger call itself failed: nothing was learned about any token in this
+		// batch. Reporting that as every token in it being missing would turn a
+		// connectivity blip into a false critical alarm that the node's money is
+		// gone, so this bubbles up as a failure of the check instead.
+		return nil, errors.WithMessagef(err, "failed querying ledger for tokens [%v]", ids)
+	}
+	if len(ledgerContent) != len(ids) {
```
AkramBitar: **`checkUnspentOneByOne` is unreachable, and `token_missing_on_ledger` is never produced.**

Its godoc says "a failed batch is resolved one token at a time", but the error branch two lines up does `return nil, err`; the fallback is gated only on this length mismatch — which neither backend produces:

- the Fabric translator (`translator.go:167`) returns an **error** when any key is absent;
- fabricx (`qe.go:145`) returns a nil entry at that position, preserving length.

So on Fabric a single absent token fails the whole Unspent Tokens Check (`check_failed`), losing every other token's comparison; on fabricx it surfaces as `token_content_mismatch` rather than `token_missing_on_ledger`.

Worse: a token spent concurrently while the sweep walks its local snapshot triggers this on a perfectly healthy busy node — a routine false `check_failed` on Fabric, a false **critical** on fabricx. The fallback should be driven by the error, not by the length.

AkramBitar (commented):
## Review: background ledger drift checks

Solid, well-tested feature — build, `go vet`, and all new tests pass at `3de2ff56c`. I verified the generated upsert SQL against both backends (`occurrences` is correctly table-qualified, `first_seen` correctly excluded from the conflict update, and the `Lt("last_seen", seenAt)` predicate correctly avoids resolving rows the same sweep just wrote), and confirmed positional comparison in `checkUnspentBatch` is safe since both backends preserve request order.

That said, I found **four defects that break the guarantee the service exists to provide** — a checker that reports "clean" when it isn't, or records nothing at all. These are inline and marked *Blocker*:

1. `db/common/checks.go:430` — a ledger that can't answer `Status` causes the next sweep to **close previously recorded critical findings**.
2. `sql/common/findings.go:74` — a sweep with >~3000 findings exceeds the bind-parameter ceiling and persists **nothing**, repeatedly.
3. `services/checks/config.go:70` — any `checks:` block without an explicit `enabled` **silently disables** the sweep, contradicting its own godoc and the docs.
4. `db/common/checks.go:901` — a transient local-DB error is reported as a **critical** `token_missing_locally`.

Plus four more inline (one-by-one fallback unreachable, timeout covering the writes, nil metrics-provider panic, `Stop()` never wired) and one minor.

### One finding with no line to anchor to

**The `auditor` sweep is not leader-elected on Postgres.** `sqlcommon.NewAuditTransactionStore` delegates to `NewOwnerTransactionStore`, which passes `recoveryLeaderFactory = nil`; `AcquireRecoveryLeadership` (`sql/common/transactions.go:388`) then returns `noopRecoveryLeadership{}, true` unconditionally. Both `services/checks/manager.go:74` and `docs/services/storage/checks.md:117` promise "only one replica sweeps a given store at a time, decided by a PostgreSQL advisory lock".

With N auditor replicas on one database, all N run the full sweep every interval — multiplying ledger traffic — and the slower replica's `ResolveFindingsNotSeenSince` can close findings the faster one just recorded. Either wire the advisory-lock factory into the audit store, or stop claiming election for that role. (The mechanism lives in code this PR doesn't touch, hence no inline anchor.)

token/services/storage/services/checks/manager.go
```diff
@@ -0,0 +1,417 @@
...
+		return err
+	}
+	m.metrics.SweepsTotal.With("role", m.role, "outcome", outcomeCompleted).Add(1)
+
+	return nil
+}
+
+// sweep runs the checks once and records what they found.
+func (m *Manager) sweep(ctx context.Context) error {
+	ctx, cancel := context.WithTimeout(ctx, m.config.Timeout)
```
AkramBitar: **The sweep timeout also covers the writes, so the longest sweeps record nothing.**

`Check`, `UpsertFindings` and `ResolveFindingsNotSeenSince` all share one `Timeout`-bounded context. A sweep whose checks consume nearly the full 30 minutes reaches `UpsertFindings` with almost no budget; when the deadline hits, the findings are discarded and the sweep is counted `failed`.

The nodes with the most drift and the longest histories are exactly the ones that will never persist a finding. Suggest bounding only `checker.Check` with `Timeout`, giving the two writes their own short deadline off the parent context.

token/sdk/db/checks.go
```diff
@@ -7,36 +7,150 @@ SPDX-License-Identifier: Apache-2.0
...
+	)
+	// custom checkers registered through dependency injection still return plain
+	// messages, so they are lifted rather than reimplemented
+	checkers = append(checkers, common.AsNamedFindingCheckers(custom)...)
+
+	manager := checks.NewManager(
+		logger,
+		storage,
+		common.NewFindingsService(checkers),
+		checks.NewMetrics(metrics.NewTMSProvider(tmsID, s.metricsProvider)),
```
AkramBitar, line 80: **The documented nil-provider guard is defeated by the wrapper.**

The comment states "A nil provider discards everything, so a caller with no metrics configured needs no special case" — but `metrics.NewTMSProvider(tmsID, nil)` returns a non-nil `*tmsProvider` holding a nil `provider`, so the guard never fires and `p.provider.NewCounter` nil-derefs.

Reachable for the hand-wired setups this file explicitly contemplates: `NewOwnerCheckServiceProvider(..., configuration, nil)` with a non-nil configuration panics inside `CheckService`, i.e. during ttx/auditor service construction. Guard `s.metricsProvider == nil` before wrapping.

token/sdk/dig/sdk.go
```diff
@@ -115,7 +116,7 @@ func (p *SDK) Install() error {
 		p.Container().Provide(ftsconfig.NewService),
 		p.Container().Provide(
 			digutils.Identity[*ftsconfig.Service](),
-			dig.As(new(cleanup.Configuration)),
+			dig.As(new(cleanup.Configuration), new(checks.Configuration)),
```
AkramBitar, line 119: **`Stop()` on the check service providers is never wired into the SDK lifecycle.**

`sweeper.Stop`, `AuditorCheckServiceProvider.Stop` and `OwnerCheckServiceProvider.Stop` all exist, but nothing invokes them and `SDK.Start` registers no corresponding shutdown.

Each running sweep therefore keeps its goroutine — and on Postgres its dedicated advisory-lock connection — past SDK shutdown. An in-process SDK restart leaves the old sweep holding the lock and the new one permanently reporting `not_leader`.

token/services/storage/db/driver/checks.go
```diff
@@ -0,0 +1,87 @@
...
+	// IncludeResolved also returns findings that are no longer observed.
+	IncludeResolved bool
+	// Limit caps the number of returned findings. Zero or negative means no cap.
+	Limit int
+}
+
+// FindingsStore stores the findings of the ledger drift checks.
+//
+
+type FindingsStore interface {
```
AkramBitar, line 64: **Minor: this doc comment is detached from the interface.**

The block ends with a bare `//` followed by a blank line, so exported `FindingsStore` has no godoc at all. AGENTS.md requires one, and `revive`'s `exported` rule isn't enabled so lint won't catch it.

Two other small things while here:
- `validateConfig` doesn't reject `Timeout > ScanInterval`, which yields back-to-back sweeping.
- `mockTransactionsStore` in `sql/{sqlite,postgres}/transactions_test.go` still builds `TableNames` without `Findings` — harmless only because those tests never call `GetSchema`, which would now emit `CREATE TABLE IF NOT EXISTS  (`.

AkramBitar (changes requested):

token/services/storage/db/common/checks.go
atharrva01: Fixed in 1ab94db60. resolvableCheckers now checks dbcommon.IsInconclusive(code) instead of just == CodeCheckFailed, so a checker reporting tx_status_unavailable stays excluded from resolution too.

token/services/storage/db/sql/common/findings.go
atharrva01: Fixed in 1ab94db60. UpsertFindings now chunks at 1000 rows per statement, all sharing the sweep's one seenAt.

token/services/storage/services/checks/config.go
atharrva01: Fixed in 1ab94db60. LoadConfig now gates Enabled on cfg.IsSet(ConfigKeyChecks + ".enabled"), same pattern already used for NotFoundGracePeriod in recovery. Left recovery/cleanup alone for this PR since they're actively touched by other open PRs right now, happy to open a follow-up for those.

token/services/storage/db/common/checks.go
atharrva01: Fixed in 1ab94db60. Added driver.ErrTokenNotFound, WhoDeletedTokens wraps it for a genuine miss, checkPresenceBatch now only treats errors.Is(err, ErrTokenNotFound) as absence and propagates anything else as a real failure.

token/services/storage/db/common/checks.go
atharrva01: Fixed in 1ab94db60. checkUnspentBatch now falls back to checkUnspentOneByOne on the batch error itself, not the length check, and a nil entry in a same-length successful response is reported as token_missing_on_ledger instead of going through the content comparison.

token/services/storage/services/checks/manager.go
atharrva01: Fixed in 1ab94db60. checker.Check is still bounded by Config.Timeout, but UpsertFindings/ResolveFindingsNotSeenSince now run under their own 2m deadline off the parent context instead of sharing it.

token/sdk/db/checks.go
atharrva01: Fixed in 1ab94db60, at the root: NewTMSProvider now returns the Provider interface and returns a true nil when given a nil provider, instead of *tmsProvider (which wrapped nil into a non-nil interface value). That covers this call site and the two driver.go ones without needing a nil check at every caller.

token/sdk/dig/sdk.go
atharrva01: Fixed in 1ab94db60. Added SDK.Stop(), which stops both check service providers (via the dig container) before delegating to the underlying SDK's Stop.

token/services/storage/db/driver/checks.go
atharrva01: Fixed in 1ab94db60: the orphaned doc comment is reattached, validateConfig now rejects Timeout > ScanInterval, and both mockTransactionsStore helpers include the Findings table name.

token/services/storage/services/checks/manager.go
```diff
@@ -0,0 +1,438 @@
...
+	case m.config.ScanInterval <= 0:
+		return errors.Errorf("invalid checks scan interval [%s]", m.config.ScanInterval)
+	case m.config.Timeout <= 0:
+		return errors.Errorf("invalid checks timeout [%s]", m.config.Timeout)
+	case m.config.BatchSize <= 0:
+		return errors.Errorf("invalid checks batch size [%d]", m.config.BatchSize)
+	case m.config.TransactionWindow < 0:
+		return errors.Errorf("invalid checks transaction window [%s]", m.config.TransactionWindow)
+	case m.config.Timeout > m.config.ScanInterval:
+		return errors.Errorf("invalid checks timeout [%s] exceeds scan interval [%s], sweeps would run back to back",
```
Effi-S, line 208: The User can only set `ScanInterval` 
Timeout is kept as default 30m 
The docs do note timeout must not exceed ScanInterval 
But the error propagates.. 
Maybe at some point this error should be logged?

token/services/storage/services/checks/manager.go
```diff
@@ -0,0 +1,438 @@
...
+	Check(ctx context.Context) ([]dbcommon.Finding, error)
+	// ResolvableCheckers names the checks whose stored findings may be closed when
+	// a sweep stops reporting them.
+	ResolvableCheckers() []string
+}
+
+// Manager runs the ledger drift checks in the background.
+//
+// The checks themselves are not new: a node has always been able to run them on
+// demand. What was missing is anyone running them on a node that is simply up,
```
Effi-S, line 78: the sweep for a TMS begins only the first time something materializes that TMS's owner/auditor Service.
not at node boot.
I think we should allow instantiating the service at `Start()`. 
Maybe not as default, but as an option.. 
Would like to hear your thoughts..

token/sdk/dig/sdk.go
```diff
@@ -317,6 +318,30 @@ func (p *SDK) Start(ctx context.Context) error {
...
+// holding the lock and the new one never acquires leadership.
+func (p *SDK) Stop() error {
+	if !p.TokenEnabled() {
+		return p.SDK.Stop()
+	}
+
+	var stopErrs []error
+	if err := p.Container().Invoke(func(a *db.AuditorCheckServiceProvider, o *db.OwnerCheckServiceProvider) error {
+		return errors2.Join(a.Stop(), o.Stop())
+	}); err != nil {
```
Effi-S, line 336: Nit: Please use FSC `errors.Join`

Effi-S (changes requested):

token/services/storage/services/checks/manager.go
atharrva01: Fixed in 37f9db22a. When only scanInterval is set, the default timeout now scales down to match it instead of failing validation. An explicit timeout is still validated against scanInterval as before.

token/sdk/dig/sdk.go
atharrva01: Good catch, switched to errors.Join in 37f9db22a.

token/services/storage/services/checks/manager.go
atharrva01: Fair point, and honestly I went back and forth on this. Starting eager at boot means we spin up sweeps for every configured TMS whether or not it's ever used, which is wasted work on nodes that only touch a subset. Lazy-on-first-use keeps that free, but yeah, it means drift can go unchecked until traffic shows up.

I'd rather not bolt an eager option onto this PR, it's already touching a lot of surface area. Let's track it as a follow-up, sounds like a reasonable knob to add.

AkramBitar:
@Effi-S

Any update with this PR? 

Regards,
Akram

token/services/storage/db/common/checks.go
```diff
@@ -242,33 +594,130 @@ func (a *DefaultCheckers) CheckUnspentTokens(ctx context.Context) ([]string, err
...
+			findings = append(findings, finding)
 		}
 	}
 
-	return errorMessages, nil
+	return findings, nil
+}
+
+// checkUnspentOneByOne resolves each token of a failed batch on its own, so that a
+// single missing token is reported as such instead of hiding the rest of the batch.
```
Effi-S, line 686: `checkUnspentOneByOne` omits the nil-ledger-entry guard that checkUnspentBatch has, so a `len-1 [nil]` response is misclassified as `token_content_mismatch` (compared against nil) instead of `token_missing_on_ledger`.

Failure scenario: If any backend returns a single-element slice with a nil entry without an error (as fabricx's `QueryTokens` does for missing tokens, `[nil]`) while a batch call errored, the fallback reaches if `len(ledgerContent) != 1` (len is 1, so false) and calls `compareTokenContent(id, local[*id], nil)`, reporting `CodeTokenContentMismatch`. The batch path explicitly special-cases `ledgerContent[i] == nil` as `CodeTokenMissingOnLedger`; the one-by-one path does not, so the same underlying condition produces a different, wrong code.

token/sdk/db/checks.go
```diff
@@ -7,40 +7,156 @@ SPDX-License-Identifier: Apache-2.0
...
+// CheckService creates a check service for the given TMS ID and databases, and
+// starts the background drift sweep over the audit store.
 // It combines default checkers with custom checkers provided during initialization.
 func (a *AuditorCheckServiceProvider) CheckService(id token.TMSID, adb *auditdb.StoreService, tdb *tokens.Service) (auditor.CheckService, error) {
-	return common.NewChecksService(append(common.NewDefaultCheckers(a.tmsProvider, a.networkProvider, adb, tdb, id, a.MaxPageSize), a.checkers...)), nil
+	if err := a.sweeper.start(a.tmsProvider, a.networkProvider, adb, adb, tdb, id, a.checkers, checks.RoleAuditor, a.MaxPageSize); err != nil {
+		return nil, err
+	}
+
+	defaultCheckers := common.AsNamedCheckers(common.NewDefaultFindingCheckers(a.tmsProvider, a.networkProvider, adb, tdb, id, common.WithMaxPageSize(a.MaxPageSize)))
```
Effi-S, line 152: The on-demand Check API now silently drops every `SeverityInfo` finding, so a Confirmed transaction whose ledger status cannot be fetched is no longer surfaced at all (previously it produced a message).

Failure scenario: `AsNamedChecker` skips `SeverityInfo` findings. `compareTransactionStatus` now returns `CodeTxStatusUnavailable (SeverityInfo)` when `l.Status `errors, where the old code appended a `'failed to get ledger transaction status'` message plus a mismatch message. Any caller or integration test asserting on the message content/count of the on-demand auditor/owner Check API (the PR notes such assertions exist) will see fewer/different messages, and a genuine ledger-connectivity problem is no longer visible through the on-demand path.

Effi-S (changes requested):

token/services/storage/db/common/checks.go
atharrva01: Fixed in e5edca36f. checkUnspentOneByOne now treats a nil entry the same as checkUnspentBatch does, reporting token_missing_on_ledger instead of falling through to compareTokenContent against nil.

token/sdk/db/checks.go
atharrva01: Fixed in 30c3de53e. compareTransactionStatus now raises severity to Warning when the local record is Confirmed, so that case survives AsNamedChecker's Info filter and reaches the on-demand path again, same as the old checker always did for it. Every other local status stays Info and gets filtered like before, since a blanket removal of the filter would bring back the itest flakiness the first review round fixed (a transient status lookup failure right after a node restart, e.g. "no such transaction ID in index", used to read as a fresh error to every legacy caller).

Effi-S:
@atharrva01, 
Please squash commits, rebase, and fix the CI Failures

token/services/storage/db/common/checks.go
```diff
@@ -242,33 +613,133 @@ func (a *DefaultCheckers) CheckUnspentTokens(ctx context.Context) ([]string, err
...
+func (a *DefaultCheckers) checkUnspentOneByOne(
+	ctx context.Context,
+	tms *token.ManagementService,
+	net *network.Network,
+	local map[token2.ID][]byte,
+	ids []*token2.ID,
+) ([]Finding, error) {
+	var findings []Finding
+	for _, id := range ids {
+		ledgerContent, err := net.QueryTokens(ctx, tms.Namespace(), []*token2.ID{id})
```
Effi-S, line 704: `checkUnspentBatch` / `checkUnspentOneByOne` are written around **fabricx** semantics — an absent token comes back as a `nil` entry at its slice position, which the code maps to a critical `CodeTokenMissingOnLedger` finding. The code comments say so explicitly.

But on the **classic Fabric** backend, `net.QueryTokens` routes to `common/rws/translator.QueryTokens`, which for an absent output (`len(bytes) == 0`) appends `"output for key [...] does not exist"` and returns `nil, error` if *any* id errored (translator.go:162–171, verified). So:

1. `checkUnspentBatch` → `net.QueryTokens(ids)` returns an error → falls back to `checkUnspentOneByOne`.
2. In `checkUnspentOneByOne`, the single-id `QueryTokens([id])` for the absent token also returns an error → `return nil, errors.WithMessagef(...)`.
3. `CheckUnspentTokens` returns an error, which `FindingsService` turns into an **inconclusive `CodeCheckFailed`**, and `CheckerUnspentTokens` is dropped from `resolvableCheckers` so prior findings **never close**.

**Failure scenario**: A classic-Fabric node holds a token locally as unspent that the ledger no longer has — the exact case this feature exists to catch, and the one that costs the owner money. Instead of a critical `token_missing_on_ledger` finding, the operator sees an inconclusive "check failed" and no finding ever ages/closes. One missing token also aborts comparison of the rest of the batch. The one-by-one fallback was clearly *intended* to handle this, but its `return nil, err` on query failure (correct for a genuine ledger-unreachable error) cannot distinguish "ledger unreachable" from "token absent" on this backend, so it can never emit the missing-token finding here.

token/services/storage/services/checks/config.go
```diff
@@ -0,0 +1,97 @@
...
+	// rather than the Go zero value. Without this gate, a checks: block that
+	// only tunes another field (e.g. scanInterval) would silently disable the
+	// sweep, contradicting the documented default of true.
+	if cfg.IsSet(ConfigKeyChecks + ".enabled") {
+		result.Enabled = loaded.Enabled
+	}
+	if loaded.ScanInterval > 0 {
+		result.ScanInterval = loaded.ScanInterval
+	}
+	if loaded.Timeout > 0 {
```
Effi-S, line 80: a checks-only misconfig fails the whole owner/auditor service (availability regression)
A parseable-but-inconsistent config (`timeout > scanInterval`, e.g. `timeout=2h`, `scanInterval=1h`) survives `LoadConfig`/`ConfigFor`: line 80 assigns `loaded.Timeout` unconditionally, and the clamp at line 82 only runs when `timeout` is unset; `ConfigFor` falls back to defaults only on a load *error*, not on a semantically bad value. `Manager.Start` then calls `validateConfig`, which errors at `manager.go:207` (`Timeout > ScanInterval`). That error propagates `sweeper.start` → `CheckService` → the lazy owner/auditor service provider returns `nil, err`.

**Failure scenario**: A single mistyped checks interval disables token transaction processing for the whole node — directly contradicting the documented "safety net, never fail the node" design. Either clamp/normalize in `LoadConfig` (as is already done for an unset timeout), or make `validateConfig` failures non-fatal (log + disable checks, keep the node up).

token/services/storage/db/common/checks.go
```diff
@@ -132,92 +402,178 @@ func (a *DefaultCheckers) CheckTransactions(ctx context.Context) ([]string, erro
...
 	}
 	var page driver2.Pagination
-	page, err = pagination.Offset(0, pageSize)
+	page, err := pagination.Offset(0, pageSize)
 	if err != nil {
-		return nil, errors.WithMessagef(err, "failed to create pagination [%s]", tms.ID())
+		return errors.WithMessagef(err, "failed to create pagination [%s]", tmsID)
 	}
+
+	seen := collections.NewSet[string]()
```
Effi-S: **unbounded in-memory txID set on large-history nodes**

`walkTransactions` accumulates every distinct `txID` seen across the whole walk into a `collections.Set`. With the default `TransactionWindow=0` (full history), this set grows with total ledger history, not with a bounded page — and **both** `CheckTransactions` and `CheckLocalCompleteness` run it on full history in the background sweep.

**Failure scenario**: A node with millions of confirmed transactions builds a set holding every distinct txID string for the duration of each sweep, so sweep memory scales with lifetime history. This is exactly the large-history node the config warning targets, yet the dedup set itself is not bounded by the window.

token/services/storage/services/checks/manager.go
```diff
@@ -0,0 +1,438 @@
...
+	for _, finding := range findings {
+		open[finding.Severity]++
+		m.metrics.FindingsObserved.With(
+			"role", m.role,
+			"checker", finding.Checker,
+			"code", finding.Code,
+			"severity", finding.Severity.String(),
+		).Add(1)
+	}
+	for severity, count := range open {
```
Effi-S, line 377: `FindingsOpen` gauge goes stale after leadership loss
The `FindingsOpen` gauge is `Set` only on a leader sweep. A replica that hits `if !acquired` (manager.go:282) returns before `report()`, so it never resets its gauge.

**Failure scenario**: Replica A is leader and sets `storage_checks_findings_open{severity=critical}=5`. Leadership moves to B (A's advisory-lock connection drops). A now exports a frozen `critical=5` forever while B reports the real number. An alert on this gauge — which the docs call out as "the one to alert on" — fires on stale data or double-counts across replicas. Reset the gauge to 0 (or clear it) on the non-leader path.

token/services/storage/services/checks/manager.go
```diff
@@ -0,0 +1,438 @@
...
+	default:
+		return nil
+	}
+}
+
+// checksLoop sweeps on the configured interval until the manager is stopped.
+func (m *Manager) checksLoop() {
+	defer m.wg.Done()
+
+	ticker := time.NewTicker(m.config.ScanInterval)
```
Effi-S: **ticker measures from construction, not from sweep completion**
`NewTicker` is created before the jitter delay and the initial sweep. On a large node the initial sweep can run for nearly the full `Timeout` (which is clamped to `ScanInterval`), so by the time the loop reaches `<-ticker.C` a tick is already buffered and the next sweep fires back-to-back with no real gap.

**Failure scenario**: `ScanInterval=1h`; initial sweep takes ~1h; the buffered tick triggers a second sweep immediately instead of one interval later. Reset/create the ticker after the initial sweep, or drain the buffered tick.

Effi-S (changes requested):

atharrva01:
hi @Effi-S , i have done the changes , you can take a look now

PR #2324, opened 2026-08-27
fix(storage): detach status notifications from the caller's context

### LINKED ISSUE #2316: recovery: timed-out attempt can commit after status notification is already dropped
Follow-up from review on #2287 (@AkramBitar, manager.go:496).

`TransactionTimeout` bounds `Handler.Recover`, but the deadline can fire partway through the handler's own commit tail rather than before it starts. In `Commit` (`token/services/ttx/finality/listener.go:219-243`):

1. `tx.SetStatus(ctx, ...)` succeeds just before the deadline.
2. `tx.Commit()` takes no context and still succeeds, the row is now durably `Confirmed`.
3. `ttxDB.NotifyStatus(ctx, ...)` then runs on the already-expired `recoverCtx`. `StatusSupport.safeSend` selects on `<-event.Ctx.Done()` (`token/services/storage/db/common/status.go:130`) and drops the event, deterministically when the listener channel buffer is full.

Meanwhile the recovery manager classifies the attempt as `DeadlineExceeded` and writes `status_message = "recovery failed: context deadline exceeded"` over a row that is actually `Confirmed`. Finality waiters only wake via fallback polling, not the notification path, until that happens.

Suggested fix (per Akram): use a non-cancellable context (`context.WithoutCancel`) for the post-commit publication path in `listener.go`, so the notification is not tied to `recoverCtx`'s deadline once the commit itself has already succeeded.

Scoped as a follow-up rather than folded into #2287: the fix lives in a different package (`token/services/ttx/finality`) than the recovery manager the PR touches.

atharrva01:
Fixes #2316

SetStatus and the finality listener's post-commit Commit both call NotifyStatus only after the status change is already durably written. NotifyStatus passed the caller's ctx straight through to the event, and safeSend's select uses that ctx purely to avoid blocking forever on a stuck listener, not to decide whether the write itself succeeded.

That conflates two unrelated lifetimes. A recovery attempt bounds its per-transaction work with TransactionTimeout. If the write finishes right before that deadline, the notification that follows inherits the same already-expired context, and safeSend drops it the moment the listener channel buffer is full, even though the write succeeded. A finality waiter then only recovers through the fallback poller instead of the push path, and the recovery manager can log a spurious "deadline exceeded" over a row that is actually Confirmed.

Fix: NotifyStatus in both ttxdb and auditdb detaches the event's context with context.WithoutCancel before notifying. The write already ran under its own context-bound path; the notification's only open question is whether a listener is still around to receive it, which the caller's deadline has nothing to do with. This also covers the Commit path in listener.go for free, since it calls through the same NotifyStatus.

## Test plan
- [x] Added a regression test in each store that fills the listener buffer, cancels the context, and asserts the notification still gets delivered once the buffer frees up. Verified both fail deterministically (5/5) against the pre-fix code and pass deterministically (5/5) with the fix.
- [x] `go test ./token/services/storage/... ./token/services/ttx/... ./token/services/network/...` with `-race`, all green.
- [x] `golangci-lint run` clean on everything touched.

### DIFF
--- a/token/core/zkatdlog/nogh/v1/benchmark/auditor_setup.go
+++ b/token/core/zkatdlog/nogh/v1/benchmark/auditor_setup.go
@@ -84,8 +84,7 @@ func NewAuditCheckSetup(conf *SetupConfiguration) (*AuditCheckSetup, error) {
 					},
 					Outputs: []*driver.IssueOutputMetadata{
 						{
-							OutputMetadata:  metaBytes,
-							OutputAuditInfo: conf.OwnerIdentity.AuditInfo,
+							OutputMetadata: metaBytes,
 							Receivers: []*driver.AuditableIdentity{
 								{
 									Identity:  conf.OwnerIdentity.ID,
--- a/token/services/storage/auditdb/store.go
+++ b/token/services/storage/auditdb/store.go
@@ -299,8 +299,16 @@ func (d *StoreService) SetStatus(ctx context.Context, txID string, status dbdriv
 // NotifyStatus pushes a status event to the in-process listeners without
 // touching the database. Used after a transaction-scoped status update,
 // which bypasses SetStatus.
+//
+// The event carries a context detached from ctx's cancellation: by the time
+// this runs, the status is already durably written, and safeSend only uses
+// the context to give up on a stuck listener. Keeping ctx as-is would let a
+// caller-side deadline that has nothing to do with the listener (for example
+// a recovery attempt's per-transaction timeout, expiring right after its
+// write commits) silently drop a notification for a status change that did
+// succeed. See issue #2316.
 func (d *StoreService) NotifyStatus(ctx context.Context, txID string, status dbdriver.TxStatus, message string) {
-	d.Notify(common.StatusEvent{Ctx: ctx, TxID: txID, ValidationCode: status, ValidationMessage: message})
+	d.Notify(common.StatusEvent{Ctx: context.WithoutCancel(ctx), TxID: txID, ValidationCode: status, ValidationMessage: message})
 }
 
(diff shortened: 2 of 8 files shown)


Effi-S (changes requested):
token/services/storage/db/common/status.go:116-121 still states that the context guard is what prevents a blocking send from hanging forever:

// ... a blocking send without a context guard could block forever if the buffer is
// full and the consumer has departed. This method handles both cases.

Effi-S (changes requested):
token/services/storage/endorserdb/store.go:162 follows the same "notify after the write is committed" pattern, but still passes the raw caller context:
```go
if err := w.Commit(); err != nil {
    return errors.Wrapf(err, "failed committing status [%s][%s]", ...)
}

// notify the listeners
d.Notify(common.StatusEvent{
    Ctx:            ctx,          // <-- not detached; same #2316 exposure
    TxID:           txID,
    ValidationCode: status,
})
```

This uses the same StatusSupport/safeSend machinery, so a caller whose context expires just after Commit() can drop the notification in exactly the way #2316 describes. If endorserdb status listeners are ever driven off a deadline-bounded path (recovery, or any WithTimeout caller), the bug reappears there.

Recommendation: Apply the same context.WithoutCancel treatment to the endorserdb notify site for consistency, or explicitly document why endorserdb is exempt (e.g. no finality/recovery listeners are registered against it). Leaving one of three stores with the old behavior invites regressions and makes the codebase harder to reason about.

AkramBitar (changes requested):
@atharrva01 

Thanks a lot for this PR, see my comments. 

Regards,
Akram

Effi-S (approved):

AkramBitar (approved):
LGTM

PR #2325, opened 2026-08-27
network-driver: empty candidate-set entry silently "succeeds" with zero endorsers in SelectEndorsersForMSPSets

### LINKED ISSUE #2049: network-driver: empty candidate-set entry silently "succeeds" with zero endorsers in SelectEndorsersForMSPSets
### Summary

A candidate-set entry that is present but lists zero required MSPs (an empty, non-nil `[]string`) is vacuously "satisfied" by `SelectEndorsersForMSPSets`: the inner loop over `requiredMSPIDs` never executes, so `satisfied` stays `true`, and the function returns success with zero selected endorsers and no error.

### Where

`token/services/network/fabric/endorsement/fsc/selection.go:39-53`

### Impact

If a degenerate/empty candidate set is ever produced upstream (a policy-evaluation bug, or a namespace with a misconfigured/empty endorsement policy), this function reports success without selecting any endorser — a transaction could proceed to be sent for endorsement to nobody, silently, rather than failing loudly on the malformed policy input.

### Reproduction

```go
candidates := [][]string{{}}
selected, err := fsc.SelectEndorsersForMSPSets([]view.Identity{org1Endorser1}, mspOf, candidates)
require.NoError(t, err)
assert.Len(t, selected, 0)
```

### Suggested fix

Treat an empty (but present) candidate-set entry as invalid input and return an error, rather than vacuous success with zero endorsers.

### Severity

LOW

Effi-S:
Fixes #2049

### Summary

A candidate-set entry that is present but lists zero required MSPs (an empty, non-nil `[]string`) is vacuously "satisfied" by `SelectEndorsersForMSPSets`: the inner loop over `requiredMSPIDs` never executes, so `satisfied` stays `true`, and the function returns success with zero selected endorsers and no error.

### Where

`token/services/network/fabric/endorsement/fsc/selection.go:39-53`

### Impact

If a degenerate/empty candidate set is ever produced upstream (a policy-evaluation bug, or a namespace with a misconfigured/empty endorsement policy), this function reports success without selecting any endorser — a transaction could proceed to be sent for endorsement to nobody, silently, rather than failing loudly on the malformed policy input.

### Reproduction

```go
candidates := [][]string{{}}
selected, err := fsc.SelectEndorsersForMSPSets([]view.Identity{org1Endorser1}, mspOf, candidates)
require.NoError(t, err)
assert.Len(t, selected, 0)
```

### Suggested fix

Treat an empty (but present) candidate-set entry as invalid input and return an error, rather than vacuous success with zero endorsers.

### Severity

LOW

### DIFF
--- a/token/services/network/fabric/endorsement/fsc/selection.go
+++ b/token/services/network/fabric/endorsement/fsc/selection.go
@@ -33,9 +33,18 @@ import (
 //
 // It returns an error if none of the candidate sets can be fully covered by distinct
 // configured endorsers. The error names the first MSP that blocked each candidate set.
+//
+// An empty (but present) candidate set — one requiring no MSP at all — is rejected as
+// malformed policy input rather than treated as vacuously satisfied: a policy satisfiable
+// by zero endorsers would let a transaction be endorsed by nobody, so it must fail loudly.
 func SelectEndorsersForMSPSets(configured []view.Identity, mspOf func(view.Identity) (string, error), candidates [][]string) ([]view.Identity, error) {
 	if len(candidates) == 0 {
 		return nil, errors.Errorf("no candidate MSP set to satisfy the namespace endorsement policy")
+	}
+	for _, candidate := range candidates {
+		if len(candidate) == 0 {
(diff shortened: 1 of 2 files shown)


AkramBitar (approved):
LGTM

PR #2329, opened 2026-08-28
fix(core): run type-gated authorizations before the wallet-based one

### LINKED ISSUE #2328: Tokens owned by a bound policy identity are stored without ownership rows and become unspendable
## Summary

`AuthorizationMultiplexer.IsMine` is first-match-wins, and both drivers (fabtoken, zkatdlog) register the generic `WalletBasedAuthorization` **before** the type-gated authorizations (`htlc.ScriptAuth`, `multisig.EscrowAuth`, `boolpolicy.EscrowAuth`). Once a boolpolicy composite identity gets bound to a wallet, the wallet-based authorization resolves it first and claims the token with the base wallet id and **no ownership ids** — so `StoreToken` persists the token without any ownership row, and the policy wallet's selector (which looks tokens up under the `policy`-prefixed wallet id) can no longer see it. The balance still reports the funds; any transfer fails with "insufficient funds".

## How the identity gets bound

`AnonymousOwnerWallet.RegisterRecipient` (token/services/identity/role/wallets.go) registers the recipient data of the transfer rest and then calls `BindIdentity(data.Identity, …, w.WalletID)`. When the rest goes back to the wallet's own boolpolicy composite identity — the normal case for a transfer that spends policy-owned inputs — this binds the composite identity to the wallet. From that moment `WalletService.OwnerWallet(composite)` succeeds, and `WalletBasedAuthorization.IsMine` outruns `boolpolicy.EscrowAuth` for every subsequent token owned by that identity.

`LongTermOwnerWallet.RegisterRecipient` accepts the policy rest recipient without binding (#2268), which is why wallets backed by long-term identities are not affected; the anonymous (idemix-style) wallet path is.

## Reproduction (deterministic)

1. Create a wallet backed by an anonymous owner identity; wrap its identity in a boolpolicy composite and receive tokens on it — ownership rows are written correctly (`EscrowAuth` answers, since the composite resolves to no wallet yet).
2. Transfer from that wallet with a rest output back to the policy identity — the transfer succeeds, and `RegisterRecipient` binds the composite identity to the wallet.
3. Receive any further token on the same policy identity (rest or plain transfer) — the token row is stored with `owner_wallet_id = <base wallet id>` and **no ownership row**.
4. Attempt another transfer — the selector finds no spendable tokens and fails with "insufficient funds", while the balance query still reports the full amount.

Step 3 is not racy: after the bind, every incoming token for that identity takes the wallet-based branch. Before the bind, the same tokens (byte-identical owner) are attributed correctly.

## Root cause

Three pieces interact:

- `WalletBasedAuthorization.IsMine` returns `(wallet.ID(), nil, true)` — no ownership ids, and the plain wallet id rather than the `policy`-prefixed one `boolpolicy.EscrowAuth` produces.
- `TokenTransaction.StoreToken` writes ownership rows only from the `owners` slice; an empty slice is accepted silently.
- `AuthorizationMultiplexer.IsMine` returns the first match, and the generic authorization is registered first.

## Suggested fix

Register the type-gated authorizations before the wallet-based one in both drivers. They reject foreign owner types up front, so the reorder is inert for plain tokens; escrow-typed tokens are always attributed by the authorization that understands their owner type. I have a patch with a regression test ready and will open a PR.

EvanYan1024:
## Motivation

`AuthorizationMultiplexer.IsMine` is first-match-wins, and both drivers registered the generic `WalletBasedAuthorization` before the type-gated authorizations. Once a boolpolicy composite identity gets bound to a wallet — `AnonymousOwnerWallet.RegisterRecipient` does this when the rest of a transfer goes back to the wallet's own policy identity — the wallet-based authorization starts resolving that identity and claims every subsequent token first, returning the base wallet id and **no ownership ids**. `StoreToken` writes ownership rows only from the ids slice, so those tokens land in the store without any ownership row, under a wallet id the policy wallet's selector never queries (`<id>` instead of `policy<id>`).

The observable failure in our test environments: an anonymous-identity policy wallet works normally until its first outgoing transfer; from then on every incoming token (rest or deposit) is invisible to the selector, so the balance reports funds while any transfer fails with "insufficient funds". The failure is deterministic — before the bind the very same owner bytes are attributed correctly through `boolpolicy.EscrowAuth`, after the bind they never are. Wallets backed by long-term identities are unaffected because their `RegisterRecipient` accepts the policy rest recipient without binding (#2268).

## Change

Extract the drivers' authorization chain into a single shared constructor, `common.NewStandardAuthorization`, which registers the type-gated authorizations (`htlc.ScriptAuth`, `multisig.EscrowAuth`, `boolpolicy.EscrowAuth`) before `common.NewTMSAuthorization`, and use it from both the fabtoken and zkatdlog drivers, so the ordering has exactly one definition. Each type-gated authorization rejects foreign owner types up front, so the reorder is inert for plain tokens; escrow-typed tokens are always attributed by the authorization that understands their owner type, whether or not the composite identity has been bound to a wallet.

The identity guide's "add a new identity type" registration example taught the generic-first order; it now shows the shared constructor and spells out that the wallet-based catch-all must stay last.

## Testing

- New regression test (`token/services/ttx/boolpolicy/auth_test.go`) exercises the same constructor the drivers use (`common.NewStandardAuthorization`) and replays the failure mode: a policy token whose composite identity resolves to a wallet must still be attributed to the member's policy wallet (`policy<id>` ownership ids), and the test also pins the wrong answer the wallet-based authorization alone would give, which the specific-first order exists to preempt. Reintroducing the generic-first order in the shared constructor turns the test red.
- `gofmt`, `go build ./token/...`, `go vet`, `go test -race` over the touched packages, and golangci-lint (repo config) all pass locally.

Fixes #2328

### DIFF
--- a/docs/services/identity.md
+++ b/docs/services/identity.md
@@ -640,17 +640,19 @@ func (a *EscrowAuth) Issued(_ context.Context, _ driver.Identity, _ *token.Token
 func (a *EscrowAuth) OwnerType(raw []byte) (driver.IdentityType, []byte, error)        { ... }
 ```
 
-Register it in **both** driver files inside `NewAuthorizationMultiplexer`:
+Register it in `common.NewStandardAuthorization` (`token/core/common/authorization.go`), the chain both drivers use. Order matters: the multiplexer is first-match-wins and `NewTMSAuthorization` resolves *any* identity bound to a wallet — including composite identities — so it would claim the token under the wrong wallet id and drop its ownership entries. Every type-gated authorization must therefore be registered **before** the trailing `NewTMSAuthorization` catch-all (regression: `TestAuthorizationOrder_BoundPolicyIdentity` in `token/services/ttx/boolpolicy/auth_test.go`).
 
 ```go
-// token/core/fabtoken/v1/driver/driver.go  (and the zkatdlog equivalent)
-authorization := common.NewAuthorizationMultiplexer(
-    common.NewTMSAuthorization(...),
-    htlc.NewScriptAuth(ws),
-    multisig.NewEscrowAuth(ws),
-    boolpolicy.NewEscrowAuth(ws),
-    mynew.NewEscrowAuth(ws),   // ← add here
-)
+// token/core/common/authorization.go
+func NewStandardAuthorization(logger logging.Logger, publicParameters driver.PublicParameters, walletService driver.WalletService) *AuthorizationMultiplexer {
+	return NewAuthorizationMultiplexer(
+		htlc.NewScriptAuth(walletService),
+		multisig.NewEscrowAuth(walletService),
+		boolpolicy.NewEscrowAuth(walletService),
+		mynew.NewEscrowAuth(walletService), // ← add here, before the wallet-based catch-all
+		NewTMSAuthorization(logger, publicParameters, walletService),
+	)
+}
 ```
 
 #### Step 6 — Add a wallet wrapper
--- a/token/core/common/authorization.go
+++ b/token/core/common/authorization.go
@@ -12,7 +12,10 @@ import (
 	"github.com/LFDT-Panurus/panurus/token"
 	"github.com/LFDT-Panurus/panurus/token/driver"
 	"github.com/LFDT-Panurus/panurus/token/services/identity"
+	"github.com/LFDT-Panurus/panurus/token/services/interop/htlc"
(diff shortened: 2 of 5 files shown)


Effi-S (approved):
LGTM

AkramBitar (approved):
LGTM

PR #2334, opened 2026-08-29
lint(ireturn): replace the T allow entry with real allows

### LINKED ISSUE #2155: lint: the 'T' entry in ireturn.allow disables the linter for all first-party code
## Problem

`.golangci.yml` allows `T` in `linters.settings.ireturn.allow`. ireturn treats allow entries as
[unanchored regexes](https://github.com/butuzov/ireturn/blob/master/analyzer/internal/config/config.go),
so `T` matches any interface path containing a capital `T`. Every first-party interface lives under
`github.com/LFDT-Panurus/...`, which contains one, so all of them are allowed.

The effect is that `ireturn` reports nothing for our own code. It only ever fires on third-party
interfaces, which is the opposite of what it is for.

## Scale

Removing the entry surfaces **330 issues across 123 files** in the root module.

Worth noting when measuring: golangci-lint caps output at 50 per linter by default, so a plain run
looks much smaller than it is.

```
golangci-lint run --enable-only=ireturn                           ->  50  (capped)
golangci-lint run --enable-only=ireturn --max-issues-per-linter=0 -> 330
```

## Notes

- `generic` is already in the allow list and covers generic type parameters, which is presumably
  what `T` was meant for, so the entry looks redundant rather than load-bearing.
- It cannot simply be deleted on its own: `make lint` then fails on all 330, so the removal needs to
  land together with whatever mix of fixes and explicit allows those 330 require.
- Context in #2114, where this was found while enabling `ireturn`. Agreed there to keep the entry for
  now and handle it separately rather than expanding that PR's scope.

Part of #1991.

atharrva01:
Fixes #2155

ireturn treats allow entries as unanchored regexes, so `T` matched any interface path containing a capital T, which is every first-party interface under `github.com/LFDT-Panurus/panurus/...`. That made ireturn fire only on third-party code and never our own.

Removing it surfaces 293 issues across 117 files in the root module, plus 4 more in the separate `integration` module (own go.mod, so its own package paths). All of them turned out to be the same shape: driver-technology packages (`token/driver` and its per-service siblings under identity/network/certifier/storage), the SQL query builder DSL under `token/services/storage/db/sql/query/...`, and a handful of dependency-inversion or swappable-strategy interfaces (`ttx/dep`, `cleanup.Leadership`, `x509/crypto.KeyStore`, the integration module's per-technology test platform abstractions, etc). Every one of these is designed to return an interface on purpose, so this is a config-only change: a real allow list in place of the blanket `T`, grouped by rationale with a comment on each group.

No source files touched.

## Test plan
- [x] `golangci-lint run --enable-only=ireturn --max-issues-per-linter=0 ./...` → 0 issues, in every module (root, `integration`, `token/services/storage/db/kvs/hashicorp`, all of `cmd/*`)
- [x] Full `golangci-lint run ./...` clean on every module
- [x] `go build ./...` clean

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -100,7 +100,6 @@ linters:
         - stdlib
         - generic
         - (or|er)$
-        - T
         - github.com/prometheus/client_golang/prometheus.Counter
         - github.com/prometheus/client_golang/prometheus.Gauge
         - github.com/prometheus/client_golang/prometheus.Histogram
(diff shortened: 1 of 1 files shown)


.golangci.yml
```diff
@@ -123,6 +122,53 @@ linters:
...
+        # (each method returns an unexported interface exposing only the next legal calls), so
+        # this matches both exported and unexported types in the tree.
+        - /db/sql/query/.*\.[A-Za-z]
+        - github.com/LFDT-Panurus/panurus/token/token.Quantity
+        - github.com/LFDT-Panurus/panurus/token/core/common/metrics.Gauge
+        - github.com/LFDT-Panurus/panurus/token/core/common/metrics.Histogram
+        # services/ttx/dep (and its auditor sub-package) exists purely to give ttx a dependency
+        # inversion seam onto storage/network/tms internals, mirroring the third-party interfaces
+        # allowed above.
+        - github.com/LFDT-Panurus/panurus/token/services/ttx/dep
```
Effi-S: `ttx/dep` also matches `ttx/deps` and `ttx/deps/mock`

Effi-S (changes requested):

.golangci.yml
```diff
@@ -123,6 +122,55 @@ linters:
         - github.com/hyperledger-labs/fabric-smart-client/integration/nwo/api.ViewClient
         - github.com/hyperledger-labs/fabric-smart-client/integration/nwo/api.Platform
         - github.com/hyperledger-labs/fabric-smart-client/integration/nwo/api.Context
+        # Our own driver-pattern packages: every package literally named "driver" (token/driver
+        # itself plus its per-service siblings under identity, network, certifier, and storage/db)
+        # exists to define the swappable interface a token technology or storage backend
+        # implements. Naming a concrete type instead of the interface here would defeat the point
+        # of the pattern, and the interface set grows with every new driver method, so this is a
+        # package match rather than an enumerated type list.
+        - /driver\.[A-Z]
```
Effi-S: minor:  /driver\.[A-Z] is unanchored: allows third-party driver packages too

.golangci.yml
```diff
@@ -123,6 +122,55 @@ linters:
...
+        # The SQL query builder DSL (token/services/storage/db/sql/query/...): conditions, table
+        # references and clause fragments compose polymorphically (And/Or combine heterogeneous
+        # Condition implementations, for example). Several of the fluent builders are staged
+        # (each method returns an unexported interface exposing only the next legal calls), so
+        # this matches both exported and unexported types in the tree.
+        - /db/sql/query/.*\.[A-Za-z]
```
Effi-S: Same here

.golangci.yml
Effi-S, line 110: Nit: Redundant now `github.com/hyperledger-labs/fabric-smart-client/platform/common/driver.Pagination `

Effi-S, line 121: Nit: redundant now
(`- github.com/hyperledger-labs/fabric-smart-client/platform/common/driver.ConfigService`)

Effi-S (approved):
Looks good! 👍

.golangci.yml
atharrva01: Good catch, fixed in 95ca0110a: anchored to our own module (^github.com/LFDT-Panurus/panurus/...). That reinstates fsc's driver.Pagination/driver.ConfigService allows as non-redundant (removing them to test breaks the build), so those two stayed.

.golangci.yml
atharrva01: Fixed the same way in 95ca0110a: anchored to our own module.

.golangci.yml
atharrva01: This one was only redundant because /driver\.[A-Z] was unanchored and happened to also match fsc's driver package. Now that it's anchored to our own module (95ca0110a), driver.Pagination is back to being the only thing covering fsc's Pagination interface, so I kept it: tested by commenting it out locally, which brought back 11 ireturn hits in token/services/storage/db/sql/query/pagination/.

atharrva01: Same as the Pagination one: kept, now non-redundant after the anchor fix in 95ca0110a.

.golangci.yml
Effi-S: Nice! 💯

PR #2340, opened 2026-08-31
fix(ttx): fix localSession data race and unblock peers on close

### LINKED ISSUE #2285: localSession: data race on info.Closed and peers left blocked on Receive after Close
### Problem

`localSession` in `token/services/ttx/session.go` backs the `LocalBidirectionalChannel` used for
self-auditing, where an issuer that is also an auditor talks to itself. `AuditingViewInitiator.startLocal`
keeps the left session and runs the auditor responder on the right session **in a separate goroutine**,
so the two sessions are used concurrently by design.

Two problems follow from that:

**1. `info.Closed` is read and written without synchronisation.**

`Close` writes `s.info.Closed`, while `Info`, `Receive` and `send` all read it, from different
goroutines. `CollectEndorsementsView.cleanupSessions` closes the initiator side on every return
path of `Call`, which can happen while the responder goroutine is still touching the session. This
is a data race: unsynchronised, and reported as such under `-race`.

**2. A peer blocked on `Receive` is never released.**

`Close` only flips the flag; it never closes the underlying channel. When one side goes away — for
example the initiator returns early with an error and `cleanupSessions` runs — the peer sitting in
`<-session.Receive()` has nothing to wake it, so it waits out its full receive timeout
(`DefaultReceiveTimeout`) before it can fail. In the self-auditing flow that is a responder
goroutine idling for the whole timeout after the conversation is already over.

The naive repair makes things worse rather than better. Adding `close(readChannel)` and
`close(writeChannel)` to `Close` introduces two panics: `readChannel` is the channel the **peer**
sends on, so closing it makes the peer panic with `send on closed channel`, and since the two
sessions share the same pair of channels, closing both from both sides panics with
`close of closed channel`.

### Impact

- A data race in a flow that runs whenever an issuer is also an auditor.
- Responder goroutines held for the length of a receive timeout after their counterpart is gone.
- No test coverage for concurrent use or for closing either side of the channel.

SurbhiAgarwal1:
Fixes #2285

### Problem
1. `info.Closed` in `localSession` was read and written without synchronization, causing a data race under `-race` during self-auditing flows (where an issuer is also an auditor).
2. When a session is closed (e.g. via `cleanupSessions`), the peer sitting in `<-session.Receive()` was never unblocked and waited out the full `DefaultReceiveTimeout` (10 seconds).

### Solution
- Added `sync.RWMutex`, `closedChan`, `inFlight` tracking, and `writeClosed` boolean to `localSession`.
- `Info()`, `Receive()`, `Send()`, and `Close()` are now synchronized and thread-safe.
- `Close()` safely closes `writeChannel` when `inFlight == 0`, instantly releasing any peer waiting on `<-peer.Receive()`.
- Added unit tests in `session_test.go` covering peer unblocking on close, concurrent usage under `-race`, and double session closure.


plan.md
AkramBitar, line 1: Please remove this file

AkramBitar (changes requested):

plan.md
SurbhiAgarwal1: Done! Removed `plan.md` in the latest push (`d3f467d28`).

token/services/ttx/session.go
```diff
@@ -93,32 +105,71 @@ func (s *localSession) SendError(ctx context.Context, payload []byte) error {
...
-		FromPKID:     s.info.RemotePKID,
+		FromEndpoint: info.RemoteEndpoint,
+		FromPKID:     info.RemotePKID,
 		Status:       status,
 		Payload:      payload,
 		Ctx:          ctx,
 	}
 
-	return nil
+	select {
```
AkramBitar, line 161: **Medium: `Send`/`SendError` is now nondeterministic when the caller's context is already cancelled.**

`writeChannel` is buffered (cap 10, set in `NewLocalBidirectionalChannel`), so with a cancelled ctx *both* `case s.writeChannel <- msg` and `case <-ctx.Done()` are ready, and Go picks pseudo-randomly. Measured on this head: 88 successes / 112 failures over 200 iterations with an already-cancelled context. Before this PR the buffered send always succeeded.

Concrete path: the self-audit responder runs via `view3.RunView(logger, context, responderView, view.AsResponder(right))` on the *initiator's* view context (`token/services/ttx/auditor.go:279`). If the initiator returns first — its 1-minute signature timeout at `auditor.go:189`, or `RunViewWithTimeout`'s `defer cancel()` — then the responder's `signAndSendBack` -> `SendTyped(context.Context(), ...)` delivers the signature about 44% of the time and returns `context cancelled while sending message` the rest, with no way for the caller to distinguish the two.

Suggested fix: test `ctx.Err()` before the `select`, or give the send case priority with a nested `select { case ch <- msg: default: }` before falling through to the three-way select.

token/services/ttx/session.go
```diff
@@ -93,32 +105,71 @@ func (s *localSession) SendError(ctx context.Context, payload []byte) error {
...
+	case s.writeChannel <- msg:
+		return nil
+	case <-s.closedChan:
```
AkramBitar, line 164: **Medium: the unblock-the-peer-on-close fix is one-directional — a peer blocked in `send` is never released.**

`send` selects only on `s.closedChan`, which is its *own* close signal. When `left.Close()` runs it closes `lr` (right's read channel), but nothing touches `rl`, so a `right.Send` blocked on a full `rl` keeps waiting.

Verified: after filling the 10-slot buffer and blocking an 11th send, calling `right.Close()` on the peer left the send blocked past a 500 ms deadline. It only returns when the sender's own ctx is cancelled or its own session is closed — i.e. the same stall class #2285 is about, just on the send side.

Suggested fix: share a close signal between the two `localSession`s, or also select on the read channel of the writing side.

token/services/ttx/session.go
```diff
@@ -93,32 +105,71 @@ func (s *localSession) SendError(ctx context.Context, payload []byte) error {
...
+		return errors.Wrap(ctx.Err(), "context cancelled while sending message")
+	}
 }
 
 func (s *localSession) Receive() <-chan *view.Message {
-	if s.info.Closed {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
+	if s.closed {
```
AkramBitar, line 177: **Medium: `Receive()` still returns a nil channel after the local `Close()`, so the closing side still burns the full 10 s `DefaultReceiveTimeout`.**

A nil channel blocks forever, so `ReceiveRawWithTimeout` (`token/services/utils/session/session.go:83`) falls through to `case <-timeout.C` and returns `ErrTimeout` only after `DefaultReceiveTimeout`. This PR now has a `closedChan` that would make it instant, but doesn't use it here — returning `s.readChannel` (which the peer closes) or a pre-closed channel would fail fast with `ErrNilMessage`.

This is pre-existing behaviour, so not a regression, but it is exactly the 10-second stall the PR is meant to remove, and as written the two sides of the same fake session end up with asymmetric close semantics.

token/services/ttx/session.go
```diff
@@ -93,32 +105,71 @@ func (s *localSession) SendError(ctx context.Context, payload []byte) error {
...
 		Ctx:          ctx,
 	}
 
-	return nil
+	select {
+	case s.writeChannel <- msg:
+		return nil
+	case <-s.closedChan:
+		return errors.New("session is closed")
+	case <-ctx.Done():
```
AkramBitar, line 168: **Low: `Send(nil, payload)` now panics with a nil pointer dereference** on `ctx.Done()`. Before this PR a nil context was only stored in `msg.Ctx` and was harmless.

No in-repo caller passes nil today, but `NewLocalBidirectionalChannel` / `LeftSession` / `RightSession` are exported, so this is a silent contract change. An `if ctx == nil { ctx = context.Background() }` guard — or dropping the ctx case per the fix suggested above — avoids it.

token/services/ttx/session.go
```diff
@@ -93,32 +105,71 @@ func (s *localSession) SendError(ctx context.Context, payload []byte) error {
...
+	s.mu.Lock()
+	if s.closed {
+		s.mu.Unlock()
+
+		return
+	}
+	s.closed = true
 	s.info.Closed = true
+	close(s.closedChan)
+	if s.inFlight == 0 && !s.writeClosed {
```
AkramBitar: **Low: the `!s.writeClosed` term is dead here.** `writeClosed` is only ever set inside `send`'s deferred block, which requires `s.closed == true`; `Close()` has already returned early above in that case, so `writeClosed` is always false at this point.

Harmless, but it obscures the invariant — worth either dropping the term or adding a short comment stating why it cannot be true.

AkramBitar (commented):
**Review summary:** the core fix is correct — I could not find a send-on-closed-channel or double-`close` path (the `closed` check and `inFlight++` share one critical section with `Close()`, and `writeClosed` guards the close), buffered messages survive the sender's `Close()`, and `go test ./token/services/ttx/... ./token/services/utils/session/... -race` is green at this head.

Two things are worth fixing before merge; the other three are nits that can be follow-ups.

1. `send` became nondeterministic when the caller's context is already cancelled (inline comment on the `select`). This one has a real caller path in the self-audit responder.
2. The close only unblocks the peer's *receive*; a peer blocked in `send` on a full buffer still stalls — the same class of hang #2285 is about, in the other direction.

Also, two test notes: `TestLocalBidirectionalChannel_ConcurrentCloseAndOperations` asserts nothing, and its receive goroutine's `select { case <-ch: default: }` almost always takes `default`, so it exercises the race detector but barely the receive path. The `time.Sleep(10 * time.Millisecond)` in `TestLocalBidirectionalChannel_CloseReleasesPeerReceive` is unnecessary — the peer gets the closed channel regardless of ordering.

AkramBitar:
@SurbhiAgarwal1 

Could you please have a look at the failures? 

Regards,
Akram

AkramBitar:
@SurbhiAgarwal1 

Could you please fix the lint issues and rebase?

Regards,
Akram

PR #2354, opened 2026-09-09
Bumped golang.org/x/crypto to v0.57.0

Effi-S:
## Vulnerability description:

Previously, a channel registered in the mux's chanList is not usable until it is established. A malicious peer was able flood the channel's incomingRequests, deadlocking the entire connection.

Now, we add an atomic established state, set when a channel becomes usable. Until such a time, handlePacket drops every packet other than the open confirmation/failure, without blocking and without tearing down the connection.

See https://osv.dev/vulnerability/GO-2026-6355


Effi-S (commented):
@adecaro, can you review?

adecaro (approved):
LGTM

Issue #1711, opened 2026-05-17
perf(bulletproof): bulletproof IPA prover doesn't use optimized verifier-style MSM


neetance:
## Description

The Bulletproofs Inner Product Argument (IPA) verifier already uses the optimization to defer generator folds and evaluate them using a single batched MSM. However, the `IPAProver` still performs sequential round-by-round generator reductions:

```go
leftGen, rightGen = reduceGenerators(leftGen, rightGen, x, xInv, p.Provider)
```

This repeatedly allocates intermediate generator slices, performs unnecessary scalar multiplications across multiple rounds, and prevents efficient use of optimized Pippenger MSM implementations.

## Proposed Solution
Refactor the prover to:
1. Keep the original generators unchanged.
2. Reuse the challenge history to compute scalar expansions (`ComputeSVector`).
3. Compute `L_k` and `R_k` directly using batched MSMs over the original generators.
4. Remove the `reduceGenerators` path entirely.

This preserves the exact same mathematics while eliminating repeated vector reductions and reducing allocation pressure.

neetance:
hi @adecaro, can you please assign me this issue?
I can start working on the pr right away once assigned.
thank you 🙏

adecaro:
Hi @neetance , go ahead. Thanks much for this effort 🙏

PR #2357, opened 2026-09-09
Upgrade FSC to v0.20.0 

### LINKED ISSUE #2351: Upgrade FSC to v0.21.0
Bump to newest FSC version 
[Release v0.21.0 · hyperledger-labs/fabric-smart-client](https://github.com/hyperledger-labs/fabric-smart-client/releases/tag/v0.21.0)
_(Ref to last bump [Issue-2173](https://github.com/LFDT-Panurus/panurus/issues/2173) [PR-2174](https://github.com/LFDT-Panurus/panurus/pull/2174))_

sid200727:
Closes #2351

Second of two PRs for the FSC bump. Stacks on #2356, which takes us from v0.18.0 to v0.19.0, merge that one first, or this diff will show both bumps.

## The runtime panic

[FSC #1700](https://github.com/hyperledger-labs/fabric-smart-client/pull/1700) removed `HasNext()` from `iterators.Empty`. It was never part of the `Iterator` interface that guarantees only `Next()` and `Close()`,but the `slice` and `permutation` iterators still have it, which is presumably how the pattern crept in here:

```go
if err == nil && it.(interface{ HasNext() bool }).HasNext() {
```

`cachedFetcher` returns a `permutation` on a cache hit and an empty iterator on a miss. After the upgrade, every cache miss panics. This is production code in `mixedFetcher.UnspentTokensIteratorBy`, and it compiles cleanly, only the test run catches it.

### The fix

`cachedFetcher` already knows the outcome: the `ok` from `cache.Get`. It now reports that alongside the iterator, through an unexported `unspentTokensIteratorBy`. `UnspentTokensIteratorBy` stays a thin wrapper over it, so the `TokenFetcher` interface and its generated mocks are untouched, and `mixedFetcher` no longer inspects an iterator it has no portable way to inspect.

I kept it to `cachedFetcher` deliberately, widening the interface method would have meant regenerating `mocks/token_fetcher.go` and updating the stubs in `manager_test.go`, `ratelimit_test.go`, `selector_test.go` and `selector_iterator_test.go`, which felt out of proportion for this.

## Other breaking changes in this range

[FSC #1711](https://github.com/hyperledger-labs/fabric-smart-client/pull/1711) (squirrel removal) also lands here. As confirmed during that PR's review, it needs no source changes on our side, `go mod tidy` just drops the stale indirect requirements.

## Testing

-> All eight modules build, `go vet` and `golangci-lint` (v2.12.2, the pinned version) clean
-> Verified the panic's provenance: `sherdlock` panics on v0.20.0 without this fix, and does not on v0.19.0, so the break arrives with this bump
-> The remaining `sherdlock` failures and the `identity/config` `TestTranslatePath` failure are pre-existing on `main` and unrelated

## Possibly worth a follow-up

The same assertion pattern may exist elsewhere in Panurus, or in other FSC consumers. I only checked `sherdlock` because that's what failed. 
Happy to open a separate issue if that seems worth sweeping for.


cmd/tokengen/go.mod
```diff
@@ -14,7 +14,7 @@ require (
 	github.com/IBM/mathlib v0.3.0
 	github.com/LFDT-Panurus/panurus v0.10.1
 	github.com/LFDT-Panurus/panurus/integration v0.0.0
-	github.com/hyperledger-labs/fabric-smart-client v0.18.0
+	github.com/hyperledger-labs/fabric-smart-client v0.20.0
 	github.com/hyperledger-labs/fabric-smart-client/integration v0.18.0
```
Effi-S: @sid200727 
I want to clarify, shouldn't `github.com/hyperledger-labs/fabric-smart-client/integration v0.18.0` Be upgraded as well?

Effi-S (changes requested):

token/services/selector/sherdlock/fetcher.go
```diff
@@ -105,9 +105,9 @@ func NewMixedFetcher(tokenDB TokenDB, m *Metrics, cacheSize int64, freshnessInte
 // UnspentTokensIteratorBy returns an iterator for unspent tokens, trying cached results first, falling back to database query.
 func (f *mixedFetcher) UnspentTokensIteratorBy(ctx context.Context, walletID string, currency token2.Type) (Iterator[*token2.UnspentTokenInWallet], error) {
 	logger.DebugfContext(ctx, "call unspent tokens iterator")
-	it, err := f.eagerFetcher.UnspentTokensIteratorBy(ctx, walletID, currency)
+	it, cached, err := f.eagerFetcher.unspentTokensIteratorBy(ctx, walletID, currency)
 	logger.DebugfContext(ctx, "fetched eager iterator")
-	if err == nil && it.(interface{ HasNext() bool }).HasNext() {
+	if err == nil && cached {
```
Effi-S, line 110: Are we sure cache can't store nil or an empty slice (or some other false)? 
`HasNext()` was self-correcting 
it seems to me `cached` is not, 
I think this warrants either a more sophisticated logical assertion

token/services/selector/sherdlock/fetcher_test.go
```diff
@@ -176,11 +176,11 @@ func TestCachedFetcher_UnspentTokensIteratorBy_CacheHit(t *testing.T) {
 	fetcher.update(ctx)
 
 	// Query cache
-	it, err := fetcher.UnspentTokensIteratorBy(ctx, "wallet1", "USD")
+	it, cached, err := fetcher.unspentTokensIteratorBy(ctx, "wallet1", "USD")
```
Effi-S, line 179: previously `it.(...).HasNext()` confirmed a cache hit actually produced tokens. 
Now a regression where a cache hit returns a broken or empty permutation while cached is still true would pass undetected. 

This is probably solved by addressing my comment on `fetcher.go`

Effi-S (changes requested):

token/services/selector/sherdlock/fetcher.go
```diff
@@ -351,11 +362,11 @@ func (f *cachedFetcher) UnspentTokensIteratorBy(ctx context.Context, walletID st
...
-		return it.NewPermutation(), nil
+		return it.NewPermutation(), true, nil
 	}
 	logger.DebugfContext(ctx, "No tokens found in cache for [%s]. Returning empty iterator.", tokenKey(walletID, currency))
 
-	return collections.NewEmptyIterator[*token2.UnspentTokenInWallet](), nil
+	return collections.NewEmptyIterator[*token2.UnspentTokenInWallet](), false, nil
 }
 
 // isCacheOverused checks if the cache has been queried too many times since the last refresh.
```
adecaro, line 383: **Design note: cache-hit invariant (`cached=true` implies non-empty) is enforced by construction elsewhere, not at this boundary.**

`cached` here is just whether `f.cache.Get` found the key — it doesn't check the iterator actually has tokens. Today that's safe because `groupTokensByKey` (fetcher.go:294-308) only appends to `m[key]` when a token is found, and `updateCache` (fetcher.go:312-334) deletes any stale key not present in the newest batch — so no key can ever map to an empty slice. But nothing at this boundary enforces that; it's implicit in two other functions.

This is the same concern @Effi-S raised on this line in the prior review round — still open at this commit, restating with the concrete trace of where the invariant actually lives.

**Suggestion**: add a short comment here documenting the invariant, or assert on it directly (e.g. treat a cache hit with an empty slice as a bug rather than trusting it can't happen).

token/services/selector/sherdlock/fetcher_test.go
```diff
@@ -176,11 +176,11 @@ func TestCachedFetcher_UnspentTokensIteratorBy_CacheHit(t *testing.T) {
...
 
 	require.NoError(t, err)
 	assert.NotNil(t, it)
```
adecaro, line 182: **Should fix: this now only asserts the `cached` flag, not that the iterator actually yields a token.**

The old assertion (`it.(interface{ HasNext() bool }).HasNext()`) confirmed the cache hit actually produced tokens. The new one (`assert.True(t, cached, ...)`) would still pass if a future change let `cached=true` coincide with an empty iterator — exactly @Effi-S's concern from the prior review round on this test.

**Suggestion**: additionally assert on the iterator's content, e.g. `tok, err := it.Next(); require.NoError(t, err); assert.NotNil(t, tok)`, so a regression in the invariant discussed on fetcher.go:362-372 would actually fail this test.

cmd/tokengen/go.mod
adecaro: **Should fix: `fabric-smart-client/integration` still pinned at v0.18.0 despite the root module bumping to v0.20.0 here.**

`integration/v0.20.0` genuinely exists upstream (confirmed via `git ls-remote --tags` against the FSC repo), so this isn't a case of the submodule not having caught up yet — it looks like an oversight in the bump. Same gap exists in every other go.mod that references this submodule. This is the same point @Effi-S raised on this line in the prior review round, still open at this commit.

**Suggestion**: bump `fabric-smart-client/integration` to v0.20.0 alongside the root module here and in the other 7 go.mod files that reference it, or note explicitly why it's deferred.

adecaro (commented):
## Review: does this fully address #2351?

@sid200727

**Approve, with non-blocking comments — the panic fix is correct and verified, two points from @Effi-S's earlier review rounds remain open at this commit.**

Checked out `fix-2351-bump-fsc-v0.20.0` at `141a87a` and verified:

```
go build ./... (each of the 9 modules individually)         -> clean, all 9
go vet ./token/services/selector/sherdlock/...               -> clean
go test -count=1 ./token/services/selector/sherdlock/...      -> pass
go test -race -count=1 ./token/services/selector/sherdlock/...-> pass, no races
golangci-lint run ./token/services/selector/sherdlock/...     -> 0 issues
revert fetcher.go fix + re-run tests                          -> panics exactly as described
```

I also read the vendored FSC v0.20.0 source directly to confirm `iterators.Empty`'s `empty[K]` struct really did drop `HasNext()` after FSC PR #1700, and grepped the repo for other callers of the same fragile `it.(interface{ HasNext() bool })` pattern — found none outside this file.

Reverting `fetcher.go`'s `mixedFetcher.UnspentTokensIteratorBy` back to the pre-fix type assertion and re-running the suite panics with `panic: interface conversion: *iterators.empty[...] is not interface { HasNext() bool }: missing method HasNext` at `fetcher.go:110`, so the existing test suite pins the bug rather than the diff inventing new coverage.

CI has not run yet on this commit — all checks except `DCO` show `pending` (fork PR awaiting workflow authorization) — so none of the above is corroborated by CI; all verification here is from local runs.

### What it fixes

| # | Finding | Verified |
|---|---|---|
| 2351.1 | FSC v0.20.0 removed `HasNext()` from the empty iterator, causing a panic on cache miss | ✅ confirmed against vendored FSC source, reproduced the panic by reverting the fix |
| 2351.2 | Fix avoids the panic without widening the public `TokenFetcher` interface/mocks | ✅ confirmed — `unspentTokensIteratorBy` is unexported, `UnspentTokensIteratorBy` signature unchanged |

Both rows are solid; the underlying fix is small, correct, and test-covered.

### Non-blocking

- **Still open from your earlier review:** `fabric-smart-client/integration` is left at `v0.18.0` in every `go.mod` that references it (e.g. `cmd/tokengen/go.mod:18`), even though the root FSC module moved to `v0.20.0` here. I confirmed `integration/v0.20.0` genuinely exists upstream via `git ls-remote --tags`, so this isn't a "can't go further" case.
- **Still open from your earlier review:** the cache-hit invariant (`cached=true` implies the cached slice is non-empty) is enforced by construction in `groupTokensByKey`/`updateCache`, not at the `unspentTokensIteratorBy` boundary itself, and the updated test (`fetcher_test.go:176-186`) now only asserts the `cached` flag rather than the iterator's actual content — so a future regression that violates the invariant would pass silently.
- **Hygiene:** PR body says "Refs #2351" rather than `Fixes #2351`/`Closes #2351`, so GitHub won't auto-link or close the issue on merge.

### Recommendation

Good to merge as-is; the two open threads from @Effi-S are worth resolving in this PR or a fast follow-up, and the `Fixes #2351` wording is a one-line fix to the PR body whenever convenient.

token/services/selector/sherdlock/fetcher.go
sid200727: You're right, and @adecaro traced it in his review: today an empty slice can't reach the cache because groupTokensByKey only appends when a token is found and updateCache drops stale keys, but nothing at this boundary enforces that. Adding an explicit guard so a cache hit with an empty iterator is treated as a bug rather than silently skipping the lazy fallback.

token/services/selector/sherdlock/fetcher_test.go
sid200727: Agreed, asserting cached alone would let a regression in that invariant pass. Adding an assertion on the iterator's contents too.

cmd/tokengen/go.mod
sid200727: Yes it should, sorry, my grep matched fabric-smart-client v0.18.0 and missed fabric-smart-client/integration v0.18.0, which is a different module path. Bumping it across every go.mod that references it.

PR #2361, opened 2026-09-12
docs(svm): propose a Solana network driver design

atharrva01:
Proposal for a Solana (SVM) network driver, a third non-Fabric backend alongside the in-progress EVM driver.

It reuses the EVM driver's trust model (off-chain Go validation via `driver.Validator`, on-chain signature-threshold checking only) rather than designing from scratch, and calls out the two places it deliberately doesn't just port that design: one PDA per token instead of one contract-wide mapping so Sealevel can actually parallelize unrelated transfers, and Solana's native commitment levels replacing the EVM driver's whole confirmation-depth/reorg-detection subsystem. It also treats the EVM driver's most expensive lesson, a multi-TMS state-sharing bug that took a dedicated bug-hunt round to find, as a day-one design requirement instead of something to discover the hard way twice.

No code, no branch work started, just the design doc for discussion, same stage the EVM driver's own design doc was at before implementation began.

### DIFF
--- a/x/token/services/network/svm/svm_network_driver_design.md
+++ b/x/token/services/network/svm/svm_network_driver_design.md
@@ -1,0 +1,469 @@
+# SVM Network Driver: Design Proposal for Panurus
+
+**Status:** Proposal, not started. No branch, no code.
+**Proposed location once work begins:** `x/token/services/network/svm/`, its own Go module, mirroring the (unmerged) `x/token/services/network/evm/` precedent.
+**Companion doc to write first:** `docs/services/network-solana.md`, a short conceptual guide in the style of the existing `docs/services/network-ethereum.md`, before a full design doc like this one gets treated as final.
+
+---
+
+## 1. Executive Summary
+
+This document proposes a Solana (SVM) network driver for Panurus: a third non-Fabric backend alongside the in-progress Ethereum/EVM driver, letting a Panurus token management service issue, transfer, and audit tokens against a Solana cluster instead of a Fabric channel.
+
+It deliberately does not start from a blank page. Panurus already has a working answer to "how does a Fabric-shaped token SDK talk to a chain with no chaincode, no endorsement policy, and no MSP": the EVM driver, built over roughly two months across 20+ PRs on `feature/evm-network-driver`, with its own 1000-line design doc at `x/token/services/network/evm/eth_network_driver_design.md`. Everything below either reuses that architecture directly or explains precisely why Solana's account model and consensus give a reason to diverge.
+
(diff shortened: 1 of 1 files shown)


AkramBitar:
## Review

This is a great idea and a well-structured design. Reusing the off-chain-validation + on-chain-signature-threshold trust model instead of starting from scratch is exactly the right call, the per-token PDA reasoning is sound (Sealevel parallelism is real and a shared-mapping design would throw it away), the finality simplification relative to the EVM driver is correctly identified, and carrying the multi-TMS isolation lesson forward as a day-one requirement rather than a later fix is the most important thing this doc gets right.

A few gaps to address before implementation starts:

**1. Durable nonce / endorsement flow interaction unspecified**
§9 identifies the blockhash-expiry problem and recommends durable nonce accounts, but doesn't specify *where* in the `RequestApproval` flow the nonce is advanced. If before collecting endorsements: the nonce is consumed even if endorsement collection fails. If after: the blockhash-expiry race is back. The EVM driver's `NonceManager.WithNonce` held a lock across the whole allocate-and-use step — the Solana equivalent needs the same explicit sequencing here.

**2. `internal_tx_id` replay protection — unbounded PDA growth risk**
§4 says `apply_state_delta` rejects a delta whose `internal_tx_id` has been seen before, and §9 also relies on Solana's blockhash-expiry window. The doc should clarify which is the *primary* mechanism. If `internal_tx_id` is primary (i.e. the driver doesn't rely on Solana's window), storing every seen ID as on-chain state creates unbounded PDA growth — how is that pruned or bounded?

**3. Transaction-size ceiling needs quantification before implementation**
§4 flags the chunked two-phase-commit as "v1.1, don't build speculatively." But a `TokenRequest` with 4 inputs + 4 outputs already needs 8 token PDAs + TMS config + system program + Ed25519 program + N endorser pubkeys in the account list — this can hit the ~1232-byte ceiling with modest-sized requests. This is a hard constraint, not a speculative one. It should be quantified ("a request with up to X inputs and Y outputs fits one transaction") before Phase 5 starts, not deferred.

**4. Multi-listener design asserted but not specified**
§7 says `AddFinalityListener` supports "more than one listener per anchor from the start." The EVM driver's fix for this was non-trivial. Stating the requirement without a concrete design (map to `[]Listener`, fanout bus, etc.) leaves the implementer to rediscover the hard parts independently.

**5. Companion `docs/services/network-solana.md` is its own prerequisite**
The doc itself says this companion doc should be written *before* this design doc is treated as final. That doc doesn't exist yet — worth creating it first, or at least in parallel, before this goes to implementation.

AkramBitar (changes requested):
@atharrva01 

Please see my comments. 

Thanks a lot,
Akram

PR #2364, opened 2026-09-13
tokens: eliminate commit-path cache barriers, struct copying, and unguarded log allocations

### LINKED ISSUE #2188: tokens: commit-path performance bundle — synchronous cache barrier, N+1 statements per transaction, missing preallocation, debug.Stack() per delete, unguarded debug logging, by-value struct copy
### Summary
A bundle of performance findings on the commit-path of `token/services/tokens`, each measured with a `-benchmem` benchmark or an assertion isolating the cost. None change externally observable behavior; all sit on a per-transaction or per-token hot path.

### Where
- **Synchronous cache barrier.** `Ristretto.Add`/`Delete` call `c.cache.Wait()` after every write, turning an async cache write into a synchronous barrier. Used by `CacheRequest`/`removeCachedTokenRequest` on every transaction. (`token/services/utils/cache/ristretto.go:66-74`)
- **N SELECT+UPDATE pairs per spend.** `DBTransaction.DeleteTokens` loops `DeleteToken`, issuing a `GetToken` (SELECT, to source the notification payload) followed by a `Delete` (UPDATE) per input — 2 statements per spent token even though the underlying store already batches deletes elsewhere. (`storage.go:169-177`)
- **One UPDATE per token instead of a batch.** `SetSpendableFlag` loops `Tx.SetSpendable`, one statement per id. (`storage.go:262-272`)
- **Missing preallocation.** `Parse`'s `toSpend`/`toAppend` (`tokens.go:407`, grown via `append` at `:411/:424/:462/:511`) and `deleteTokens`'s `toDelete` (`tokens.go:335`, grown at `:339`) are declared as nil slices and grown by repeated `append`, even though final sizes are knowable upfront from `is.Count()`/`os.Count()` and `len(tokens)` respectively.
- **`debug.Stack()` on every delete.** `Service.DeleteTokens` (`tokens.go:210-212`) captures `string(debug.Stack())` unconditionally on every call, purely to build an audit label, regardless of whether it is ever read.
- **Unguarded debug logging in hot loops.** `deleteTokens` (`tokens.go:315`) formats an entire `[]*token2.UnspentToken` slice with `%v` via an unconditional `logger.DebugfContext` call, and its per-token loop (`tokens.go:338/:341`) does the same per item — none guarded by `logger.IsEnabledFor(zapcore.DebugLevel)`. `Parse` shows the correct guarded pattern for its more expensive log lines (`tokens.go:470`, `:513`) but several of its own per-input/output calls are unguarded (`:410`, `:419`, `:423`, `:432`, `:437`). Go boxes the variadic arguments into `[]any` at the call site unconditionally, before `DebugfContext` is ever entered, so only the guarded form skips that allocation when debug logging is disabled.
- **Struct copied by value.** `AppendToken` takes `TokenToAppend` by value (a struct with 13 fields), copying it once per output per transaction. (`storage.go:180`)

### Impact
Each of these sits on the per-transaction commit path (`AppendValid` → `AppendToken`/`DeleteTokens`) or the per-transaction caching path (`CacheRequest`), so their cost is paid once per token per committed transaction across the whole node.

### Reproduction
```
BenchmarkAdd_WithWait-<n>        ...  allocs/op, ns/op
BenchmarkAdd_WithoutWait-<n>     ...  (lower ns/op, isolating Wait()'s cost)
BenchmarkDelete_WithWait-<n>     ...
BenchmarkDelete_WithoutWait-<n>  ...

TestDeleteTokens_TwoStatementsPerInput   — PASS (asserts GetTokenCallCount == DeleteCallCount == n, one SELECT + one UPDATE per input)
TestSetSpendableFlag_OneStatementPerToken — PASS (asserts SetSpendableCallCount == n)

BenchmarkParse_Allocations-<n>       ...  allocs/op
BenchmarkBufferFlush_NilReset-<n>    ...  (higher allocs/op — regrows from zero capacity every flush)
BenchmarkBufferFlush_SliceReset-<n>  ...  (lower allocs/op — retains backing array capacity)

BenchmarkDebugStack-<n>  ...  ns/op, B/op  (cost of the debug.Stack() primitive itself)

TestDebugfContext_UnguardedAllocatesEvenWhenDisabled — PASS
  unguardedAllocs > guardedAllocs, guardedAllocs < 1.0
  (testing.AllocsPerRun over 1000 runs, debug logging disabled — mirrors the
  existing TestLoggerDebugAllocs technique in token/services/logging)

BenchmarkAppendToken_ByValueCopy/ByValue-<n>    ...  B/op
BenchmarkAppendToken_ByValueCopy/ByPointer-<n>  ...  (lower B/op, isolating the value-copy cost)
```
(Exact numbers depend on hardware; all benchmarks live in `token/services/tokens/tokens_bench_test.go` and `token/services/utils/cache/ristretto_bench_test.go` and can be re-run with `go test <pkg> -bench=. -benchmem -run='^$'`.)

### Severity
Low individually; cumulative cost across a busy node's transaction throughput.

Part of #2112.

SurbhiAgarwal1:
## Summary

Optimizes performance and memory allocations on the transaction commit path in `token/services/tokens` and `token/services/utils/cache`:

- **Cache Barrier Removal (`ristretto.go`)**: Removed synchronous `c.cache.Wait()` calls from write paths (`Add`/`Delete`/`Clear`), unlocking asynchronous Ristretto cache throughput. Exposed explicit `Wait()` helper for testing.
- **Pointer Pass Optimization (`storage.go` & `tokens.go`)**: Changed `AppendToken` parameter to pointer (`*TokenToAppend`) and updated `CacheEntry.ToAppend`, `getActions`, `extractActions`, and `Parse` to pass `[]*TokenToAppend`, eliminating 13-field struct copying (~200+ bytes per output).
- **Event Buffer Capacity Retention (`storage.go`)**: Reset event buffer via `t.pending = t.pending[:0]` in `FlushEvents` and `Rollback` to retain allocated backing array capacity across transactions.
- **Upfront Slice Preallocation (`tokens.go`)**: Preallocated capacities for `toSpend`, `toAppend`, and `toDelete` slices.
- **Stack Trace Overhead Elimination (`tokens.go`)**: Guarded `debug.Stack()` in `Service.DeleteTokens` behind `logger.IsEnabledFor(zapcore.DebugLevel)`.
- **Zero-Allocation Variadic Logging (`tokens.go` & `storage.go`)**: Wrapped hot-path `DebugfContext` log calls with `if logger.IsEnabledFor(zapcore.DebugLevel)` to eliminate Go variadic `[]any` heap boxing allocations when debug logging is disabled.

Fixes #2188


token/services/ttx/session.go
```diff
@@ -93,32 +105,71 @@ func (s *localSession) SendError(ctx context.Context, payload []byte) error {
...
+	defer s.mu.RUnlock()
+
+	if s.closed {
 		return nil
 	}
 
 	return s.readChannel
 }
 
 func (s *localSession) Close() {
```
AkramBitar, line 160: **medium — `Close()` does not release a peer blocked in `send()`.**

`Close()` is asymmetric: it closes the local *write* channel, which releases a peer blocked in `Receive()`, but nothing releases a peer blocked in `send()`. `send` selects only on its own `closedChan` and `ctx.Done()`, so a peer stuck on a full `writeChannel` is not woken by the other side closing.

Confirmed with a probe test: fill the 10-slot buffer, block an 11th `left.Send`, then call `right.Close()` — the sender stays blocked (only `left.Close()` or ctx cancellation frees it). Because the closed side's `Receive()` now returns `nil`, nothing will ever drain that buffer, so the blocked sender's goroutine leaks permanently — the same failure class #2285 set out to fix.

A close signal shared by the session pair (or selecting on the peer's `closedChan`) would cover both directions.

token/services/utils/cache/ristretto.go
```diff
@@ -65,16 +65,19 @@ func (c *ristrettoCache[T]) Get(key string) (T, bool) {
 
 func (c *ristrettoCache[T]) Add(key string, value T) {
```
AkramBitar, line 66: **low/medium — dropping `Wait()` removes read-your-write for *new* keys.**

ristretto's `SetWithTTL` only applies immediately via `storedItems.Update` when the key already exists (v2.4.2 `cache.go:350`); a brand-new key lives only in `setBuf` until the background `processItems` goroutine picks it up.

`sherdlock`'s fetcher got a compensating `Wait()` in `updateCache`, but the other production consumer — `tokens.Service.RequestsCache` (`manager.go:63`) — did not, and the `tokens.Cache` interface has no `Wait`. So `CacheRequest` (`tokens.go:190`) followed shortly by `getActions`/`GetCachedTokenRequest` can now miss and fall back to a full `extractActions` re-parse, or to a ttxdb read plus `ProcessTokenRequest` (`finality/listener.go:121`).

Both fallbacks are correct, so this is not a correctness break — but it silently erodes the very commit-path optimization this commit is about, and the asymmetry with the fetcher looks unintentional.

token/services/tokens/storage.go
```diff
@@ -275,9 +280,11 @@ func (t *DBTransaction) Notify(ctx context.Context, topic string, tmsID token.TM
 // Calling it more than once is safe: the buffer is empty after the first call.
 func (t *DBTransaction) FlushEvents(ctx context.Context) {
 	pending := t.pending
-	t.pending = nil
+	t.pending = t.pending[:0]
```
AkramBitar, line 283: **low — `pending[:0]` aliases the slice being iterated.**

`pending := t.pending; t.pending = t.pending[:0]` makes the slice being iterated alias `t.pending`'s backing array. FSC's `simple.EventBus.Publish` invokes subscribers **synchronously** (`platform/view/services/events/simple/eventbus.go:30`), so any subscriber that reaches back into `Notify` on the same `DBTransaction` appends over `pending[0]` mid-loop: the original event is lost and the newly recorded one is published in its place.

The previous `t.pending = nil` had no such hazard, and it only costs one allocation per transaction.

Same at `storage.go:296` (`Rollback`), where `[:0]` additionally keeps the discarded events reachable from the backing array instead of dropping them.

token/services/ttx/session.go
```diff
@@ -93,32 +105,71 @@ func (s *localSession) SendError(ctx context.Context, payload []byte) error {
...
+		FromEndpoint: info.RemoteEndpoint,
+		FromPKID:     info.RemotePKID,
 		Status:       status,
 		Payload:      payload,
 		Ctx:          ctx,
 	}
 
-	return nil
+	select {
+	case s.writeChannel <- msg:
```
AkramBitar, line 140: **low — `send` returns `nil` after the peer closed, silently dropping the message.**

After the peer has closed, `send` still returns `nil`: the local session's `closed` flag is false and the peer's read buffer has room, so the payload is accepted into a channel whose reader (`peer.Receive()`) now returns `nil` and will never drain it.

Concretely, in `startLocal` (`auditor.go:236`) `cleanupSessions` closes `left` at the end of `CollectEndorsementsView` while the `AuditApproveView` goroutine is still running (it does `Append` + `CacheRequest` after `signAndSendBack`); any further `right.Send` reports success while the message is silently dropped, instead of surfacing "session is closed".

token/services/ttx/session.go
```diff
@@ -93,32 +105,71 @@ func (s *localSession) SendError(ctx context.Context, payload []byte) error {
...
+		return nil
+	case <-s.closedChan:
+		return errors.New("session is closed")
+	case <-ctx.Done():
```
AkramBitar, line 144: **low — `case <-ctx.Done()` races an immediately-ready buffered send.**

Go picks uniformly among ready cases, so a `Send` made with an already-cancelled context but free buffer space delivers ~50% of the time and returns `"context cancelled while sending message"` the other ~50% — nondeterministic, and a latent flake source.

Attempting a non-blocking send first (`select { case ch <- msg: default: }`) before falling through to the blocking select would make it deterministic.

token/services/tokens/tokens.go
```diff
@@ -223,7 +223,12 @@ func (t *Service) DeleteTokensBy(ctx context.Context, deletedBy string, ids ...*
 
 // DeleteTokens marks the tokens as spent in the database, attributed to the caller's stack trace.
 func (t *Service) DeleteTokens(ctx context.Context, ids ...*token2.ID) (err error) {
-	return t.DeleteTokensBy(ctx, string(debug.Stack()), ids...)
+	deletedBy := "Service.DeleteTokens"
```
AkramBitar, line 226: **low — `spent_by` records a constant string when debug logging is off.**

With debug logging off (the production default), `Service.DeleteTokens` now writes the literal string `"Service.DeleteTokens"` into the `spent_by` column for every token it removes. `WhoDeletedTokens` therefore returns a constant with no attribution value for tokens deleted via `PruneInvalidUnspentTokens` or the public `DeleteTokens` API, and the stored value differs between debug and non-debug deployments.

The commit-path attribution is unaffected — `AppendValid` still passes the real `txID` — and `integration/token/fungible/tests.go:814` only asserts on that path, so no test breaks.

If the field is meant to be diagnostic, a stable caller identifier would be better than a constant.

AkramBitar (commented):
Review of `541179e8c..2d0ca3c53` (the `localSession` race fix + the tokens commit-path optimizations).

Verified before commenting: `go build`, `go vet` and `go test` are clean on all touched packages, plus `-race -count=200` on the new session tests. I could **not** reproduce a `close of closed channel` / `send on closed channel` panic: `writeChannel` is closed only when `closed && inFlight == 0`, and `inFlight` is incremented under the same lock that rejects new sends, so that window is properly closed. Left and right also own disjoint write channels, so cross-side double-close is impossible.

The diff is largely sound. Six findings inline — one medium, the rest low. The medium (`session.go` `Close()` asymmetry) is worth addressing before merge; it leaks a goroutine in the same failure class #2285 set out to fix.

PR #2367, opened 2026-09-14
tokens: GetCachedTokenRequest nil-derefs on a (nil, true) cache hit on the finality hot path

### LINKED ISSUE #2185: tokens: GetCachedTokenRequest nil-derefs on a (nil, true) cache hit on the finality hot path
### Summary
`GetCachedTokenRequest` dereferences the cache lookup result without checking it for `nil`, unlike `getActions` reading the very same cache a few lines away, which does guard it. A `Cache` implementation that can return `(nil, true)` — present-but-nil — makes `GetCachedTokenRequest` panic instead of behaving like a cache miss.

### Where
`token/services/tokens/tokens.go:186-193` (unguarded):
```go
func (t *Service) GetCachedTokenRequest(txID string) (*token.Request, []byte) {
	res, ok := t.RequestsCache.Get(txID)
	if !ok {
		return nil, nil
	}
	return res.Request, res.MsgToSign
}
```
vs. `tokens.go:354-355` (guarded, same cache):
```go
entry, ok := t.RequestsCache.Get(string(anchor))
if ok && entry != nil {
```

### Impact
`GetCachedTokenRequest` is called from the finality path per committed transaction. A `Cache` that returns `(nil, true)` for an evicted-but-tracked or explicitly-nil-valued entry turns a cache lookup into a panic in a listener goroutine, instead of the graceful fallback the `ok`-only check was presumably meant to provide.

### Reproduction
```go
mockCache.GetReturns(nil, true)

assert.Panics(t, func() {
	svc.GetCachedTokenRequest("tx1")
})
```
```
=== RUN   TestGetCachedTokenRequest_PanicsOnNilEntryCacheHit
--- PASS: TestGetCachedTokenRequest_PanicsOnNilEntryCacheHit (0.00s)
PASS
```
(The test asserts the panic happens — i.e. it confirms the current, undesired behavior.)

### Severity
Medium — panics in a finality listener goroutine; blast radius depends on how that goroutine's panic is recovered upstream.

Part of #2112.

Effi-S:
Fixes #2185

### Summary
`GetCachedTokenRequest` dereferences the cache lookup result without checking it for `nil`, unlike `getActions` reading the very same cache a few lines away, which does guard it. A `Cache` implementation that can return `(nil, true)` — present-but-nil — makes `GetCachedTokenRequest` panic instead of behaving like a cache miss.

### Where
`token/services/tokens/tokens.go:186-193` (unguarded):
```go
func (t *Service) GetCachedTokenRequest(txID string) (*token.Request, []byte) {
	res, ok := t.RequestsCache.Get(txID)
	if !ok {
		return nil, nil
	}
	return res.Request, res.MsgToSign
}
```
vs. `tokens.go:354-355` (guarded, same cache):
```go
entry, ok := t.RequestsCache.Get(string(anchor))
if ok && entry != nil {
```

### Impact
`GetCachedTokenRequest` is called from the finality path per committed transaction. A `Cache` that returns `(nil, true)` for an evicted-but-tracked or explicitly-nil-valued entry turns a cache lookup into a panic in a listener goroutine, instead of the graceful fallback the `ok`-only check was presumably meant to provide.

### Reproduction
```go
mockCache.GetReturns(nil, true)

assert.Panics(t, func() {
	svc.GetCachedTokenRequest("tx1")
})
```
```
=== RUN   TestGetCachedTokenRequest_PanicsOnNilEntryCacheHit
--- PASS: TestGetCachedTokenRequest_PanicsOnNilEntryCacheHit (0.00s)
PASS
```
(The test asserts the panic happens — i.e. it confirms the current, undesired behavior.)

### Severity
Medium — panics in a finality listener goroutine; blast radius depends on how that goroutine's panic is recovered upstream.

Part of #2112.


adecaro (commented):
## Review: does this fully address #2185?

@Effi-S

**Approve on substance — the fix is minimal, matches the sibling guard on the same cache a line away, and the new test genuinely pins the bug.**

Checked out `fix-2185` at `0cbc8613` and verified:

```
$ go test ./token/services/tokens/... -run TestGetCachedTokenRequest -v
--- PASS: TestGetCachedTokenRequest (0.00s)

# temporarily reverted the guard (`if !ok || res == nil` -> `if !ok`) and re-ran:
--- FAIL: TestGetCachedTokenRequest
panic: runtime error: invalid memory address or nil pointer dereference
    .../tokens.go:207 GetCachedTokenRequest(...)
# restored the fix; git status clean, no diff vs. head
```

I also grepped every caller of `GetCachedTokenRequest` (`token/services/ttx/finality/listener.go:121` and `recovery.go:98`) — both already branch on `tr == nil` to reload the request from `ttxDB`, so this fix changes observed behavior from "panic" to exactly the fallback path those callers were already written to expect. No other production `Cache` implementation (`ristrettoCache`, `NoCache`) can currently produce `(nil, true)`, matching the triage on #2185 that this is a defensive/consistency fix rather than a live production bug today.

CI is fully green at this SHA (`checks`, `lint`, `utest` race+regression legs, `cgo-check`, `bench-check`, CodeQL, the full `itest` matrix, DCO) and the single commit is signed off.

### What it fixes

| # | Finding | Verified |
|---|---|---|
| 2185.1 | `GetCachedTokenRequest` dereferenced a cache hit without a nil guard, unlike `getActions`'s `ok && entry != nil` on the same cache | ✅ `tokens.go:203` now matches that pattern |
| 2185.2 | A `(nil, true)` cache response should behave like a miss, not panic | ✅ confirmed by reverting the fix and re-running the new test case |

### Recommendation

Nothing to change — this is good to go as-is.

adecaro (approved):
LGTM

PR #2370, opened 2026-09-14
feat(evm): EVM network driver

### LINKED ISSUE #2376: Add an EVM-backed network driver
## Problem

Panurus currently supports only a Fabric-backed `driver.Network` implementation. There is no way to run token services against an EVM-compatible chain (e.g. Besu, or a fabric-x-evm gateway), which blocks interoperability scenarios and adoption by teams standardized on EVM infrastructure.

## Goal

Provide an alternative `driver.Network` implementation backed by an EVM chain, on par with the existing Fabric driver, so token services can run against Besu directly or through a fabric-x-evm gateway.

## Scope

- JSON-RPC client, EIP-712 signing/recovery, ABI encode/decode
- Endorsement flow (initiator/responder/registry)
- State-delta translation, nonce/txid handling
- Public-params provider/watcher
- NWO test-network wiring and integration test topology/suites (Besu direct and via fabric-x-evm gateway, for both zkatdlog and fabtoken token drivers)
- Fuzz coverage for components parsing untrusted bytes (JSON-RPC responses, ABI decoding, EIP-712 recovery, state-delta, endorsement decoding)
- Documentation of the driver's design, deployment, and internals

## Impact

Enables running Panurus token services on EVM chains, expanding interoperability options beyond Fabric.

adecaro:
## Summary

Graduates the EVM network driver from `feature/evm-network-driver` into `main`: a full alternative `driver.Network` implementation backed by an EVM chain (Besu, or a fabric-x-evm gateway), alongside the existing Fabric driver.

- New `x/token/services/network/evm` Go module: JSON-RPC client, EIP-712 signing/recovery, ABI encode/decode, endorsement flow (initiator/responder/registry), state-delta translator, nonce/txid handling, public-params provider/watcher, and NWO test-network wiring.
- Integration test topology and suites under `integration/nwo/token/evm` and `integration/token/fungible/{evm,evmfabtoken,evmgw,evmgwfabtoken}`, covering the driver against a Besu network directly and via a fabric-x-evm gateway, with both zkatdlog and fabtoken token drivers.
- `docs/services/network-ethereum*.md` documents the driver's design, deployment, and internals; `docs/services/network.md` updated to reflect the current `driver.Network` interface.
- Fuzz coverage for every EVM component that parses untrusted bytes (JSON-RPC responses, ABI decoding, EIP-712 recovery, state-delta, endorsement decoding), wired into `.github/workflows/nightly-fuzz.yml`.
- CI: `itest-evm` previously ran only on the feature branch; this PR updates its gate so it keeps running once the driver lives on `main`.

## Verification

- `make checks-fast` and `make checks-heavy` pass across all Go modules, including the new `x/token/services/network/evm` module.
- `go test ./...` passes in the `x/token/services/network/evm` module.
- `.github/workflows/tests.yml` already builds/tests/lints the evm module (it's part of `GO_MODULES` in the `Makefile`) and runs the dedicated `itest-evm` job (Besu + fabric-x-evm gateway suites) on both `main` and the feature branch.

## Test plan

- [ ] CI green on this PR: `checks`, `lint`, `utest`, `itest`, `itest-evm`, `bench-check`, `cgo-check`
- [ ] `docs` workflow builds cleanly (mkdocs strict mode)
- [ ] Markdown link check passes

Fixes #2376


cmd/artifactgen/go.mod
```diff
@@ -225,3 +226,5 @@ require (
 	modernc.org/memory v1.12.1 // indirect
 	modernc.org/sqlite v1.58.0 // indirect
 )
+
```
arner, line 229: This is probably not intended to be committed?

docs/services/network-ethereum-internals.md
```diff
@@ -0,0 +1,268 @@
...
+## How it compares to the other drivers
+
+| Concern | Fabric | FabricX | EVM (this driver) |
+|---|---|---|---|
+| Validation | chaincode on-chain | FSC off-chain + endorse | FSC off-chain + endorse |
+| Backend artifact | RWSet | RWSet (+ read deps) | typed `StateDelta` |
+| Endorsement transport | Fabric endorser protocol | FSC views | FSC views |
+| Spent semantics | delete/SN keys | delete/SN keys + MVCC read-set | spentRefs list + explicit on-chain checks |
+| Re-validation at commit | MVCC | MVCC | the contract does it explicitly — there is no read-set |
+| Finality signal | push committer | notification queue | receipt + `eth_getTransactionByHash` polling, primarily; a fabric-x-evm gateway can add its own `isPending` lifecycle as a faster signal where available |
```
arner, line 54: What do you mean with the fabric-x-evm gateway isPending lifecycle? We're close to supporting websockets, would that help?

docs/services/network-ethereum-deployment.md
```diff
@@ -0,0 +1,253 @@
...
+2. Collect a quorum of endorsements over the setup delta. The delta asserts the parameters it
+   supersedes, so it must be built against the *current* version - if another update lands first, it
+   is stale and reverts with `StalePublicParams`.
+3. Submit it through a funded account and wait for the receipt. A revert here is much easier to read
+   than the timeouts it otherwise causes on every node afterwards.
+
+[`nwo.SetupUpdater`](../../x/token/services/network/evm/nwo/setup.go) does exactly this for a harness
+that holds every endorser key. In production the signatures come from the endorsers themselves.
+
+Nodes pick the change up on their own. Each one polls the contract's version counter (`pp.Watcher`,
```
arner, line 192: Perhaps this can be optimized in the future by using ethereum events (logs)?

docs/services/network-ethereum-deployment.md
```diff
@@ -0,0 +1,253 @@
...
+- **A node that logs `public parameters updated to version N` has reloaded.** Updating a TMS *evicts*
+  the cached management service so the next caller builds a new one, so anything still holding the old
+  service keeps the old parameters. If requests are rejected against parameters a node has already
+  logged as updated, that is the shape of the bug.
+- **An update that authorises a new issuer takes effect per node, as each notices.** Until every node
+  in the endorsement path has, a request from the new issuer can be endorsed by some and rejected by
+  others.
+
+Recovery for transactions left `Pending` across a restart is not configured under `evm` at all: settings
+load from `services.network.fabric.recovery`, the same key the Fabric and FabricX drivers read, with the
```
arner, line 207: @adecaro perhaps this is a sign that it's not so network specific?

x/token/services/network/evm/client/evmclient.go
```diff
@@ -0,0 +1,119 @@
...
+// CallMsg describes a read-only call or a gas estimation request.
+type CallMsg struct {
+	From  *Address
+	To    *Address
+	Data  []byte
+	Value *big.Int
+}
+
+// EVMClient abstracts the JSON-RPC surface the network driver depends on. It is intentionally
+// minimal and backend-agnostic so the driver works against any EVM node, including fabric-x-evm.
```
arner, line 86: please let us know if there are any features you would implement if fabric-x-evm would support them.

arner (commented):
Hi @adecaro , I focused on the docs only as you asked. Of course it's hard to grasp the full extent, so I left just a couple of comments (for what it's worth). At a high level the approach looks good, and it's cool to see an ethereum-style backend on Panurus! Plus not relying on geth for the client is an achievement.

cmd/artifactgen/go.mod
adecaro: it is required due to the import of the integration module.

x/token/services/network/evm/client/evmclient.go
adecaro: we will, thanks.

docs/services/network-ethereum-internals.md
adecaro: I'll remove that. It does add anything. eth_getTransactionByHash is what is used.

docs/services/network-ethereum-deployment.md
adecaro: an indexer could fire these events and panurus can pick them up?

docs/services/network-ethereum-deployment.md
adecaro: it is worth fixing this, indeed.

adecaro: I have opened Issue https://github.com/LFDT-Panurus/panurus/issues/2396

KElkhiyaoui (approved):

arner (approved):
Too large for a full review. Docs and general approach look good, I trust that you reviewed all the code in earlier rounds :)

x/token/services/network/evm/network.go
```diff
@@ -0,0 +1,726 @@
...
+	}
+
+	return computeAnchor(id.Nonce, id.Creator)
+}
+
+// SetupPublicParams collects endorsements for a public-parameters update. The endorsers translate the
+// setup request into a setup delta, so the flow matches RequestApproval; it exists separately because
+// it takes a TMSID rather than a management service, which is what makes first-time setup of a
+// namespace with no public parameters possible.
+func (n *Network) SetupPublicParams(
```
AkramBitar, line 480: **`SetupPublicParams` can never succeed.**

This sends `publicParamsRaw` as `EndorseRequest.TokenRequest`, so the responder's `DeltaFactory.Build` (`endorsement/delta.go:96`) hands public parameters to `validator.UnmarshallAndVerifyWithMetadata`, which cannot decode them as a `token.Request`. The request is declined with `ErrValidation` and the initiator returns `ErrInsufficientEndorsements`.

Because no endorsement path ever yields a `translator.SetupAction`, `StateDelta.IsSetup` / `SetupParameters` are unreachable outside `nwo/setup.go` — an on-chain PP update isn't possible through the driver. Fabric handles this with a *separate* responder (`network/fabric/endorsement/fsc/responder.go`, `setupBehaviour` + a synthetic setup action).

The suites don't catch it because the NWO handler submits the setup delta itself via `evmnwo.SetupUpdater`.

AkramBitar (commented):
@adecaro Thanks a lot for this great work. Added some comments.

x/token/services/network/evm/driver.go
```diff
@@ -0,0 +1,607 @@
...
+	d.watchersMu.Lock()
+	defer d.watchersMu.Unlock()
+	for tokenState, tmsIDs := range byContract {
+		key := network + ":" + channel + ":" + tokenState.Hex()
+		if _, running := d.watchers[key]; running {
+			continue
+		}
+		cfg := configByContract[tokenState]
+
+		watcher, err := pp.NewWatcher(
```
AkramBitar, line 268: **Block-tag mismatch stalls all endorsement after a PP update.**

The watcher reads at `cfg.Finality.BlockTag` (default `finalized`), while the endorsement `ChainProvider` at line 369 deliberately reads at `latest`. Once a setup delta is mined, `DeltaFactory.Build` sees the new params at `latest` but this node's TMS still holds the old ones, so the `chainHash != localHash` check refuses with `ErrStalePublicParams` for the whole finalization lag — `MinFinalizedTagTimeout` puts that at ~13 minutes.

Nothing retries: `Initiator.Collect` surfaces it as `ErrInsufficientEndorsements` and the transaction aborts. `ErrStalePublicParams`' own comment ("retried rather than fatal") understates both the duration and the absence of any retry. Invisible in the suites, which set `blockTag: latest`.

x/token/services/network/evm/contracts/README.md
```diff
@@ -0,0 +1,53 @@
...
+re-run — never edit expected values by hand.
+
+## Build and test
+
+```bash
+# one-time: fetch forge-std (vendored as a git submodule, pinned by foundry.lock)
+git submodule update --init --recursive
+
+forge build
+forge test
```
AkramBitar, line 43: **The Solidity tests never run in CI.**

`forge test` is documented here as a manual step and isn't invoked by any make target or workflow — `grep -rn "forge test"` over `Makefile`, `*.mk` and `.github` finds only comments. So the ~950 lines under `contracts/test/` never execute, including `EIP712.t.sol`, the cross-implementation golden-digest gate this README calls "the gate everything else builds on". `e2e_anvil_test.go` (1,253 lines) also skips, because the `utest` job installs no anvil or forge.

Net effect: a change to `StateDelta.sol` / `EIP712.sol` that desynchronises the Solidity digest from the Go one lands green, and the first symptom is `InvalidSignatures` against a live chain.

Makefile
```diff
@@ -75,6 +77,9 @@ GO_PACKAGES = $(shell go list ./... | grep -v '/integration/' | grep -v 'regress
 unit-tests:
 	@go test $(GO_TEST_PARAMS) $(GO_PACKAGES)
 	cd token/services/storage/db/kvs/hashicorp/; go test -cover ./...
+	cd x/token/services/network/evm; go test -coverpkg=./... -coverprofile=profile.cov ./...
```
AkramBitar, line 80: **`unit-tests-race` doesn't cover the new module.**

`unit-tests` was extended here, but `unit-tests-race` just below still enumerates `go list ./...` from the root module only, so `x/token/services/network/evm` is never race-tested.

That module is where the concurrency lives: `NonceManager`, `pp.Watcher` (goroutine plus two mutexes), `finality.Manager` (detached goroutines plus `sync.Once`), and `conflictWatch` — exactly what `-race` is for.

.github/workflows/tests.yml
```diff
@@ -251,7 +283,21 @@ jobs:
       - name: Set up tools
         run: make install-tools
 
+      - name: Set up foundry
+        if: startsWith(matrix.tests, 'evm')
+        uses: foundry-rs/foundry-toolchain@v1
```
AkramBitar, line 288: **Only unpinned action in the workflow.**

Every other `uses:` here is pinned as `@<sha> # vX.Y.Z`; this one is a mutable tag. `version: v1.7.1` pins *forge*, not the action — a retag of `v1` runs new third-party code in the itest jobs.

x/token/services/network/evm/envelope.go
```diff
@@ -0,0 +1,75 @@
...
+}
+
+// Compile-time assertion that Envelope satisfies the driver contract.
+var _ driver.Envelope = (*Envelope)(nil)
+
+// NewApprovedEnvelope builds the envelope RequestApproval returns once endorsement has assembled a
+// quorum: the anchor, the endorsed delta, and the signatures over its digest. RawTx and EthTxHash
+// stay empty until Broadcast submits it. Kept in primitive terms so the root package does not depend
+// on the endorsement package; the driver maps an endorsement result onto these arguments.
+func NewApprovedEnvelope(anchor string, delta *statedelta.StateDelta, endorsements [][]byte) *Envelope {
```
AkramBitar, line 46: **`Namespace` is never set, so the result can't be broadcast.**

`Broadcast` (`network.go:382`) fails with `no configuration bound for namespace []` for anything this constructor produces. The doc comment says it builds the envelope `RequestApproval` returns, but `RequestApproval` constructs the struct *with* `Namespace`. Currently only tests call this, so the divergence is invisible.

integration/nwo/token/evm/nwo.go
```diff
@@ -0,0 +1,401 @@
...
+			continue
+		}
+		identity, err := funder.FundedIdentity(ctx, keyDir, node.Name+"-submitter", DefaultFunding)
+		gomega.Expect(err).NotTo(gomega.HaveOccurred(), "failed to fund a submitter for [%s]", node.Name)
+		entry.SubmitterOf[node.Name] = identity
+	}
+}
+
+// startNode boots the chain this TMS settles on, reusing one already started for the same network.
+func (p *NetworkHandler) startNode(tms *topology2.TMS) Node {
```
AkramBitar, line 237: **`startNode` ignores `tms.Network`.**

It returns the first entry with a non-nil `Node`, though the comment says "one node per network". Since one `NetworkHandler` is registered per topology *type* (`integration/nwo/token/factory.go:42`), a suite standing up two EVM networks via `NewTopologyWithName` would silently settle both TMSs on the same chain — both `TokenState` clones and both `ReservePort` allocations, one of which is then never used.

x/token/services/network/evm/config.go
```diff
@@ -0,0 +1,437 @@
...
+		if _, dup := named[b.FSCIdentity]; dup {
+			return errors.Errorf(
+				"evm config: endorser identity [%s] is bound twice; one node cannot supply two of the "+
+					"distinct signatures a quorum needs", b.FSCIdentity)
+		}
+		seen[addr] = struct{}{}
+		named[b.FSCIdentity] = struct{}{}
+	}
+
+	if c.Endorser.Enabled {
```
AkramBitar, line 331: **Endorser address isn't checked against the endorsers set.**

With `endorser.enabled` set, this validates that the address parses and that the allowlist is non-empty, but never that the address actually appears in `endorsement.endorsers`. `verifyEndorsementPolicy` only checks the configured set against the chain, not this node against the set.

A node whose own address is missing or typo'd therefore starts cleanly and signs endorsements that `Initiator.verify` discards as `ErrUnknownSigner`. With the common `threshold == len(endorsers)`, that's every transaction failing, with no startup error and only a debug log.

integration/nwo/token/evm/nwo.go
```diff
@@ -0,0 +1,401 @@
...
+
+	backend := &ForgeBackend{Endpoint: entry.Node.Endpoint()}
+	deployment, err := backend.Deploy(evmnwo.DeploySpec{
+		TMS: evmnwo.TMSRef{
+			Network:   tms.Network,
+			Channel:   tms.Channel,
+			Namespace: tms.Namespace,
+		},
+		Endorsers:    addresses,
+		Threshold:    entry.Threshold,
```
AkramBitar, line 181: **`DeploySpec.GraphHiding` is never populated.**

The `Deploy` call omits it and `NetworkHandler` exposes no knob, so every TMS deploys with `graphHiding = false`. A graph-hiding driver emits serial-number `spentRefs` (`keys.SpentRefForSerial`), which the contract would then read as content-bound markers and revert `InputMissingOrSpent` on every spend. Harmless only while `zkatdlognoghv1` is the sole zkatdlog driver.

Makefile
```diff
@@ -92,9 +97,15 @@ unit-tests-regression:
 install-softhsm:
 	./ci/scripts/install_softhsm.sh
 
+# The EVM integration suites run against this Besu image.
+BESU_IMAGE ?= hyperledger/besu:24.3.0
```
AkramBitar, line 101: **These overrides don't reach the suites.**

`BESU_IMAGE` and `FABRICX_EVM_IMAGE` feed only the `docker pull` targets in `evm.mk`. The suites boot whatever `evm.DefaultBesuImage` / `defaultGatewayImage` hard-code, because `NetworkHandler.Image` is never set from the environment.

So `make integration-tests-evm BESU_IMAGE=hyperledger/besu:25.x` pulls one image and then fails in `CheckImagesExist` for the other — two independent sources of truth for the same value.

adecaro: This is true for the other flags as well. I would treat this as a more general issue, if one really wants to address it.

integration/nwo/token/evm/nwo.go
adecaro: graph hiding is not supported explicitly and indeed, panurus does not have yet a driver with such a support.

integration/nwo/token/evm/nwo.go
adecaro: not sure, it looks a stale comment.

The comment on startNode says: // startNode boots the chain this TMS settles on, reusing one already started for the same network.

x/token/services/network/evm/config.go
adecaro: but if this node is an endorser, then, it might not even have the list of the endorsers because it does not need to contact them at all. No?
What would the check buy us?
We could check if this address is in named in the endorsement policy later. 
What do you think?

AkramBitar: Agree.

Makefile
AkramBitar: Do you want to open a fellow up issue for this one?

.github/workflows/tests.yml
adecaro: all right, it might be helpful then to have a rule for the AI agents about this.

adecaro:
Thanks for the thorough review, @AkramBitar. I've collected these into #2412 to track and address as follow-up work: https://github.com/LFDT-Panurus/panurus/issues/2412

AkramBitar (approved):
LGTM

x/token/services/network/evm/network.go
adecaro: will be fixed in the context of https://github.com/LFDT-Panurus/panurus/issues/2412

PR #2372, opened 2026-09-14
ci: integrate govulncheck vulnerability scanning

### LINKED ISSUE #2371: Integrate govulncheck vulnerability scanning into CI
## Problem

CI currently runs `golangci-lint`, `staticcheck`, `go vet`, and license/format checks, but nothing scans dependencies for known vulnerabilities. A vulnerable transitive dependency can land and stay unnoticed unless someone manually runs a scanner or waits for GitHub's Dependabot alerts.

## Impact

- No automated signal when a new dependency (or a dependency bump) introduces a package with a published CVE/advisory that Panurus's code actually calls.
- Relies entirely on external tooling (Dependabot) rather than a build-time gate developers see in the same `make checks` / CI run as everything else.

## Proposal

Add `govulncheck` (the official Go vulnerability scanner) as part of the standard checks, restricted to vulnerabilities reachable from actually-called code (its call-graph analysis, not just "any vulnerable dependency in go.sum"). Findings with no upstream fix need a documented, reviewable allowlist mechanism so a single unfixable transitive advisory doesn't permanently block CI.

adecaro:
## Summary
- Add `govulncheck` (official Go vulnerability scanner) to `make checks-heavy`, so it runs alongside `staticcheck`/`golangci-lint` in CI's `lint` job without gating the integration-test critical path (it also needs network access to fetch the vuln DB).
- Restrict failures to vulnerabilities actually reachable from called code (govulncheck's call-graph analysis), not just "any vulnerable package in go.sum".
- Since govulncheck has no built-in suppression mechanism, `ci/scripts/govulncheck.sh` wraps it: only findings *not* listed in `ci/govulncheck-allowlist.txt` fail the build. The allowlist currently has one justified entry, `GO-2024-3218` (`go-libp2p-kad-dht`, no upstream fix, reached only through FSC's libp2p-based peer discovery in a permissioned Fabric network — not exposed to the public IPFS DHT the advisory concerns).
- `tools/tools.go` + `tools/go.mod`/`go.sum` now track `golang.org/x/vuln/cmd/govulncheck` so `make install-tools` installs it.
- Docs updated: `docs/development/tools.md`, `docs/development/makefile.md`, and `docs/development/update-fsc.md` (the FSC-bump runbook now calls out reviewing the allowlist entry when `go-libp2p-kad-dht`'s version changes).

## Test plan
- [x] `make govulncheck` passes cleanly across all 9 `GO_MODULES` (verified locally).
- [x] `make checks-heavy` passes end-to-end with the new target wired in.
- [x] Verified the wrapper actually fails a module when a reachable finding isn't in the allowlist (temporarily emptied the allowlist, confirmed exit 1, restored it).
- [x] `make tidy-check` / `misspell` clean on the modified `tools` module.

Fixes #2371


Effi-S (approved):
LGTM

PR #2373, opened 2026-09-15
test(zkatsnark): add regression tests and fix transfer action validation

neetance:
### Summary
This PR adds an end-to-end regression test suite for the `zkatsnark` token driver and resolves key transfer action validation and signature verification requirements:
1. **End-to-End Regression Test Suite (`regression_test.go`)**:
   - Tests complete token lifecycle flows (`Issue` and `Transfer`) end-to-end using real SNARK provers (`SpendProver`, `OutputProver`), circuits, and validators.
   - Verifies single-input and multi-input transfer requests, ensuring ledger state resolution, SNARK proof verification, and signature consumption behave as expected.
2. **Transfer Signature Validation & Consumption (`validator_transfer_sig.go`)**:
   - Implemented `TransferSignatureValidate` to satisfy the common token framework validation contract.
   - Validates that each spend input is signed by its respective owner identity and that signatures are properly consumed from the signature provider (`EnsureExhausted`).
3. **Input Token State Matching in Transfer Actions**:
   - Updated `TransferAction` to hold and serialize original input `OutputDescription`s in `InputTokens`, ensuring ledger lookup and token key verification succeed during transfer validation.

### Testing

All tests pass
```
 go test ./...
?       github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark  [no test files]
ok      github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark/circuit  148.514s
ok      github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark/circuit/gadgets  0.043s
ok      github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark/crypto/jubjub    0.121s
ok      github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark/crypto/mimc      0.045s
?       github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark/crypto/params    [no test files]
?       github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark/driver   [no test files]
ok      github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark/pp       0.039s
ok      github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark/prover   152.515s
ok      github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark/regression       7.358s
ok      github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark/setup    155.601s
ok      github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark/token    0.044s
ok      github.com/LFDT-Panurus/panurus/x/token/core/zkatsnark/validator        183.331s
```


x/token/core/zkatsnark/validator/validator_transfer_sig.go
```diff
@@ -0,0 +1,134 @@
...
+
+	inputIDs := action.GetInputs()
+	if len(inputIDs) != len(action.Inputs) {
+		return errors.Wrapf(ErrMissingInputIDs, "expected [%d] input IDs, got [%d]", len(action.Inputs), len(inputIDs))
+	}
+
+	var signatures [][]byte
+	var inputTokens []*snarktoken.Input
+
+	for i, inputID := range inputIDs {
```
Effi-S, line 58: The validation chain (`TransferSignatureValidate` + `TransferZKValidate`) never checks that the SNARK-proven `Inputs[i].CommitmentIn` corresponds to the token identified by the new `InputIDs[i]/InputTokens[i]` fields. Authorization and ledger deletion are decoupled from what the proof actually spends.

**Failure scenario:** An attacker owning tokens A (value 100) and B (value 1) builds a ZK-valid transfer whose `SpendDescription` proves the opening of A, but sets `InputIDs/InputTokens` to reference `B. TransferSignatureValidate` loads B from the ledger, finds the attacker as owner, and accepts the signature; `TransferZKValidate` verifies the (valid) proof over A's commitment; the RWS translator deletes B (value 1). Value conservation is enforced only over A (value 100), so the attacker mints ~99 units. Neither validator checks that `Inputs[i].CommitmentIn == OutputDescription(InputTokens[i]).CommitmentOut`

x/token/core/zkatsnark/validator/validator_transfer_sig.go
```diff
@@ -0,0 +1,134 @@
...
+		// Load the on-chain output (OutputDescription) to get the owner
+		raw, err := ctx.Ledger.GetState(*inputID)
```
Effi-S, line 60: A crafted `TokenRequest` with a transfer action JSON 
```json
{"InputIDs":[null],"Inputs":[{...one spend...}]}
```
deserializes to a `[]*token2.ID` slice containing a nil pointer.
The `len(inputIDs)==len(action.Inputs)` guard passes (both 1), then `ctx.Ledger.GetState(*inputID)` dereferences the nil pointer and panics, crashing the validating node on untrusted input.

x/token/core/zkatsnark/transfer.go
```diff
@@ -148,6 +148,15 @@ func (s *TransferService) Transfer(
 		return nil, nil, errors.Wrapf(err, "failed to generate zk transfer")
 	}
 
+	// Populate InputIDs and InputTokens so the validator can look up input token owners
+	// and correctly generate osn keys for ledger validation.
+	action.InputIDs = tokenIDs
+	var inputTokens [][]byte
+	for _, loadedToken := range loadedTokens {
+		inputTokens = append(inputTokens, loadedToken.Token)
```
Effi-S, line 156: nit: append in a loop -> best to pre-allocate and index assign
Something like this:
```go 
inputTokens := make([][]byte, len(loadedTokens))
for i, token := range loadedTokens {
    inputTokens[i] = token.Token
}
```

x/token/core/zkatsnark/validator/validator_transfer_sig.go
```diff
@@ -0,0 +1,134 @@
...
+		if err != nil {
+			return errors.Wrapf(err, "failed loading token [%s] from ledger for input [%d]", inputID, i)
+		}
+
+		var outputDesc snarktoken.OutputDescription
+		if err := json.Unmarshal(raw, &outputDesc); err != nil {
```
Effi-S, line 66: This PR adds new fields to `TransferAction.Deserialize` (`InputIDs`/`InputTokens`/`Issuer`) and adds a validator that `json.Unmarshals` ledger tokens, but adds no fuzz target and no `{name,pkg,func}` entry to `.github/workflows/nightly-fuzz.yml`
The new parsing surface is never exercised under `-fuzztime`.

x/token/core/zkatsnark/token/transfer.go
```diff
@@ -56,14 +59,16 @@ func (a *TransferAction) NumInputs() int {
...
 
-// GetInputs returns nil; zkatsnark inputs are identified by commitment,
-// not by ledger token IDs on the action itself.
+// GetInputs returns the ledger token IDs of the tokens being spent.
 func (a *TransferAction) GetInputs() []*token2.ID {
-	return nil
+	return a.InputIDs
 }
 
 // GetSerializedInputs returns the serialized inputs of the action.
```
Effi-S, line 67: `GetSerializedInputs` falls back to serializing `SpendDescriptions` when InputTokens is empty, producing `OutputSN` keys that can never match the on-ledger token keys.

**Failure scenario:**  If a `TransferAction` has `InputID`s set but `InputTokens` empty (an action built/reconstructed without the new population step, or a wire action omitting `InputTokens`), `GetSerializedInputs` returns marshaled `SpendDescriptions`. `translator.checkInputs` then computes `CreateOutputSNKey(TxId, Index, <SpendDescription bytes>)`, which never equals the key stored from the token's `OutputDescription`, so `StateMustExist` fails with 'input must exist'... 
A hard-to-diagnose failure driven by which field happens to be populated.

x/token/core/zkatsnark/regression/regression_test.go
```diff
@@ -0,0 +1,545 @@
...
+
+	// Store the issued token on the test ledger so the validator can look up owner
+	ledger := &testLedger{tokens: make(map[string][]byte)}
+	storeLedgerToken(t, ledger, "issue-tx-0", 0, issueAction.Outputs[0])
+
+	// Build token request with the transfer action
+	transferRaw, err := transferAction.Serialize()
+	require.NoError(t, err)
+
+	tr := &driver.TokenRequest{
```
Effi-S, line 377: Signatures here are never used. 
so the test does not exercise the common framework's signature-scoping/EnsureExhausted path it claims to regression-test.

Effi-S (changes requested):

x/token/core/zkatsnark/validator/validator_transfer_sig.go
neetance: Thanks for picking this up.
I added a verification step in `TransferSignatureValidate` checking `bytes.Equal(action.Inputs[i].CommitmentIn, outputDesc.CommitmentOut)`, returning a new error `ErrCommitmentMismatch` on mismatch. This binds the snark proof directly to the on-ledger token being spent.

x/token/core/zkatsnark/validator/validator_transfer_sig.go
neetance: Fixed. Added a nil check before dereferencing `inputID`, returning `ErrNilInputID` if any entry is nil.

x/token/core/zkatsnark/transfer.go
neetance: Thanks for the suggestion 🙏 
Pre-allocated `inputTokens` with `make([][]byte, len(loadedTokens))` and index assignment.

x/token/core/zkatsnark/validator/validator_transfer_sig.go
neetance: Added `FuzzActionDeserializerNoPanic` and `FuzzTransferActionValidateNoPanic` in validator_fuzz_test.go
Both targets have also been added to the matrix in `.github/workflows/nightly-fuzz.yml`.

x/token/core/zkatsnark/token/transfer.go
neetance: Addressed. Removed the fallback to `SpendDescriptions`
`GetSerializedInputs` now returns an explicit error if `InputTokens` is not populated (or nil, nil if there are no inputs). Updated unit tests in `token_actions_test.go` to verify this.

x/token/core/zkatsnark/regression/regression_test.go
neetance: Refactored the regression tests (`SignaturesConsumed`, `TwoInputs_SignaturesConsumed`, and `MissingInputIDs_Fails`) to call `VerifyTokenRequestFromRaw` end-to-end.

Effi-S (changes requested):
@neetance Is `zkatsnark` is meant to be a first-class driver like `zkatdlog`? 

- `make checks` / `fmt` / `lint` / `license`:  all loop over `GO_MODULES`, and `x/token/core/zkatsnark` isn't in that list.

- `nightly-fuzz.yml`: the new entries use pkg: `./x/token/core/zkatsnark/validator` and run `go test ${pkg}` from repo root. That path is a different module, so it won't resolve either. the fuzz targets won't actually run!

So the suite passes locally but is invisible to the CI

Issue #2377, opened 2026-09-16
Split gocognit complexity fixes (#2327) into per-subsystem PRs


atharrva01:
Tracking issue for splitting up #2327 (enable the `gocognit` linter) into reviewable, per-subsystem PRs.

## Background

#2327 enables `gocognit` (min-complexity 15) across every Go module and fixes every violation it found. The diff ended up touching 100+ files across most subsystems in one PR, which made it hard to review closely, and a real regression (a dropped signature-signing call in a mechanical extraction) got through review as a result and only showed up in the itest matrix, not in unit tests or lint.

Per @adecaro and @Effi-S's feedback on #2327, this work is being split: #2327 becomes the base PR (enables the linter repo-wide, keeps the small number of permanent, security/crypto-justified `//nolint:gocognit` waivers), and every function that needs an actual complexity-reducing extraction gets a temporary `//nolint:gocognit // TODO(#2377): tracked in issue #N` marker instead, removed one subsystem at a time by the sub-PRs below.

## Sub-issues

- [ ] #2378 token/core/common
- [ ] #2379 token/core/fabtoken
- [ ] #2380 token/core/zkatdlog
- [ ] #2381 token/services/identity
- [ ] #2382 token/services/ttx
- [ ] #2383 token/services/selector
- [ ] #2384 token/services/storage
- [ ] #2385 token/services/network
- [ ] #2386 token/services/benchmark
- [ ] #2387 token/services misc (tokens, auditor, certifier, interop/htlc, driver)
- [ ] #2388 integration module
- [ ] #2389 cmd/* tooling modules

Each sub-issue lists the exact files in scope.

atharrva01:
First sub-PR is up: #2390 for #2382 (token/services/ttx).

atharrva01:
#2327 is closed. Dropping the "base PR" step from the plan since there's nothing left to shrink, each sub-issue below now just lands directly as its own PR against `main` (like #2390 for #2382).

atharrva01:
Next sub-PR up: #2394 for #2378 (token/core/common).

PR #2390, opened 2026-09-16
lint(ttx): split gocognit fixes out of #2327

### LINKED ISSUE #2382: gocognit split: token/services/ttx
Part of #2377.

Re-apply the gocognit complexity-reduction extraction for `token/services/ttx`, removing the temporary `//nolint:gocognit // TODO(#2377)` markers #2327 leaves on this package.

Note: this package had the one real regression found on #2327 (a dropped attestation-signing call in `recipients.go`'s `exchangeRemotely` extraction, caught by the itest matrix). The corrected version of that extraction is already worked out and itest-verified on the `lint/enable-gocognit` branch; re-review it carefully given the history.

Files in scope:
- token/services/ttx/collectendorsements.go
- token/services/ttx/recipients.go
- token/services/ttx/marshaller.go
- token/services/ttx/finality.go
- token/services/ttx/endorse.go
- token/services/ttx/endorse_opts.go
- token/services/ttx/finality/recovery.go
- token/services/ttx/finality/listener.go
- plus associated `*_test.go` files touched only for split subtests (endorse_test.go, finality_test.go)

atharrva01:
Fixes #2382.

First slice of the gocognit split agreed on #2327 (tracking issue #2377): pulls just `token/services/ttx`'s complexity-reduction extraction out into its own reviewable PR, ahead of #2327 itself enabling the linter.

Splits:
- `collectendorsements.go`: the signer-strategy branches in `requestSignatures` (multi-sig, policy, external wallet, local, remote) into named `tryXxxSignature` helpers, and the distribution-list construction in `prepareDistributionList` into `unwrapDistributionIDs`/`addPartyToDistributionList`/`addAuditorToDistributionList`/`resolveLongTermIdentity`.
- `recipients.go`: the recipient-identity request/respond/exchange handshakes into per-case helpers.
- `endorse.go`, `endorse_opts.go`, `finality.go`, `finality/listener.go`, `finality/recovery.go`, `marshaller.go`: smaller, single-purpose extractions.

A couple of view handshakes keep a justified `//nolint:gocognit`: splitting the decision logic from the response send risks a response that doesn't match what was actually decided, so those stay as one function.

Also fixes a real bug that only showed up in #2327's itest matrix, not in unit tests or lint: the first pass at `ExchangeRecipientIdentitiesView.Call`'s `exchangeLocally`/`exchangeRemotely` split dropped the initiator's attestation-signing call, so every cross-node recipient exchange (swap, DVP, HTLC) went out with an empty signature and got rejected. That's restored here and itest-verified on #2327 before this split.

Verified: `go build ./...`, `golangci-lint run` (default set) on `token/services/ttx/...`, `go test ./token/services/ttx/...`, `gofmt -l -s`, `go vet`.


atharrva01:
@Effi-S PTAL

token/services/ttx/recipients.go
```diff
@@ -209,44 +209,54 @@ func RequestPolicyIdentity(context view.Context, policy string, ids []view.Ident
...
+			return nil, false, errors.Wrapf(err, "failed to get recipient identity")
+		}
+
+		return identity, false, nil
+	}
+
+	if isRemoteRecipient := recipient.RecipientData != nil; isRemoteRecipient {
+		return recipient.RecipientData.Identity, true, nil
+	}
+	if w == nil {
```
Effi-S: nit: unreachable (b/c of line 222)

Effi-S (approved):

token/services/ttx/recipients.go
atharrva01: fixed in b430fcf48 (also rebased onto main, which had drifted since I opened this). Removed the dead `if w == nil` check.

AkramBitar:
@atharrva01 

Could you please rebase and squash all the commits to one before I will merge it?

Regards,
Akram

Effi-S (approved):

Issue #2395, opened 2026-09-21
sherdlock selector: hot-token lock contention causes false insufficient-funds under load


adecaro:
## Summary

During a stress test, PostgreSQL logs showed massive `zkat_token_locks_pkey` constraint-violation traffic concentrated on a very small number of tokens: 60 out of 1276 tokens experienced lock collisions, and just 5 tokens (0.47% of the wallet's tokens) accounted for ~85% of all observed collisions. One token was contested across roughly a 6-minute window (aggregated across many separate, repeated selection attempts hitting that same token — not one continuous selector call, since a single `Select()` invocation cannot itself run anywhere near that long under the default retry/backoff config), despite belonging to a transaction that eventually reached `SETTLED`. These collisions coincide with spurious "insufficient funds" selection errors on 1 EUR payments even though the paying wallets held multiple tokens.

## Impact

- Payment attempts fail with `insufficient funds` even when the wallet has enough total value, because the funds are transiently (or unnecessarily long) locked by other in-flight selections.
- Database load/log volume balloons under concurrency due to repeated `INSERT` attempts on `token_locks` that are known in advance to collide.
- Selection is effectively serialized on a handful of tokens per wallet whenever concurrent debits exceed the wallet's UTXO count, amplified by the fact that token selection order has no size-based priority (a payment for a small amount can just as easily draw the wallet's single large token as its single small one, since candidate order is randomized, not smallest-fit).
- It's currently unclear from application code/logs alone whether a contested token's lock is released promptly once its owning transaction settles, or whether it lingers until the periodic lease-expiry sweep — this needs to be diagnosable without a manual DB/log correlation exercise like the one done for this stress run.

## Ask

1. Investigate and fix the token-selector/locking behavior in `token/services/selector/sherdlock` (and the lock store in `token/services/storage/db/sql/...`) so that:
   - Tokens already locked by another in-flight selection are not repeatedly re-offered as candidates to competing selections.
   - Locks are released promptly once their owning transaction reaches a terminal state, rather than only through the periodic lease-expiry cleanup. (This is the most load-bearing gap of the ones identified so far — confirm there is genuinely no existing hook on the successful/settled commit path, not just on the error path, before designing the fix.)
[...]

PR #2397, opened 2026-09-21
sherdlock: add lock-contention diagnostics and a reproducible baseline test (Phases 1-2 of #2395)

adecaro:
## Summary

Part of #2395 — a stress load test showed 0.47% of tokens absorbing 95.2% of
`zkat_token_locks_pkey` violations under the sherdlock token selector. This
PR covers Phases 1-2 of the fix plan: diagnostics and a reproducible
contention baseline, no behavioral change to selection itself, so later
phases (blacklist, anti-join, lock release on settlement) have a real
yardstick to measure against.

### Phase 1 — Diagnostics

- **Read API for held locks.** `driver.TokenLockStore` gains `ListLocks`.
  Fixed a latent cross-dialect bug while adding it: `modernc.org/sqlite`
  only auto-converts `DATE`/`DATETIME`/`TIMESTAMP` columns to `time.Time` on
  `Scan`, not `TIMESTAMPTZ` (which the shared schema deliberately uses for
  timezone-consistent comparison against Postgres's `NOW()`). Added a
  `scannableTime` custom `sql.Scanner` that handles both dialects' wire
  formats. Also adds `cond.NotExists` to the query DSL, a prerequisite
  building block for the Phase 4 anti-join.
- **`tokendiag locks` CLI.** New `cmd/tokendiag` module (mirrors
  `cmd/skicleanup`'s pattern) with a `locks` subcommand that connects
  directly to an existing Panurus database and reports every held lock with
  its age and consumer tx status, flagging locks whose consumer is already
  terminal (`Confirmed`/`Deleted`/`Orphan`) as **leaked** — this is
  mechanism 4 from the issue: nothing releases a lock on settlement, so it
  sits until the next lease sweep.
- **Contention metrics + structured logging.** `sherdlock/metrics.go` gains
  `LockConflicts` (deliberately unlabeled by token/wallet id — unbounded
  cardinality) and `DistinctTokensAttempted` (per `Select()` call), to tell
  apart "one hot token retried many times" from "many tokens each
  contended once". `selector.go` now distinguishes a lost lock race
  (`driver.ErrTokenAlreadyLocked`) from a genuine store error at the
  `TryLock` call site, promotes the conflict log line from `Debug` to
  `Info` with the token id, and the in-memory locker maps its own
  `AlreadyLockedError` to the same driver sentinel so both backends behave
  alike.

### Phase 2 — Reproducible contention baseline

- **`testutils.TestHotTokenContention`.** A workload shaped like the
  reported incident: a wallet with a few small tokens plus one large,
  rotating hot token, and far more concurrent requests than tokens across
  multiple replicas.
- **`countingLocker` decorator + `startManagersWithLockCounters`** in
  `sherdlock/manager_test.go`, wrapping the real Postgres-backed `Locker` to
  record per-token lock attempts and conflicts (`driver.ErrTokenAlreadyLocked`)
  without duplicating the manager wiring.
- **Baseline recorded** (3 replicas × 100 requests, real Postgres via
  testcontainers): 300 distinct tokens attempted, ~7500 total lock attempts,
  **~96% conflict rate**, no single token ID absorbing a large share of
  conflicts by itself — because `deleteTokensAndStoreChange` mints a fresh
  token ID each time the hot token is spent, so it is the *lineage* that
  stays hot, not one static ID. The only hard assertion is the functional
  invariant that must hold regardless of how contention is distributed:
  total demand exactly equals total wallet balance, so no error can be a
  genuine insufficient-funds — any error observed would be spurious,
  contention-induced. This baseline is the yardstick Phases 3-5 should
  improve on.

## Test plan

- [x] `go build ./...` and `cd cmd/tokendiag && go build ./...`
- [x] `make lint-auto-fix` — 0 issues across every module
- [x] `make checks` — passes (the one flagged vuln, `GO-2024-3218` in
      `go-libp2p-kad-dht`, is pre-existing and allow-listed)
- [x] `go test ./token/services/storage/db/...` — `ListLocks` covered on
      both sqlite and postgres via the shared `dbtest` suite
- [x] `go test ./token/services/selector/...` and `go test -race
      ./token/services/selector/sherdlock/...` — including the new
      `TestHotTokenContention` against real Postgres
- [x] Docs updated: `docs/README.md` tools table, `docs/development/`
      index, new `docs/development/tokendiag.md`, and the sherdlock metrics
      table in `docs/development/metrics.md`

Part of #2395


token/services/selector/sherdlock/selector.go
```diff
@@ -204,9 +211,21 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
 			if errors.Is(lockErr, token.SelectorRateLimited) {
 				return nil, nil, immediateRetries, lockErr
 			}
-			s.logger.DebugfContext(ctx, "Tried to lock token [%v], but it was already locked by another process", t)
+			attempted.Add(t.Id)
+			if errors.Is(lockErr, driver.ErrTokenAlreadyLocked) {
+				// Lost the race: someone else holds this token. This is the
+				// expected, common case under contention, not a DB error.
+				s.metrics.LockConflicts.Add(1)
+				s.logger.Infof("Lost lock race on token [%s:%d]: already locked by another process", t.Id.TxId, t.Id.Index)
```
Effi-S: nit: replacing `DebugfContext` with `Infof` can lose the trace. 
suggested fix: 
```go
if errors.Is(lockErr, driver.ErrTokenAlreadyLocked) {
    s.metrics.LockConflicts.Add(1)
    s.logger.DebugfContext(ctx, "Lost lock race on token [%s:%d]: already locked by another process", t.Id.TxId, t.Id.Index)
} else {
    s.logger.WarnfContext(ctx, "Failed to lock token [%s:%d]: %v", t.Id.TxId, t.Id.Index, lockErr)
}
```

token/services/selector/sherdlock/selector.go
```diff
@@ -204,9 +211,21 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
...
+			} else {
+				// A real store error (not a lock conflict) collapsed into the
+				// same !locked branch by TryLock. Log it distinctly so a DB
+				// outage does not read as ordinary contention.
```
Effi-S: This comment claims the split achieves more than it does.
A DB outage still reads as ordinary contention to every caller, only the log line differs. 

We can fix the behavior: on the non-conflict, non-rate-limit error, `return nil, nil, immediateRetries, errors.Wrapf(lockErr, "failed to lock token [%s]", t.Id)` so the real error propagates. 
OR, 
We can soften the comment.

Effi-S (approved):
Some low priority notes:
1. Run docstring over-claims a "hot-token ranking" section `cmd/tokendiag/cobra/locks/runner.go:50-57`

The docstring lists four report sections, including "a hot-token ranking, oldest lock first" as distinct from "every held lock … with its age". The implementation prints only one oldest-first list plus a summary; there is no separate ranking/aggregation. Either trim the docstring to match, or clarify that the oldest-first list is the ranking.

2. `Stores.db` field is dead state (`cmd/tokendiag/cobra/locks/stores.go:24`). `NewStores` stores the raw `*sql.DB` in `Stores.db`, but `Close()` only calls` s.TokenLock.Close()` (which closes the same handle via common2.Close), and nothing else ever reads db. Remove the field, as written it suggests to a reader that Close is incomplete.

3. Duplicated `attempted.Add(t.Id)` (`selector.go:214` and `:228`). Every token reaching TryLock is "attempted" regardless of outcome, so a single `attempted.Add(t.Id)` just after obtaining t (before the TryLock branch) is simpler and can't be forgotten if a third outcome branch is added later.

4. Nil-Status (LEFT JOIN) path is documented but untested. LockRecord.Status can be nil when a lock's consumer has no requests row; TestListLocks only covers Pending/Confirmed. A one-line case with an orphaned lock would lock in the documented behavior and the statusName→"unknown" / isTerminal(nil)==false handling.

5. `DistinctTokensAttempted` skips rate-limited attempts. The token.SelectorRateLimited branch returns before attempted.Add(t.Id) (selector.go:212), so a token denied by rate-limiting isn't counted as "attempted." Defensible (it wasn't really attempted), just worth a one-line comment so the histogram's semantics are unambiguous.

adecaro:
# Issue #2395 — sherdlock lock-contention fix: PR-by-PR analysis

Source data: CERT stress-test log `Liste_60_tokens_collisions_CERT_20260903.csv`
(60/1276 tokens with a `zkat_token_locks_pkey` collision; 5 tokens ≈ 0.39% of
the wallet's 1276 tokens absorbed 85.0% of all collisions; 8 tokens
(0.63%) absorbed 95.2%; token `f8a27fc4...` alone accounted for 1296
collisions (23.0%) spread from 14:06:18.515 to 14:13:00.453 — the "over six
minutes" in the report). Issue: [#2395](https://github.com/LFDT-Panurus/panurus/issues/2395).
PR stack (base → tip):

| PR | Phase | Title | Base |
|----|-------|-------|------|
| [#2397](https://github.com/LFDT-Panurus/panurus/pull/2397) | 1–2 | diagnostics + reproducible contention baseline | `main` |
| [#2398](https://github.com/LFDT-Panurus/panurus/pull/2398) | 3 | per-`Select`-call lock blacklist | #2397 |
| [#2399](https://github.com/LFDT-Panurus/panurus/pull/2399) | 4 | anti-join locked tokens + size-ordered candidates | #2398 |
| [#2400](https://github.com/LFDT-Panurus/panurus/pull/2400) | 5 | release locks on transaction settlement | #2399 |
| [#2402](https://github.com/LFDT-Panurus/panurus/pull/2402) | 6 | configurable Postgres lock-acquisition strategies | #2400 |

All five mechanisms named in the issue map 1:1 onto Phases 3–6:

1. Repeated re-offering of an already-lost-race token within one `Select` call → **Phase 3**
2. No size-based candidate ordering (small payment could draw the largest token) → **Phase 4b**
3. No lock-aware candidate query (a selector could still start a race it was bound to lose) → **Phase 4a**
4. No lock release on successful settlement (locks lived until the lease sweep) → **Phase 5**
5. Lost races surfacing as server-side unique-constraint errors (the log storm itself) → **Phase 6**

---

## PR #2397 (Phase 1–2) — diagnostics, no behavioral change

**Goal:** make the next four PRs falsifiable instead of guesswork. Landed with *zero* change
to selection or locking logic.

- `driver.TokenLockStore.ListLocks` — read API over held locks, joined with the consumer
  transaction's status, so "is this lock leaked" becomes a query instead of a manual
  Postgres/log correlation exercise. This directly answers the issue's ask #2 (a
  diagnostic tool) and its sub-question "has its availability status been correctly
  updated" for `f8a27fc4...`.
- `cmd/tokendiag locks` CLI — surfaces `ListLocks`, flags a lock whose consumer is already
  `Confirmed`/`Deleted`/`Orphan` as **leaked** — i.e. exactly the state the CERT report's
  6-minute token was suspected to be in.
- `sherdlock/metrics.go`: `LockConflicts` counter + `DistinctTokensAttempted` histogram,
  distinguishing "one hot token retried many times" from "many tokens each contended once" —
  the CSV's own Pareto shape (5 tokens = 85% of collisions) is exactly the first pattern.
- `selector.go` conflict log line promoted `Debug` → `Info`, tagged with the token ID, and
  both backends (Postgres store, in-memory locker) now map a lost race to the same
  `driver.ErrTokenAlreadyLocked` sentinel.
- **Baseline test**: `testutils.TestHotTokenContention` — few small tokens + one rotating
  large "hot" token, far more concurrent requests than tokens, against real Postgres.
  Baseline: **300 distinct tokens attempted, ~7500 lock attempts, ~96% conflict rate**, no
  single token ID dominating — because `deleteTokensAndStoreChange` mints a fresh output ID
  every time the hot token is spent, so the *lineage* is hot, not a static ID. Hard assertion:
  total demand == total wallet balance, so any observed selection error is provably spurious.

**Before/after:** none — this PR is instrumentation and a yardstick, not a fix.

## PR #2398 (Phase 3) — per-call lock blacklist

**Root cause fixed:** issue mechanism 1 — `selectInternal` had no memory of a lock it had
already lost within the same `Select` call, so a refetch could re-propose and re-lose the
same hot token repeatedly. This is the direct mechanism behind the CSV's `f8a27fc4...` row:
one token, 1296 collisions, a single first/last-occurrence window of ~6m42s.

**Change:** a per-call (not process-global) blacklist of tokens this `Select` invocation has
already lost a race on; skipped on later scans of the same call. Cleared if an entire scan
yields no non-blacklisted candidate, so a genuinely tight wallet doesn't get turned into a
permanent false `insufficient funds`. A token freed by another process is still reconsidered
on the *next* `Select` call.

**Measured (`TestHotTokenContention`, before → after):**

| metric | Phase 2 baseline | Phase 3 |
|---|---|---|
| total lock attempts | 7468 | 5096 (−32%) |
| total conflicts | 7168 | 4796 |
| conflict rate | 0.96 | 0.94 |
| distinct tokens attempted | 300 | 300 |
| max single-token conflict share | 0.04 | 0.06 |

Fewer wasted attempts within a call; aggregate conflict rate barely moves because this only
stops re-attempts *within* one call — the cross-call race for the same rotating hot token is
untouched. That's explicitly deferred to Phase 4.

## PR #2399 (Phase 4) — anti-join + size-ordered candidates

**Root causes fixed:** issue mechanisms 2 and 3.

- **4a anti-join:** the spendable-tokens query now excludes tokens already held by a lock,
  via a dialect-independent `NOT EXISTS` against `TokenLocks` (`cond.NotExists`, built on
  Phase 1's DSL primitive). Stops a selector from *starting* a race it's bound to lose,
  rather than losing it and recovering (Phase 3's job). Because the anti-join can hide every
  remaining candidate from a wallet that still has funds (they're just all locked elsewhere),
  a new lock-ignoring `TokenFetcher.HasAnySpendableTokens` disambiguates a genuinely-empty
  scan from a fully-locked-but-solvent one before reporting `SelectorInsufficientFunds` —
  this is precisely the issue's reported symptom ("insufficient funds" on a 1 EUR payment
  despite multiple tokens in the wallet).
- **4b size ordering:** candidates ordered ascending by amount, shuffled only within
  same-amount runs (`bucketedIterator.NewPermutation`). Directly answers the issue's Xavier
  test: 1 EUR request, 1 EUR + 200 EUR available → previously could draw the 200 EUR token
  (unordered); now the 1 EUR token sorts first. Deliberately *not* strict smallest-fit — that
  would relocate all contention onto the single smallest token; shuffling within a bucket
  keeps contention spread across equally-good candidates.
- Only `sherdlock` is touched; `simple` (unordered, no anti-join) is explicitly unchanged.

**Measured (`TestHotTokenContention`, Phase 3 → Phase 4):**

| metric | Phase 3 (blacklist only) | Phase 4 |
|---|---|---|
| total lock attempts | 4646 | 3414 (−26%) |
| total conflicts | 4346 | 3114 (−28%) |
| conflict rate | 0.94 | 0.91 |
| distinct tokens conflicted | 296 | 212 |
| max single-token conflict share | 0.07 | 0.10 |

First measurable drop in *distinct tokens contended at all* (296→212) — the anti-join is
keeping already-locked tokens out of candidate sets across calls, not just within one.

## PR #2400 (Phase 5) — release locks on settlement

**Root cause fixed:** issue mechanism 4 — the exact question the CERT report raised about
`f8a27fc4...`: "why does this token stay locked well after its transaction settled?" Answer,
confirmed by code inspection: **no production path unlocked on success.**
`Transaction.Release` (the only production `Unlock` caller) was wired solely to
`context.OnError`; `view.Context` has no success hook. A settled transaction's locks
therefore sat until the `leaseExpiry` sweep (minutes, by config) — during that entire window
the tokens were both spent-for *and* invisible to the Phase 4 anti-join, so they kept
colliding.

**Change:** both places that already compute terminal transaction status now call
`SelectorManager.Unlock` — `finality.Listener.runOnStatus` (live subscription) and
`TTXRecoveryHandler.applyFinalityLogic` (crash recovery) — for **both** `Confirmed` and
`Deleted` (a failed tx never spends its selected tokens either, so no reason to hold the
lock). Best-effort: a failed `Unlock` is logged, never fails settlement/recovery; the lease
sweep remains the backstop for `Orphan` (which the finality path never observes) and for any
failed release call. Wired at all three production sites (`ttx.Service`, `auditor.Service`,
Fabric recovery-manager).

**Measured:** *unchanged* — `TestHotTokenContention` calls `Select` and deletes/spends tokens
synchronously against the store, never routing through TTX finality, so it is structurally
blind to this mechanism. Verification is at the unit level instead: `Unlock` fires exactly
once per terminal transaction (`TestOnStatus_ReleasesLocksAfterTerminalStatus`,
`TestTTXRecoveryHandler_Recover_ReleasesLocksOnConfirmed/OnDeleted`), and a failing `Unlock`
doesn't fail the settlement/recovery path
(`TestTTXRecoveryHandler_Recover_LockReleaseErrorDoesNotFailRecovery`). This is a real,
important limitation of the benchmark harness — see gaps below.

## PR #2402 (Phase 6) — configurable Postgres lock-acquisition strategies

**Root cause fixed:** issue mechanism 5 — the collision *symptom itself*: every lost race
under the prior plain-`INSERT` acquisition surfaced as a server-side
`zkat_token_locks_pkey` unique-constraint violation, which is what actually filled the
Postgres logs the CERT report was built from.

**Change:** `token.storage.db.lockStrategy` (`insert` default / `onConflict` / `skipLocked`)
on the Postgres `TokenLockStore`; SQLite reads and ignores the key, `simple` never reaches
this config path — both untouched by construction.
- `onConflict`: `INSERT ... ON CONFLICT DO NOTHING RETURNING` — a lost race becomes a clean
  zero-row result, not a server error.
- `skipLocked`: same, plus `BatchLocker.LockBatch` — claims a whole covering window of
  candidates in one statement under `FOR UPDATE OF <tokens> SKIP LOCKED`, joined with the
  same `ON CONFLICT DO NOTHING` backstop.
- `HasEnoughSpendableTokens` lets the Phase 4 empty-anti-join fallback fail fast on a
  genuinely unpayable wallet instead of burning the full backoff budget first.

**Measured, precisely scoped (not inferred):**
- Aggregate conflict *rate* does **not** move across strategies — confirmed, not assumed:
  `SKIP LOCKED` only helps against a rival mid-claim at the exact same instant, not against
  an already-committed lock, which is the dominant conflict mode under sustained load.
- Server-side unique-constraint errors go to **zero**, but only on the single-token `Lock`
  path — `sherdlock` always uses `LockBatch` once a store implements `BatchLocker` (true for
  Postgres under every strategy), so under normal operation `insert`'s error-surfacing path
  is already dead code from the selector's own call site. `insert` only matters for a
  non-`BatchLocker` `Locker` (custom backend, or a rolling deploy with some replicas not yet
  upgraded). Isolated directly by forcing the single-token path
  (`TestHotTokenContention_SingleTokenLockPath`): `insert` → thousands of real
  unique-violations per run; `onConflict`/`skipLocked` → zero.
- Round-trip reduction comes from batching the claim into one statement per covering window —
  a property of `LockBatch` itself, uniform across all three strategies — not from strategy
  choice.
- `SKIP LOCKED` mechanism isolated deterministically against real Postgres
  (`TestTokenLockStore_LockBatch_SkipLocked_SkipsRowLockedByConcurrentTx`): a row a rival
  transaction is mid-claim on is skipped, not blocked; a plain `FOR UPDATE` on the same row
  genuinely blocks (control case).

---

## Baseline → tip, end to end

| CERT-observed symptom | Mechanism | Fixed in | Verified by |
|---|---|---|---|
| One token (`f8a27fc4...`) re-collided for >6 min, 1296 times | no per-call memory of a lost race | #2398 | lock-attempt count −32% on the fixed baseline harness |
| 1 EUR payment could draw a 200 EUR token instead of a 1 EUR one | no size-based candidate ordering | #2399 (4b) | doc/algorithm change; not independently benchmarked — see gaps |
| Selector started races against tokens it had no chance to win | no lock-aware candidate query | #2399 (4a) | distinct tokens *conflicted* 296→212 |
| Settled transaction's tokens stayed locked and collided for minutes after | no unlock on success, only `OnError` + lease sweep | #2400 | unit tests only — contention harness is structurally blind to this (see gaps) |
| Postgres logs flooded with `zkat_token_locks_pkey` violations | lost races surfaced as server errors | #2402 | zero unique-violations on `LockBatch` path (always used by `sherdlock`); single-token path isolated separately |

Net effect on the one benchmark that spans Phases 2–4 (`TestHotTokenContention`, 3 replicas ×
100 requests): **total lock attempts 7468 → 3414 (−54%), total conflicts 7168 → 3114 (−57%)**,
distinct tokens actually contended 300 → 212. The aggregate *conflict rate* stays high (0.91)
throughout — expected, since the workload is deliberately shaped to have more concurrent
demand than tokens; what moves is wasted work and, as of #2402, whether that residual
contention is expressible as a server error at all.

---

## Further testable gaps

1. **Phase 5 (lock release on settlement) has no contention-level regression test.**
   `TestHotTokenContention` deletes/spends tokens synchronously against the store and never
   goes through `finality.Listener`/TTX settlement, so it cannot detect a regression that
   reintroduces the "locked-until-lease-sweep" window. Worth adding a harness variant that
   drives real `ttx` settlement (or at least calls the finality listener) so a future change
   to `runOnStatus`/`applyFinalityLogic` is caught by the same yardstick used for Phases 3–4,
   not only by the four narrow unit tests.
2. **No end-to-end timing assertion for the 6-minute-lock scenario itself.** All current
   coverage is either unit-level (`Unlock` fires once) or aggregate-conflict-rate. Nothing
   asserts *lock age* — e.g. "no lock outlives its transaction's terminal status by more than
   N seconds" — which is the literal quantity the CERT report flagged (`ListLocks`'s "leaked"
   flag from Phase 1 could be reused as the assertion primitive here).
3. **Phase 4b's size-ordering claim ("1 EUR request no longer draws the 200 EUR token") is
   documented and code-reviewed but not covered by a dedicated regression test** reproducing
   the exact Xavier scenario (wallet with a small and a large token, small payment, assert the
   small token is selected). `TestHotTokenContention` exercises ordering only incidentally
   through its rotating-hot-token shape.
4. **`token.SelectorRateLimited` hard-abort path is unexercised by the contention suite.**
   `selectInternal` treats a lock failure wrapping this sentinel as a hard abort rather than a
   skip-and-continue; there's no test establishing this differs correctly from
   `ErrTokenAlreadyLocked` under actual contention (only that the sentinel exists).
5. **`simple` driver has no equivalent contention baseline.** All of `TestHotTokenContention`,
   the blacklist, the anti-join, and the lock-strategy tests are `sherdlock`-only by design
   (per the docs, `simple` is unordered, no anti-join, releases all locks between retries).
   There is currently no data on how badly `simple` degrades under the same CERT-shaped
   workload, so operators choosing between drivers have no comparable numbers — worth at
   least running `TestHotTokenContention`-equivalent against `simple` once, even without
   fixing it, purely as a documented baseline (mirrors what #2397 did for `sherdlock`).
6. **Rolling-deploy mixed-strategy scenario is untested.** Phase 6's docs explicitly flag that
   plain `insert`'s error-surfacing path only matters "for a rolling deploy where some
   replicas have not yet upgraded" — i.e. some replicas on `insert`, others on `skipLocked`
   against the same table concurrently. No test currently exercises that mixed-strategy
   window.
7. **`leaseExpiry: 0` / `leaseCleanupTickPeriod: 0` misconfiguration is documented as
   unsupported but not guarded.** The docs call out that setting either to zero silently
   disables the cleanup goroutine, permanently orphaning `Orphan`-consumer locks and any
   failed release-on-settlement call. Worth a config-validation test/startup warning rather
   than relying solely on documentation.
8. **Integration tests (fabtoken/dlog T1) are marked "not yet run" in PR #2399, #2400, and
   #2402's own test plans.** These are the closest thing to a true end-to-end reproduction of
   the CERT scenario (real Fabric network, real finality latency) and should be run — with
   `TEST_FILTER="T1"` — before merging the stack, per `AGENTS.md`'s own integration-test
   guidance.
9. **No metric/alert threshold tied to Phase 1's `LockConflicts` / `DistinctTokensAttempted`
   or Phase 6's `RoundTrips()`/`UniqueViolations()` counters.** They exist for diagnosis but
   there's no test or documented runbook asserting an operator would actually get paged before
   the next CERT-style incident reaches the same 85%-on-5-tokens shape.

AkramBitar:
## Review

### Summary
Good diagnostics foundation — the `ListLocks` API, `tokendiag` CLI, contention metrics, and baseline test are exactly the right building blocks for the later fix phases. Two blockers need addressing before merge.

---

### Blocker 1 — `ListLocks` uses INNER JOIN (`tokenlock.go`)

The join against `requests` on `consumer_tx_id = tx_id` drops any lock whose consumer transaction row does not exist yet — which includes every in-flight transfer and every leaked lock. The tool reports 0 locks against a full table. Needs a LEFT JOIN so that `Status == nil` is actually reachable as documented.

### Blocker 2 — `tokendiag` resolves table names without `TableNameParams` (`stores.go`)

`GetTableNamesWithConfig` produces names like `fsc_tkn_locks_<prefix>_<params>`. Without the params component the CLI queries a non-existent table and errors out on any real node deployment.

---

### Bug — `DistinctTokensAttempted` is recorded per retry, not per `Select()` call

The `attempted` set lives inside `selectInternal`, which is called once per retry iteration. The histogram records fan-out per retry, not per user-facing `Select()` call, so it does not measure what it claims. Fix: move the set to the `Select()` caller.

---

### Nit — `fmt.Errorf` in `cmd/tokendiag/cobra/locks/`

`config.go`, `runner.go`, and `stores.go` use `fmt.Errorf` to wrap errors. Project rule requires `github.com/hyperledger-labs/fabric-smart-client/pkg/utils/errors` instead (`errors.Wrapf`, `errors.Errorf`).

AkramBitar (changes requested):
@adecaro 
See my comments.
Thanks a lot,
Akram

AkramBitar:
What it fixes: under high concurrency, a small number of "hot" tokens absorb the vast majority of lock contention in the sherdlock selector, causing spurious insufficient-funds errors. The PR also surfaces a second bug: locks on settled transactions are never released, so they accumulate until a lease-age sweep.

What the fix is: Phases 1–2 only — no behavioral change to selection yet. The PR adds:

Diagnostics — ListLocks API on TokenLockStore, LockConflicts and DistinctTokensAttempted metrics, and a tokendiag locks CLI that connects directly to the database and flags leaked locks (held but consumer already settled).
Contention baseline — TestHotTokenContention reproduces the incident against real Postgres so phases 3–5 (blacklist, anti-join, lock release on settlement) have a measured yardstick to improve against.

AkramBitar (approved):
What we added on top of this PR:

1. **`cmd/tokendiag/cobra/locks/config.go`** — `CORE_*` overrides were silently ignored. `Unmarshal` walks `AllKeys()`, which never contains env-only keys, so `CORE_DATASOURCE` tripped the empty-dataSource check and `CORE_TABLEPREFIX` / `CORE_TABLENAMEPARAMS` resolved the wrong tables. Each overridable key is now bound up front.
2. **`token/services/storage/db/sql/common/tokenlock.go`** — `ListLocks` returned a truncated slice with a `nil` error on a mid-scan failure, because the shared `rowIterator` never consults `rows.Err()`. Now checked.
3. **`token/services/selector/sherdlock/selector.go`** — `DistinctTokensAttempted` was observed once per backoff round instead of once per `Select()`, inflating `_count` and understating fan-out. The accumulator now lives in the caller and is unioned across rounds.

Tests: `TestLoadConfigEnvOverrides`, `TestDistinctTokensAttemptedObservedOncePerSelect`.

AkramBitar:
@Effi-S 

Could you please have addition review round on latest fixes improvements? 

Regards,
Akram

PR #2398, opened 2026-09-21
fix(selector): sherdlock per-attempt lock blacklist (Phase 3 of #2395)

adecaro:
## Summary

Part of #2395. `selectInternal` had no memory of a lost lock race: after a
refetch, the same hot token could be re-proposed and re-lost repeatedly
within a single `Select` call — matching the incident's >6-minute
re-proposal loop on a single token.

This PR adds a per-call blacklist: every token this `Select` call has
already lost a lock race on is remembered and skipped on subsequent scans,
instead of being re-attempted and re-losing the same race. The blacklist is
scoped to the single `Select` call, not process-global, so a token
genuinely freed by another process is reconsidered on the caller's next
`Select` call. If an entire scan (since the last refetch) produces no
non-blacklisted candidate, the blacklist is cleared — otherwise a
genuinely-contended wallet with few tokens could turn a lost race into a
permanent false insufficient-funds.

This PR is stacked on #2397 (Phases 1-2: diagnostics + reproducible
contention baseline) and should be reviewed/merged after it.

## Before / after (Phase 2 baseline, `sherdlock.TestHotTokenContention`)

| metric | before (Phase 2 baseline) | after (this PR) |
|---|---|---|
| total lock attempts | 7468 | 5096 (-32%) |
| total conflicts | 7168 | 4796 |
| conflict rate | 0.96 | 0.94 |
| distinct tokens attempted | 300 | 300 |
| max single-token conflict share | 0.04 | 0.06 |

Total lock attempts drop substantially — far fewer wasted attempts against
tokens a call already knows it will lose. The aggregate conflict rate
barely moves because this fix only prevents re-attempts *within* a single
`Select` call, not across the many concurrent `Select` calls all racing for
the same rotating hot token; that cross-call contention is what Phase 4's
SQL anti-join and size-aware ordering address.

## Test plan
- [x] `go test ./token/services/selector/...` (full suite, incl. new
  `TestHotTokenContention`)
- [x] `go test -race ./token/services/selector/sherdlock/...`
- [x] `make checks` — pass (same pre-existing allow-listed `GO-2024-3218`)

Part of #2397


token/services/selector/sherdlock/selector.go
```diff
@@ -206,17 +227,26 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
 
 			immediateRetries++
```
AkramBitar, line 230: **[Medium — blocker]** A blacklist-only scan (every candidate was blacklisted, no `TryLock` called) still increments `immediateRetries`. On a wallet with a single hot token this means you burn 5 retries without ever winning a race — same DB refetch cost, half the chances. Fix: only increment `immediateRetries` when `sawNonBlacklistedCandidate` is true, or skip the increment on the blacklist-clear path.

token/services/selector/sherdlock/selector.go
```diff
@@ -206,17 +227,26 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
...
+			s.logger.DebugfContext(ctx, "Skipping blacklisted token [%v]: already lost a lock race on it this call", t.Id)
+			tokensLockedByOthersExist = true
 		} else if locked, lockErr := s.locker.TryLock(ctx, &t.Id, owner.ID()); !locked {
 			// A rate-limit denial from the locker is a hard stop: abort instead of retrying.
 			if errors.Is(lockErr, token.SelectorRateLimited) {
 				return nil, nil, immediateRetries, lockErr
 			}
 			attempted.Add(t.Id)
+			sawNonBlacklistedCandidate = true
 			if errors.Is(lockErr, driver.ErrTokenAlreadyLocked) {
```
AkramBitar: **[Medium]** `(false, nil)` from `TryLock` (contention without an error value) falls into the store-error `else` branch — logged as `Warnf` and never added to `blacklisted`. The blacklist only works when the locker wraps `driver.ErrTokenAlreadyLocked`. Document on the `Locker` interface that a lost race **must** return `driver.ErrTokenAlreadyLocked`, or treat `!locked && lockErr == nil` as a lost race here.

AkramBitar:
.

token/services/selector/sherdlock/selector.go
```diff
@@ -184,6 +199,12 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
 				)
 			}
 
+			if !sawNonBlacklistedCandidate && !blacklisted.Empty() {
+				s.logger.DebugfContext(ctx, "Blacklist excluded every candidate this scan; clearing it so freed tokens can be retried.")
+				blacklisted = collections.NewSet[token2.ID]()
```
AkramBitar, line 206: **Skip-only scans still consume the retry budget.**

When a scan blacklists every candidate, the next scan skips them all, clears the blacklist here, and still falls through to `immediateRetries++` (L228). So a scan that attempted no lock costs a retry.

Measured with a single always-contended token, same config both sides:

| | TryLock attempts | refetches | outcome |
|---|---|---|---|
| base (#2397) | 6 | 6 | `SelectorSufficientButLockedFunds` |
| this PR | 3 | 6 | same |

Same six DB refetch round-trips and the same give-up point, but half as many chances to acquire the token. For a wallet whose only viable token is the hot one that lowers the success rate per unit of DB work, and part of the "-32% total lock attempts" figure is this rather than eliminated waste.

Not incrementing `immediateRetries` when the scan attempted nothing would keep the six real attempts while still preventing the back-to-back re-attempt this PR is about.

token/services/selector/sherdlock/selector.go
```diff
@@ -184,6 +199,12 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
...
+			}
+			sawNonBlacklistedCandidate = false
+
 			if immediateRetries > maxImmediateRetries {
```
AkramBitar, line 210: No test asserts this boundary: that after `maxImmediateRetries` the selector returns `SelectorSufficientButLockedFunds` specifically. `MaxRetriesExceeded` (`selector_test.go:151`) stops at `require.Error`, with an outer budget of 1, so neither the error identity nor the 5-retry boundary is exercised.

Pre-existing, but worth pinning while the retry logic is being touched.

token/services/selector/sherdlock/selector.go
```diff
@@ -206,17 +227,26 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
...
 				return nil, nil, immediateRetries, lockErr
 			}
 			attempted.Add(t.Id)
+			sawNonBlacklistedCandidate = true
 			if errors.Is(lockErr, driver.ErrTokenAlreadyLocked) {
 				// Lost the race: someone else holds this token. This is the
 				// expected, common case under contention, not a DB error.
 				s.metrics.LockConflicts.Add(1)
 				s.logger.Infof("Lost lock race on token [%s:%d]: already locked by another process", t.Id.TxId, t.Id.Index)
+				blacklisted.Add(t.Id)
```
AkramBitar: This line is reached only if the Locker wraps `driver.ErrTokenAlreadyLocked`.

A Locker that signals contention as `(false, nil)` - which the `TryLock` signature permits, and which this package's own test mocks do - falls into the `else` below, logs a nil error via `Warnf`, and is never blacklisted, so the fix silently does not apply to it.

Worth documenting on the `Locker` interface that a lost race must wrap the sentinel, or treating `!locked && lockErr == nil` as a lost race.

AkramBitar (commented):
Reviewed by building and running the full selector suite on the PR head (passes), plus an instrumented head-vs-base comparison.

The mechanism is sound and not a no-op: `tokenlock.go:91` maps `UniqueKeyViolation` to `ErrTokenAlreadyLocked`, `locker.TryLock` passes it through untouched, and the inmemory locker normalizes to the same sentinel, so the blacklist really does populate on both backends. The clear-on-full-exclusion guard correctly avoids turning a lost race into a false insufficient-funds, and the loop stays bounded.

Three inline comments. Two findings have no diff line to anchor to:

**Test coverage.** This PR touches only `selector.go`, and neither new behaviour is pinned by a test: that a token lost this call is skipped on the next scan, and that the blacklist clears when it excludes every candidate. The nearest existing test does not reach the new code - `MaxRetriesExceeded` stubs `TryLockReturns(false, nil)`, so `lockErr` is nil, it routes to the store-error branch and never reaches `blacklisted.Add`. Separately, `TestHotTokenContention` states that Phases 3-5 should reduce the numbers it reports, but it only `t.Logf`s them, so the -32% in the description is enforced nowhere and can regress silently.

**Docs.** `docs/services/selector.md` documents exactly the semantics this changes - "a candidate already locked by another process is skipped and the loop moves on", the immediate-retry/backoff layer breakdown, and the sherdlock-vs-simple lock-retention contrast - and is not updated.

AkramBitar (changes requested):

adecaro:

PR #2399, opened 2026-09-21
fix(selector): anti-join locked tokens and order candidates by amount

adecaro:
## Summary

Stacked on #2398. Implements Phase 4 of #2395 — the remaining two contributing
mechanisms identified in the stress load-test incident:

- **4a — anti-join against locked tokens**: the spendable-tokens query now excludes
  tokens currently held by a lock, via a dialect-independent `NOT EXISTS` subquery
  against `TokenLocks` (`cond.NotExists`, added to the DSL). This stops a selector
  from *starting* a race it will lose; the `INSERT`-based lock acquisition remains
  the race-safe backstop, since the anti-join is read-then-act.
  - Because the anti-join can hide every remaining candidate from a wallet that
    still has funds (they're just all locked by someone else), an empty scan is
    disambiguated via a new lock-ignoring `TokenFetcher.HasAnySpendableTokens`
    check before reporting `SelectorInsufficientFunds`.
  - `TokenStore.GetSchema` now also creates the `TokenLocks` table itself
    (idempotently, alongside `TokenLockStore.GetSchema`), since the anti-join makes
    it a hard dependency of the token store, not just of the locker.
- **4b — size-aware, bucket-shuffled ordering**: candidates are ordered ascending by
  amount (new `ORDER BY`), then shuffled only within runs of equal amount
  (`bucketedIterator.NewPermutation`). This answers the incident's "why did a small
  request grab a large token instead of a same-size one" question, without
  introducing a new hot spot the way a strict smallest-fit rule would.

Only the `sherdlock` driver is affected; `simple` is unchanged (unordered, no
anti-join).

## Contention baseline (`TestHotTokenContention`)

Before Phase 4 (per-attempt blacklist only, PR #2398):
`distinct tokens attempted=300, total lock attempts=4646, total conflicts=4346, conflict rate=0.94, distinct tokens conflicted=296, max single-token conflict share=0.07`

After Phase 4 (anti-join + bucketed ordering active):
`distinct tokens attempted=300, total lock attempts=3414, total conflicts=3114, conflict rate=0.91, distinct tokens conflicted=212, max single-token conflict share=0.10`

Fewer distinct tokens are contended at all (212 vs 296), and total lock attempts/conflicts
drop ~26-28%, consistent with the anti-join keeping already-locked tokens out of the
candidate set entirely.

## Docs

Updated `docs/services/selector.md` (selection algorithm, diagram, driver differences
table) and `docs/development/sql-query-dsl.md` (`cond.NotExists`) to reflect the new
`sherdlock` behaviour.

## Test plan

- [x] `make checks` clean
- [x] `go test ./token/services/selector/... ./token/services/storage/db/sql/...` (incl. `-race`) — all pass
- [x] `TestHotTokenContention` baseline recaptured post anti-join wiring (above)

Part of #2395


token/services/selector/sherdlock/selector.go
```diff
@@ -190,13 +190,25 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
...
-					tokenType,
-					quantity.Decimal(),
-				)
+				// The candidate query excludes already-locked tokens (#2395,
+				// mechanism 3), so an empty scan that never saw a lock conflict is
+				// ambiguous: it may mean this wallet truly has no more funds, or
+				// that every remaining token is currently locked by someone else
+				// and was hidden from us entirely. Disambiguate with a direct,
+				// lock-ignoring existence check before giving up.
+				hasAny, hasAnyErr := s.fetcher.HasAnySpendableTokens(ctx, owner.ID(), tokenType)
```
AkramBitar, line 201: **`HasAnySpendableTokens` also ignores *our own* locks, so genuine insufficient funds no longer fast-fails.**

The check is lock-ignoring by design, but that includes the locks this very `Select` call just acquired — tokens we locked are still `spendable = true, is_deleted = false`.

Scenario: wallet holds two tokens worth 3 and 2, caller requests 10. Pass 1 locks both, `sum = 5 < 10`, the cache exhausts, `tokensLockedByOthersExist == false`. `HasAnySpendableTokens` sees our own two rows, returns `true`, and the `SelectorInsufficientFunds` return below is skipped. The refetch then returns nothing (the new `notLocked` anti-join hides the tokens *we* hold), so the loop burns all `maxImmediateRetries` — 6 extra spendable queries plus 6 extra `HasAnySpendableTokens` queries — and returns `token.SelectorSufficientButLockedFunds`.

Two consequences:
- For the plain `Selector` (backoff `< 0`) that is a *different sentinel* than before this PR, and `integration/token/fungible/views/transfer.go:293-295` branches on exactly those two sentinels.
- For `StubbornSelector` it additionally costs `numRetries+1 = 4` backoff rounds of up to `retryInterval = 5s` each (~10s p50) before finally reporting insufficient funds, with the misleading message "aborted too many times and no other process unlocked or added tokens".

So the most common user error — asking to spend more than you have — goes from milliseconds to seconds, and reports the wrong reason. Suggest gating the check on `selected.Empty()`, or excluding locks whose `consumer_tx_id` is this consumer tx.

token/services/storage/db/sql/common/tokens.go
```diff
@@ -391,18 +397,47 @@ func (it *dedupedTokenRowsIterator) Next() (*token.UnspentToken, error) {
...
+		Where(cond.And(
+			HasTokenDetails(driver.QueryTokenDetailsParams{
+				WalletID:           walletID,
+				TokenType:          typ,
+				Spendable:          driver.SpendableOnly,
+				LedgerTokenFormats: db.getSupportedTokenFormats(),
+			}, nil),
+			notLocked(tokenTable, tokenLocksTable),
+		)).
+		OrderBy(q.Asc(common3.FieldName("amount"))).
```
AkramBitar, line 415: **Ascending `ORDER BY amount` can deterministically exceed `limits.MaxInputs`.**

The docs note input count may increase, but not that it can now hard-fail. `token/driver/limits.go:60` sets `MaxInputs: 256`, enforced in `token/core/fabtoken/v1/actions/transfer.go:214` and `token/core/zkatdlog/nogh/v1/issue/action.go:237` with `ErrTooManyInputs`.

Scenario: a wallet holding 1000 tokens of 1 unit plus 5 tokens of 1000 units transfers 1000 units. Ascending order picks 1000 one-unit tokens and the action is rejected. Under the previous uniform shuffle the expected draw was ~10.9 units, so ~92 inputs — comfortably under the cap.

That turns a statistically-succeeding transfer into a deterministically-failing one for dust-heavy wallets. Worth capping the ascending preference, e.g. skipping to a larger bucket once the selected count approaches `MaxInputs`.

token/services/storage/db/sql/common/tokens.go
```diff
@@ -1518,6 +1591,20 @@ func (db *TokenStore) GetSchema() string {
...
+
+		-- TokenLocks: created here too (idempotently, alongside
+		-- TokenLockStore.GetSchema) because buildSpendableTokensIteratorByQuery's
+		-- notLocked anti-join (#2395, mechanism 3) makes this table a hard
+		-- dependency of the token store itself, not just of the locker.
+		CREATE TABLE IF NOT EXISTS %s (
+			tx_id TEXT NOT NULL,
+			idx INT NOT NULL,
+			consumer_tx_id TEXT NOT NULL,
+			created_at TIMESTAMPTZ NOT NULL,
```
AkramBitar, line 1603: **This DDL disagrees with `TokenLockStore.GetSchema` on `created_at`, so the effective column type depends on init order.**

Here: `created_at TIMESTAMPTZ NOT NULL`. In `token/services/storage/db/sql/common/tokenlock.go:97`: `created_at TIMESTAMP NOT NULL`. Both are `CREATE TABLE IF NOT EXISTS`, and `Driver.TokenLock` / `Driver.Token` are independent `lazy.Provider`s with no ordering guarantee — the sibling `newWalletStoreProvider` comment at `sqlite/driver.go:120` calls exactly this out for the Identity/Wallet FK. Whichever store initializes first wins.

On Postgres this changes lock expiry: `cond.OlderThan` renders `created_at < NOW() - INTERVAL 'N seconds'` (`postgres/conditions.go:24`), and `TIMESTAMP` is cast to `timestamptz` using the session `TimeZone` while the values were written as `time.Now().UTC()` — so on a non-UTC session stale locks expire off by the zone offset, whereas with `TIMESTAMPTZ` they expire correctly.

Also worth noting: `dbtest/tokenlock.go:28` creates the token store *first*, so CI now exercises `TIMESTAMPTZ` while a deployment that touches the lock store first gets `TIMESTAMP` — the tests no longer cover production's schema. Best to make the two statements byte-identical, or keep the DDL in one place.

token/services/storage/db/sql/common/tokens.go
```diff
@@ -391,18 +397,47 @@ func (it *dedupedTokenRowsIterator) Next() (*token.UnspentToken, error) {
...
+// concurrent selectors stop racing to lock a token they can already see is
+// held by someone else (#2395, mechanism 3). The row-level INSERT into
+// TokenLocks remains the race-safe backstop for the window between this read
+// and that INSERT; this anti-join only stops selectors from starting a race
+// they are very likely to lose.
+//
+// This is deliberately not folded into HasTokenDetails: balance and audit
+// queries need to see locked tokens too, only the spendable-tokens query
+// used by the selector should exclude them.
+func notLocked(tokenTable, tokenLocksTable common3.Table) cond.Condition {
```
AkramBitar, line 429: **The anti-join silently no-ops when `tokendb` and `tokenlockdb` resolve to different persistences.**

`notLocked` joins `db.table.TokenLocks` inside the token store's own database and table-name space, but the two stores are configured through independent keys — `tokendb.persistence` (`token/services/storage/tokendb/store.go:28`) and `tokenlockdb.persistence` (`token/services/storage/tokenlockdb/store.go:18`) — each resolving its own `fscSqlite.Config`/Postgres DSN and its own `TablePrefix` via `GetTableNamesWithOverrides`.

If an operator points them at different databases, or the same database with different prefixes, or one at `memory` and one at `postgres`, then `TokenStore.GetSchema` happily creates a *second, permanently empty* `TokenLocks` table in the token DB, the `NOT EXISTS` matches nothing, and the whole mitigation disappears with no error, warning or log line — while `HasAnySpendableTokens` still fires on every exhausted scan. This new hard dependency needs either a startup validation that the two share a persistence, or an explicit documented constraint.

token/services/storage/db/sql/common/tokens.go
```diff
@@ -391,18 +397,47 @@ func (it *dedupedTokenRowsIterator) Next() (*token.UnspentToken, error) {
 // can compare the dynamic path against a prepared-once path using identical
```
AkramBitar, line 397: **No index supports the new sort on the hottest query in the service.**

`buildSpendableTokensIteratorByQuery` is what the cached fetcher runs across *all* wallets and types (`fetcher.go:376`, `SpendableTokensIteratorBy(ctx, "", "")`) at `defaultCacheFreshnessInterval = 1 * time.Second` / every 5 queries. The tokens schema (`tokens.go:1421-1426`) has `idx_owner_wallet_part (owner_wallet_id, token_type) WHERE is_deleted = false AND owner = true` but nothing covering `amount`, so Postgres must add a full sort of the matched set on every refresh and every lazy fetch, on top of the new correlated `NOT EXISTS`.

Adding `amount` to the partial index would avoid it — or drop the global `ORDER BY` and sort the grouped slices in Go, since `groupTokensByKey` already materializes them.

token/services/selector/sherdlock/fetcher.go
```diff
@@ -183,15 +191,84 @@ func (f *lazyFetcher) UnspentTokensIteratorBy(ctx context.Context, walletID stri
...
+// ascending by amount overall, but with each run of equal-Quantity tokens
+// independently shuffled, so equally-good candidates are still randomized
+// against each other while the size ordering across runs survives.
+func (b *bucketedIterator) NewPermutation() iterators.Iterator[*token2.UnspentTokenInWallet] {
+	shuffled := make([]*token2.UnspentTokenInWallet, len(b.items))
+	copy(shuffled, b.items)
+
+	for start := 0; start < len(shuffled); {
+		end := start + 1
+		for end < len(shuffled) && shuffled[end].Quantity == shuffled[start].Quantity {
```
AkramBitar, line 261: **Buckets are detected on the `Quantity` string while the ordering comes from the `amount` column.**

This compares the `quantity TEXT` column, but the run structure it assumes was produced by `ORDER BY amount` on a separate `NUMERIC(78,0)` column. The two are written independently — `token/services/tokens/storage.go:208-210`: `Quantity: tta.Tok.Quantity` verbatim from the action, `Amount: q.ToBigInt().Uint64()` derived.

Any non-canonical spelling difference between two equal-valued tokens (e.g. `0x0a` vs `0xa`, or a decimal vs hex encoding from a different driver/peer version) splits one amount bucket into singleton buckets that are never shuffled against each other, restoring exactly the deterministic hot-spot the bucketing exists to prevent — and silently, since the ordering still looks correct. Comparing the numeric amount (or carrying it on `UnspentTokenInWallet`) would make the bucket boundary agree with the `ORDER BY` by construction.

token/services/selector/sherdlock/fetcher.go
```diff
@@ -183,15 +191,84 @@ func (f *lazyFetcher) UnspentTokensIteratorBy(ctx context.Context, walletID stri
...
+func newBucketedIterator(items []*token2.UnspentTokenInWallet) *bucketedIterator {
+	return &bucketedIterator{items: items}
+}
+
+// Next returns items in the order they were stored, and (nil, nil) once
+// exhausted: sherdlock.Iterator's contract (see selectInternal's t == nil
+// refetch branch) signals exhaustion with a nil element and nil error, not
+// io.EOF — returning io.EOF here made every lazy-fetch refetch cycle look
+// like a hard failure instead of "cache exhausted, fetch more" (#2395).
+func (b *bucketedIterator) Next() (*token2.UnspentTokenInWallet, error) {
```
AkramBitar, line 239: **Nit: this Godoc justifies itself with a bug that never existed on this path.**

The comment says returning `io.EOF` "made every lazy-fetch refetch cycle look like a hard failure instead of 'cache exhausted, fetch more' (#2395)". The code being replaced was `collections.NewPermutatedIterator` → FSC `iterators.Permutate` → `iterators.Slice`, whose `Next` returns `(zero, nil)` — i.e. `(nil, nil)` for a pointer element — at exhaustion (`fabric-smart-client@v0.16.0/.../iterators/slice.go:43-46`).

The `(nil, nil)` contract is right; only the stated rationale is wrong, and it will mislead the next reader into thinking a regression was fixed here.

token/services/selector/sherdlock/fetcher_test.go
```diff
@@ -44,6 +44,12 @@ func (m *mockTokenDB) SpendableTokensIteratorBy(ctx context.Context, walletID st
 	return args.Get(0).(driver.SpendableTokensIterator), args.Error(1)
 }
 
+func (m *mockTokenDB) HasAnySpendableTokens(ctx context.Context, walletID string, typ token2.Type) (bool, error) {
```
AkramBitar, line 47: **None of the new behaviour has a test.**

The only test-file changes in this PR are mock/stub methods to satisfy the widened interfaces. Nothing asserts that:
- the emitted SQL contains the `NOT EXISTS` clause or the `ORDER BY amount` — `TestSpendableTokensIteratorByPreparedReuse` only compares the dynamic and prepared paths *to each other*, so it would pass identically if either clause were dropped;
- a locked token is actually excluded end-to-end;
- `bucketedIterator.NewPermutation` preserves cross-bucket ordering while permuting within a bucket;
- `HasAnySpendableTokens` ignores locks.

Per AGENTS.md's testing conventions each of these is a cheap table-driven or SQL-text assertion, and the `HasAnySpendableTokens` / `MaxInputs` / `Quantity`-bucketing issues flagged elsewhere in this review are all the kind of regression such a test would have caught.

AkramBitar (commented):
Review of the anti-join + amount-ordering change. I diffed against the PR's actual base (`fix/2395-sherdlock-per-attempt-blacklist`, since this is stacked on #2398), 14 files.

**Bottom line:** one blocker, one behaviour change that needs a decision, two things to settle before this reaches Postgres, and no test coverage for any of the new behaviour.

- **Blocker** — `HasAnySpendableTokens` counts the locks this same `Select` call just took, so a genuine over-spend now returns `SelectorSufficientButLockedFunds` after burning every retry (~10s on `StubbornSelector`) instead of fast-failing with `SelectorInsufficientFunds`. Wrong sentinel and wrong latency on the most common user error; `integration/token/fungible/views/transfer.go:293-295` branches on those sentinels.
- **Needs a decision** — ascending `ORDER BY amount` makes dust-heavy wallets deterministically exceed `MaxInputs: 256`, so transfers that previously succeeded will now fail hard.
- **Before Postgres** — the two `TokenLocks` DDL copies disagree on `created_at` (`TIMESTAMPTZ` vs `TIMESTAMP`) with no init ordering guarantee, and the anti-join silently does nothing if `tokendb` and `tokenlockdb` are configured onto different persistences.
- **Performance / correctness nits** — no index covers the new sort on the service's hottest query; bucket boundaries compare the `quantity` string while the ordering comes from the `amount` column.

Details inline.

adecaro:

PR #2400, opened 2026-09-21
fix(selector): release token locks on transaction settlement

adecaro:
## Summary

Stacked on #2399. Implements Phase 5 (final phase) of #2395 — the fourth
contributing mechanism identified in the stress load-test incident: no
production code path released a transaction's selection locks on success.

`Transaction.Release` (the only production `Unlock` caller) was registered
solely against `context.OnError` — `view.Context` exposes no success hook. So
a settled transaction's locks sat until the `sherdlock` lease-expiry sweep
(several minutes, by config), during which its tokens stayed invisible to the
Phase 4 anti-join and kept colliding.

- Both places that already compute terminal transaction status now release
  locks once that status is known: `finality.Listener.runOnStatus` (the live
  finality subscription) and `TTXRecoveryHandler.applyFinalityLogic`
  (recovery on restart) — for **both `Confirmed` and `Deleted`**, since a
  failed transaction will never spend the tokens it selected either.
- A new `finality.SelectorManagerProvider` resolves the `token.SelectorManager`
  bound to a fixed TMS, re-resolving it on every call rather than caching it —
  mirroring the existing `TokenRequestHasher.ProcessTokenRequest` pattern.
- Release is best-effort: a failed `Unlock` is logged and does not fail the
  settlement/recovery path (mirrors `Transaction.Release`'s existing style);
  the lease-expiry sweep remains the backstop for `Orphan` consumers (which the
  finality path never observes) and for any release call that failed.
- Wired at all three production construction sites: `ttx.Service.Append`,
  `auditor.Service.Append` (a harmless no-op unlock for the auditor role, which
  never itself locks tokens), and the Fabric recovery-manager wiring.

## Why the contention benchmark shows no change here

`TestHotTokenContention` (the harness used to baseline Phases 3-4) is
structurally blind to this mechanism: it calls `Select` and then immediately
deletes/spends the tokens synchronously against the store, never routing
through the TTX finality settlement path where this PR's `Unlock` hook lives.
Re-running it before/after this change on the same environment gives identical
numbers:

```
distinct tokens attempted=300, total lock attempts=3353, total conflicts=3053,
conflict rate=0.91, distinct tokens conflicted=202, max single-token conflict share=0.10
```

This is expected, not a regression — mechanism 4 is a settlement-timing bug
(locks outliving a *settled* transaction by minutes), not a selection-time
contention bug, so it needs a harness that actually waits on finality to
observe. The correct verification for this PR is at the unit level: new tests
assert `Unlock` fires exactly once per transaction, for both `Confirmed` and
`Deleted`, and that a failing `Unlock` does not fail settlement/recovery
(`TestOnStatus_ReleasesLocksAfterTerminalStatus`,
`TestTTXRecoveryHandler_Recover_ReleasesLocksOnConfirmed`,
`TestTTXRecoveryHandler_Recover_ReleasesLocksOnDeleted`,
`TestTTXRecoveryHandler_Recover_LockReleaseErrorDoesNotFailRecovery`).

## Docs

Updated `docs/services/selector.md` — new "Release on settlement" section, and
reframed "Lease expiry" as the backstop it now is rather than the sole release
mechanism.

## Test plan

- [x] `make checks` clean
- [x] `go test ./token/services/ttx/... ./token/services/auditor/... ./token/services/network/fabric/...` (incl. `-race`) — all pass
- [x] New unit regression tests for lock release on Confirmed/Deleted, and for release-error not failing settlement

Fixes #2395


token/services/ttx/finality/listener.go
```diff
@@ -169,11 +183,30 @@ func (t *Listener) runOnStatus(ctx context.Context, txID string, status int, mes
 	} else {
 		t.metrics.DeletedTransactions.Add(1)
 	}
+	releaseLocks(ctx, t.logger, t.selectorManagerProvider, txID)
```
AkramBitar, line 199: Release runs *after* `Commit`, which ends in `NotifyStatus`. The woken client can return from `Finality` and start its next `Select` while these lock rows still exist, so the Phase-4 anti-join keeps hiding the tokens — the contention this PR removes.

Release before `NotifyStatus`, or `defer` it at the top of `runOnStatus`.

token/services/ttx/finality/listener.go
```diff
@@ -169,11 +183,30 @@ func (t *Listener) runOnStatus(ctx context.Context, txID string, status int, mes
...
+	sm, err := sp.SelectorManager()
+	if err != nil {
+		logger.WarnfContext(ctx, "failed to get selector manager to release locks for tx [%s]: [%s]", txID, err)
+
+		return
+	}
+	if sm == nil {
+		return
+	}
+	if err := sm.Unlock(ctx, txID); err != nil {
```
AkramBitar, line 218: `ctx` is the caller's cancellable context. `recovery.Manager.callHandler` abandons this goroutine when `TransactionTimeout` fires, but `tx.Commit()` takes no ctx and `NotifyStatus` uses `WithoutCancel` — so execution reaches here with a canceled ctx, the `DELETE` fails instantly, and the lock waits out the full `leaseExpiry`.

Use `context.WithoutCancel(ctx)`, as `Transaction.Release` already does ("we need to unlock even if t.Context is canceled").

AkramBitar (commented):
Logic is correct and tests/vet are clean. Two issues undercut the fix, plus the self-flagged `[ ]` integration tests.

AkramBitar (changes requested):

adecaro:

PR #2402, opened 2026-09-22
feat(selector): configurable Postgres lock-acquisition strategies (Phase 6, #2395)

adecaro:
## Summary

Stacked on #2400. Phase 6 of #2395 — attacks the lock-acquisition
*mechanism* on top of the 4-PR correctness stack. Every replica still reads
the same anti-join snapshot, picks the same smallest candidate, and races a
plain `INSERT`, surfacing a lost race as a server-side unique-constraint
violation (the log storm the CERT report captured).

- `token.storage.db.lockStrategy` selects `insert` (default, unchanged) /
  `onConflict` / `skipLocked` on the Postgres `TokenLockStore`. SQLite reads
  and ignores the key; `simple` never reaches this config path, so both are
  untouched by construction — **`simple` is not modified anywhere in this
  PR.**
- `onConflict` and `skipLocked` replace the plain `INSERT` with
  `INSERT ... ON CONFLICT DO NOTHING RETURNING`: a lost race becomes a clean
  zero-row result, not a server error.
- `skipLocked` additionally implements `BatchLocker.LockBatch`: a covering
  window of candidates is claimed in one statement against the `Tokens` rows
  under `FOR UPDATE OF <tokens> SKIP LOCKED`, so a claimant walks past a row
  a concurrent claimant is already mid-claim on. `selectInternal`
  type-asserts for `BatchLocker` and claims the whole window in one round
  trip when available, falling back to the one-at-a-time `Lock` path
  otherwise.
- `HasEnoughSpendableTokens` (`common/tokens.go`) lets the empty-anti-join
  fallback fail immediately on a genuinely unpayable wallet instead of
  burning the full backoff budget first.

## What actually moves, verified against real Postgres — not inferred

The literal lock-*conflict rate* does **not** improve across strategies
(`TestHotTokenContention`/`TestHotTokenContentionWideWindow`, all three
strategies, same numbers). This is expected, not a regression:
`FOR UPDATE SKIP LOCKED` only helps against a rival mid-claim at the exact
same instant, not against an already-committed lock — the dominant conflict
mode under load.

What is real and measured, via new round-trip/unique-violation
instrumentation (`RoundTrips()`/`UniqueViolations()` on `TokenLockStore`):

1. **Server-side unique-constraint errors are eliminated, scoped to the
   single-token `Lock` path.** `sherdlock`'s selector always uses
   `LockBatch` once a store implements `BatchLocker` (true for Postgres
   under every strategy), so plain `insert`'s error-surfacing behavior is
   dead code from the selector's own call path and only matters for a
   non-`BatchLocker` `Locker`.
   `TestHotTokenContention_SingleTokenLockPath` forces the single-token path
   directly (via a `singleTokenOnlyLocker` adapter that hides `LockBatch`
   through Go's interface-embedding rules) and confirms it: `insert`
   produces thousands of real unique-constraint violations in one run;
   `onConflict`/`skipLocked` produce exactly zero.
2. **Round-trip reduction comes from batching the claim into one statement
   per covering window, uniform across all three strategies once
   `LockBatch`'s per-token loop was fixed — not from strategy choice.**
3. `TestTokenLockStore_LockBatch_SkipLocked_SkipsRowLockedByConcurrentTx`
   proves the `SKIP LOCKED` mechanism directly against real Postgres: a row
   a concurrent transaction is mid-claim on is skipped, not blocked or
   collided with, while a plain `FOR UPDATE` on the same row genuinely
   blocks (control case).

## Docs

`docs/services/selector.md` gets a new "Lock-acquisition strategies"
section, scoping the unique-violation-elimination benefit to the
single-token `Lock` path and the round-trip saving to batching itself, plus
the `token.storage.db.lockStrategy` config key in the Configuration example.

## Test plan

- [x] `make checks` clean (only the pre-existing allow-listed libp2p
      govulncheck finding, unrelated)
- [x] Full `sherdlock` and `postgres` package suites pass against real
      Postgres (testcontainers)
- [x] `TestHotTokenContention`/`TestHotTokenContentionWideWindow` extended
      across all three strategies
- [x] New: `TestHotTokenContention_SingleTokenLockPath`,
      `TestTokenLockStore_LockBatch_SkipLocked_SkipsRowLockedByConcurrentTx`

Fixes #2395


token/services/selector/sherdlock/selector.go
```diff
@@ -374,12 +470,34 @@ func (l *locker) UnlockAll(ctx context.Context) error {
...
+
+	return won, err
+}
+
 func NewSherdSelector(txID transaction.ID, fetcher TokenFetcher, lockDB Locker, precision uint64, backoff time.Duration, maxRetriesAfterBackoff int, m *Metrics) TokenSelectorUnlocker {
 	logger := logger.Named("selector-" + txID)
-	locker := &locker{txID: txID, Locker: lockDB}
+	base := &locker{txID: txID, Locker: lockDB}
+	var tokenLocker TokenLocker = base
+	if bl, ok := lockDB.(BatchLocker); ok {
```
AkramBitar, line 495: **Blocker: this assertion can never succeed for the `Locker` the SDK actually passes, so the whole batch path is dead in production.**

`LockBatch` is declared only on `*postgres.TokenLockStore`. It is *not* a method on `driver.TokenLockStore` (`token/services/storage/db/driver/token.go:377`). The `Locker` reaching `NewSherdSelector` is `*tokenlockdb.StoreService` (`sherdlock/service.go:139`), and that type is `struct{ driver.TokenLockStore }` — it embeds the *interface*, so only the interface's method set is promoted. `lockDB.(BatchLocker)` therefore fails on a real node, `batchLocker` is never constructed, `supportsBatch` at line 193 is always false, and every replica stays on the one-token-at-a-time `TryLock` path.

I confirmed this by compiling an assertion against the real types: raw `*postgres.TokenLockStore` satisfies `BatchLocker`; `*tokenlockdb.StoreService` does not.

This is the same interface-embedding rule the PR description invokes deliberately for `singleTokenOnlyLocker` — production hits it by accident. So the test written to model the *fallback* path is actually modelling production, and the batch numbers in the description describe a path no deployed node takes: the round-trip reduction, `skipLocked`'s only real mechanism, and "unique violations eliminated" are all inert outside the tests.

Fix by adding `LockBatch` to `driver.TokenLockStore`, or by having `StoreService` forward it explicitly.

token/services/storage/db/sql/common/tokens.go
```diff
@@ -492,6 +492,54 @@ func (db *TokenStore) HasAnySpendableTokens(ctx context.Context, walletID string
 	return has, nil
 }
 
+// buildHasEnoughSpendableTokensQuery builds the SQL query and args for
+// HasEnoughSpendableTokens without executing it: the same spendable, lock-ignoring
+// predicate as buildHasAnySpendableTokensQuery, but summing amount instead of capping at
+// one row, so the wallet's total is computed in SQL rather than by fetching every token.
+func buildHasEnoughSpendableTokensQuery(db *TokenStore, walletID string, typ token.Type) (string, []any) {
+	return q.Select().FieldsByName("SUM(amount)").
```
AkramBitar, line 500: **Blocker: `SUM(amount)` over a `NUMERIC(78,0)` column is not portable, and this query fails on SQLite.**

Verified against the repo's own `modernc.org/sqlite` driver, two independent failure modes:

1. Two amounts that each fit in int64 but whose *total* does not → the query fails with `SQL logic error: integer overflow (1)`.
2. Any single amount above int64 (reachable at the default 64-bit precision) is stored with REAL affinity, so `SUM` returns `float64` and `BigInt.Scan` fails with `cannot scan type float64 into BigInt`.

Either error propagates out of `selectInternal` (`sherdlock/selector.go:216`) as `failed to check for locked tokens for [...]`. So a wallet that previously got a clean `SelectorInsufficientFunds` — or a retry — now gets a hard selection error instead. The `HasAnySpendableTokens` this replaces used `SELECT 1 ... LIMIT 1` and was immune by construction.

There is also no store-level test for the new SQL on either backend; the selector-side coverage is all mocks, which is why this passes CI. A `common`-level test that runs on SQLite and Postgres would have caught it.

token/services/selector/sherdlock/selector.go
```diff
@@ -248,34 +261,117 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
...
+			window := []*token2.UnspentTokenInWallet{t}
+			windowSum, err := token2.ToQuantity(t.Quantity, s.precision)
+			if err != nil {
+				return nil, nil, immediateRetries, errors.Wrapf(err, "invalid token [%s] found", t.Id)
+			}
+			remaining, err := quantity.Sub(sum)
+			if err != nil {
+				return nil, nil, immediateRetries, errors.Wrapf(err, "failed to compute remaining amount")
+			}
+			for windowSum.Cmp(remaining) < 0 {
```
AkramBitar, line 279: **Blocker: the covering window grows without a cap, so a dust-heavy wallet exceeds Postgres's bind-parameter limit and then cannot recover.**

This loop pulls candidates until `windowSum >= remaining` with no maximum window size. Downstream:

- `tryInsertOnConflict` binds 4 parameters per candidate (`consumer_tx_id, tx_id, idx, created_at`) → fails above ~16,383 candidates with `extended protocol limited to 65535 parameters`.
- `tryLockSkipLocked` binds 2 per candidate plus 2 → fails above ~32,766.

A wallet of many small tokens paying a large amount hits this deterministically, and the recovery path makes it worse: the error isn't per-token contention, so the batch-error branch marks the *whole* window `attempted`, refetches, rebuilds the same oversized window, and fails identically on every scan until `maxRetriesAfterBackoff` is exhausted. No progress is possible.

A max-window-size cap (claim in chunks, then continue accumulating) fixes both the hard failure and the retry livelock.

token/services/storage/db/sql/common/config.go
```diff
@@ -87,8 +118,24 @@ func LoadStorageConfig(cfg driver2.Config) (StorageConfig, error) {
...
 
+	lockStrategy := LockStrategyInsert
+	if cfg != nil && cfg.IsSet(ConfigKeyLockStrategy) {
+		if err := cfg.UnmarshalKey(ConfigKeyLockStrategy, &lockStrategy); err != nil {
+			return StorageConfig{}, err
+		}
+		switch lockStrategy {
+		case LockStrategyInsert, LockStrategyOnConflict, LockStrategySkipLocked:
+		default:
+			return StorageConfig{}, errors.Errorf(
```
AkramBitar, line 129: **An invalid `lockStrategy` silently discards every table-name override.**

Returning `StorageConfig{}` alongside the error is only safe if callers stop. Neither does: `postgres/driver.go:61-65` and `sqlite/driver.go:56-60` both `logger.Warnf(...)` and then carry on using `storageConfig.TableNames`, which at that point is the zero `TableNamesConfig`.

So a single typo — `skiplocked`, `skip_locked` — makes the node fall back to default short table codes *and* lose `skipPrefix`, i.e. it opens a completely different set of tables. It comes up looking at an empty database, with one warning line as the only explanation.

Either fail construction on an unrecognised value, or validate the strategy without clobbering the rest of the config (return the populated struct alongside the error).

token/services/selector/sherdlock/selector.go
```diff
@@ -198,11 +202,20 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
...
+				if remainingErr != nil {
+					return nil, nil, immediateRetries, errors.Wrapf(remainingErr, "failed to compute remaining amount for [%s:%s]", owner.ID(), tokenType)
 				}
-				if !hasAny {
+				// HasEnoughSpendableTokens is the sum-aware counterpart of HasAnySpendableTokens:
+				// a wallet whose total spendable balance cannot cover the remaining amount can
+				// never satisfy this Select call no matter how the rest gets unlocked, so fail
+				// immediately instead of spending the immediate-retry/backoff budget on a request
+				// that can never succeed.
+				hasEnough, hasEnoughErr := s.fetcher.HasEnoughSpendableTokens(ctx, owner.ID(), tokenType, remaining.ToBigInt())
```
AkramBitar, line 214: **The fast-fail threshold double-counts tokens this call has already locked, so it mostly doesn't fire.**

`HasEnoughSpendableTokens` sums *all* spendable tokens ignoring locks — including the ones this very `Select` already locked and folded into `sum` — but it's compared against `remaining = quantity - sum` (line 205). The effective predicate is `total >= quantity - sum`, not `total >= quantity`.

Concretely: a wallet holds two CHF10 tokens (total 20) and 30 is requested. The first scan locks both (`sum=20`); after the refetch the anti-join hides them, so the scan is empty, `remaining=10`, and `20 >= 10` reports "enough". The request then burns the full immediate-retry plus backoff budget on a payment that can never succeed — exactly what this check was added to prevent. The benefit only materialises when `quantity` exceeds roughly twice the balance.

The argument should be `quantity.ToBigInt()`. It errs safe as written (no false insufficient-funds), so this is correctness-of-intent rather than a wrong answer.

Separately: with this call site converted, `HasAnySpendableTokens` is now dead at every selector call site — worth removing or noting why it stays.

token/services/selector/sherdlock/selector.go
```diff
@@ -248,34 +261,117 @@ func (s *Selector) selectInternal(ctx context.Context, owner token.OwnerFilter,
...
+			}
+			for windowSum.Cmp(remaining) < 0 {
+				next, nextErr := s.next()
+				if nextErr != nil {
+					return nil, nil, immediateRetries, errors.Wrapf(nextErr, "failed to get tokens for [%s:%s]", owner.ID(), tokenType)
+				}
+				if next == nil {
+					break
+				}
+				if blacklisted.Contains(next.Id) {
```
AkramBitar, line 287: **This `continue` should set `tokensLockedByOthersExist`, like the outer loop's blacklist branch does.**

The outer loop's equivalent (line 263) sets `tokensLockedByOthersExist = true` when it skips a blacklisted candidate; this inner one just consumes it and moves on. A post-refetch scan whose *only* blacklisted candidates were swallowed here therefore ends with the flag false and takes the "genuinely insufficient funds" branch at line 198 — a wrong `SelectorInsufficientFunds` where the caller should have kept retrying.

Right now the lock-ignoring sum from the `HasEnoughSpendableTokens` finding is the only thing preventing that, which is a fragile coupling between two separate bugs. The two blacklist branches should agree independently.

(For the record, `sawNonBlacklistedCandidate` is *not* affected: reaching this branch requires a non-blacklisted `t`, and every non-returning path below sets the flag, so the blacklist cannot clear incorrectly.)

token/services/storage/db/sql/postgres/tokenlock.go
```diff
@@ -70,9 +100,159 @@ func NewTokenLockStore(dbs *common2.RWDB, tableNames common5.TableNames) (*Token
...
+
+	// #nosec G202 -- db.Table.Tokens/TokenLocks are trusted table names derived from
+	// this process's own config at construction time, never from request input; the
+	// only per-request values (tokenIDs, consumerTxID, createdAt) are passed as
+	// placeholders in args, never concatenated into the query text.
+	query := "WITH candidates(tx_id, idx) AS (VALUES " + strings.Join(values, ", ") + "), " +
+		"claimed AS (" +
+		"SELECT t.tx_id, t.idx FROM " + db.Table.Tokens + " t " +
+		"JOIN candidates c ON c.tx_id = t.tx_id AND c.idx = t.idx " +
+		"FOR UPDATE OF t SKIP LOCKED" +
```
AkramBitar, line 230: **`SKIP LOCKED` here also skips rows locked for reasons unrelated to claiming, and the selector reports those as lost lock races.**

`FOR UPDATE OF t SKIP LOCKED` skips a `Tokens` row locked by *any* concurrent transaction. That includes the `FOR KEY SHARE` lock another replica's lock `INSERT` takes via `FOREIGN KEY (tx_id, idx) REFERENCES <Tokens>`, and a `TokenTransaction` mid-update on that row. The caller can't distinguish these from a genuine conflict: the token is simply absent from `won`, so `selectInternal` increments `LockConflicts`, blacklists it for the rest of the call (`selector.go:295-298`), and can end up returning `SelectorSufficientButLockedFunds` for tokens that have no lock row at all.

Worth stating in the strategy's doc comment alongside the caveats already there.

Second, smaller point: the batch path does `LockConflicts.Add(1)` per non-won token in the window, which inflates the counter by roughly the window size relative to the single-token path. The metric isn't comparable across an upgrade, or between strategies — which matters given this PR uses those numbers as evidence.

docs/services/selector.md
```diff
@@ -202,6 +202,65 @@ lock; SQLite is non-distributed and always runs it locally.
...
+    `FOR UPDATE OF <tokens> SKIP LOCKED` against the `Tokens` rows, joined with the same
+    `INSERT ... ON CONFLICT DO NOTHING` backstop, so a claimant walks past a row a
+    concurrent claimant is already mid-claim on instead of colliding with it.
+
+**What each strategy actually changes, precisely — the mechanism has a narrower effect
+than "reduces lock contention" might suggest:**
+
+*   **Server-side unique-constraint errors are eliminated only for callers that take the
+    single-token `Lock` path directly.** `Selector.selectInternal` type-asserts the
+    configured `Locker` for `BatchLocker` and, when present (i.e. under Postgres
```
AkramBitar, line 229: **This paragraph asserts behaviour that the wiring bug above makes false.**

"`Selector.selectInternal` type-asserts the configured `Locker` for `BatchLocker` and, when present (i.e. under Postgres regardless of strategy ...), always claims its covering window through `LockBatch`" does not hold for the store the SDK actually wires in — `*tokenlockdb.StoreService` embeds `driver.TokenLockStore` as an interface and so never satisfies `BatchLocker`.

A reader configuring Postgres will conclude batch claiming is active and tune or measure against it, while production always runs the single-token path. The same claim appears in items 1 and 2 of the PR description.

Best fixed by making it true (add `LockBatch` to the driver interface); otherwise both this section and the description need correcting.

AkramBitar (commented):
Review of the Phase 6 lock-strategy work. `go build ./token/...` and `go vet` are clean on the touched packages; findings 1 and 2 below were verified empirically rather than inferred, and the notes say how.

Three of these look blocking to me:

- **The batch path is unreachable through the SDK's own wiring**, so the feature is inert in production and only the tests ever see it (finding 1). This also makes the docs and PR-description claims about `LockBatch` untrue as written (finding 8).
- **`HasEnoughSpendableTokens` fails outright on SQLite** (finding 2), turning what used to be a clean `SelectorInsufficientFunds` into a hard selection error.
- **The covering window has no size cap**, so a dust-heavy wallet deterministically exceeds Postgres's 65535 bind-parameter limit and then cannot make progress on any retry (finding 3).

The remaining five are smaller but worth folding in while this is open.

One thing that is **not** a defect, in case it comes up: the `FOR UPDATE ... SKIP LOCKED` claim mechanism itself checks out. `tryLockSkipLocked` sends a single statement (locking select in the `claimed` CTE, `INSERT ... SELECT FROM claimed` as the outer statement), so Postgres's implicit per-statement transaction holds the row locks across the INSERT — there is no window where they are released early, and wrapping the call in an explicit transaction would change nothing. `TestTokenLockStore_LockBatch_SkipLocked_SkipsRowLockedByConcurrentTx` passes against real Postgres here (3.57s), and it would fail if the locks were dropped before the INSERT. The genuine limits on `skipLocked` are finding 1 (it never runs in production), finding 7 (it skips rows locked for unrelated reasons), and the flat conflict rates this PR already reports honestly.

Not verified: `make checks`, `make unit-tests`, and the integration tests still unchecked in the test plan.

adecaro:

